Issue #13, opened 2023-08-09
Upload failed with success message because of timeout


Kuermel:
If the upload fails due to 30s timeout it will still be shown as successful. 
The error message should be adjusted and a higher timeout should be set for the upload. 
![image](https://github.com/SpecterOps/BloodHound/assets/6574969/26819d4d-3583-4cd0-842a-07780e7eb9dd)

This happens e.g. with a slow network connection or large JSON files. 

Via API the timeout value can be increased to 60s, but not higher than the statically defined "maximumTimeout" value.
https://github.com/SpecterOps/BloodHound/blob/main/cmd/api/src/api/middleware/middleware.go#L101

zinic:
There is a bit of technical debt in our request cancellation logic for eager timeouts that needs review. We'll track that effort and prioritization here once we get some internal discussion around it.

zinic:
I've been kicking the tires with a custom client that just takes a really long time to upload and ran into no issues. The PR should effectively remove the limitation you're running into @Marshall-Hallenbeck.

zinic:
This will roll into the next release of BloodHound CE - please feel free to reopen this issue if, after the release, you still experience this behavior.

PR #21, opened 2023-08-10
fix: remove aggressive timeout context setting middleware

zinic:
## Description

Remove a bunch of experimental machinery around controlling request context timeouts. This style of control leads to strange double-commit scenarios for the `http.Response` type and results in incorrect `504` responses where the client is behaving but behind a slow connection.

This changeset addresses https://github.com/SpecterOps/BloodHound/issues/13.

## Motivation and Context

There's a long history of debug attempts related to the code being cleaned up here. To save everyone a long read, the gist is that this was an experiment in aggressively trying to preserve UX for multi-user, single-tenant instances that back BloodHound Enterprise.

## How Has This Been Tested?

Integration test passes to validate that request handling operates as expected. One negative test harness written in Python that won't be committed.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/middleware/middleware.go
+++ b/cmd/api/src/api/middleware/middleware.go
@@ -1,24 +1,23 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-// 
+//
 // SPDX-License-Identifier: Apache-2.0
 
 package middleware
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -38,6 +37,11 @@ import (
 	"github.com/unrolled/secure"
 )
 
+const (
+	// Default timeout for any request is thirty seconds
+	defaultTimeout = 30 * time.Second
+)
+
 // Wrapper is an iterator for middleware function application that wraps around a http.Handler.
 type Wrapper struct {
 	middleware []mux.MiddlewareFunc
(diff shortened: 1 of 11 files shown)


cmd/api/src/api/middleware/middleware.go
```diff
@@ -93,113 +72,37 @@ func getScheme(request *http.Request) string {
...
-	maximumTimeout = time.Minute
-)
-
 // ContextMiddleware is a middleware function that sets the BloodHound context per-request. It also sets the request ID.
 func ContextMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
 		var (
 			startTime = time.Now()
-			requestID string
+			requestID = "ERROR"
```
superlinkx, line 80: Is there a reason to initialize requestID here? It seems like there's never a code path that will activate with it set to `ERROR` and if it did go off the rails, the zero value should be sufficient

superlinkx (commented):
Just a quick question on a variable initialization change. Planning to pull this one down and test locally before approval.

cmd/api/src/api/middleware/middleware.go
zinic: Not really. I'll revert this. It was more just my preference.

zinic: Reverted :)

juggernot325:
It looks like this middleware is the only place that used the `Prefer` header. Currently every endpoint in our OpenAPI documentation has this parameter specified. We should probably remove that to reflect this change in functionality. The question is, should we do it in this PR or a separate one?

cmd/api/src/api/middleware/middleware.go
superlinkx: Oh, I misread the error path. For some reason I read the `log.Errorf()` and assumed that was the end of the story. I still like this approach, but now it makes sense. I think either way it's written is fine now that I'm actually reading it properly.

zinic:
The comment about the `Prefer` header is a good question. I have a sneaking suspicion that it still might be worth supporting. We could pull the `Prefer` header logic out in its own middleware so that it can set user deadline expectations in the request context for further consumption downstream without each handler having to parse or inspect the `Prefer` header.

I think a follow up discussion around use of the header and its documentation is worth having with the team. This bug is painful enough, however, that I'm not keen on blocking merge of this change for long.

zinic:
After further internal team discussion, I'm going to rewrite portions of this PR to include better plumbing for the `Prefer` header so that users can bypass default transaction time-outs per-request.

zinic:
This changeset has been updated with follow-up work:
* A new type in the BH context package has been added to track user timeout preferences.
* `Prefer` header is no longer tossed out but instead is now parsed and stored in the request `context.Context` value for future introspection. The context timeout is now correctly preserved and may be relied on for downstream context-aware function calls.
* Cypher queries now honor the `Prefer` header correctly.
* Early cancellation of responses has been removed.

cmd/api/src/queries/graph.go
```diff
@@ -382,20 +386,23 @@ func (s *GraphQuery) RawCypherSearch(ctx context.Context, rawCypher string) (mod
...
+			availableRuntime := bhCtxInst.Timeout.Value
 
-				// The weight of the query is divided by 5 to get a runtime reduction factor. This means that query weights
-				// of 5 or less will get the full 30 seconds of runtime.
-				reductionFactor := preparedQuery.complexity.Weight / 5
+			log.Infof("Available timeout for query is set to: %.2f seconds", availableRuntime.Seconds())
 
-				// Reduce the available runtime
-				availableRuntime /= reductionFactor
+			if !s.DisableCypherQC && !bhCtxInst.Timeout.UserSet {
```
juggernot325, line 394: Do we have any automated tests that verify this timeout value is being exercise correctly? I did a quick skim and did not see one, but I could have missed it. Whether we do or don't, do you think it would be valuable to add a test that also considers the `Prefer` header use case?

superlinkx (commented):
Other than James' testing question, nothing really sticks out to me. Seems like a good way to replumb the Prefer header

cmd/api/src/queries/graph.go
zinic: Let me jam on this and update the changeset accordingly.

zinic: I've added three test cases that validate the inner QC timeout and user-preference overrides.

juggernot325 (approved):
LGTM! Thank you for adding those tests 💪

PR #26, opened 2023-08-14
feat: enable literal scrubbed user query logging

zinic:
## Description

This changeset improves the openCypher front end by adding an emitter type that can contain output configuration information. This type is used to add a switch in the format functions for stripping literals when outputting openCypher from the query model.

Stripped literals are replaced with a parameter named `$STRIPPED`. This allows the query to remain syntactically correct while removing any sensitive information contained in the literal values.

Property names and other syntax elements are currently not considered sensitive.

### Output Example

```json
{"level":"info","query":"match p = ()-[:HasSession]->(n)-[]->() where n.pii = $STRIPPED return p","time":"2023-08-14T09:38:22.575719651-07:00","message":"Executing user cypher query"}
```

### Other Changes

* Refactor the emitter type to output to the `io.Writer` type for flexibility.
* Refactor GraphQuery to output user queries into the application structured log.

## Motivation and Context

It's very hard to trace what queries may be problematic from the BH application log. In order to enable log output for user queries we need to first scrub any literal data in it.

## How Has This Been Tested?

Added unit tests to cover the new path.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/queries/graph.go
+++ b/cmd/api/src/queries/graph.go
@@ -1,17 +1,17 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


superlinkx (approved):
After a proper code review session, this all checks out. Excited to have this diagnostic tool in our bag.

PR #48, opened 2023-08-18
fix: move dev docker compose to better name

superlinkx:
## Description

Moves our development `docker-compose.yml` to `docker-compose.dev.yml` to more clearly mark that it's meant for development and to prevent `docker compose up` from even starting any containers by default in the project root.

This now also adds an FAQ item for how to remove the neo4j-data volume without restoring to a full wipe (`docker compose down -v`)

## Motivation and Context

There is sometimes confusion between our root `docker-compose.yml` (which is used for development and requires more setup to use) and our `examples/docker-compose/docker-compose.yml` (which is the way to run BloodHound CE as an end user). In addition, running `docker compose up` without specifying profiles doesn't really do anything helpful, as it will only start the common containers and not the API/UI, so making it harder to run without `just` seems warranted.

## How Has This Been Tested?

* Ran through our just commands for development
* Attempted to run `docker compose up` to ensure an error is presented rather than starting dev mode containers

## Types of changes

## Checklist:


zinic (approved):
:+1:

zinic (approved):
:+1: reads good, pull

superlinkx:
Added a small fix to our `just init` command to be idempotent (your config files will no longer be overwritten if they exist) and to ensure it can clean up after running a docker compose command without the default config file existing.

ddlees (approved):

PR #50, opened 2023-08-18
fix: incorrect wrapping with log

superlinkx:
## Description

Fixes places where we accidentally used %w in error logs

## Motivation and Context

`%w` is not a valid verb for zerolog, so while it feels the same as `fmt.Errorf` we can't actually use it. This will ensure that when these log lines get hit, we get proper errors logged.

## How Has This Been Tested?

This is a cosmetic change. All lines were audited and ensured there were no build/syntax errors afterward

## Types of changes

## Checklist:


juggernot325 (approved):
LGTM

PR #54, opened 2023-08-23
Update issue templates

elikmiller:


zinic (approved):
Fantastic place to start.

slokie-so:
Can we add the label "GitHub",  with the Jira integration, it allows us to track those through

juggernot325 (approved):
Thank you for making that change! I'm excited to get these in place :)

PR #55, opened 2023-08-23
feat: add gzip compression support

codydbentley:
## Description

- created CompressionMiddleware to add transparent support for gzip requests and responses
- added CompressionMiddleware to FOSS middleware stack
- modified file upload service to do naive content type checking on ingest files
- added new integration tests to check compressed file handling and ingest accuracy
- added new methods to API Client to support compressed requests

## Motivation and Context

As a BloodHound User, the Sharp/AzureHound systems generate uncompressed data and transmit it to the API.  I’d like Sharp/AzureHound systems to compress the data prior to send and decompress and validate prior to ingestion. 

This will allow clients to set a "Content-Encoding" header to "gzip" so the server can recognize and handle the compressed ingest data. As a bonus, clients can use "Accept-Encoding" to request gzip responses. This will happen transparently for all API requests initiated by the browser UI. 

## How Has This Been Tested?

Manual testing of file uploads via Postman (no UI support yet), and new integration tests.

## Types of changes

## Checklist:


cmd/api/src/api/middleware/compression.go
```diff
@@ -0,0 +1,115 @@
...
+				combinedEncodings := strings.Join(headerValues, ",") // "Content-Encoding: gzip, deflate; Content-Encoding: br;" = "gzip, deflate, br"
+				encodings := strings.Split(combinedEncodings, ",")
+
+				for _, encoding := range encodings {
+
+					request.Body, err = WrapBody(encoding, request.Body)
+					if err != nil {
+						errMsg := fmt.Sprintf("failed to create reader for %s encoding: %v", encoding, err)
+						log.Warnf(errMsg)
+						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("Error trying to read compressed request: %s", errMsg), request), responseWriter)
```
ddlees, line 72: The appropriate return code would be `415 Unsupported Media Type`

cmd/api/src/api/middleware/compression.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func CompressionMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(responseWriter http.ResponseWriter, request *http.Request) {
+		var (
+			gw  *GzipResponseWriter
+			err error
+		)
+
+		for headerName, headerValues := range request.Header {
```
ddlees, line 59: `request.Header` is a `map[string][]string` so there is no need to loop through the whole map to get the encoding values and what you have here might be a little overkill. Consider simplifying to something like:

``` go
if encodings := strings.Split(request.Header.Get(headers.ContentEncoding.String()), ","); len(encodings) > 0 {}
```

cmd/api/src/api/middleware/compression.go
```diff
@@ -0,0 +1,115 @@
...
+						errMsg := fmt.Sprintf("failed to create reader for %s encoding: %v", encoding, err)
+						log.Warnf(errMsg)
+						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("Error trying to read compressed request: %s", errMsg), request), responseWriter)
+						return
+					}
+				}
+
+			}
+
+			if headerName == "Accept-Encoding" {
```
ddlees, line 79: Consider simplifying to something like:

``` go
if acceptEncodings := strings.Split(request.Header.Get(headers.AcceptEncoding.String()), ","); len(encodings) > 0 {}
```

cmd/api/src/api/middleware/compression.go
```diff
@@ -0,0 +1,115 @@
...
+	switch strings.TrimSpace(encoding) {
+	case "gzip", "x-gzip":
+		newBody, err = gzip.NewReader(body)
+	case "deflate":
+		newBody, err = zlib.NewReader(body)
+	default:
+		log.Infof("unsupported encoding detected: %s\n", encoding)
+		err = errors.New("content encoding is not supported")
+	}
+	return newBody, err
```
ddlees, line 114: `gzip.NewReader` only returns a `Reader` you will need to set a no-op close method.

``` go
return io.NopCloser(newBody), err
```

cmd/api/src/api/v2/apiclient/ingest.go
```diff
@@ -71,6 +75,39 @@ func (s Client) SendFileUploadData(payload map[string]any, id int64) error {
...
+		return errors.New("zero bytes written to compressed body payload")
+	} else if err = gw.Close(); err != nil {
+		return fmt.Errorf("failed to close gzip writer: %w", err)
+	}
+
+	request, err := s.NewRequest(http.MethodPost, fmt.Sprintf("api/v2/file-upload/%d", id), nil, io.NopCloser(&body))
+	if err != nil {
+		return fmt.Errorf("failed to create compressed ingest request: %w", err)
+	}
+	request.Header.Set("Content-Encoding", "gzip")
```
ddlees, line 96: nit: encouraging the use of `headers.ContentEncoding.String()` to avoid potential typos.

cmd/api/src/api/v2/file_uploads.go
```diff
@@ -126,8 +127,12 @@ func (s Resources) ProcessFileUpload(response http.ResponseWriter, request *http
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
 	} else if fileUploadJob, err := fileupload.GetFileUploadJobByID(s.DB, int64(fileUploadJobID)); err != nil {
 		api.HandleDatabaseError(request, response, err)
-	} else if fileName, err := fileupload.SaveToTempFile(s.Config.TempDirectory(), request.Body); err != nil {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("Error saving ingest file: %v", err), request), response)
+	} else if fileName, err := fileupload.SaveIngestFile(s.Config.TempDirectory(), request.Body); err != nil {
+		if errors.Is(err, fileupload.InvalidIngestFileType) {
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("Error saving ingest file: %v", err), request), response)
```
ddlees, line 132: We might want to return a status code 415 here but I'm not totally sure because we're checking the mime type of the persisted file and not the payload itself. Unless someone else has some feels I say it's your call.

cmd/api/src/api/middleware/compression.go
codydbentley: Good call out, I've made this change

cmd/api/src/api/middleware/compression.go
codydbentley: `Header.Get()` won't work here because it will only return the first header value and ignore any others. I opted to use `Header.Values()`, which still requires joining and splitting again, but it has been collapsed.

cmd/api/src/api/middleware/compression.go
codydbentley: This has been simplified using `Header.Values()`

cmd/api/src/api/middleware/compression.go
codydbentley: At a glance it looks like `gzip.NewReader()` returns `io.Reader`, but it's actually a custom `&Reader` type from the `gzip` package that does implement `io.ReadCloser`.

codydbentley: I made sure the `wrapBody()` function explicitly returns the `io.ReadCloser` interface to enforce this in case we ever decide to support other encodings.

cmd/api/src/api/v2/apiclient/ingest.go
codydbentley: Done!

cmd/api/src/api/v2/file_uploads.go
codydbentley: Since this part of the request is concerned with the contents of the request rather than the transport, I opted to handle this how we handle other places where a marshal failure occurs, which is consider it a bad request. We had a team discussion which resulted in creating an internal ticket to track improving the validation step beyond just checking for plain text content.

cmd/api/src/api/middleware/compression.go
codydbentley: Coming back to this, I have decided to implement the 415 code in only one specific case. I based my decision on following the semantic of "400 means the request is wrong due to incorrect/invalid input", and "415 means the request seems valid, but we don't support the type sent". This means that most cases I tested previously would remain 400, but I added a new case for a 415 response where the user sends a content encoding we don't support.

codydbentley:
@ddlees thanks for the review, I have updated this and replied to the conversations.

cmd/api/src/services/fileupload/file_upload.go
```diff
@@ -82,13 +85,22 @@ func GetFileUploadJobByID(db FileUploadData, jobID int64) (model.FileUploadJob,
...
-	} else if _, err := io.Copy(tempFile, fileData); err != nil {
-		return "", fmt.Errorf("error copying temp file: %w", err)
+func SaveIngestFile(location string, fileData io.ReadCloser) (string, error) {
+	var typeBuf = make([]byte, 512)
+	tempFile, err := os.CreateTemp(location, "bh")
+	if err != nil {
+		return "", fmt.Errorf("error creating ingest file: %w", err)
+	}
+	defer tempFile.Close()
+	if _, err := io.Copy(tempFile, fileData); err != nil {
```
computator, line 95: Not sure if it's worth the effort depending on how often we expect invalid files, but you could use `io.CopyN()` or `io.ReadFull()` or something to get the first 512 bytes of the file and do the type detection immediately instead of having to pipe the entire request out to disk before you can go back and check if it's valid or not. Then if it's valid you can just write the buffer out to the file and then use `io.Copy()` to copy the rest like before.

This would also allow us to not even create the temporary file until after the content type checking, thus avoiding some extra files on disk (similar to the last comment about deleting temporary files)

cmd/api/src/services/fileupload/file_upload.go
```diff
@@ -82,13 +85,22 @@ func GetFileUploadJobByID(db FileUploadData, jobID int64) (model.FileUploadJob,
...
-		return "", fmt.Errorf("error creating temp file: %w", err)
-	} else if _, err := io.Copy(tempFile, fileData); err != nil {
-		return "", fmt.Errorf("error copying temp file: %w", err)
+func SaveIngestFile(location string, fileData io.ReadCloser) (string, error) {
+	var typeBuf = make([]byte, 512)
+	tempFile, err := os.CreateTemp(location, "bh")
+	if err != nil {
+		return "", fmt.Errorf("error creating ingest file: %w", err)
+	}
+	defer tempFile.Close()
```
computator, line 94: We are closing the file with the moved defer here, which is good, but we aren't deleting the invalid temporary file if an error occurs.

cmd/api/src/api/middleware/compression.go
```diff
@@ -0,0 +1,113 @@
...
+			for _, encoding := range strings.Split(contentEncodingString, ",") {
+				encoding = strings.TrimSpace(encoding)
+				request.Body, err = wrapBody(encoding, request.Body)
+				if err != nil {
+					errMsg := fmt.Sprintf("failed to create reader for %s encoding: %v", encoding, err)
+					log.Warnf(errMsg)
+					if errors.Is(err, unsupportedEncodingError) {
+						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusUnsupportedMediaType, fmt.Sprintf("Error trying to read request: %s", errMsg), request), responseWriter)
+					} else {
+						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("Error trying to read request: %s", errMsg), request), responseWriter)
```
computator, line 72: Shouldn't this actually be a 5xx? A failure to process the request is only a client error if the request was malformed or something. If the issue is with our code or server or something then that's still a server error.

ddlees: This case would indicate a failure to instantiate an instance of the compression lib's `ReadCloser`. We would have to dig into the error to derive if it was because of a client issue or not. I'm inclined to agree with @computator  and say 500 is likely the the more appropriate response code.

cmd/api/src/api/v2/apiclient/ingest.go
```diff
@@ -71,6 +75,39 @@ func (s Client) SendFileUploadData(payload map[string]any, id int64) error {
 	}
 }
 
+func (s Client) SendCompressedFileUploadData(jsonFile io.Reader, id int64) error {
+	var (
+		body bytes.Buffer
+		gw   = gzip.NewWriter(&body)
+	)
+
+	if n, err := io.Copy(gw, jsonFile); err != nil {
```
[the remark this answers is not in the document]
ddlees: The json file is an embedded file living in `embed.FS` so I want to say the `io.Reader` in this case is just a `[]byte`. That is to say all the data is already living in memory. Because of that, I think @computator's proposed optimization would not have that huge of a performance impact.

That said, if we wanted to load data directly from disk in addition to what's in `embed.FS` then including the change would definitely grant us that flexibility.

That said, it's a small change but I don't think it's necessary if you'd rather not include it in this changeset.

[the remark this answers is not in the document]
codydbentley: Using a pipe here could introduce a bit of complexity here to solve a problem that doesn't seem to exist currently. I think I will skip this change until we have a need for it.

cmd/api/src/api/v2/apiclient/ingest.go
codydbentley: Good catch, I found the other places and fixed them

cmd/api/src/services/fileupload/file_upload.go
codydbentley: I think it's fine for now. We currently accept all files already (we have an orphan cleanup process), and since this is more of a client feedback improvement and less of an ingest validation improvement (which we we have a ticket open), I believe we can leave as is.

cmd/api/src/services/fileupload/file_upload.go
codydbentley: We have an orphan removal process that takes care of this

cmd/api/src/api/middleware/compression.go
codydbentley: After inspecting how the different readers are instantiated in the stdlib, all of the errors raised are the result of bad data (empty/unexpected EOF, bad header, etc). This means all the failure cases would be the result of a bad request because bad data was sent. A 500 also signals to the client that it cannot do anything different to fix the request, which in this case would be incorrect.

computator (approved):

cmd/api/src/api/v2/apiclient/ingest.go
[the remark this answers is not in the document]
ddlees: @computator FWIW AzureHound does not use this.

ddlees (approved):

PR #58, opened 2023-08-24
chore: add St Bernard to help with module syncing in the workspace

superlinkx:
## Description

Adds a new Go based tool as a sort of Swiss Army Knife of useful mini-commands for the project. Currently, the only tool implemented is `modsync` which simplifies syncing all modules in the Go workspace, allowing for an easy way to fix issues with local dev when dependencies change.

This tool is wired up with `just modsync` and is run instead of `just build` during `just init`. Note that this tool may be merged with Beagle in the future, but the implementation was needed now to reduce pain when helping devs reset their environment.

## Motivation and Context

Solves the problem of `go.work.sum` being out of sync and preventing `just bh-dev` from starting properly, by giving a small command to run. The previous solution was to use `just build`, but it does _a lot_ of additional work and always breaks docker build caches for the UI image, so it was making `just init` unbearably slow. This was done using Go because it was faster to prototype, and may eventually merge into Beagle instead. Please use `just modsync` rather than using St Bernard directly, as the `just` recipe will be updated if the command moves to the Beagle project later.

## How Has This Been Tested?

Local runs of `just init` were significantly faster with `just modsync` compared to `just build` and because the Docker cache layers aren't disturbed, subsequent runs typically saw an improvement of around 5x compared to first run, making `just init` significantly more bearable.

## Types of changes

## Checklist:


justfile
```diff
@@ -37,13 +38,28 @@ init wipe="":
...
+  fi
+
+  echo "Start integration testing services"
+  if [[ "{{wipe}}" == "clean" ]]; then
+    echo "Clear volumes and restart testing services without cache"
+    just bh-testing-clear-volumes
+    just bh-testing build --no-cache
+  fi
+
+  just bh-testing
```
ddlees, line 60: Consider removing this line.

I'm not huge fan of spinning up services for project initialization. IMO we should leave that to the developer to decide when those testing services should be started especially since there are devs on the team that prefer to run those services on their host machine or elsewhere.

superlinkx: Fair enough, I'm fine with removing it. Goal was to reduce mental load with making sure to restart bh-testing (or know to start it at all), but I think you're right that it would be better to let devs decide on their own

ddlees (approved):

PR #47, opened 2023-08-18
feat: add api key mgmt to user profile

ddlees:
## Description

- Adds a button to launch the API Key Management Dialog from the User Profile view
- Migrates the `UserProfile` view into `bh-shared-ui` to reduce duplication of code
- Migrates several reusable components into `bh-shared-ui`
- Decoupled global state to create `NotificationProvider` and `AppNotifications` components to support the above Migrations
- Added response types to relevant client APIs

## Motivation and Context

Currently, only users with the Administrator role can create/revoke API keys. This change allows users to manage their own API keys from the My Profile view. Additionally, several components have been duplicated to support the enterprise variant of BloodHound. This change includes several refactors to reduce the unnecessary duplication and to better keep the enterprise and community editions in sync.

## Screenshots (if appropriate):

image

image

image

image

image

image

## Types of changes

-   [ ] Chore (a change that does not modify the application functionality)
-   [ ] Bug fix (non-breaking change which fixes an issue)
-   [x] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

-   [ ] Documentation updates are needed, and have been made accordingly.
-   [ ] I have added and/or updated tests to cover my changes.
-   [x] All new and existing tests passed.
-   [ ] My changes include a database migration.

PR #61, opened 2023-08-28
Search Current Results

maffkipp:
## Description
- Adds a button in the bottom left corner of the Explore view that allows a user to search through all Nodes. The button opens a search panel with the following behavior:
    - searches nodes in the currently displayed graph by label or objectid
    - highlights the node and displays the associated information in the entity panel when a search result is clicked
    - closes when the user clicks outside of the search panel
- Corrects some behavior around layouts that caused the displayed graph to zoom in and out when a new layout was selected
- Adjusts the size of the search and entity panels at larger screen widths

## Motivation and Context
- Allows the user to search by text in the currently displayed graph, which should make parsing through large graph results easier and quicker

## How Has This Been Tested?
- Added unit tests for the search panel

https://github.com/SpecterOps/BloodHound/assets/16313351/64e76885-1fac-4b4a-8e96-e5dcaa914036

## Types of changes

## Checklist:


cmd/ui/src/components/GraphButton/GraphButton.tsx
```diff
@@ -46,18 +45,16 @@ const useStyles = makeStyles((theme) => ({
...
+const GraphButton: FC<GraphButtonProps> = ({ onClick, displayText, disabled }) => {
     const styles = useStyles();
-    const { reset } = useCamera();
 
     return (
         <Button
-            onClick={() => {
-                onClick();
-                reset();
-            }}
```
urangel, line 60: Thanks for surfacing and cleaning up this reset call :raised_hands:

urangel (changes requested):

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -225,8 +226,15 @@ export const GraphEvents: FC<GraphEventProps> = ({
...
+        if (selectedEdge) setHighlightedNode(null);
     }, [selectedEdge]);
 
+    useEffect(() => {
+        if (selectedNode?.graphId) {
+            setSelectedEdge(null);
+            setHighlightedNode(selectedNode.graphId);
+        }
+    }, [selectedNode])
+
```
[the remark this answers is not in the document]
maffkipp: From my understanding, this is not the case -- previous to these changes, we were not watching the selected node in the redux store for changes, we were just highlighting the node and updating the store in the case of a click event. 

We were also previously not storing the node's ID in redux (just the object ID), which is the identifier sigma needs, so I had to add that `graphId` property as a workaround.

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -139,6 +141,8 @@ const GraphView: FC = () => {
     const opts: GlobalOptionsState = useSelector((state: AppState) => state.global.options);
     const formIsDirty = Object.keys(useSelector((state: AppState) => state.tierzero).changelog).length > 0;
     const [graphologyGraph, setGraphologyGraph] = useState<MultiDirectedGraph<Attributes, Attributes, Attributes>>();
+    const [currentNodes, setCurrentNodes] = useState<GraphNodes>({});
```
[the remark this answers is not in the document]
maffkipp: That is the case, but to get it in the format we want and filter out non-nodes, we'd have to call `transformFlatGraphResponse` on it again before passing it in (or something like it). Since we were already doing that in the first `useEffect` I just grabbed the nodes from there, but if you think its better to handle that logic separately I can do that too.

packages/javascript/bh-shared-ui/src/components/SearchCurrentNodes/types.ts
```diff
@@ -0,0 +1,12 @@
...
+export type GraphNode = {
+    label: string;
+    kind: string;
+    objectId: string;
+    lastSeen: string;
+    isTierZero: boolean;
+    descendent_count?: number | null;
+};
+
+export type GraphNodes = Record<string, GraphNode>;
```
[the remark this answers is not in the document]
maffkipp: We are using them once in `js-client-library`, in the `GraphData` type. My thought was that there is a change to pull the client library into the shared UI [being worked on here](https://github.com/SpecterOps/BloodHound/pull/47/files#diff-71c2c6df779e612c283a755cc2e47bb9a5f6c8f6b3a079344dd393dacd31cdd9R20), and this could be a temporary solution until we can just use those types directly. Otherwise I agree it is a bad idea to have type definitions duplicated across projects.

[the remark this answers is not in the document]
urangel: Ah, I see. Yea, I'll leave this to your discretion. Mostly wanted to point out that we had it defined somewhere else already.

cmd/ui/src/views/Explore/GraphView.tsx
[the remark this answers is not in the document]
urangel: I see what you mean. This does make it so we don't have to filter that out again. Thanks for pointing that out.

urangel (approved):
Nice work!

packages/javascript/bh-shared-ui/src/components/SearchCurrentNodes/SearchCurrentNodes.test.tsx
```diff
@@ -0,0 +1,114 @@
+import {  fireEvent, render } from "@testing-library/react";
```
elikmiller, line 1: New files like this one need to include the license information.

packages/javascript/bh-shared-ui/src/components/SearchCurrentNodes/SearchCurrentNodes.tsx
```diff
@@ -0,0 +1,99 @@
...
+            const filteredNodes = flatNodeList.filter(node => {
+                const label = node.label.toLowerCase();
+                const objectId = node.objectId.toLowerCase();
+                const lowercaseInputValue = inputValue?.toLowerCase() || '';
+
+                if (inputValue === '') return false;
+
+                return label.includes(lowercaseInputValue) || objectId.includes(lowercaseInputValue);
+            });
+            setItems(filteredNodes);
```
elikmiller, line 48: There is some, in my opinion, noticeable lag that can occur when changing this value. Do you reckon there is any easy fix we could consider?

cmd/ui/src/views/Explore/GraphView.tsx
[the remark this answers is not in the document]
elikmiller: transformFlatGraphResponse makes me sad. Longer term we should consider a refactor to homogenize how we store graph information in the global store.

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -194,20 +200,33 @@ const GraphView: FC = () => {
...
+        }
+    ];
+    
     return (
         <div style={{ position: 'relative', height: '100%', width: '100%', overflow: 'hidden' }} data-testid='explore'>
             <SigmaChart
                 rankDirection={RankDirection.LEFT_RIGHT}
                 options={options}
                 graph={graphologyGraph}
+                currentNodes={currentNodes}
```
elikmiller, line 225: To me it feels somewhat strange to pass the graphology graph (which details the nodes) and pass `currentNodes` as a separate prop.

maffkipp: in order to support both regraph and sigma in the search component, we are passing in the transformed data from the redux store (the output of `transformFlatGraphResponse`), as opposed to data directly from the graph. This is similar to the issue I discussed with Uli above; since we are already running this transformation in the parent component, this prevents us from having to run it in the child as well. 

Definitely open to alternative solutions, I just didn't have a good idea that wasn't messy in other ways!

elikmiller: We'll chalk it up as an indicator that this component has some good refactor opportunities

packages/javascript/bh-shared-ui/src/components/SearchCurrentNodes/SearchCurrentNodes.tsx
elikmiller: The changes you made make a night and day difference. Thanks for digging in!

elikmiller (approved):

PR #68, opened 2023-08-30
User auth method update fix

maffkipp:
## Description
- When a user's auth method is updated to "username/password", the form still sends the previously selected `samlProviderId` in the request body, which prevents the auth method from actually being updated. In order to make the update correctly, we need to set this field to an empty string when "username/password" is chosen. 

## Motivation and Context
- Corrects a bug which prevents administrators from updating a user's auth method from SAML.

## How Has This Been Tested?
- Manual testing to confirm users can be switched freely between the two auth methods.

## Types of changes

## Checklist:


elikmiller:
The changes seem solid however I would love to see some accompanying tests.

urangel (approved):
I like how much of the repetition was cleared out by putting it in the setup function in the unit tests. The changes worked out for me locally.

PR #76, opened 2023-09-05
Fix azure owns owner

rvazarkar:
## Description
Changes certain edges created during ingest from Owns to Owner to reflect the correct portion of azure management these require.

## Motivation and Context
Some of our azure edges were incorrectly created as a different type due to confusion between owner and owns. This corrects the issue.

## Types of changes

## Checklist:


cmd/api/src/migrations/manifest.go
```diff
@@ -1,35 +1,68 @@
...
 )
 
+func Version_508_Migration(db graph.Database) error {
+	defer log.Measure(log.LevelInfo, "Migrating Azure Owns to Owner")()
+
+	return db.BatchOperation(context.Background(), func(batch graph.Batch) error {
+		return batch.Relationships().Filterf(func() graph.Criteria {
+			return query.And(
+				query.Kind(query.Start(), azure.Entity),
+				query.KindIn(query.End(), azure.ManagementGroup, azure.ResourceGroup, azure.Subscription, azure.KeyVault),
```
ddlees, line 41: I believe you'll also need to query for these end node kinds in the migration:
- `azure.AutomationAccount`
- `azure.ContainerRegistry`
- `azure.LogicApp`
- `azure.VM`
- `azure.VMScaleSet`
- `azure.WebApp`
- `azure.FunctionApp`
- `azure.ManagedCluster`

ddlees: Actually, it may not be necessary but I don't think it hurts anything to include these in the migration.

packages/go/ein/azure.go
ddlees, line 1: I think we also need to do this for `ConvertAzureKeyVaultOwnerToRels`

ddlees (approved):

zinic (approved):
Looks good. Gave the migration a closer look as well.

PR #82, opened 2023-09-07
draft: built in groups get filtered out 

brandonshearin:
## Description

some of the built in groups were not searchable from the explore page because of the way the dawgs query was written in `queries.SearchNodesByName()`.  The dawgs query had a filter that excluded any nodes with a label of `:Group`, but there are cases when we want to return nodes with the labels `:Group:ADLocalGroup`.  This PR updates the dawgs query to express this logic

the equivalent cypher we want to express in dawgs is:
```MATCH (n) WHERE NOT (n:ADLocalGroup AND NOT n:Group)```

this query will match nodes like `(n:AdLocalGroup:Group)` but WON'T match `(n:Group)`

i also updated the query formatter in the cypher frontend package to wrap negation clauses in parens.  

## How Has This Been Tested?

the test suite for the cypher parser has been updated so that test cases with negation clauses are now correctly wrapped in parens

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/queries/graph.go
+++ b/cmd/api/src/queries/graph.go
@@ -276,12 +276,15 @@ func searchNodeByKindAndEqualsName(kind graph.Kind, name string) graph.Criteria 
 			query.Equals(query.NodeProperty(common.Name.String()), name),
 			query.Equals(query.NodeProperty(common.ObjectID.String()), name),
 		),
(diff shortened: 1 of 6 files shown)


zinic (approved):
Nothing stands out. Exception for parenthetical expressions in a negation looks reasonable to me.

PR #90, opened 2023-09-11
fix asset_group_selectors table and data

irshadaj:
## Description
This migration cleans up the data in the `asset_group_selectors` table in order to allow us to add a uniqueness constraint on the `name` column.
* Rows that have duplicated names and selectors will be removed so only one remains
* Rows that have duplicated names but different selectors, will be updated to append a `_2` at the end of their `name` data.

Following the above 2 updates, the uniqueness constraint will be re-added on `name`.

## How Has This Been Tested?
Manual testing on the database using dummy data

First I set up the database to have a total of 5 duplicates with both name and selector (2 on `test2` and 3 on `test3`). This simulates having a database WITHOUT the uniqueness constraint, and data already violating said constraint that we want to add

Then I ran the first migration query, this deletes the rows with duplicated names AND selectors, so there are 3 total deleted rows:

This leaves us with no more rows with the same name and selector:

Now on to the second query. I run this and expect the 2 duplicated rows with name `test3` to be updated, and one left as is:

Querying the table again, we notice the appended numbers to the names for what used to be `test3`:

Now all of the data is cleaned up, we can proceed to add the constraint:

In order to verify the constraint, I try to add a duplicate name+selector record and get a failure:

I try to add a duplicate name-only record, and get a failure again:

This verifies all of the functionality pertaining to this change.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/v2/cypher_search.go
+++ b/cmd/api/src/api/v2/cypher_search.go
@@ -1,17 +1,17 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-// 
+//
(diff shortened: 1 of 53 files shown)


superlinkx:
Validated with the following divergence:

After setting up my table to match the test rows given above, I skipped running the queries by hand. Instead, I changed the migration table to have the following version: 5.0.8. This meant that when I restarted the server, I could validate that the migration for 5.0.9 ran successfully. I then validated the end result as above, getting the same dataset (looks like it's somewhat non-deterministic about which name gets to be the primary name, but I don't believe this will be an issue). I then tried adding a new row with `INSERT INTO asset_group_selectors (name, selector) VALUES ('test', 'test')` and got the expected constraint failure.

image

superlinkx (approved):

PR #96, opened 2023-09-13
Export Graph as JSON

brandonshearin:
## Motivation and Context

UPDATE:
- in order to ensure that we export the same shape of data across search experiences, i call the existing `transformToFlatGraph` utility in the saga for the single node search.  The goal of this addition is to massage the raw API response from that saga into the same shape that we get from other searches, such as pathfinding and cypher search.
- i have kept the export field in the explore reducer so that we have the option to decouple the data we export from the data we render to the canvas.  

This PR adds the ability to export your graph data to a JSON file.  The format of the JSON file will be identical to the API response of the search that was performed.  This "export to JSON" feature will be enabled for any search that is possible from the explore page, whether you search using the single node search, the pathfinding search, or the cypher editor.

This PR also includes a slight UX improvement.  It collapses all layout-related buttons into a "Layout" menu in the bottom left.  Likewise, all export-related buttons will be placed into the "Export" menu.

EDIT: i moved `<GraphMenu/>` and `<GraphButton/>` to the bh-shared-ui.  i also moved the `exportToJSON()` func to bh-shared-ui

## How Has This Been Tested?

Unit tests have been added to the GraphButtons.test.tsx suite.  Here are some instructions about how to manually tests these changes:

- Perform any kind of search from the explore page: single node, pathfinding, or cypher
- If a graph is rendered to the canvas, then the JSON button will be enabled
- Clicking the JSON button will download a file named `bh-graph.json`.  If you open the file, the contents will be the exact same as the raw API response from the related api call.

New UX:  layout + export menus
Screenshot 2023-09-13 at 2 15 45 PM

opening the layout menu:
Screenshot 2023-09-13 at 2 15 56 PM

opening the export menu:
Screenshot 2023-09-13 at 2 16 12 PM

clicking the export button downloads a file named `bh-graph.json`:
Screenshot 2023-09-13 at 2 17 03 PM

## Types of changes

## Checklist:


cmd/ui/src/components/GraphButtons/GraphButtons.tsx
```diff
@@ -91,4 +107,61 @@ const GraphButtons: FC<GraphButtonsProps> = ({ rankDirection, options, nonLayout
     );
 };
 
+const GraphMenu: FC<{ label: string; children: ReactNode }> = ({ children, label }) => {
```
maffkipp, line 110: I think the `GraphMenu` component and the `exportToJson` functionality would be good candidates to move to the `bh-shared-ui` library, so that they don't need to be duplicated for enterprise

brandonshearin: gotcha. moved both of these!

packages/javascript/bh-shared-ui/src/components/GraphMenu/GraphMenu.tsx
```diff
@@ -0,0 +1,54 @@
...
+                aria-expanded={open ? 'true' : undefined}
+                displayText={label}></GraphButton>
+            <Menu
+                id={`${label}-menu`}
+                open={open}
+                anchorEl={anchorEl}
+                onClose={handleClose}
+                MenuListProps={{
+                    'aria-labelledby': `${label}-button`,
+                }}>
```
urangel, line 45: It may look nicer to pass positioning props for the popover element so that the popover doesn't overlay on the menu button when it opens.

brandonshearin: thanks for this idea! updated the popover position:

Screenshot 2023-09-19 at 9 50 35 AM

packages/javascript/bh-shared-ui/src/utils/exportGraphData.ts
elikmiller, line 1: This file needs the license

elikmiller (commented):
I see some errors during build:

```
src/components/SigmaChart.tsx:34:34 - error TS2307: Cannot find module './GraphButton/GraphButton' or its corresponding type declarations.

34 import { GraphButtonProps } from './GraphButton/GraphButton';
                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/views/Explore/GraphView.tsx:27:34 - error TS2307: Cannot find module 'src/components/GraphButton/GraphButton' or its corresponding type declarations.

27 import { GraphButtonProps } from 'src/components/GraphButton/GraphButton';
                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  src/components/SigmaChart.tsx:34
     1  src/views/Explore/GraphView.tsx:27
```

elikmiller:
And looks like you have a couple of merge conflicts, they seem related to the build issues above

urangel (dismissed):
The `GraphButtonProps` import in GraphView.tsx and SigmaChart.tsx need to be updated to come from the bh-shared-ui

cmd/ui/src/ducks/explore/reducer.ts
```diff
@@ -45,6 +46,8 @@ const graphDataReducer = (state = initialGraphDataState, action: types.GraphActi
             // handle delete nodes (action.ids)
         } else if (action.type === types.GRAPH_INIT) {
             draft.init = action.payload;
+        } else if (action.type === types.SAVE_RESPONSE_FOR_EXPORT) {
+            draft.export = action.payload;
```
elikmiller, line 50: Out of curiosity - why do we want to store data for export separately? Can we not use whatever the current chartProps/items are to populate the JSON file to be exported?

brandonshearin: the data that we export is supposed to be a 1:1 copy of the response we get from the API.  we cant use chartProps for this, because the data that gets saved to chartProps has been massaged into a new shape for rendering purposes, so I added the "export" field as the place where we can save that raw API response.

elikmiller: That makes sense - thank you for the clarification.

elikmiller: Depending on the specific API call being made the resulting JSON structure could be quite different - is that accurate?

elikmiller: So based on our internal conversation we decided that the desired end state is to transform our JSON before producing a file so that the exported files have a consistent structure.

Issue #9, opened 2023-08-09
Security Bad Practice - Hardcoded Credentials in bloodhound.config.json


Daniel-CS-Team:
I consider it a very bad practice to hardcode any kind of credentials into config-files.

Maybe you can add a little entrypoint.sh wrapper, that retrieves credentials from environment variables, so that one can set them at runtime. I am no pro-developer, but something like this should do the job:

```
echo "{\"version\":1,\"bind_addr\":\"0.0.0.0:8080\",\"metrics_port\":\":2112\",\"root_url\":\"http://127.0.0.1:8080/\",\"work_dir\":\"/opt/bloodhound/work\",\"log_level\":\"${LOG_LEVEL:-INFO}\",\"log_path\":\"bloodhound.log\",\"features\":{\"enable_auth\":true},\"tls\":{\"cert_file\":\"\",\"key_file\":\"\"},\"database\":{\"connection\":\"user=${DB_USER:-bloodhound} password=${DB_PASSWORD:-bloodhoundcommunityedition} dbname=${DB_NAME:-bloodhound} host=${DB_HOST:-app-db}\"},\"neo4j\":{\"connection\":\"${NEO_CONNECTION_STRING:-neo4j://neo4j:bloodhoundcommunityedition@graph-db:7687/}\"},\"collectors_base_path\":\"/etc/bloodhound/collectors\"}" > /bloodhound.config.json
exec bloodhound -config /bloodhound.config.json
```

BTW: Thanks for your great work. Hopefully, we can make it even better.

Acebond:
Having creds stored in config files is a correct way to keep/store credentials.

Acebond:
How about instead of just saying "very bad practice" and having a whinge you demonstrate and explain why storing credentials in a config file is bad?

Acebond:
Also your proposed solutions to a problem and doesn't exist has multiple naïve errors. 
If any of a environment variables contain JSON escape characters your bash script breaks imminently and would crash.

zinic:
At minimum, we can provide additional documentation and justification for our practice here even if the risk is indeed low.

The intent of the issue appears genuine and requires an equally genuine response from the BloodHound team, even if we don't take on the author's suggested fixes.

zinic:
Thanks for the additional information. It will help frame our internal discussion about this issue.

PR #99, opened 2023-09-14
Env fixes and docs

superlinkx:
## Description

- Fix a problem with configuration values with underscores when provided as environment variables.
- Document our environment variables and how to use them to provide sensitive properties instead of putting those properties in the `bloodhound.config.json`.

## Motivation and Context

https://github.com/SpecterOps/BloodHound/issues/9

While we have always supported configuring BHCE through environment variables, there is a distinct lack of documentation. Furthermore, while preparing the documentation, a bug was found in how we parse our environment variables that prevented a few of them from actually working. This PR seeks to resolve both gaps.

## How Has This Been Tested?

- Manually tested by adding various environment variables to our dev docker compose and ensuring they worked properly
- Added a new regression test case for the patched environment variable parsing bug

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/config/config_test.go
+++ b/cmd/api/src/config/config_test.go
@@ -1,17 +1,17 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
 // distributed under the License is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-// 
+//
 // SPDX-License-Identifier: Apache-2.0
 
 package config_test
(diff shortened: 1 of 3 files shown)


cmd/api/src/database/file_upload.go
```diff
@@ -64,6 +64,10 @@ func (s *BloodhoundDB) GetAllFileUploadJobs(skip int, limit int, order string, f
 		result = s.db.Model(model.FileUploadJob{}).Count(&count)
 	}
 
+	if result.Error != nil {
+		return nil, 0, CheckError(result)
+	}
```
superlinkx, line 69: This is to fix an issue the linter was warning me about. We weren't doing anything with the result value after doing a count, so I added error checking after the count step.

cmd/api/src/config/config_test.go
```diff
@@ -24,66 +24,156 @@ import (
 )
 
 func TestSetValuesFromEnv(t *testing.T) {
```
superlinkx, line 26: I updated this test to use subtests to make it easier to see which test cases are being used

cmd/api/src/config/config_test.go
```diff
@@ -24,66 +24,156 @@ import (
...
+			"bhe_database_secret=bhe4eva",
+		}))
+
+		assert.Equal(t, "neo4j://neo4j:neo4jj@localhost:7070/neo4j", cfg.Neo4J.Neo4jConnectionString())
+		assert.Equal(t, "postgresql://bhe:bhe4eva@localhost:5432/bhe", cfg.Database.PostgreSQLConnectionString())
+	})
+
+	// This test ensures that fields that could be considered sensitive are configurable through expected environment
+	// variables. Not all fields in a given struct are necessarily sensitive and should not be included here.
+	t.Run("all sensitive fields are configurable", func(t *testing.T) {
```
superlinkx, line 79: This subtest was added to ensure that all fields we deem "potentially sensitive" are properly covered with working environment variables. This is a regression test to prevent future issues with environment variable parsing affecting this critical subset that is most likely to be configured with environment variables in the wild.

cmd/api/src/config/reflect.go
```diff
@@ -253,8 +253,9 @@ func SetValue(target any, path, value string) error {
 		cursor    = indirectOf(target)
 	)
 
-	for idx, nextPathPart := range pathParts {
+	for idx := 0; idx < len(pathParts); idx++ {
```
superlinkx, line 256: The changes to this loop and associated logic allows us to parse field names that contain underscores while still using underscores as path separators. Previous logic did allow for underscores in field names, but only when it was the last field in a path. Fields with underscores at the beginning or middle of a path would be ignored previously.

examples/docker-compose/docker-compose.yml
```diff
@@ -64,6 +64,11 @@ services:
     image: docker.io/specterops/bloodhound:${BLOODHOUND_TAG:-latest}
     environment:
       - bhe_disable_cypher_qc=${bhe_disable_cypher_qc:-false}
+      ### Add additional environment variables you wish to use here.
```
superlinkx, line 67: This seemed like the best way to help users know how to get their environment variables to be respected within the BloodHound container (without reading about it in the Wiki)

PR #104, opened 2023-09-15
Saved user cypher queries (GET)

irshadaj:
## Description

This MR adds the ability for users to CRUD custom cypher queries that persist through server restarts (but not through cleared DBs). Access to these queries will be limited to self for now.

## How Has This Been Tested?
Added unit tests for model, handler

Manual testing:

Set up a custom query in the database as follows
pgadmin

Then tried different combinations of params and verified that each one returns the expected response:

sort error

sorted

invalid filter column

invalid predicate

success

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/slices/slices.go
+++ b/packages/go/slices/slices.go
@@ -82,7 +82,6 @@ func UniqueBy[T any, U comparable](slice []T, fn func(T) U) []T {
 	return out
 }
 
-// TODO delete this function, use utils.Contains instead
 // Contains returns true if a slice contains an element that is equal to the given value
 func Contains[T comparable](slice []T, value T) bool {
 	for _, sliceValue := range slice {


packages/go/slices/slices.go
```diff
@@ -82,6 +82,7 @@ func UniqueBy[T any, U comparable](slice []T, fn func(T) U) []T {
 	return out
 }
 
+// TODO delete this function, use utils.Contains instead
```
elikmiller, line 85: Is this something to cleanup before a merge?

elikmiller:
Looks like there is an issue when returning the `count` field in the response.

Screenshot from 2023-09-21 09-54-37

Screenshot from 2023-09-21 09-55-02

I've highlighted the rows that are associated with the authenticated user. I would expect the `count` returned to be `3` rather than the given `6`

cmd/api/src/model/queries.go
```diff
@@ -0,0 +1,66 @@
+package model
+
+import "fmt"
+
+type SavedQuery struct {
+	UserID string `json:"user_id"`
+	Name   string `json:"name"`
```
brandonshearin, line 7: where do we typically define sql-level constraints for a model, such as the fact that there is a uniqueness constraint on the `name` field?

cmd/api/src/api/registration/v2.go
```diff
@@ -148,7 +148,10 @@ func NewV2API(cfg config.Configuration, resources v2.Resources, routerInst *rout
 		routerInst.GET("/api/v2/graphs/shortest-path", resources.GetShortestPath).Queries(params.StartNode.String(), params.StartNode.RouteMatcher(), params.EndNode.String(), params.EndNode.RouteMatcher()).RequirePermissions(permissions.GraphDBRead),
 		// TODO discuss if this should be a post endpoint
 		routerInst.GET("/api/v2/graph-search", resources.GetSearchResult).RequirePermissions(permissions.GraphDBRead),
+
+		// Cypher Queries API
 		routerInst.POST("/api/v2/graphs/cypher", resources.CypherSearch).RequirePermissions(permissions.GraphDBRead),
+		routerInst.GET("/api/v2/saved-queries", resources.ListSavedQueries).RequirePermissions(permissions.AuthManageSelf),
```
elikmiller, line 154: Is the `AuthManageSelf` permission required for this endpoint?

cmd/api/src/model/saved_queries.go
```diff
@@ -0,0 +1,67 @@
...
+import "fmt"
+
+type SavedQuery struct {
+	UserID string `json:"user_id" gorm:"primaryKey"`
+	Name   string `json:"name" gorm:"primaryKey"`
+	Query  string `json:"query"`
+
+	ID int64 `json:"id"`
+	Basic
+}
```
elikmiller, line 12: Looks like with this you can have rows with duplicate `id` fields which is probably not what we want

Screenshot from 2023-09-22 10-04-22

cmd/api/src/api/registration/v2.go
superlinkx: I'd argue that AuthManageSelf makes more sense for ListPermissions and ListRoles because those are both related to auth management. I do think we need a permission here, but maybe it would be worth having a short discussion on what that permission should be (either an existing permission or a new one)

packages/go/slices/slices.go
superlinkx: I'm not really sure what this TODO means in this case. Afaik, we don't have a `utils.Contains` function, if we ever did it was moved to this package to move us away from a collect-all utils package. Furthermore, if we were to move away from this implementation, we would almost certainly want to switch to the stdlib `slices` package which has its own implementation of Contains and most of our other utilities in this current package. We will have to wait until we adopt Go 1.21 though

superlinkx:
Other than the concern that the permission used doesn't quite fit the use case and the TODO being ambiguous, this all tests well. Validated the schema and usage against the ticket and everything checks out otherwise.

packages/go/slices/slices.go
superlinkx: Regardless, the utils package is considered "deprecated", so if anything should move it's things in `utils` moving to more distinct packages. In this case though, we're planning on migrating our usage of our `slices` package to the newer stdlib `slices` package to reduce maintenance, so we're going to be moving away from both this package and utils wherever possible.

superlinkx, line 85: I think it would be best to just remove this `TODO` for now and make a case for the change you propose outside of this PR.

superlinkx (approved):

PR #122, opened 2023-09-28
saved queries DELETE

irshadaj:
## Description

Adding the delete endpoint to saved queries

## How Has This Been Tested?
Added unit tests for handler

Manual testing by creating 2 saved queries for the current user, and a third one for another user. 

db

First I checked that the delete succeeds as expected for the right `ID` specified. 

sucess

db after delete

Then I checked that upon trying to delete a query that belongs to another user, or sending in an invalid or malformed ID, we get an HTTP 400.

other user's query

id malformed

Also checked that the documentation is updated:

doc

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/v2/saved_queries.go
+++ b/cmd/api/src/api/v2/saved_queries.go
@@ -6,8 +6,8 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/specterops/bloodhound/src/api"
 	ctx2 "github.com/specterops/bloodhound/src/ctx"
+	"github.com/specterops/bloodhound/src/database"
 	"github.com/specterops/bloodhound/src/model"
-	"gorm.io/gorm"
 	"gorm.io/gorm/utils"
 	"net/http"
 	"strconv"
@@ -113,16 +113,19 @@ func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.
 
 	if savedQueryID, err := strconv.Atoi(rawSavedQueryID); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
-	} else if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(userID, savedQueryID); err != nil {
+	} else if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(userID, savedQueryID); errors.Is(err, database.ErrNotFound) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "query does not exist", request), response)
+	} else if err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
 	} else if !savedQueryBelongsToUser {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "invalid saved_query_id supplied", request), response)
-	} else if err := s.DB.DeleteSavedQuery(savedQueryID); err != nil {
-		if errors.Is(err, gorm.ErrRecordNotFound) {
-			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "query does not exist", request), response)
-		} else {
-			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
-		}
+	} else if err := s.DB.DeleteSavedQuery(savedQueryID); errors.Is(err, database.ErrNotFound) {
+		// This is an edge case and can only occur if the database has a concurrent operation that deletes the saved query
+		// after the check at s.DB.SavedQueryBelongsToUser but before getting here.
+		// Still, adding in the same check for good measure.
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "query does not exist", request), response)
+	} else if err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
 	} else {
 		response.WriteHeader(http.StatusNoContent)
 	}
--- a/cmd/api/src/api/v2/saved_queries_test.go
+++ b/cmd/api/src/api/v2/saved_queries_test.go
@@ -10,12 +10,12 @@ import (
 	v2 "github.com/specterops/bloodhound/src/api/v2"
 	"github.com/specterops/bloodhound/src/auth"
 	"github.com/specterops/bloodhound/src/ctx"
+	"github.com/specterops/bloodhound/src/database"
 	"github.com/specterops/bloodhound/src/database/mocks"
 	"github.com/specterops/bloodhound/src/model"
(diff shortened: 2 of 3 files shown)


cmd/api/src/api/v2/saved_queries_test.go
```diff
@@ -479,3 +479,178 @@ func TestResources_CreateSavedQuery(t *testing.T) {
 	router.ServeHTTP(response, req)
 	require.Equal(t, http.StatusCreated, response.Code)
 }
+
+func TestResources_DeleteSavedQuery_IDMalformed(t *testing.T) {
```
urangel, line 483: Is it possible to refactor theses tests using the `apitest` package?

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -100,3 +102,32 @@ func (s Resources) CreateSavedQuery(response http.ResponseWriter, request *http.
...
+		for _, savedQuery := range savedQueries {
+			if savedQuery.ID == int64(savedQueryID) {
+				savedQueryIDMatched = true
+				break
+			}
+		}
+
+		if !savedQueryIDMatched {
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "invalid saved_query_id supplied", request), response)
+		} else {
```
superlinkx, line 128: I feel like we could just use SQL to manage this complexity rather than pulling "all" (technically only the first 10000, which is likely to be all in most real world scenarios, but importantly isn't guaranteed to be all) queries for the current user and manually sifting. This would mean adding another DB method for this purpose, but it'll be a lot cleaner. Something like:

```go
func (s *BloodhoundDB) UserOwnsSavedQuery(userID uuid.UUID, id int) (bool, error) {
	var count int64
	if result := s.db.Where("user_id = ?", userID).Where("id = ?", id).Find(model.SavedQuery{}).Count(&count); result.Error != nil {
		return false, result.Error
	} else {
		return count == 1, nil
	}
}
```

You can then do error checking and a bool check, then proceed with the delete without any additional application level handling.

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -100,3 +102,32 @@ func (s Resources) CreateSavedQuery(response http.ResponseWriter, request *http.
...
+			s.DB.DeleteSavedQuery(savedQueryID)
+			response.WriteHeader(http.StatusNoContent)
```
superlinkx, line 130: This is ineffective, as you can see that a successful delete gives a 200 response currently, likely due to a middleware seeing that we haven't sent a response and somehow overriding our status header

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -100,3 +102,32 @@ func (s Resources) CreateSavedQuery(response http.ResponseWriter, request *http.
...
+			if savedQuery.ID == int64(savedQueryID) {
+				savedQueryIDMatched = true
+				break
+			}
+		}
+
+		if !savedQueryIDMatched {
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "invalid saved_query_id supplied", request), response)
+		} else {
+			s.DB.DeleteSavedQuery(savedQueryID)
```
superlinkx, line 129: We need to emit an error if the transaction fails, and handle it here accordingly

cmd/api/src/api/v2/saved_queries.go
superlinkx: The accepted solution is going to be similar to this, but much better than my quick scratch fix. One thing we want to handle with the equivalent of `UserOwnsSavedQuery()` is to check the error here at the API level and have a special case for `errors.Is(err, database.ErrNotFound)`, returning a 404 to the user to let the know the record has already been deleted (or doesn't exist)

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -100,3 +104,26 @@ func (s Resources) CreateSavedQuery(response http.ResponseWriter, request *http.
...
+
+func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.Request) {
+	var (
+		rawSavedQueryID = mux.Vars(request)[api.URIPathVariableSavedQueryID]
+		userID          = ctx2.FromRequest(request).AuthCtx.Session.UserID
+	)
+
+	if savedQueryID, err := strconv.Atoi(rawSavedQueryID); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
+	} else if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(userID, savedQueryID); err != nil {
```
superlinkx, line 116: In most cases, we won't hit the `database.ErrNotFound` in the delete call, we'll hit it here first since if the id was deleted, this query will fail. We need to check for `database.ErrNotFound` and handle it in this section of code. We could optionally handle it below as well just for thoroughness, but it would be an edge case. You can chain multiple  error checks without nesting as well:

```suggestion
	} else if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(userID, savedQueryID); errors.Is(err, database.ErrNotFound) {
	// 404
	} else if err != nil {
	// 500
	} else if !savedQueryBelongsToUser {
	//400
	} else if err := s.DB.DeleteSavedQuery; errors.Is(err, database.ErrNotFound) {
	// 404
	} else if err != nil {
	// 500
	} else {
	// 204
	}
```

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -100,3 +104,26 @@ func (s Resources) CreateSavedQuery(response http.ResponseWriter, request *http.
...
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
+	} else if !savedQueryBelongsToUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "invalid saved_query_id supplied", request), response)
+	} else if err := s.DB.DeleteSavedQuery(savedQueryID); err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
```
superlinkx, line 121: We do not want gorm leaking into the API layer, so this should be doing a check against `database.ErrNotFound`. It's still not ideal to have database types in our API layer either, but it's less of a problem than leaking `gorm` and is why `CheckError()` is used in the first place

cmd/api/src/database/saved_queries.go
```diff
@@ -39,3 +39,18 @@ func (s *BloodhoundDB) CreateSavedQuery(userID uuid.UUID, name string, query str
...
 }
+
+func (s *BloodhoundDB) DeleteSavedQuery(id int) error {
+	return CheckError(s.db.Delete(&model.SavedQuery{}, id))
+}
+
+func (s *BloodhoundDB) SavedQueryBelongsToUser(userID uuid.UUID, savedQueryID int) (bool, error) {
+	var savedQuery model.SavedQuery
+	if result := s.db.First(&savedQuery, savedQueryID); result.Error != nil {
+		return false, result.Error
```
superlinkx, line 50:
```suggestion
		return false, CheckError(result)
```
This will ensure we properly wrap the error if it's a `gorm.ErrRecordNotFound` to prevent needing to leak `gorm` outside of the database layer.

superlinkx (changes requested):
A few error handling issues we need to address, but otherwise I think this looks solid.

superlinkx (approved):
All scenarios check out locally. Approved!

PR #125, opened 2023-09-29
fix: add migration to delete data between specified times

urangel:
## Description
Includes a migration targeted for v5.0.11

## Motivation and Context
This cleans up some inaccurate data.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/database/migration/migrations/v5.0.11.sql
+++ b/cmd/api/src/database/migration/migrations/v5.0.11.sql
@@ -1,0 +1,26 @@
+-- Copyright 2023 Specter Ops, Inc.
+--
+-- Licensed under the Apache License, Version 2.0
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


cmd/api/src/database/migration/migrations/v5.0.11.sql
```diff
@@ -0,0 +1,26 @@
...
+DELETE
+FROM risk_posture_stats
+WHERE created_at > TIMESTAMP WITH TIME ZONE '2023-08-30 19:00:00+00'
+  AND created_at < TIMESTAMP WITH TIME ZONE '2023-08-31 21:30:00+00';
+
+
+DELETE
+FROM risk_counts
+WHERE created_at > TIMESTAMP WITH TIME ZONE '2023-08-30 19:00:00+00'
+  AND created_at < TIMESTAMP WITH TIME ZONE '2023-08-31 21:30:00+00';
```
superlinkx, line 26: Neither of these exist in BHCE, which makes these migrations likely to break when upgrading a live version of BHCE. We will only want to run the DELETE statements _if_ the tables exist. I'm not entirely sure the best way to do this, so we can discuss outside of the PR to find the best solution here.

elikmiller:
I think this PR can be closed since the affected tables are not in BHCE

PR #137, opened 2023-10-10
Better hygiene on asset group members endpoint

brandonshearin:
## Description

The /api/v2/asset-groups/:assetGroupId/members endpoint was returning an empty response in some environments due to a panic caused by type conversions of the form 

`somevar := someval.(string)` .  

The offending type conversions in the `GetAssetGroupNodes` function are covered by this PR

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/v2/agi.go
+++ b/cmd/api/src/api/v2/agi.go
@@ -441,7 +441,7 @@ func parseAGMembersFromNodes(nodes graph.NodeSet, selectors model.AssetGroupSele
 
 		if node.Kinds.ContainsOneOf(azure.Entity) {
 			if tenantID, err := node.Properties.Get(azure.TenantID.String()).String(); err != nil {
-				log.Warnf("%s is missing for node %d, skipping AG Membership...", azure.Tenant.String(), node.ID)
+				log.Warnf("%s is missing for node %d, skipping AG Membership...", azure.TenantID.String(), node.ID)
 				continue
 			} else {
 				agMember.EnvironmentKind = azure.Tenant.String()
--- a/cmd/api/src/api/v2/agi.go
+++ b/cmd/api/src/api/v2/agi.go
@@ -1,17 +1,17 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0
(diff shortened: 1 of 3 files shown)


cmd/api/src/api/v2/agi.go
```diff
@@ -403,27 +404,52 @@ func parseAGMembersFromNodes(nodes graph.NodeSet, selectors model.AssetGroupSele
...
 				isCustomMember = true
 			}
 		}
 
+		var (
+			memberObjectId string
+			memberName     string
+		)
+
+		if objectId, ok := node.Properties.Map[common.ObjectID.String()].(string); ok {
```
superlinkx, line 419: It is preferred to consistently handle the negative case first and the happy path second

cmd/api/src/api/v2/agi.go
```diff
@@ -403,27 +404,52 @@ func parseAGMembersFromNodes(nodes graph.NodeSet, selectors model.AssetGroupSele
...
+			memberObjectId = objectId
+		} else {
+			log.Warnf("objectid is missing for node %d", node.ID)
+			memberObjectId = ""
+		}
+
+		if name, ok := node.Properties.Map[common.Name.String()].(string); !ok {
```
superlinkx, line 426: The conditions appear to be flipped here. You're checking if there was an error and then setting the value that isn't ok, but warning and handling the error case in the case it is ok.

cmd/api/src/api/v2/agi.go
```diff
@@ -403,27 +404,52 @@ func parseAGMembersFromNodes(nodes graph.NodeSet, selectors model.AssetGroupSele
...
 				isCustomMember = true
 			}
 		}
 
+		var (
+			memberObjectId string
+			memberName     string
+		)
+
+		if objectId, ok := node.Properties.Map[common.ObjectID.String()].(string); !ok {
```
superlinkx, line 419: So it turns out, we do have a better solution to this and that's why we couldn't find other portions of the code that were doing these unsafe conversions. You can use `node.Properties.Get(common.ObjectID.String()).String()` which will return `(string, error)` and assert on the error and otherwise use the string. This should make all of these a bit cleaner and will mean we don't need to do type casting at the web API level.

PR #139, opened 2023-10-11
fix: ensure azexecutecommand is only set for intune admins

ddlees:
## Description

- Removes processing devices owners for AZExecuteCommand edge
- Adds tests

## Motivation and Context

AZExecuteCommand should be processed for those with the Intune Admin role over a device, however, the current implementation includes device ownership for processing. As device ownership already has its own edge (AZOwns), this edge is duplicative and unnecessary.

## How Has This Been Tested?

New tests for foundational units that aid in the processing of AZExecuteCommand. More unit tests to follow in separate PR.

## Types of changes

## Checklist:


packages/go/analysis/azure/post.go
```diff
@@ -272,6 +273,7 @@ func roleMembers(tx graph.Transaction, tenantRoles graph.NodeSet, additionalRela
 		} else {
 			// TODO: This could be more optimal by iterating in place instead of aggregating all results
 			members.AddSet(paths.AllNodes())
+			members.Remove(tenantRole.ID)
```
computator, line 276: I'm not seeing anything obvious, but I just wanted to make sure nothing calling `RoleMembers()` or `RoleMembersWithGrants()` depends on the Role ID being included in the returned list of members.

packages/go/dawgs/graph/mocks/graph.go
```diff
@@ -15,20 +15,272 @@
 // SPDX-License-Identifier: Apache-2.0
 
 // Code generated by MockGen. DO NOT EDIT.
-// Source: github.com/specterops/bloodhound/dawgs/graph (interfaces: Batch,Transaction,Database)
+// Source: graph.go
```
computator, line 18: This new comment seems a lot less helpful. Just wanted to make sure MockGen isn't misconfigured in your environment or something

Edit:
This seems to be on purpose actually so you can see it's from `graph.go`. Is there a way to include the other info or a path too?

computator (approved):
This all seems reasonable afaict. I had two questions about things, but if they aren't concerns then it seems ready to merge to me!

packages/go/dawgs/graph/mocks/graph.go
ddlees: That comment is autogenerated by mockgen. After looking through the docs, I don't see a way to include additional information without reverting to reflection mode and forcing maintainers to list every declared interface.

As a note, there are definitely trade-offs between generating mocks via source instead of using reflection. We might want to consider pulling this issue into its own discussion and make a PR afterward.

packages/go/analysis/azure/post.go
```diff
@@ -273,13 +279,14 @@ func roleMembers(tx graph.Transaction, tenantRoles graph.NodeSet, additionalRela
...
 			// TODO: This could be more optimal by iterating in place instead of aggregating all results
 			members.AddSet(paths.AllNodes())
-			members.Remove(tenantRole.ID)
 		}
 	}
 
 	return members, nil
 }
 
+// RoleMembersWithGrants returns the NodeSet of members for a given set of roles, including those members who may be able to grant themselves one of the give roles
```
computator, line 288: > ...grant themselves one of the *given* roles

packages/go/analysis/azure/post.go
computator: The updated logic looks good!

PR #164, opened 2023-10-24
added owned node capability in API

irshadaj:
## Description

The “Owned” functionality shall manifest in the API as a new Asset Group. All of its functionality and endpoints remain the same as with the T0 asset group, the only exception being that when items are added or removed from the Owned group, analysis should not be triggered.

This deliverables for this story are:

A database migration to add the Owned asset group to all existing installations, and its inclusion in any new builds

Addition of a conditional so as to bypass the analysis trigger [here](https://github.com/SpecterOps/BloodHound/blob/main/cmd/api/src/api/v2/agi.go#L262) when the asset group ID corresponds to the Owned asset group

## How Has This Been Tested?
For the change in analysis trigger logic, an additional unit test has been added.

Testing on an existing local environment to make sure the migration is applied to it (note the different timestamps indicating that the Owned asset group was added after the initial setup):

pg existing

Then I wiped the database and containers, and built everything from scratch to ensure that new installs will have the new asset group set up correctly (note the same timestamps here):

pg new install

Now updating the tier zero asset group (ID is 1 from the previous screenshots), I checked that analysis is triggered:

t0

Updating the owned asset group does NOT kick off analysis:

owned

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/database/migration/agi.go
+++ b/cmd/api/src/database/migration/agi.go
@@ -65,12 +65,12 @@ func (s *Migrator) updateAssetGroups() error {
 			}
 		}
 
-		if _, hasOwned := systemAssetGroups.FindByName(model.OwnedAsserGroupTag); !hasOwned {
+		if _, hasOwned := systemAssetGroups.FindByName(model.OwnedAssetGroupTag); !hasOwned {
 			log.Infof("Missing the default Owned asset group. Creating it now.")
 
 			ownedAG := model.AssetGroup{
 				Name:        model.OwnedAssetGroupName,
-				Tag:         model.OwnedAsserGroupTag,
+				Tag:         model.OwnedAssetGroupTag,
 				SystemGroup: true,
 			}
 
--- a/cmd/api/src/database/migration/migrations/v5.1.1.sql
+++ b/cmd/api/src/database/migration/migrations/v5.1.1.sql
@@ -1,3 +1,19 @@
+-- Copyright 2023 Specter Ops, Inc.
+--
+-- Licensed under the Apache License, Version 2.0
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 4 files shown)


cmd/api/src/model/agi.go
```diff
@@ -221,6 +221,10 @@ type AssetGroupSelectorSpec struct {
 const (
 	SelectorSpecActionAdd    = "add"
 	SelectorSpecActionRemove = "remove"
+	TierZeroAssetGroupName   = "Admin Tier Zero"
+	TierZeroAssetGroupTag    = "admin_tier_0"
+	OwnedAssetGroupName      = "Owned"
+	OwnedAsserGroupTag       = "owned"
```
superlinkx, line 227: This is misspelled, but consistently. Should be OwnedAssetGroupTag

cmd/api/src/database/migration/agi.go
```diff
@@ -70,6 +65,20 @@ func (s *Migrator) updateAssetGroups() error {
...
+
+			ownedAG := model.AssetGroup{
+				Name:        model.OwnedAssetGroupName,
+				Tag:         model.OwnedAsserGroupTag,
+				SystemGroup: true,
+			}
+
+			if result := tx.Create(&ownedAG); result.Error != nil {
+				return result.Error
+			}
```
superlinkx, line 79: Since we aren't doing anything with the error, we can actually shortcut this bit:

```go
return tx.Create(&ownedAG).Error
```

superlinkx (commented):
Preliminary code check looks good, just a couple of suggestions. I'm going to pull this tonight to validate the migration, but I don't expect to run into any issues.

superlinkx (approved):
I have validated the migrations work as documented locally. Once the suggestions have been reviewed, this is good to go in.

cmd/api/src/database/migration/migrations/v5.1.1.sql
```diff
@@ -0,0 +1,3 @@
+INSERT INTO asset_groups (name, tag, system_group)
```
elikmiller, line 1: Needs license

PR #166, opened 2023-10-25
feat: handle dc reg values not existing

JonasBK:
## Description

Handle when DC registry values do not exist.

It is possible that the registry values we want to collect do not exist. Sharphound will return -1 for integer values in this case.
We will store the value as is and make the user aware that the value does not exist: 
image

## How Has This Been Tested?

Uploaded positive int values, 0 values, and -1 values.

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/ein/ad.go
+++ b/packages/go/ein/ad.go
@@ -566,6 +566,8 @@ const (
 
 // Prettified definitions for DCRegistryData
 const (
+	RegValNotExisting                               = "Registry value does not exist"
+
 	PrettyCertMappingManyToOne                      = "0x01: Many-to-one (issuer certificate)"
 	PrettyCertMappingOneToOne                       = "0x02: One-to-one (subject/issuer)"
 	PrettyCertMappingUserPrincipalName              = "0x04: User principal name (UPN/SAN)"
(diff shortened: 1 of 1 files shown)


packages/go/ein/ad.go
```diff
@@ -581,40 +583,50 @@ func ParseDCRegistryData(computer Computer) IngestibleNode {
...
-		case 0:
-			propMap[ad.StrongCertificateBindingEnforcement.String()] = PrettyStrongCertBindingEnforcementDisabled
-		case 1:
-			propMap[ad.StrongCertificateBindingEnforcement.String()] = PrettyStrongCertBindingEnforcementCompatibility
-		case 2:
-			propMap[ad.StrongCertificateBindingEnforcement.String()] = PrettyStrongCertBindingEnforcementFull
+		if computer.DCRegistryData.StrongCertificateBindingEnforcement.Value == -1 {
+			propMap[ad.StrongCertificateBindingEnforcement.String()] = RegValNotExisting
+
+		} else if computer.DCRegistryData.StrongCertificateBindingEnforcement.Value >= 0 {
```
juggernot325, line 621: I think we can remove the check on L621 and add the `-1` case to the switch statement. That should help simplify this a bit.

packages/go/ein/ad.go
```diff
@@ -581,40 +583,50 @@ func ParseDCRegistryData(computer Computer) IngestibleNode {
 	var ()
 	propMap := make(map[string]any)
 
-	if computer.DCRegistryData.CertificateMappingMethods.Collected && computer.DCRegistryData.CertificateMappingMethods.Value >= 0 {
+	if computer.DCRegistryData.CertificateMappingMethods.Collected {
 		propMap[ad.CertificateMappingMethodsRaw.String()] = computer.DCRegistryData.CertificateMappingMethods.Value
 
-		var prettyMappings []string
+		if computer.DCRegistryData.CertificateMappingMethods.Value == -1 {
+			propMap[ad.CertificateMappingMethods.String()] = RegValNotExisting
```
juggernot325, line 590: I'm not sure if this really matters, but in this case the value for the `CertificateMappingMethods` property will be a `string`, but in the case on L611 it will be an array. This might be annoying for API consumers?

Maybe we could remove the `>= 0` check on L592 and add a new `if` statement for this case in that block?

juggernot325 (changes requested):
I found a couple of small places for simplification of the code.

juggernot325 (approved):
Thanks for making those changes!

PR #179, opened 2023-10-30
fix: stepwise was making new migrations on every restart

superlinkx:
## Description

This will prevent the stepwise migrator from adding additional entries for the same version

## Motivation and Context

Due to how the stepwise migrator runs, we were stamping a new version in the migrations database on every app restart, even if that version already existed. This change will ensure it only stamps the new version if the last migration and the current version aren't the same.

## How Has This Been Tested?

Validated locally with test harness

## Types of changes

## Checklist:


zinic (approved):
lgtm

juggernot325 (approved):

PR #184, opened 2023-11-01
increase dawgs memory limit to 2gb

irshadaj:
## Description

Increase dawgs query limit to 2GB since enterprise customers are running into the limits too often.

## Types of changes

## Checklist:

### DIFF
--- a/local-harnesses/build.config.json.template
+++ b/local-harnesses/build.config.json.template
@@ -22,6 +22,5 @@
         "first_name": "BloodHound",
         "last_name": "Dev",
         "expire_now": false
-    },
-    "traversal_memory_limit": 1
+    }
 }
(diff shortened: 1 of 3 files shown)


urangel (changes requested):
The configuration file templates, e.g., build.config.json.template, should probably be updated as well in order to keep things aligned.

urangel (approved):

local-harnesses/build.config.json.template
```diff
@@ -23,5 +23,5 @@
         "last_name": "Dev",
         "expire_now": false
     },
-    "traversal_memory_limit": 1
+    "traversal_memory_limit": 2
```
superlinkx, line 26: I would like us to remove this piece from the template. In dev, we should probably strive to use application defaults wherever possible. If this entry exists in `integration.config.json.template`, yank it out of there too

superlinkx (approved):

PR #186, opened 2023-11-01
fix: add missing edges to filter

JonasBK:
## Description

Add four ADCS edges to path finding filter selection.

## How Has This Been Tested?

Edges are filtered out if they are deselected as expected.

image

## Types of changes

-   [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:


cmd/ui/src/views/Explore/ExploreSearch/edgeTypes.tsx
```diff
@@ -89,13 +89,17 @@ export const AllEdgeTypes: Category[] = [
             {
                 name: 'Active Directory Certificate Services',
                 edgeTypes: [
-                    ActiveDirectoryRelationshipKind.RootCAFor,
-                    ActiveDirectoryRelationshipKind.PublishedTo,
-                    ActiveDirectoryRelationshipKind.ManageCertificates,
-                    ActiveDirectoryRelationshipKind.ManageCA,
                     ActiveDirectoryRelationshipKind.DelegatedEnrollmentAgent,
                     ActiveDirectoryRelationshipKind.Enroll,
```
urangel, line 93: I believe we only list transitive edges in this filter so the enroll edge, for instance, would not be something that should be populated here. You'll notice getChanges and getChangesAll are absent from any of the AD filters.

juggernot325:
The failing tests here should be fixed by #187. I just merged that into `adcs-poc`, so merging that into your branch should resolve this.

cmd/ui/src/views/Explore/ExploreSearch/edgeTypes.tsx
JonasBK: Oh yeah, you are right.. Good catch. It is actually only GoldenCert, and the ESCX edges that are transitive. So maybe we should remove all the others. I will ask Andy if he agrees

urangel: Yea that sounds correct to only include GoldenCert and the ESCX edges. Thanks for confirming with Andy

rvazarkar (approved):

PR #196, opened 2023-11-07
draft: context menu with node actions

brandonshearin:
## Types of changes

## Checklist:


cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -76,11 +79,15 @@ export const GraphEvents: FC<GraphEventProps> = ({
...
-                }, 150);
+                // a button value of `0` indicates that the main button of the mouse device was pressed
+                // https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value
+                if (event.event.original.button === 0) {
+                    setDraggedNode(event.node);
+
+                    dragTimerRef.current = setTimeout(() => {
+                        isLongPress.current = true;
+                    }, 150);
+                }
```
elikmiller, line 90: It looks like react-sigma offers both `clickNode` and `rightClickNode` events - would those provide a cleaner solution?

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -127,6 +134,10 @@ export const GraphEvents: FC<GraphEventProps> = ({
                     }, 200);
                 }
             },
+            rightClickNode: (event) => {
+                event.preventSigmaDefault();
+                onRightClickNode?.(event);
+            },
```
elikmiller, line 140: Oh, I see you use rightClickNode here

cmd/ui/src/ducks/searchbar/reducer.ts
```diff
@@ -52,86 +51,90 @@ const initialSearchState: types.SearchState = {
...
-function isTargetedActionType(action: types.SearchbarActionTypes): action is types.SearchbarTargetedActionTypes {
-    return (action as types.SearchbarTargetedActionTypes).target !== undefined;
-}
-
 const searchReducer = (state = initialSearchState, action: types.SearchbarActionTypes) => {
-    if (action.type === types.SEARCH_RESET) {
-        return cloneDeep(initialSearchState);
-    }
+    switch (action.type) {
+        case types.SEARCH_RESET: {
```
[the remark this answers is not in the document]
brandonshearin: the `produce()` function is from the [immer library](https://immerjs.github.io/immer/).  i actually didn't know what it was before this feature, but its pretty cool.  it provides a developer-friendly way to make changes to immutable objects, like a redux store.

my thinking at the time when i split the reducer logic into two switch blocks was that one set of the case blocks was using immer, the other wasn't.

cmd/ui/src/views/Explore/ExploreSearch/NodeSearch.tsx
```diff
@@ -14,55 +14,29 @@
...
-interface NodeSearchProps {
-    labelText: string;
-    searchType: typeof PRIMARY_SEARCH | typeof SECONDARY_SEARCH;
-}
+const NodeSearch = () => {
+    const dispatch = useAppDispatch();
 
-const NodeSearch = ({ searchType, labelText }: NodeSearchProps) => {
-    const dispatch = useDispatch();
+    const { primary } = useSelector((state: AppState) => state.search);
```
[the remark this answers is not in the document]
brandonshearin: hey! i didn't know that. i updated the components that useSelector() following this suggestion. thank youu

cmd/ui/src/components/GraphEvents.tsx
brandonshearin: is this comment resolved?

we need to have the logic that is currently defined in downNode() in a separate, discrete step from the logic defined in clickNode().  this is because sigma fires downNode() before it fires clickNode(), so our dragging-related logic goes into downNode().  it is mandatory to include the `event.original.button === 0` check in the downNode() handler because we only want to drag nodes on left clicks, not right clicks.

benwaples (approved):

PR #202, opened 2023-11-09
feat: update azresetpassword processing to start from role nodes inst…

ddlees:
## Description

- Update AZResetPassword post-processing to start at AZRole nodes and end at AZUser nodes
- Misc. performance refactors and others
- Updated post-processing tests

## Motivation and Context

The existing approach in which AZResetPassword edges are processed creates an unwieldy number of edges in the graph which impacts analysis time significantly and creates a poor user experience in the case where misconfiguration leads to every user having an AZResetPassword edge to every other user.

## How Has This Been Tested?

Updated the existing post-processing tests to account for the new requirements.

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/analysis/azure/post.go
+++ b/packages/go/analysis/azure/post.go
@@ -1033,33 +1033,9 @@ func resetPassword(ctx context.Context, db graph.Database, operation analysis.St
 			return err
 		} else {
 			for _, role := range pwResetRoles {
-				if roleTemplateIDProp := role.Properties.Get(azure.RoleTemplateID.String()); roleTemplateIDProp.IsNil() {
-					log.Errorf("unable to process azresetpassword for role node %d - missing property %s", role.ID, azure.RoleTemplateID)
-				} else if roleTemplateID, err := roleTemplateIDProp.String(); err != nil {
-					log.Errorf("unable to process azresetpassword for role node %d - property %s is not a string", role.ID, azure.RoleTemplateID)
+				if targets, err := resetPasswordEndNodeBitmapForRole(role, roleAssignments); err != nil {
+					return fmt.Errorf("unable to continue processing azresetpassword for tenant node %d", tenant.ID, err)
 				} else {
-					targets := roaring.New()
-					switch roleTemplateID {
-					case constants.GlobalAdministratorRoleID, constants.PrivilegedAuthenticationAdministratorRoleID, constants.PartnerTier2SupportRoleID:
-						targets.Or(roleAssignments.Users())
-					case constants.UserAdministratorRoleID:
-						targets.Or(roleAssignments.UsersWithoutRoles())
-						targets.Or(roleAssignments.UsersWithRolesExclusive(UserAdministratorPasswordResetTargetRoles()...))
-					case constants.HelpdeskAdministratorRoleID:
-						targets.Or(roleAssignments.UsersWithoutRoles())
-						targets.Or(roleAssignments.UsersWithRolesExclusive(HelpdeskAdministratorPasswordResetTargetRoles()...))
-					case constants.AuthenticationAdministratorRoleID:
-						targets.Or(roleAssignments.UsersWithoutRoles())
-						targets.Or(roleAssignments.UsersWithRolesExclusive(AuthenticationAdministratorPasswordResetTargetRoles()...))
-					case constants.PasswordAdministratorRoleID:
-						targets.Or(roleAssignments.UsersWithoutRoles())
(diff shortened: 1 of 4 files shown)


packages/go/analysis/azure/tenant.go
```diff
@@ -62,3 +65,22 @@ func PopulateTenantEntityDetailsCounts(tx graph.Transaction, node *graph.Node, d
 
 	return details, nil
 }
+
+// TenantPrincipals returns the complete set of User, Group, and Service Principal nodes contained by the given Tenant node
+func TenantPrincipals(tx graph.Transaction, tenant *graph.Node) (graph.NodeSet, error) {
+	if !tenant.Kinds.ContainsOneOf(azure.Tenant) {
```
superlinkx, line 71: This could use `!IsTenantNode(azure.Tenant)` for consistency

packages/go/analysis/azure/post_test.go
```diff
@@ -42,14 +43,21 @@ var (
...
+	roleMap := map[string]*roaring.Bitmap{
+		constants.GlobalAdministratorRoleID:   roaring.New(),
+		constants.ReportsReaderRoleID:         roaring.New(),
+		constants.HelpdeskAdministratorRoleID: roaring.New(),
+		constants.PartnerTier1SupportRoleID:   roaring.New(),
+	}
+	roleMap[constants.GlobalAdministratorRoleID].Add(user.ID.Uint32())
+	roleMap[constants.ReportsReaderRoleID].Add(group.ID.Uint32())
+	roleMap[constants.HelpdeskAdministratorRoleID].Add(group.ID.Uint32())
+	roleMap[constants.PartnerTier1SupportRoleID].Add(app.ID.Uint32())
```
superlinkx, line 55: In this same file, we're using `uint32(user.ID)` and this method. Honestly, not a huge fan that we made a method that isn't any shorter than just doing an explicit type conversion, it feels like it's an unnecessary indirection, but we should probably at least be consistent in how we approach it per file

packages/go/analysis/azure/post.go
```diff
@@ -993,26 +1027,63 @@ func ExecuteCommand(ctx context.Context, db graph.Database) (*analysis.AtomicPos
...
-func resetPassword(ctx context.Context, db graph.Database, operation analysis.StatTrackedOperation[analysis.CreatePostRelationshipJob], roleAssignments RoleAssignments) error {
-	usersWithoutRoles := roleAssignments.UsersWithoutRoles()
+func resetPassword(ctx context.Context, db graph.Database, operation analysis.StatTrackedOperation[analysis.CreatePostRelationshipJob], tenant *graph.Node, roleAssignments RoleAssignments) error {
+	return operation.Operation.SubmitReader(func(ctx context.Context, tx graph.Transaction, outC chan<- analysis.CreatePostRelationshipJob) error {
+		if pwResetRoles, err := TenantRoles(tx, tenant, ResetPasswordRoleIDs()...); err != nil {
+			return err
+		} else {
+			for _, role := range pwResetRoles {
+				if roleTemplateIDProp := role.Properties.Get(azure.RoleTemplateID.String()); roleTemplateIDProp.IsNil() {
+					log.Errorf("unable to process azresetpassword for role node %d - missing property %s", role.ID, azure.RoleTemplateID)
```
superlinkx, line 1037: I feel like if we're going to use our pipeline pattern, we should really early return in each error branch (`return nil`), to prevent surprises if something happens after the pipeline in a refactor.

packages/go/analysis/azure/post.go
```diff
@@ -993,26 +1027,63 @@ func ExecuteCommand(ctx context.Context, db graph.Database) (*analysis.AtomicPos
...
-	} else {
-		for _, userID := range roleAssignments.Nodes.AllNodeIDs() {
-			if err := resetPasswordCases(roleAssignments, operation, userID, usersWithoutRoles, securityGroupOwners); err != nil {
-				log.Errorf("Unable to process AZResetPassword for node %d: %v", userID, err)
+						if !channels.Submit(ctx, outC, nextJob) {
+							return nil
+						}
+
+					}
+				}
```
superlinkx, line 1076: This entire pipeline should probably be broken out into a separate function to help reduce nesting and make the logic a little easier to track

superlinkx (changes requested):
Overall logic seems good as far as I can see. A few consistency notes and a suggestion for helping keep the complex logic a little more readable, but otherwise code looks good. Attempting to poke it locally while these comments are addressed.

superlinkx:
Completed some test runs with and without this change and noted that data was consistent between runs, but the time it takes to process azresetpassword is significantly reduced in our test dataset.

superlinkx (approved):

PR #210, opened 2023-11-14
allow asset group selectors with the same name to exist under different asset groups

irshadaj:
## Description

Currently, the name field of an asset group selector needs to be unique regardless of the asset group. This prevents custom “High Value” nodes from being added to the “Owned” group and vice versa in BHCE, and would cause further problems if we want to add additional groups in the future. Trying to add a duplicate selector causes an error.

This story drops that constraint and creates a new uniqueness constraint on both the name and asset_group_id columns, thereby allowing a selector to exist under both Owned as well as HVT asset groups.

## How Has This Been Tested?
I spun up a database on the main branch to simulate an existing DB. The selectors table has a constraint on the name column here:

existing constraints

Then I restarted the environment to have the migration applied:

existing migration execution

Verified that the constraint has been updated (the new constraint name contains both column names in it):

existing constraints after

Then I created 2 selectors with the same name but belonging to the 2 different asset groups via postman:

existing owned

existing t0

And verified through the database that both the selectors were created:

existing db

Next, I wiped out my database and re-created it to verify that any fresh installs will have the right constraint set up, and then repeated the selecor creation via postman to verify that multiple selectors can exist with the same name:

fresh owned

fresh t0

fresh db

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/database/migration/migrations/schema.sql
+++ b/cmd/api/src/database/migration/migrations/schema.sql
@@ -715,7 +715,7 @@ ALTER TABLE ONLY asset_group_collection_entries
 ALTER TABLE ONLY asset_group_collections
     ADD CONSTRAINT asset_group_collections_pkey PRIMARY KEY (id);
 ALTER TABLE ONLY asset_group_selectors
-    ADD CONSTRAINT asset_group_selectors_name_assetgroupid_key UNIQUE (name, asset_group_id);
+    ADD CONSTRAINT asset_group_selectors_name_key UNIQUE (name);
 ALTER TABLE ONLY asset_group_selectors
     ADD CONSTRAINT asset_group_selectors_pkey PRIMARY KEY (id);
 ALTER TABLE ONLY asset_groups
--- a/cmd/api/src/database/migration/migrations/v5.2.1.sql
+++ b/cmd/api/src/database/migration/migrations/v5.2.1.sql
@@ -14,8 +14,8 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
-ALTER TABLE ONLY asset_group_selectors
-  DROP CONSTRAINT asset_group_selectors_name_key;
+ALTER TABLE IF EXISTS ONLY asset_group_selectors
+  DROP CONSTRAINT IF EXISTS asset_group_selectors_name_key;
 
-ALTER TABLE ONLY asset_group_selectors
+ALTER TABLE IF EXISTS ONLY asset_group_selectors
   ADD CONSTRAINT asset_group_selectors_name_assetgroupid_key UNIQUE (name, asset_group_id);


cmd/api/src/database/migration/migrations/v5.2.1.sql
```diff
@@ -0,0 +1,21 @@
...
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+ALTER TABLE ONLY asset_group_selectors
```
zinic, line 17: Consider using `IF EXISTS` or otherwise replicating its functionality to avoid lack of the constraint from failing the migration.

zinic (dismissed):
Left a comment on the `DROP CONSTRAINT` SQL statement. Changes otherwise look good.

cmd/api/src/database/migration/migrations/schema.sql
```diff
@@ -715,7 +715,7 @@ ALTER TABLE ONLY asset_group_collection_entries
 ALTER TABLE ONLY asset_group_collections
     ADD CONSTRAINT asset_group_collections_pkey PRIMARY KEY (id);
 ALTER TABLE ONLY asset_group_selectors
-    ADD CONSTRAINT asset_group_selectors_name_key UNIQUE (name);
+    ADD CONSTRAINT asset_group_selectors_name_assetgroupid_key UNIQUE (name, asset_group_id);
```
elikmiller, line 718: According to @zinic we should not be modifying this file

cmd/api/src/database/migration/migrations/v5.2.1.sql
```diff
@@ -0,0 +1,21 @@
...
+  DROP CONSTRAINT asset_group_selectors_name_key;
+
+ALTER TABLE ONLY asset_group_selectors
+  ADD CONSTRAINT asset_group_selectors_name_assetgroupid_key UNIQUE (name, asset_group_id);
```
superlinkx, line 21: To be clear, both `ALTER TABLE` statements should be `ALTER TABLE IF EXISTS ONLY ...`, and the `DROP CONSTRAINT` should be `DROP CONSTRAINT IF EXISTS ...`. These are all painless changes that make the migration more idempotent.

Unfortunately, for adding the new constraint, we end up with something that's a bit more tricky. We can't simply run `ADD CONSTRAINT IF NOT EXISTS` and the alternative is to either try dropping it first (which could have performance implications in the future) or doing a pretty gross transaction that checks for existence and then runs the add if it doesn't exist. This one I'm less sure about how we should tackle it and whether it's worth doing so. Having all of our migrations being idempotent would be really good in theory, but it's also limited in practicality so how far we're willing to push the concept should probably be a point of further discussion.

Source for what it would take to make our constraint addition idempotent: https://stackoverflow.com/questions/6801919/postgres-add-constraint-if-it-doesnt-already-exist

zinic: This migration would only run once. The assumption is that prior to this version `asset_group_selectors_name_assetgroupid_key`. That is to say this isn't a schema assertion that must be 100% idempotent. I think it's safe-enough to just have the constraint drop be idempotent.

superlinkx (approved):

superlinkx:
Looks good, feel free to pull

zinic (approved):
Pull!

PR #211, opened 2023-11-14
Fix import of relationships with wrong casing

FrodeHus:
## Description

Added function to check if a resource belongs to a scope either by full ID or inherited by resource group.
This check uses case-insensitive matching due to reasons explained below.
It also checks if the scope is at resource group level and will grant inherited relationship if the resource is prefixed with the scope.

## Motivation and Context

When importing data from AzureHound, the virtual machine ID sometimes does not match the casing of the scope which results in relationships such as AZVMAdminLogin is not registered.

Additionally, relationships were only registered if a direct role assignment to the resource was added and ignored resource group inheritance.

From [docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules):
> Resource and resource group names are case-insensitive unless specifically noted in the valid characters column.
> 
> When using various APIs to retrieve the name for a resource or resource group, the returned value may have different casing than what you originally specified for the name. The returned value may even display different case values than what is listed in the valid characters table.
> 
> Always perform a case-insensitive comparison of names.

## How Has This Been Tested?

This change only abstracts away the resource<->scope matching and will be tested by existing tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/ein/azure.go
+++ b/packages/go/ein/azure.go
@@ -42,6 +42,8 @@ const (
 	KeyVaultPermissionGet string = "Get"
 )
 
+var resourceGroupLevel, _ = regexp.Compile("^[\\w\\d\\-\\/]*/resourceGroups/[0-9a-zA-Z]+$")
+
 func ConvertAZAppToNode(app models.App) IngestibleNode {
 	return IngestibleNode{
 		PropertyMap: map[string]any{
@@ -1488,8 +1490,8 @@ func ResourceWithinScope(resource, scope string) bool {
 	if strings.EqualFold(resource, scope) {
 		return true
 	}
-	resourceGroupLevel, _ := regexp.MatchString("^[\\w\\d\\-\\/]*/resourceGroups/[0-9a-zA-Z]+$", resource)
-	if resourceGroupLevel && strings.HasPrefix(strings.ToLower(resource), strings.ToLower(scope)) {
+
+	if resourceGroupLevel.MatchString(scope) && strings.HasPrefix(strings.ToLower(resource), strings.ToLower(scope)) {
 		return true
 	}
 	return false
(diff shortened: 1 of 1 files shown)


FrodeHus:

packages/go/ein/azure.go
```diff
@@ -1482,3 +1483,14 @@ func getKeyVaultPermissions(data models.KeyVaultAccessPolicy) []graph.Kind {
 	}
 	return relationships
 }
+
+func ResourceWithinScope(resource, scope string) bool {
+	if strings.EqualFold(resource, scope) {
+		return true
+	}
+	resourceGroupLevel, _ := regexp.MatchString("^[\\w\\d\\-\\/]*/resourceGroups/[0-9a-zA-Z]+$", resource)
```
zinic, line 1491: Compile this regex and assign it to an outer `var` - it will avoid dumping the error since `MatchString` on the compiled regex object will only return `bool`.

zinic (changes requested):

packages/go/ein/azure.go
FrodeHus: Thanks - pushed the requested change

packages/go/ein/azure.go
```diff
@@ -41,6 +42,8 @@ const (
 	KeyVaultPermissionGet string = "Get"
 )
 
+var resourceGroupLevel, _ = regexp.Compile("^[\\w\\d\\-\\/]*/resourceGroups/[0-9a-zA-Z]+$")
```
zinic, line 45: One last nit, apologies. Please rewrite `regexp.Compile` to `regexp.MustCompile` as the latter will force a panic if the regex is invalid.

zinic (changes requested):

packages/go/ein/azure.go
FrodeHus: Makes sense :)

zinic (approved):
Looks good to me. Thanks for the edits and the PR!

PR #222, opened 2023-11-15
BED-3893: stepwise migration rework

codydbentley:
## Description
- Removed some unnecessary steps from stepwise migration process (GetFilenames, etc)
- Remodeled `Migration` and `Manifest`
- Added `Source`s to capture multiple migration sources
- Modified Migrator to hold multiple `fs.FS` sources
- Modified manifest generation to enable a multi-source migration strategy
- Reworked migration execution to run based on new multi-source Manifest model
- Improved migration tracking (no more 999.999.999 version trickery in local dev)
- Improved migration failure handling through more granular tracking and transactions
- Improved migration integration testing and added some new unit tests

## Motivation and Context
Our enterprise product has a need for pulling in specific SQL migrations from different sources, so we are updating our base migration tool to support this in both BHCE and BHE

## How Has This Been Tested?
- New unit tests
- New integrations tests
- Tons of manual testing

## Types of changes

## Checklist:


cmd/api/src/database/migration/agi.go
```diff
@@ -42,7 +42,7 @@ func SelectorToObjectID(rawSelector string) string {
 }
 
 func (s *Migrator) updateAssetGroups() error {
-	return s.db.Transaction(func(tx *gorm.DB) error {
+	return s.Db.Transaction(func(tx *gorm.DB) error {
```
superlinkx, line 45: Our usual standard for exporting the name `db` is `DB`, which follows the naming style used in other libraries

cmd/api/src/database/migration/migrations/schema.sql
superlinkx, line 1: Adding this for other reviewers:
Normally we won't want to change `schema.sql`, but Cody wanted to make sure migrations handled creating their own table as part of the actual logic, rather than relying on the main `schema` to be involved. It likely does make sense to split things this way, and there's no danger of breaking existing instances here (or new ones)

cmd/api/src/database/migration/stepwise_internal_integration_test.go
superlinkx, line 1: This file was rewritten in the `migrations_integration_test.go` file without the internal package shenanigans and with tests for the concept of having more than one migration folder

cmd/api/src/database/migration/test_migrations/v0.1.1.sql
superlinkx, line 1: I would have liked to see the tests I wrote ported over to our new tests. There's a reason I created some of these more complex migrations and tested the outputs carefully, and they helped avoid migration bugs that would have otherwise not been caught.

superlinkx (commented):
Initial comments ahead of team review

cmd/api/src/database/migration/test_migrations/v0.1.1.sql
codydbentley: The reason I didn't keep this is because it seemed overly specific and engineered more towards testing specific SQL logic. I couldn't determine how it helped test the logic of the Migrator code and decided to replace it with simpler tests specific to testing the Migrator's correctness. I'm happy to be wrong here, but I'd need more context about what this migration was meant to test so I can understand it.

superlinkx: The idea was that it provided something substantial for the migrator to do, ensuring it would successfully run complex operations in the right order. I suppose looking back, the newer simpler tests are probably sufficient for that, so I'll withdraw this comment

superlinkx (approved):

juggernot325 (approved):
LGTM! Thank you for taking on this big effort. It's going to make our lives a lot easier going forward.

PR #225, opened 2023-11-17
tag ag selectors on update

irshadaj:
## Description

This MR is to tag asset group selectors as soon as they are added or removed via the `api/v2/asset-groups/{id}/selectors` endpoint. This circumvents the need to wait for an analysis run to complete in order for the updated results to show.

## How Has This Been Tested?

Spun up the local environment, note the quiet period of 11 minutes:

quiet period

Grabbed a random user from the graph database, note that this is not a T0 user:

before

Added a selector for that user:

post

Refreshed the graph immediately, and the chosen user shows up as T0 even before analysis runs:

updated

Now to test removal, first I started off with picking SteveDraper00002 which is already marked as a T0:

before

Hit the endpoint and removed this selector:

remove post

Immediately querying via the UI shows the node has been removed from T0:

after

## Types of changes

## Checklist:


cmd/api/src/api/v2/agi.go
```diff
@@ -258,7 +259,29 @@ func (s Resources) UpdateAssetGroupSelectors(response http.ResponseWriter, reque
...
 			api.HandleDatabaseError(request, response, err)
 		} else {
-			// When T0 asset group selectors are modified we must trigger analysis
+			// Update system tags for the updated selectors
+			nodeProperties := graph.NewProperties()
+			if assetGroup.Tag == model.TierZeroAssetGroupTag {
+				nodeProperties.Set(common.SystemTags.String(), model.TierZeroAssetGroupTag)
+			} else {
+				nodeProperties.Set(common.SystemTags.String(), model.OwnedAssetGroupTag)
+			}
```
elikmiller, line 268: I'm not sure if you are soliciting feedback on this MR yet however I think we should simplify this logic to just apply or remove `assetGroup.Tag` based on if the selector is being created or deleted. There may be a future where we introduce additional asset groups in which case defaulting to `model.OwnedAssetGroupTag` is not the desired behavior.

cmd/api/src/api/v2/agi.go
```diff
@@ -258,7 +259,29 @@ func (s Resources) UpdateAssetGroupSelectors(response http.ResponseWriter, reque
...
+
+			for _, selectorSpec := range selectorSpecs {
+				if err := s.Graph.WriteTransaction(request.Context(), func(tx graph.Transaction) error {
+					if err := tx.Nodes().Filterf(func() graph.Criteria {
+						return query.Equals(query.NodeProperty(common.ObjectID.String()), selectorSpec.EntityObjectID)
+					}).Update(nodeProperties); err != nil {
```
elikmiller, line 274: We also want to not necessarily overwrite all existing tags. If a node is already a part of tier zero, and we are adding it to owned via a selector, we don't want to overwrite the existing tier zero tag.

elikmiller (changes requested):
I added a node to an asset group via a selector and saw that the asset group label was appended to the "system_tags" property of that node. However, when removing the same selector I noted that the node was still tagged as being part of that asset group.

codydbentley (approved):
I'm not super familiar with asset tagging logic, but I don't see anything that stands out codewise. As long as the other issues are resolved, lgtm

cmd/api/src/api/v2/agi.go
```diff
@@ -258,8 +258,17 @@ func (s Resources) UpdateAssetGroupSelectors(response http.ResponseWriter, reque
...
 		} else {
-			// When T0 asset group selectors are modified we must trigger analysis
+			// update the asset group tags immediately
+			if err = s.GraphQuery.ClearSystemTags(request.Context()); err != nil {
+				log.Warnf("failed clearing asset group tags; will be retried upon next analysis run: %v", err)
+			}
+
+			if err = s.GraphQuery.UpdateAssetGroupIsolationTags(request.Context(), s.DB); err != nil {
+				log.Warnf("failed updating asset group tags; will be retried upon next analysis run: %v", err)
+			}
```
elikmiller, line 268: I notice that any modification to an asset group causes all members of other asset groups to disappear. I think only the nodes associated with a selector need to be modified/tagged.

elikmiller (changes requested):

rvazarkar (approved):

PR #255, opened 2023-12-05
Update the 'Computers where Domain Users can read LAPS passwords' Query

Pri3st:
## Description
It is better to leave the "Computers where Domain Users can read LAPS passwords" Query unconstrained regarding Domain Group membership to avoid false negative results.

## Motivation and Context
The previous query limits the results in cases where Domain Users belong to Domain Groups that are not members of the "Domain Users" Group.

## How Has This Been Tested?
I have experienced engagements where Users belong to the default "Domain Users" Group and a custom "LAPS" Domain Group. The default query would first find membership to the "LAPS" Domain Group and then stop querying for membership in the "Domain Users" Group, resulting in showing no results. However, those Users had the target permission.

## Types of changes
Removed `WHERE m.objectid ENDS WITH "-513"\n` from the query.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/commonSearches.tsx
+++ b/packages/javascript/bh-shared-ui/src/commonSearches.tsx
@@ -82,7 +82,7 @@ export const CommonSearches: CommonSearchType[] = [
             },
             {
                 description: 'Computers where Domain Users can read LAPS passwords',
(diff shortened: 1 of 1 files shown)


Pri3st:

Pri3st:
recheck

StephenHinck:
@Pri3st - this saved query aims to show explicitly where the Domain Users group has permissions to RDP to systems (hence "Domain Users" is capitalized in the query name). The change suggested will show anywhere any user has RDP access, which, although an accurate and potentially interesting query, does not meet the intent of this saved query.

PR #259, opened 2023-12-07
Tag/untag nodes from context menu

brandonshearin:
## Description

This PR adds two new options to the context menu on the explore graph:

**Add to High Value**
**Add to Owned**
Screenshot 2023-12-07 at 10 27 37 AM

A node can be in both 'high value' and 'owned' asset groups.

When a node has been added to an asset group, you will see the menu item change from 'Add to X' -> 'Remove from X'.

Nodes that are default members of the 'High Value' asset group will not provide the remove option.

When you add/remove a node from the 'High Value' asset group, a confirmation dialog will appear before you can apply the change.  Applying the change successfully will optimistically update the diamond glyph on the selected node.

## How Has This Been Tested?

Two components were extracted from the`ContextMenu.tsx` component-- `CopyMenuItem` and `AssetGroupMenuItem`.  All three components are under unit test

confirmation dialog when adding to high value:
Screenshot 2023-12-07 at 10 34 27 AM

menu options when both options are "Remove from X":
Screenshot 2023-12-07 at 10 35 59 AM

## Types of changes

-   [x ] New feature (non-breaking change which adds functionality)

## Checklist:

-   [ x] I have added and/or updated tests to cover my changes.
-   [x ] All new and existing tests passed.


cmd/ui/src/views/Explore/ContextMenu/CopyMenuItem.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+                },
+            },
+        });
+    });
+
+    it('handles copying a display name', async () => {
+        const user = userEvent.setup();
+
+        const copyOption = screen.getByRole('menuitem', { name: /copy/i });
+        await user.hover(copyOption);
```
benwaples, line 41: very cool!!

dockerfiles/bloodhound.Dockerfile
```diff
@@ -17,8 +17,8 @@
 ########
 # Global build args
 ################
-ARG SHARPHOUND_VERSION=v2.0.2
-ARG AZUREHOUND_VERSION=v2.1.6
+ARG SHARPHOUND_VERSION=v2.0.1
+ARG AZUREHOUND_VERSION=v2.1.5
```
benwaples, line 21: heads up - looks like you are missing some commits from main, unless this is intentional?

benwaples (approved):

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <>
+                <MenuItem onClick={isMenuItemForTierZero ? handleOpenConfirmation : handleAddToAssetGroup}>
+                    Add to {assetGroupName}
+                </MenuItem>
+                {isMenuItemForTierZero ? (
+                    <ConfirmationDialog
+                        handleCancel={handleCloseConfirmation}
+                        handleApply={handleAddToAssetGroup}
+                        open={open}
+                        dialogContent={`Are you sure you want to add this node to ${assetGroupName}? This action will initiate an analysis run to update group membership.`}
```
maffkipp, line 111: I think we might want to change the confirmation messages here, since updating membership no longer requires waiting on an analysis run

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+                },
+            ]);
+        },
+        onSuccess: () => {
+            if (selectedNode?.graphId && isMenuItemForTierZero) {
+                dispatch(toggleTierZeroNode(selectedNode.graphId));
+            }
+
+            addNotification(
+                'Update successful. Please check back later to view updated Asset Group.',
```
maffkipp, line 53: Same as the other comment -- these changes should be available immediately at the /members endpoint for the asset group

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItem.tsx
brandonshearin: we should leave this message, because this message is explicitly when adding/removing a node to the high value asset group.

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItem.tsx
brandonshearin: I agree with your wording in this case.  I changed this message to just say "Update successful."

PR #270, opened 2023-12-13
fix: gorm override column names for domain_collection_results

urangel:
## Description
The column names in the domain_collection_results table are set to be overridden.

## Motivation and Context
Gorm uses snake case to relate structs to tables but our migration was not in that form so writing the data point resulted in an error. 

## How Has This Been Tested?
A collection was run with the updated changes and the error observed originally was no longer logging.

## Types of changes

## Checklist:


juggernot325:
The code here looks good, but one thing I thought of while reviewing is why didn't our tests catch this problem? I know we mock most of our DB calls, but do we have a way to verify the columns match what the code is expecting?

juggernot325 (approved):

PR #284, opened 2024-01-02
pathfinding bug fix

brandonshearin:
## Description

the explore page contained a logical error that could be seen if a user had previously performed a pathfinding search, and then switched to a single node search. after searching for a new node, the graph would render a pathfinding search from the new node to the old end node.

this was due to a logical error in the redux saga that slipped in during the development of the right-click menu.  The saga logic has been updated with a new parameter named "doPathfindSearch, " so that the right-click menu can still issue a pathfinding search when the user clicks "set as starting node."  This scenario is possible because a user can have already provided a destination node before clicking "set as starting node." 

## Types of changes

## Checklist:


benwaples (approved):

elikmiller (approved):
Looks sensible to me

PR #292, opened 2024-01-05
fix: prevent analysis failure when entity properties are not found

urangel:
## Description
Checks were added for property accessing. A warning is logged when a property is not found rather than returning an error.

## Motivation and Context
Returning an error when properties are not found breaks analysis. We should be continuing analysis without further processing if properties are not found rather than erroring out.

## How Has This Been Tested?
The ADCS feature flag was enabled in local development and collection data that was not sourced with any of the new collection options selected was ingested into BloodHound. A failed analysis run was observed. After adding checks for non-existing properties, the same data was sent through ingest and analysis only logged property not found messages rather than failing. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/analysis/ad/adcs.go
+++ b/packages/go/analysis/ad/adcs.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	"github.com/specterops/bloodhound/analysis"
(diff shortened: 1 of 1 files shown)


ddlees:
Cherry-picked this change into [stage/v5.4.1](/SpecterOps/BloodHound/tree/stage/v5.4.1) to follow the org's hotfix process.

PR #296, opened 2024-01-08
ParseKind/ParseKinds Refactor and Testing

superlinkx:
## Description

UPDATE: This now also fixes an issue with meta nodes, since they aren't a technically known node type in BHCE. The fix is to fake them. This fixes an issue introduced into BHE after the previous PR.

The logic in these functions was getting a little hairy, so I took some time to reduce the problems a bit. New tests were added for functions that were touched before refactoring to help ensure these changes have no logical side effects.

## Motivation and Context

The goal is to keep our logic in these functions a bit easier to trace and ensure we are creating more reusable abstractions. The original fix for unknown kinds in `GetNodeKinds()` showed a gap in how we can validate that nodes are "known". Rather than overburden `ParseKind()` for this, I decided to split the known nodes list to a separate function. This allows us to check if a node is known much more easily and we can share the implementation with `ParseKind()` as well. While evaluating this fix, it also became apparent that `ParseKind()` could be simplified significantly.

`ParseKinds()` is only moderately more simple. I've moved the type negotiation to a generic function in our slices package for future reuse (it's a common problem, and one we're likely to run into more as we work on larger refactor efforts in the API view layer), so for now it's adding more code than before, but in the long run I'm expecting that function to save us a decent amount of boilerplate.

## How Has This Been Tested?

- New tests were added prior to changes and passed.
- Changes all pass the same tests.

## Types of changes

## Checklist:


packages/go/slices/slices.go
superlinkx, line 1: I'm constantly wishing for an easier way to map slices of one type to another, so I decided to write a generic function for it. This allows us to apply `ParseKind` against a slice of strings in this MR, but could clean up other areas where we need to convert entire slices to another type

packages/go/analysis/analysis_test.go
superlinkx, line 1: Additional tests added to validate existing behavior before the refactor

packages/go/analysis/analysis.go
```diff
@@ -104,16 +102,24 @@ func ClearSystemTags(ctx context.Context, db graph.Database) error {
...
+		kinds   = make([]graph.Kind, 0, lenCalc)
+	)
+
+	kinds = append(kinds, ad.Nodes()...)
+	kinds = append(kinds, ad.Relationships()...)
+	kinds = append(kinds, azure.NodeKinds()...)
+	kinds = append(kinds, azure.Relationships()...)
+
+	return kinds
+}
```
superlinkx, line 117: Moving our list of valid kinds out of `ParseKind` helps us reuse this important bit without having to do weird stuff with the `ParseKind` function elsewhere.

packages/go/analysis/analysis.go
```diff
@@ -125,17 +131,7 @@ func ParseKinds(rawKinds ...string) (graph.Kinds, error) {
...
-	for idx, rawKind := range rawKinds {
-		if parsedKind, err := ParseKind(rawKind); err != nil {
-			return nil, err
-		} else {
-			kinds[idx] = parsedKind
-		}
-	}
-
-	return kinds, nil
+	return slices.MapWithErr(rawKinds, ParseKind)
```
superlinkx, line 134: Generics are neat, I need to evaluate using them in more situations like this

packages/go/analysis/analysis.go
```diff
@@ -104,16 +102,24 @@ func ClearSystemTags(ctx context.Context, db graph.Database) error {
...
 
-	for _, azureKind := range append(azure.NodeKinds(), azure.Relationships()...) {
-		if azureKind.String() == rawKind {
-			return azureKind, nil
+func ParseKind(rawKind string) (graph.Kind, error) {
+	for _, kind := range ValidKinds() {
+		if kind.String() == rawKind {
+			return kind, nil
 		}
 	}
```
superlinkx, line 124: We had no different logic for Azure or AD, so made sense to just have a bulk list of valid kinds outside the function and iterate over them

packages/go/analysis/analysis.go
```diff
@@ -58,21 +59,18 @@ func GetNodeKindDisplayLabel(node *graph.Node) string {
...
 		if kind.Is(ad.Entity, azure.Entity) {
 			baseKind = kind
 		} else if kind.Is(ad.LocalGroup) {
-			// Allow ad.LocalGroup to overwrite NodeKindUnknown, but nothing else
+			// If we run into a local group kind, we want to only consider it valid if no other known kinds have been found yet
 			if resultKind.String() == NodeKindUnknown {
 				resultKind = kind
 			}
-		} else if _, err := ParseKind(kind.String()); err == nil {
+		} else if slices.Contains(ValidKinds(), kind) {
```
superlinkx, line 73: All this work to avoid `err == nil`

urangel (approved):
Nice generics usage 👍 I've pulled and tried this out in my local and it lgtm. Thanks for the support!

PR #297, opened 2024-01-08
New audit log fields

juggernot325:
## Description

This PR adds additional fields to the `audit_logs` model so additional information is available for tracking actions. Three new fields were added:

* `actor_email` - The email address of the user who performed the action. This field is only present on the model and is pulled from the associated record in the users table.
* `source` - The IP address of the requester. This field lives in the `audit_logs` table.
* `status` - Captures the `success` or `failure` of the action performed. This field lives in the `audit_logs` table.

All existing audit logs will have a `status` of `success` as we previously have not tracked audit logs for failed actions. With this PR we still only track successes, so it was hardcoded into newly created audit logs. Support for failing audit actions will be added in a PR in the near future.

The new `actor_email` attribute, as well as the existing `actor_name` will now be pulled directly from the `users` table when querying audit logs. The `actor_name` column was removed from the `audit_logs` table to adjust for this. Also, the current implementation is less efficient than desired due to complexities with associations using `gorm`. It is planned to re-write this to a single `JOIN` query when `gorm` is removed.

Note: This PR is the first of a series related to audit logging. It can be merged on its own and shouldn't cause any issues. But it might be better to wait for the others to be completed as they will all build on each other.

## Motivation and Context

* Provide additional information for tracking actions within the BloodHound app.

## How Has This Been Tested?

* Tested the base migration in an empty DB
* Verified the new columns are created with the correct type
* Verified audit logs are successfully created as before. `source` and `actor_email` are empty on new records, but `status` is populated with `success`
* Verified pre-migration `audit_log` entries are successfully populated with the user's email address or `unknown`

## Types of changes

## Checklist:


superlinkx (changes requested):
We discussed in slack whether we should change the fields to be joined from the users table, rather than duplicating user's names and email addresses into the audit log table. The API output should still include them as is, but we'd like to use a join to prevent unnecessary duplication in the database. This review is on hold until the changes are ready.

cmd/api/src/database/audit.go
```diff
@@ -78,5 +81,27 @@ func (s *BloodhoundDB) ListAuditLogs(before, after time.Time, offset, limit int,
 		result = cursor.Find(&auditLogs).Count(&count)
 	}
 
+	// Populate the actor name and email from the users table. Uses a map as a cache to avoid looking up the same user multiple times.
```
superlinkx, line 84: We should probably do a `CheckError(result)` after running the first db query, since it doesn't make sense to continue if we already had an error.

cmd/api/src/database/audit.go
```diff
@@ -78,5 +81,27 @@ func (s *BloodhoundDB) ListAuditLogs(before, after time.Time, offset, limit int,
...
+	// NOTE: This is intended to be a temporary solution for including this data directly in the returned AuditLog. This is due to a lack
+	// of flexibility with gorm's associations/JOINs.
+	// TODO: When gorm is removed, select these values with a JOIN on the users table.
+	for i, log := range auditLogs {
+		var actor model.User
```
superlinkx, line 89: There's no real need to declare actor here since we aren't worried about shadowing and we have to do `actor, ok := ...` anyway. It doesn't hurt anything but is a little redundant

superlinkx (changes requested):
I'd like us to handle a potential query error before the loop and I had one minor nit that you can evaluate to see if it's worth changing, but otherwise lgtm

cmd/api/src/database/audit.go
juggernot325: Oh good catch. That was leftover from a small refactor. I'll clean it up.

superlinkx (approved):
LGTM, go ahead and pull this in while we work on the other subtasks, no reason to hold it up since it's a safe change on its own

juggernot325:
We ended up abandoning this refactored version of this PR in favor of #310. The other version does duplicate the `actor_name` and `actor_email` columns across two DB tables, but as we got further into the the refactor we kept discovering new landmines that were increasing the scope of work. 

This branch should eventually be picked back up. When it is, we should re-write the gorm queries to use raw SQL and properly join the `actor_name` and `actor_email` fields with the `users` table. We also need to ensure that at least the `actor_name` field is sortable and filterable to avoid breaking API contracts.

PR #306, opened 2024-01-10
Docker Compose QoL

superlinkx:
## Description

- Removes database connection strings from our default configs
- Adds database connection strings to environment variables in Docker Compose, allowing for individual changes to env vars to be reflected for both database configuration and BloodHound's connection to the databases.
- Adds initial support for new `docker compose watch` based commands. This requires a pretty new version of Docker Compose (v2.22.0+) to use and is set up to be optional. (This is designed to be used in conjunction with normal dev operations like `just bh-dev`)

## Motivation and Context

Changing the default database password was a chore, and not hard coding the password in multiple places will make future QoL updates easier.

There are several UI workflows that don't trigger hot reloads, because they aren't part of the module system. Things like findings files in `cmd/ui/public` were the primary impetus.

Over the weekend I found out about the new `docker compose watch` configuration and command, and realized there was an easy way to make UI dev just a little bit easier. This initial configuration is very tightly scoped. It looks for changes in `cmd/ui/public` to trigger a sync and restart (pulls files into the running container and then restarts the service commands, changes in any of the `package.json` files to trigger a full container rebuild (can take anywhere from a couple seconds to almost a minute, depending on if a full new yarn build is needed), takes a few seconds), and changes to any folders in `packages/javascript` other than `node_modules` directories triggers a full container rebuild (again takes anywhere from a few seconds to a minute)

### Docker Compose Example

Tested docker compose with defaults, docker compose with modified passwords in env vars, and running docker compose with a locally built BloodHound image. Everything worked as intended.

### Docker Compose Watch

Expected usage was done with the following:

- First terminal: `just all-watch`
- Second terminal: `just all-dev`

Changes to `cmd/ui/public` triggered an in-place service restart with the files getting synced prior to restart

Note that this may not save changes made between `just all-dev` runs. A `just all-dev up --build` may not be the worst idea to ensure that previous changes are copied into a newly started service, and that recommendation will be updated after gathering initial feedback

## Types of changes

## Checklist:


zinic (approved):
Needs a rebase on mainline but the changes look good from where I sit.

PR #310, opened 2024-01-11
Add new audit log fields

juggernot325:
## Description

This PR adds additional fields to the audit_logs model and DB tables so additional information is available for tracking actions. Three new fields were added:

`source` - The IP address of the requester
`actor_email` - The email address of the user who performed the action
`status` - Captures the success or failure of the action performed

Existing audit logs were also migrated to include the email address of the actor who completed the action. If that user record no longer exists the column will be updated unknown instead.

Also, all existing audit logs will have a status of `success` as we previously have not tracked audit logs for failed actions. With this PR we still only track successes, so it was hardcoded into newly created audit logs. Support for failing audit actions will be added in a PR in the near future.

Note: This PR is the first of a series related to audit logging. It can be merged on its own and shouldn't cause any issues. But it might be better to wait for the others to be completed as they will all build on each other.

## Motivation and Context

* Provide additional information for tracking actions within the BloodHound app.

## How Has This Been Tested?

* Tested the base migration in an empty DB
* Verified the new columns are created with the correct type
* Verified audit logs are successfully created as before. `source` and `actor_email` are empty on new records, but `status` is populated with `success`
* Verified pre-migration `audit_log` entries are successfully populated with the user's email address or `unknown`

## Types of changes

## Checklist:


superlinkx (approved):

superlinkx:
These changes have been through testing before, so we're good to pull when ready

PR #325, opened 2024-01-18
fix: Filter out group nodes from GPO affected T0 queries

mistahj67:
## Description

GPOs don’t apply to groups.  Groups can be used as filters for GPOs but you can’t take control over a group through a GPO.  Our entity panel for GPOs incorrectly shows Groups when expanding the affected objects accordion, then expanding Tier Zero objects.  The query for this should only account for the following objects: Domains, Users, Computers, OUs. 

## Motivation and Context

Fix query to properly filter groups.

## How Has This Been Tested?

Before and after change was compared against local data that reproduced this behavior.
A unit test verifying logic paths was added.

## Types of changes

## Checklist:


superlinkx (approved):
LGTM

PR #337, opened 2024-01-23
Revert ip bind to 0.0.0.0

nurfed1:
https://github.com/SpecterOps/BloodHound/commit/7787b6baf17b54536411b626fd40ba517c27f023 breaks the development build (just bh-dev). 
The traefik proxy can't connect to the api when the service is only listening on localhost.


superlinkx:
We actually ran into this as well this morning. I'll remove it from my Docker QoL PR and we'll get this one merged in

superlinkx (approved):

PR #347, opened 2024-01-24
chore: remove unnecessary complexity in the data pipeline

zinic:
## Description

This changeset removes several unnecessary components of the data pipeline. The original intent for the in-memory retention of completed file upload jobs was to trigger analysis immediately upon completion of file-upload. However, addition of configuration for the loop ticker and a clearer path for triggering a data pipeline loop run removes the need for this complexity.

This also removes a gap where a timed out file upload job could have its status overwritten later on during analysis.

## Motivation and Context

The data pipeline is already complex enough, as-is. Let's not make it harder on ourselves to maintain this part of the system.

## How Has This Been Tested?

Updated unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/daemons/datapipe/datapipe.go
+++ b/cmd/api/src/daemons/datapipe/datapipe.go
@@ -40,22 +40,20 @@ const (
 )
 
 type Tasker interface {
-	NotifyOfFileUploadJobStatus(task model.FileUploadJob)
 	RequestAnalysis()
 	GetStatus() model.DatapipeStatusWrapper
 }
 
 type Daemon struct {
-	db                            database.Database
-	graphdb                       graph.Database
-	cache                         cache.Cache
-	cfg                           config.Configuration
-	analysisRequested             bool
-	tickInterval                  time.Duration
-	status                        model.DatapipeStatusWrapper
-	ctx                           context.Context
-	fileUploadJobIDsUnderAnalysis []int64
-	completedFileUploadJobIDs     []int64
+	db                              database.Database
+	graphdb                         graph.Database
+	cache                           cache.Cache
+	cfg                             config.Configuration
+	analysisRequested               bool
+	tickInterval                    time.Duration
+	status                          model.DatapipeStatusWrapper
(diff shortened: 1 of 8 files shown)


cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -104,7 +102,17 @@ func (s *Daemon) setAnalysisRequested(requested bool) {
 	s.analysisRequested = requested
 }
 
+func (s *Daemon) finishAnalysis() {
+	// Ensure that the user-requested analysis switch is flipped back to false
+	s.setAnalysisRequested(false)
```
elikmiller, line 107: My understanding is that analysis can be requested again while already in progress (like when a file ingest or collection completes while analysis is currently underway.)

Does this changeset account for that scenario?

elikmiller: In my brain you'd need to turn this switch off as soon as analysis _begins_ processing, and allow it to be flipped back on midway through.

superlinkx: We talked about this one before and the previous location was at the end of the function which was also incorrect and meant that we didn't toggle to false if analysis was set to disabled. Of course, we ended up simplifying even more for FOSS since there's no real reason anyone would want to disable analysis here.

I think your proposal makes sense though, and perhaps we should be clearing it at the start to ensure that any analysis requests that get queued during an analysis run aren't wiped. The goal here was to make the logic more visible, and I think we found a few areas that could potentially be better as a result. I'll defer to @zinic on what the correct course of action is for this PR, but I think it will at least be worth the discussion in a separate ticket.

zinic: I hadn't considered the case where the user flips the latch during an analysis run. I'll update this accordingly.

zinic: This has been patched as per Eli's suggestion

superlinkx (approved):
LGTM

PR #354, opened 2024-01-25
ESC9a Edge Composition

rvazarkar:
## Description

Adds edge composition to ESC9a

## Motivation and Context
See description

## How Has This Been Tested?

Tested in lab environment

image

## Types of changes

## Checklist:


JonasBK (approved):

packages/go/dawgs/traversal/traversal.go
```diff
@@ -49,14 +49,18 @@ type PatternMatchDelegate = func(terminal *graph.PathSegment) error
...
+	InboundWithDepth(min, max int, criteria ...graph.Criteria) PatternContinuation
 	Do(delegate PatternMatchDelegate) Driver
 }
 
 // expansion is an internal representation of a path expansion step.
 type expansion struct {
 	criteria  []graph.Criteria
 	direction graph.Direction
+	minDepth  int
+	maxDepth  int
```
urangel, line 63: This might be overly cautious but defining these depths as unsigned integers types might prevent someone from using this incorrectly unless we are handling depths of less than zero.

rvazarkar: We're not handing it currently, but I think a more correct solution here is to simply set the min/max to the default if its < 0. I'll make the appropriate change

urangel (approved):
Thanks for updating the other edge compositions with their outbound depths!

PR #357, opened 2024-01-26
tasks 6 and 7 for #345

irshadaj:
* IP address to not be read from `X-Real-IP` header as this is easy to spoof.
* `X-Forwarded-For` returns a list of IPs, grab the originating IP (first one) from the list 
* Check that Audit Context Action conforms to the DB constraint (`success` or `failure`)

### DIFF
--- a/cmd/api/src/api/middleware/middleware.go
+++ b/cmd/api/src/api/middleware/middleware.go
@@ -130,20 +130,21 @@ func ContextMiddleware(next http.Handler) http.Handler {
 			requestCtx, cancel := context.WithTimeout(request.Context(), requestedWaitDuration.Value)
 			defer cancel()
 			// Insert the bh context
-			if ipAddress, err := parseUserIP(request); err != nil {
-				log.Errorf(err.Error())
-			} else {
-				requestCtx = ctx.Set(requestCtx, &ctx.Context{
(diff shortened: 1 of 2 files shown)


cmd/api/src/api/middleware/middleware.go
```diff
@@ -147,19 +152,16 @@ func ContextMiddleware(next http.Handler) http.Handler {
...
-		}
+func parseUserIP(r *http.Request) (string, error) {
+	if ipAddress := r.Header.Get("X-Forwarded-For"); ipAddress != "" {
+		return strings.Split(ipAddress, ",")[0], nil
+	} else if parsedUrl, err := url.Parse(r.RemoteAddr); err != nil {
+		return "", fmt.Errorf("error parsing IP address from RemoteAddr: %s", err)
+	} else if hostName := parsedUrl.Hostname(); hostName == "" {
+		return "", fmt.Errorf("hostname not found in URL: %s", parsedUrl.String())
+	} else {
+		return parsedUrl.Hostname(), nil
```
superlinkx, line 163: We already have the `hostName` var, but otherwise this logic seems good. I'd like to find out for sure if we need to attach any additional information on how the IP was derived for audit purposes, but that can be cleaned up in a follow up

PR #361, opened 2024-01-29
chore: slices cleanup

superlinkx:
## Description

Refocuses our internal `slices` module as an extension of the newly available Go 1.21+ stdlib `slices` package. This involved renaming our slices module to `slicesext` and removing functions that stdlib now provides. All features that our `slicesext` module provide that the stdlib doesn't were kept as is.

This also backports an upcoming stdlib `slices` function into our extension module: Append! 

## Motivation and Context

- Our `slices` module became a bit unwieldy after the upgrade to Go 1.21, since it directly collides with the name of the stdlib package with a different set of features, but some overlap
- Removing functions that we maintain in favor of stdlib variants seemed like good hygiene
- We've been needing a proper slice append function for a good long time, and with it right around the corner, it seemed prudent to start moving towards its use. `slicesext` uses the current version of this function since it's a fairly stable piece of code and isn't expected to have any API changes between now and Go 1.22 full release. When we eventually upgrade to Go 1.22, we can delete our copy as well as the copied test from our extension, but for now it lets us start using Append with a clean refactor path.

## How Has This Been Tested?

All tests continue to pass. New tests added for new function (stolen and slightly tweaked from Go source).

## Types of changes

## Checklist:


zinic (approved):
No comments. Merge when passing.

cmd/api/src/api/v2/pathfinding.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/specterops/bloodhound/graphschema/ad"
 	"github.com/specterops/bloodhound/graphschema/azure"
 	"github.com/specterops/bloodhound/params"
-	"github.com/specterops/bloodhound/slices"
+	slicesext "github.com/specterops/bloodhound/slicesext"
```
codydbentley, line 29: Nit: redundant alias?

superlinkx: Thanks, I thought I had autocleaned all those

packages/go/analysis/ad/esc_shared.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/specterops/bloodhound/dawgs/util/channels"
 	"github.com/specterops/bloodhound/graphschema/ad"
 	"github.com/specterops/bloodhound/log"
-	"github.com/specterops/bloodhound/slices"
+	slicesext "github.com/specterops/bloodhound/slicesext"
```
codydbentley, line 32: Another possible redundant alias, likely from rename shenanigans

superlinkx: That's exactly what it was. Was doing a multi-step refactor and it just didn't get updated properly after

packages/go/slicesext/slicesext.go
```diff
@@ -125,12 +118,20 @@ func Init[T any](list []T) []T {
 	return list[:len(list)-1]
 }
 
-// Reverse reverses the order of a slice by doing an in place reverse
-// This will reorder the provided slice in place, and uses zero allocations
-func Reverse[T any](list []T) []T {
-	for low, high := 0, len(list)-1; low < high; low, high = low+1, high-1 {
-		list[low], list[high] = list[high], list[low]
+// Concat returns a new slice concatenating the passed in slices.
+// This was ripped from go1.22 source and should be replaced with the stdlib
```
codydbentley, line 122: Nice lol

codydbentley: Do we need to do any license stuff here if we're admitting to taking this directly from go source?

superlinkx: Hmm, that's a great question that I probably should have thought through. A link to the original is likely due at a minimum

codydbentley (approved):
Lgtm

packages/go/slicesext/slicesext.go
superlinkx: Fixed up

PR #362, opened 2024-01-29
redo IP parsing for changed requirements

irshadaj:
* Return all IPs in the X-Forwarded-For header
* Always return Hostname field, whether or not the X-Forwarded-For is populated
* Return all of these as a single string

The result of the parseUserIP function will be in the format:
"X-Forwarded-For: 192.168.1.1:8080,192.168.1.2,192.168.1.3; Hostname: www.google.com"

### DIFF
--- a/cmd/api/src/api/middleware/middleware.go
+++ b/cmd/api/src/api/middleware/middleware.go
@@ -130,10 +130,6 @@ func ContextMiddleware(next http.Handler) http.Handler {
 			requestCtx, cancel := context.WithTimeout(request.Context(), requestedWaitDuration.Value)
 			defer cancel()
 			// Insert the bh context
-			var ipAddress string
-			if ipAddress, err = parseUserIP(request); err != nil {
-				log.Errorf("requestIP not set: %v", err)
-			}
 
 			requestCtx = ctx.Set(requestCtx, &ctx.Context{
 				StartTime: startTime,
@@ -143,7 +139,7 @@ func ContextMiddleware(next http.Handler) http.Handler {
 					Scheme: getScheme(request),
 					Host:   request.Host,
 				},
-				RequestIP: ipAddress,
+				RequestIP: parseUserIP(request),
 			})
 
 			// Route the request with the embedded context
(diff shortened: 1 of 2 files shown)


superlinkx (approved):

juggernot325 (approved):

cmd/api/src/api/middleware/middleware.go
```diff
@@ -152,16 +148,23 @@ func ContextMiddleware(next http.Handler) http.Handler {
...
-func parseUserIP(r *http.Request) (string, error) {
+func parseUserIP(r *http.Request) string {
+	res := ""
 	if ipAddress := r.Header.Get("X-Forwarded-For"); ipAddress != "" {
-		return strings.Split(ipAddress, ",")[0], nil
-	} else if parsedUrl, err := url.Parse(r.RemoteAddr); err != nil {
-		return "", fmt.Errorf("error parsing IP address from RemoteAddr: %s", err)
+		res += "X-Forwarded-For: " + ipAddress + "; "
+	} else {
+		log.Errorf("No data found in X-Forwarded-For")
```
computator, line 156: I think we'll get a lot of these in CE? Since most of the time it won't have the header

cmd/api/src/api/middleware/middleware_internal_test.go
```diff
@@ -80,35 +80,51 @@ func TestRequestWaitDuration(t *testing.T) {
...
-func TestParseUserIP_XForwardedFor(t *testing.T) {
+func TestParseUserIP_XForwardedForMissing(t *testing.T) {
 	req, err := http.NewRequest("GET", "/teapot", nil)
 	require.Nil(t, err)
 
-	ip1 := "192.168.1.1:8080"
-	ip2 := "192.168.1.2"
-	ip3 := "192.168.1.3"
-	req.Header.Set("X-Forwarded-For", strings.Join([]string{ip1, ip2, ip3}, ","))
+	req.RemoteAddr = "http://www.google.com/0.0.0.0:3000"
```
computator, line 87: The `http.Request.RemoteAddr` field says it's just `<host>:<port>` so I believe this should just be `'0.0.0.0:3000'`

cmd/api/src/api/middleware/middleware_internal_test.go
```diff
@@ -80,35 +80,51 @@ func TestRequestWaitDuration(t *testing.T) {
...
 	require.Nil(t, err)
 
-	_, err = parseUserIP(req)
-	require.Contains(t, err.Error(), "hostname")
+	ip1 := "192.168.1.1:8080"
+	ip2 := "192.168.1.2"
+	ip3 := "192.168.1.3"
+	req.Header.Set("X-Forwarded-For", strings.Join([]string{ip1, ip2, ip3}, ","))
+
+	req.RemoteAddr = "http://www.google.com/0.0.0.0:3000"
```
computator, line 120: The `http.Request.RemoteAddr` field says it's just `<host>:<port>` so I believe this should just be `'0.0.0.0:3000'`

computator (commented):
Added comments for a few small issues.

I believe the current code is fine with IPv6 but perhaps the tests should incorporate some IPv6 addresses to prevent future regressions.

If it's helpful, the `2001:db8::/32` network for IPv6 and the `192.0.2.0/24` network for IPv4 are reserved for documentation and tests like this. https://en.wikipedia.org/wiki/Reserved_IP_addresses

The IPv6 format for `RemoteAddr` will be like `[2001:db8::abcd:45]:12345` (I.E. URL Format. [Verified here](https://stackoverflow.com/questions/57563049/whats-the-cleanest-way-to-obtain-the-ip-address-from-net-http-request-remoteadd)) and in `X-Forwarded-For` it will usually just be plain addresses such as: `X-Forwarded-For: 192.0.2.25, 2001:db8::efg:23, 192.0.2.84`

cmd/api/src/api/middleware/middleware.go
superlinkx: This has been changed to a warning level with a little more context in another PR. Shouldn't be too noisy, since most will probably just turn warnings off in most cases (or filter)

PR #363, opened 2024-01-29
Datapipe State Transition Hardening and ADCS Post Processing Fixes

zinic:
## Description

This adds an ingest task lookup before transitioning file upload jobs from the `INGESTING` state to the `ANALYZING` state. There is a potential race where a job may transition into `INGESTING` after `ingestAvailableTasks(...)` is called. In this rare case, ingest tasks would remain until the next datapipe loop tick. Upon ingestion of the remaining tasks analysis would not be triggered, leaving the system in a inconsistent state until the next analysis run.

Additionally, this changeset contains a change in the ADCS post-processing logic that defers creation of a parallel operation context. The functions that previously ran between creation of the parallel operation context would spin up their own parallel operation contexts resulting in creation of 12+ concurrent connections to the database. The system is designed to limit concurrent connections to 10 leaving some of the parallel operations hung until their association context is canceled.

## Motivation and Context

State transition hardening will ensure there are no graph inconsistency gaps. The ADCS post-processing change will increase resiliency of the associated parallel operation code paths. 

## How Has This Been Tested?

Minor changes are covered by existing integration tests.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/daemons/datapipe/jobs.go
+++ b/cmd/api/src/daemons/datapipe/jobs.go
@@ -18,6 +18,7 @@ package datapipe
 
 import (
 	"context"
+	"github.com/specterops/bloodhound/src/database"
 	"os"
 
 	"github.com/specterops/bloodhound/dawgs/graph"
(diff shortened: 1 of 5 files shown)


computator (approved):
I don't know the details of the ADCS logic but the rest looks good!

packages/go/analysis/ad/adcs.go
```diff
@@ -412,29 +412,25 @@ func PostADCS(ctx context.Context, db graph.Database, groupExpansions impact.Pat
...
-		operation.Done()
 		return &analysis.AtomicPostProcessingStats{}, fmt.Errorf("failed fetching domain nodes: %w", err)
 	} else if step1Stats, err := postADCSPreProcessStep1(ctx, db, enterpriseCertAuthorities, rootCertAuthorities); err != nil {
-		operation.Done()
 		return &analysis.AtomicPostProcessingStats{}, fmt.Errorf("failed adcs pre-processing step 1: %w", err)
 	} else if step2Stats, err := postADCSPreProcessStep2(ctx, db, certTemplates); err != nil {
-		operation.Done()
 		return &analysis.AtomicPostProcessingStats{}, fmt.Errorf("failed adcs pre-processing step 2: %w", err)
 	} else {
+		operation := analysis.NewPostRelationshipOperation(ctx, db, "ADCS Post Processing")
```
urangel, line 429: Nice

cmd/api/src/daemons/datapipe/jobs.go
```diff
@@ -51,12 +51,16 @@ func (s *Daemon) completeJobsUnderAnalysis() {
...
-		for _, ingestedFileUploadJob := range ingestedFileUploadJobs {
-			if err := fileupload.UpdateFileUploadJobStatus(s.db, ingestedFileUploadJob, model.JobStatusAnalyzing, "Analyzing"); err != nil {
-				log.Errorf("Error updating fileupload job %d: %v", ingestedFileUploadJob.ID, err)
+		for _, ingestingFileUploadJob := range ingestingFileUploadJobs {
+			if remainingIngestTasks, err := s.db.GetIngestTasksForJob(ingestingFileUploadJob.ID); err != nil {
+				log.Errorf("Failed looking up remaining ingest tasks for file upload job %d: %v", ingestingFileUploadJob.ID, err)
+			} else if len(remainingIngestTasks) == 0 {
+				if err := fileupload.UpdateFileUploadJobStatus(s.db, ingestingFileUploadJob, model.JobStatusAnalyzing, "Analyzing"); err != nil {
+					log.Errorf("Error updating fileupload job %d: %v", ingestingFileUploadJob.ID, err)
+				}
```
urangel, line 63: Not sure how difficult it would be to write an integration test for this but it would be nice to prevent regressions. Feel free to make that call, I'll go ahead and stamp this

urangel (approved):
Thanks for the adcs post cleanup!

cmd/api/src/daemons/datapipe/jobs.go
zinic: I don't actually think this needs an integration test. I need to think about this for a moment but let me see what I can come up with. The state transitions should be relatively easy to mock out.

Thanks for calling this out!

PR #368, opened 2024-01-30
chore: edge composition check from list of constants

urangel:
## Description
I was updating BHE and wanted to make some changes that required updates to CE since this list of composition edges will be shared.

## Motivation and Context
Makes edge checking less verbose and more robust.

## Types of changes

## Checklist:


packages/javascript/bh-shared-ui/src/constants.ts
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { ActiveDirectoryRelationshipKind } from '.';
+
+export const EdgesWithComposition = [
+    ActiveDirectoryRelationshipKind.GoldenCert,
```
rvazarkar, line 20: Can we generate this via the cue files?

rvazarkar: I guess if its only used in one place, potentially not necessary

urangel: I thought about that and had that working but then I realized we would only use it in the UI since the backend needs mapping to the handler and I don't want to define the handlers from the go code in the cue schema

rvazarkar (approved):

packages/javascript/bh-shared-ui/src/constants.ts
urangel: Actually I might have to pull these from a list defined in ad.cue to avoid the circular dependency that's causing tests to fail ;s

PR #373, opened 2024-01-31
more auth handlers for audit log

irshadaj:
Audit logging for database methods:
* CreateUser
* DeleteUser
* DeleteAuthSecret

Also added unit tests for the uncovered endpoint to delete users.

### DIFF
--- a/cmd/api/src/api/v2/auth/auth.go
+++ b/cmd/api/src/api/v2/auth/auth.go
@@ -493,9 +493,9 @@ func (s ManagementResource) updateUser(response http.ResponseWriter, request *ht
 	}
 }
 
-func (s ManagementResource) ensureUserHasNoAuthSecret(context ctx.Context, user model.User) error {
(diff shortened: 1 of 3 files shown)


cmd/api/src/api/v2/auth/auth.go
```diff
@@ -495,7 +495,7 @@ func (s ManagementResource) updateUser(response http.ResponseWriter, request *ht
 
 func (s ManagementResource) ensureUserHasNoAuthSecret(context ctx.Context, user model.User) error {
```
superlinkx, line 496: This function wasn't using the `ctx.Context` before anyway, so I'd prefer us to change the signature to `context.Context` to avoid the unnecessary `context.ConstructGoContext()`

superlinkx (commented):
Small nit to fix a previously broken window. Doesn't block my approval though, merge when ready

superlinkx (approved):

PR #382, opened 2024-02-02
Generate audit logs on login

juggernot325:
## Description

This PR updates the `authenticator` to generate audit logs on login attempts. It follows a similar system as database-level audit logs where an `intent` record is created when the login attempt starts and a second record is generated after completion to record the actual status. These records were added to the `authenticator` to ensure they would be logged anywhere a password is checked.

One item to note is that the username will be logged in the `fields` section for the `intent` record. This is because it is created prior to verification that the user is valid. If the user is found in the database by the principal name then it will be populated in the `actor_` fields of the audit log.

Lastly, these logs are only created for login attempts using a password. Access via an API token will not create audit log entries and will need to be investigated in the application logs.

## Motivation and Context

* Create a reliable record of each user that accesses or attempts to access BloodHound

## How Has This Been Tested?

* Tested various successful and failure login cases

## Types of changes

## Checklist:


cmd/api/src/api/middleware/middleware.go
```diff
@@ -149,13 +149,16 @@ func ContextMiddleware(next http.Handler) http.Handler {
...
 	} else {
-		result += "," + r.RemoteAddr
-		return result
+		log.Warnf("No data found in X-Forwarded-For, possible upstream misconfig?")
 	}
+
+	res += "Remote Address: " + r.RemoteAddr
+
+	return res
 }
```
superlinkx, line 162: I think this is a result of the parent branch getting merged before you could rebase off of the changes made. Needs to be reverted

cmd/api/src/api/middleware/middleware_internal_test.go
superlinkx, line 1: This file also appears to have been reverted accidentally

cmd/api/src/api/router/router.go
```diff
@@ -95,6 +95,11 @@ func (s *Route) CheckFeatureFlag(db database.Database, flagKey string) *Route {
 	return s
 }
 
+// func (s *Route) CreateAuditLog(db database.Database, action string) *Route {
+// 	s.handler.Use(middleware.AuditLogMiddleware(db, action))
+// 	return s
+// }
+
```
superlinkx, line 102: Another unintentional revert, let's make sure it gets deleted

superlinkx (changes requested):
Just some unexpected shifts caused by the upstream branch getting last minute changes and then merging before we could pull them in (each of these were in my PR too). Once fixed, this looks good to merge

cmd/api/src/api/middleware/middleware.go
juggernot325: I updated the functionality here to strip the port from `r.RemoteAddr`. It's just a randomly selected port that the client used to connect with and doesn't really add anything useful. I figured stripping it would avoid possible confusion.

superlinkx: Good thinking on removing the port, I agree that it'll take away useless information that just adds confusion

superlinkx (approved):

PR #387, opened 2024-02-05
Group management filtering

benwaples:
## Description
Adds a Filter section to Group Managment tab where one can filter on Node Type and custom member.

## Motivation and Context
This should help with narrowing search results in the group management tab.

## How Has This Been Tested?
Manual Tests. Adding unit and integration tests in a later PR

https://github.com/SpecterOps/BloodHound/assets/66393111/a7cee46c-c273-4195-973c-2b405d0a63b1

## Types of changes

## Checklist:


packages/javascript/bh-shared-ui/src/components/AssetGroupEdit/AssetGroupEdit.tsx
```diff
@@ -151,6 +161,7 @@ const FilteredMemberCountDisplay: FC<{
     const hasValidCount = !isLoading && !isError && count && count > 0;
 
     if (hasValidCount) {
+        handleAvailableNodeKind?.();
```
vermentor, line 164: Today I learned

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.tsx
```diff
@@ -0,0 +1,133 @@
...
+interface Props {
+    filterParams: AssetGroupMemberParams;
+    handleFilterChange: (
+        key: keyof Pick<AssetGroupMemberParams, 'primary_kind' | 'custom_member'>,
+        value: string
+    ) => void;
+    availableNodeKinds: Array<ActiveDirectoryNodeKind | AzureNodeKind>;
+}
+
+const AssetGroupFilters: FC<Props> = (props) => {
```
vermentor, line 71: Should we destructure these props?  Could make it easier for anyone else using this component to see what props it takes per intellisense.

packages/javascript/bh-shared-ui/src/components/GroupManagementContent/GroupManagementContent.tsx
```diff
@@ -80,6 +83,24 @@ const GroupManagementContent: FC<{
...
 
+    const handleFilterChange = (
+        key: keyof Pick<AssetGroupMemberParams, 'primary_kind' | 'custom_member'>,
+        value: string
+    ) => {
+        // Custom Member filter displays custom members, or all members.
+        // If we want to also display only non customer members, change this:
+        if (key === 'custom_member' && value.toLowerCase().includes('false')) {
+            setFilterParams(({ custom_member, ...removedCustomMemberFilter }) => removedCustomMemberFilter);
+            return;
```
vermentor, line 94: Is this backwards?  The logic escapes me a bit, but should we be setting filter params to `custom_member` without the `removedCustomMemberFilter`?

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/FilterAutocomplete.tsx
vermentor, line 1: Is this file intentionally empty?

packages/javascript/bh-shared-ui/src/components/AssetGroupEdit/AssetGroupEdit.tsx
benwaples: yess very handy with functions that are optional params!

packages/javascript/bh-shared-ui/src/components/GroupManagementContent/GroupManagementContent.tsx
benwaples: Great question! In the case that the value of the checkbox is `'false'` (not just falsey 🙄 ) then we want to remove the filter param completely because the two states with this filter are: 1. only show custom members, or 2. show all members. If we set the `custom_member` param to 'false', then it will return all non-custom members. Which is not the intended use case according to product.

So here we are deconstructing the existing `filterParams` object pulling `customer_member` out and renaming the rest of the params to `removedCustomMemberFilter`. That way we can return `removedCustomMemberFilter` without the customer_member param! Now that im looking at it again, its not a very good name :/ 

Ill break this line up into a couple lines to make is a little more explicit as to whats going on!

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/FilterAutocomplete.tsx
benwaples: ohhhh great catch! I totally made this and then then search component got removed from scope and forgot to delete the file. Thanks!

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.tsx
benwaples: The props do get deconstructed just below this but it was a force of habit to not deconstruct in line with the function definition... however you are 100% right and most other component files deconstruct in line so Ill update to match! Thank you!

definitelynotagoblin (approved):

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+            expect(handleFilterChange).toHaveBeenCalledWith('primary_kind', `eq:${availableNodeKinds[0]}`);
+        });
+    });
+
+    describe('Custom Member checkbox filter', () => {
+        it("displays the checkbox as checked if the filter params value is 'true'", async () => {
+            const { screen } = await setup({ filterParams, availableNodeKinds });
+            const checkbox = screen.getByTestId('asset-groups-custom-member-filter');
+
+            expect((checkbox.firstChild as HTMLInputElement)?.checked).toBe(true);
```
definitelynotagoblin, line 120: Should we explicitly set the filterParams value to true here instead of trusting that the mock factory provides this value?

definitelynotagoblin (approved):

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.test.tsx
benwaples: Yeah great suggestion! And it's more explicit

PR #397, opened 2024-02-07
feat: Add /count endpoint to /asset-group/:assetGroupId/members endpoint

mistahj67:
## Description

The group management page makes duplicate /members requests in order to get the total counts across primary kinds. This is resource intensive and redundant db hits for the same resources.

## Motivation and Context

Decrease the DB load on neo4j as only 1 request needs to be made to obtain the group member counts

## How Has This Been Tested?

Tested against test db
Integration test added

image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/v2/agi.go
+++ b/cmd/api/src/api/v2/agi.go
@@ -430,6 +430,49 @@ func (s Resources) ListAssetGroupMembers(response http.ResponseWriter, request *
 	}
 }
 
+func (s Resources) ListAssetGroupMemberCountsByKind(response http.ResponseWriter, request *http.Request) {
+	var agMembers api.AssetGroupMembers
+
+	queryParameterFilterParser := model.NewQueryParameterFilterParser()
+	if queryFilters, err := queryParameterFilterParser.ParseQueryParameterFilters(request); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsBadQueryParameterFilters, request), response)
+		return
+	} else {
+		for name, filters := range queryFilters {
+			if validPredicates, err := agMembers.GetValidFilterPredicatesAsStrings(name); err != nil || name == "primary_kind" {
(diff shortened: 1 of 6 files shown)


cmd/api/src/api/v2/agi.go
```diff
@@ -430,6 +430,49 @@ func (s Resources) ListAssetGroupMembers(response http.ResponseWriter, request *
...
+func (s Resources) ListAssetGroupMemberCountsByKind(response http.ResponseWriter, request *http.Request) {
+	var agMembers api.AssetGroupMembers
+
+	queryParameterFilterParser := model.NewQueryParameterFilterParser()
+	if queryFilters, err := queryParameterFilterParser.ParseQueryParameterFilters(request); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsBadQueryParameterFilters, request), response)
+		return
+	} else {
+		for name, filters := range queryFilters {
+			if validPredicates, err := agMembers.GetValidFilterPredicatesAsStrings(name); err != nil || name == "primary_kind" {
```
mistahj67, line 442: I wanted to point out the hardcode primary_kind, wasn't sure if there was a better way to prevent just this filter or not.

cmd/api/src/api/v2/agi.go
```diff
@@ -430,6 +430,49 @@ func (s Resources) ListAssetGroupMembers(response http.ResponseWriter, request *
 	}
 }
 
+func (s Resources) ListAssetGroupMemberCountsByKind(response http.ResponseWriter, request *http.Request) {
```
mistahj67, line 433: This uses a lot of the above list assetgroupmember fn, did we want to abstract this out a bit or good to keep the barriers up across endpoints?

cmd/api/src/api/v2/agi.go
mistahj67: We discussed in slack but the object being counted is the `AssetGroupMembers` so we can't make a new model as there's nothing to filter / count there and we didn't want to create a hidden dependency between the two models to keep in sync. Instead we opted to abstract the two to ensure these endpoints stay in-sync

PR #408, opened 2024-02-09
feat: fixup migration visiblity for ordering

zinic:
## Description

Fixups for migration visibility to allow for conditional ordering of migration application.

## Motivation and Context

To enable better migration logic elsewhere in the application.

## How Has This Been Tested?

Local testing.

## Types of changes

## Checklist:


packages/go/dawgs/drivers/neo4j/driver.go
```diff
@@ -135,6 +135,10 @@ func (s *driver) AssertSchema(ctx context.Context, schema graph.Schema) error {
 	return assertSchema(ctx, s, schema)
 }
 
+func (s *driver) SetDefaultGraph(ctx context.Context, schema graph.Graph) error {
+	return nil
```
codydbentley, line 139: something sniffs funny here, I think I need context

codydbentley:
i'm a little fuzzy on the default graph setting vs asserting schema logic, but it sniffs good (pending one comment)

packages/go/dawgs/drivers/neo4j/driver.go
zinic: Ah, yeah, this needs an inline comment. Neo4j doesn't support graph namespaces so this is literally a `nop`. Will add documentation.

codydbentley (approved):

PR #409, opened 2024-02-12
Group management node counts

benwaples:
## Description
Implement a new endpoint, `/members/count`, to get the count per node kinds in a environment+member group. This replaces calling `/members` for every node kind only to use the count value in the response. 

## Motivation and Context
Calling `/members` for every node kind was expensive and slow in large environments. 

## How Has This Been Tested?
Manual tests, additional unit tests, and updated unit tests.

## Types of changes

## Checklist:


benwaples:
This PR builds off of: https://github.com/SpecterOps/BloodHound/pull/397. 

Once we get that merged in, Ill move this PR out of draft.

packages/javascript/bh-shared-ui/src/components/AssetGroupEdit/AssetGroupEdit.tsx
```diff
@@ -15,27 +15,26 @@
...
-} from '../../graphSchema';
 import { useNotifications } from '../../providers';
 
 const AssetGroupEdit: FC<{
     assetGroup: AssetGroup;
     filter: AssetGroupMemberParams;
-    makeNodeFilterable: (node: ActiveDirectoryNodeKind | AzureNodeKind) => void;
-}> = ({ assetGroup, filter, makeNodeFilterable }) => {
+    memberCounts: AssetGroupMemberCounts | undefined;
+}> = ({ assetGroup, filter, memberCounts }) => {
```
definitelynotagoblin, line 37: Is this intended to remove the filterable logic we added last week?  Looks like this might have been build atop an older commit.

definitelynotagoblin: This simplifies filtered counts logic it's intended 👍

benwaples: Great question!

In this case its not removing the filtering logic because the new `memberCounts` value replaces the need for `makeNodeFilterable`!

packages/javascript/bh-shared-ui/src/components/GroupManagementContent/GroupManagementContent.tsx
```diff
@@ -74,6 +72,24 @@ const GroupManagementContent: FC<{
...
+            'getAssetGroupMembersCount',
+            filterParams.environment_id,
+            filterParams.environment_kind,
+            selectedAssetGroup,
+        ],
+        enabled: !!selectedAssetGroupId,
+        queryFn: ({ signal }) =>
+            apiClient
+                .getAssetGroupMembersCount(
+                    selectedAssetGroupId?.toString() ?? '', // TODO: this sucks, but dependent queries are better in react-query v4 and we're on v3
```
definitelynotagoblin, line 86: Will this log net request errors when selectedAssetGroupId is undefined?  Do we care?

definitelynotagoblin (approved):

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.tsx
```diff
@@ -30,46 +29,45 @@ import {
...
 
 interface Props {
     filterParams: AssetGroupMemberParams;
     handleFilterChange: (key: (typeof FILTERABLE_PARAMS)[number], value: string) => void;
-    availableNodeKinds: Array<ActiveDirectoryNodeKind | AzureNodeKind>;
+    memberCounts: AssetGroupMemberCounts | undefined;
 }
 
-const AssetGroupFilters: FC<Props> = ({ filterParams, handleFilterChange, availableNodeKinds }) => {
+const AssetGroupFilters: FC<Props> = ({ filterParams, handleFilterChange, memberCounts }) => {
```
definitelynotagoblin, line 70: Would it be easier to make this an optional param with default value {}?

packages/javascript/bh-shared-ui/src/components/GroupManagementContent/GroupManagementContent.tsx
benwaples: This query should only run when the selectedAssetGroup is not falsey because we have it enabled depending on `!!selectedAssetGroup`. So in theory selectedAssetGroup shouldnt be undefined when this network request happens.  Ill update this comment to better describe whats going on here!

There way to do this that is [cleaner and more explicit in the v5](https://tanstack.com/query/v5/docs/framework/react/guides/dependent-queries#usequeries-dependent-query) version of react-query, but this enabled field is whats available to us in v4!

Thanks!

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.tsx
benwaples: Good suggestion! We will need a default value of `{ counts: {}}` but I think that is cleaner than the fallback below! Then if we add more logic around memberCounts.counts we dont always need to fallback.

Updating!

PR #413, opened 2024-02-12
Create `usePermissions` UI hook

maffkipp:
## Description
- Adds a reusable `usePermissions` hook that checks permissions of the current user against a provided list of permissions
- Creates a custom `renderHook` method override for testing library set up in the same way as our custom `render` method
- Adds permission definition enums to the shared UI project

## Motivation and Context
- We want a reusable solution for restricting UI elements based on the current user's permissions

## How Has This Been Tested?
- Added tests for `usePermissions` hook

## Types of changes

## Checklist:


cmd/ui/src/hooks/usePermissions/usePermissions.tsx
```diff
@@ -0,0 +1,46 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { useCallback, useEffect, useState } from 'react';
+import { useSelector } from 'react-redux';
+import { getSelfResponse } from 'src/ducks/auth/types';
+import { PermissionsSpec } from 'bh-shared-ui';
+
+const usePermissions: (permissions: PermissionsSpec[]) => boolean = (permissions) => {
+    const [hasAllPermissions, setHasAllPermissions] = useState<boolean>(false);
+    const auth: { user: getSelfResponse } = useSelector((state: any) => state.auth);
```
benwaples, line 24: Is it possible to replace this `any` with `AppState` here?

Someting that I think we could do (in a separate PR) is replace useSelectors with `useAppSelector`, essentially a typed `useSelector` where you dont have to call out state every single time.

https://react-redux.js.org/using-react-redux/usage-with-typescript#define-typed-hooks

cmd/ui/src/test-utils.jsx
```diff
@@ -27,75 +27,103 @@ import createSagaMiddleware from 'redux-saga';
...
 };
 
+const customRenderHook = (
+    hook,
+    {
+        initialState = {},
+        queryClient = createDefaultQueryClient(),
+        history = createMemoryHistory(),
+        theme = createTheme(defaultTheme),
+        store = createDefaultStore(initialState),
```
benwaples, line 116: I never thought about using a param as the default value of another param. very neat!

cmd/ui/src/hooks/usePermissions/usePermissions.tsx
maffkipp: I've replaced it with `useAppSelector` after your recent PR

benwaples (approved):

elikmiller:
It might be more versatile to have this hook return the list of permissions for the currently logged in user alongside a few helper functions to help determine if a user has a list of required permissions or at least one from a list of permissions.

What are your thoughts?

maffkipp:
@elikmiller I don't think thats a bad idea at all, we may not even need to return the user's current permissions -- just a set of helper functions that can be called with a list of permissions to match against as the argument.

mistahj67 (approved):
Thanks for cleaning up the permissions callback! 🧹 🧼

cmd/ui/src/hooks/usePermissions/usePermissions.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const usePermissions = () => {
+    const auth = useAppSelector((state) => state.auth);
+    const [userPermMap, setUserPermMap] = useState<Record<string, boolean>>({});
+
+    const formatKey = useCallback((p: { authority: string; name: string }) => `${p.authority}-${p.name}`, []);
+
+    useEffect(() => {
+        const userPermissions = auth.user?.roles.map((role) => role.permissions).flat();
```
benwaples, line 34: I think we also need an optional chain for `roles?.map`!

cmd/ui/src/hooks/usePermissions/usePermissions.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+        if (userPermissions) {
+            const newPermMap: Record<string, boolean> = {};
+            userPermissions.forEach((perm) => (newPermMap[formatKey(perm)] = true));
+            setUserPermMap(newPermMap);
+        }
+    }, [auth.user, formatKey]);
```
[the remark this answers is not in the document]
maffkipp: This setup was based on some previous feedback so we aren't iterating through the user's permissions every time we check a new set, just when the authenticated user changes in our store. I was originally just checking directly against the permissions from `auth`, if that approach is preferable I can rework it!

cmd/ui/src/test-utils.jsx
```diff
@@ -27,75 +27,103 @@ import createSagaMiddleware from 'redux-saga';
...
-                    </StyledEngineProvider>
-                </QueryClientProvider>
-            </Provider>
-        );
-    };
+    const AllTheProviders = ({ children }) => createProviders({ queryClient, history, theme, store, children });
     return render(ui, { wrapper: AllTheProviders, ...renderOptions });
 };
 
+const customRenderHook = (
```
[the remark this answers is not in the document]
maffkipp: I _think_ this is rolled into testing-library as of react 18, and that the `renderHook` function we are wrapping here is the same thing (at least, I am not needing to manually set up a harness component for calling that hook when writing tests).

packages/javascript/bh-shared-ui/src/utils/permissions.ts
```diff
@@ -0,0 +1,124 @@
...
+    AUTH_MANAGE_USERS,
+    CLIENTS_MANAGE,
+    CLIENTS_READ,
+    CLIENTS_TASKING,
+    COLLECTION_MANAGE_JOBS,
+    GRAPH_DB_READ,
+    GRAPH_DB_WRITE,
+    SAVED_QUERIES_READ,
+    SAVED_QUERIES_WRITE,
+    FAKE_PERMISSION,
```
elikmiller, line 35: Maybe we can remove this one? Doesn't seem to be used in a test or anything

elikmiller (approved):
Looks good!

packages/javascript/bh-shared-ui/src/utils/permissions.ts
maffkipp: done!

PR #414, opened 2024-02-13
useAppSelector

benwaples:
## Description
This PR mostly improves types so that we dont have to specify `AppState` every time we use `useSelector`. There are a couple of changes to support the more accurate types but nothing that changes functionality.

## Motivation and Context
Using `useAppSelector` instead of the generic `useSelector` reduces imports, slightly increases readability, and helps make sure we use types from the redux store. There are a couple places in the app where we use the `any` type for redux state which isnt ideal.

## How Has This Been Tested?
Regression testing any non type-only changes and updated unit tests.

## Types of changes
- [x] Chore (a change that does not modify the application functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] Documentation updates are needed, and have been made accordingly.
- [x] I have added and/or updated tests to cover my changes.
- [ ] My changes include a database migration.

### DIFF
--- a/cmd/ui/src/views/GroupManagement/GroupManagement.tsx
+++ b/cmd/ui/src/views/GroupManagement/GroupManagement.tsx
@@ -20,21 +20,19 @@ import { SelectedNode } from 'src/ducks/entityinfo/types';
 import { useState } from 'react';
 import { AssetGroup, AssetGroupMember } from 'js-client-library';
 import { faGem } from '@fortawesome/free-solid-svg-icons';
-import { useSelector } from 'react-redux';
-import { Domain } from 'src/ducks/global/types';
 import { setSelectedNode } from 'src/ducks/entityinfo/actions';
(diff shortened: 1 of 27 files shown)


cmd/ui/src/views/GroupManagement/GroupManagement.tsx
```diff
@@ -72,7 +70,8 @@ const GroupManagement = () => {
             };
         });
     };
-
+    console.log('globalDomain', globalDomain);
+    if (!globalDomain) return null;
```
benwaples, line 74: `globalDomain` is initiated as null so it doesnt hurt to have this check. Also TS wants it to be able to pass to globalDomain as a prop to `GroupManagementContent`.

maffkipp (approved):

PR #415, opened 2024-02-13
Init property map if nil on ingest node

definitelynotagoblin:
## Description

Initialize property map on node ingest if undefined in JSON input file.

## Motivation and Context

https://specterops.atlassian.net/browse/BED-4076
This error occurs because we assume that the field "Properties" is defined on the ingest object.
We should either initialize it if it does not exist on processing or otherwise raise an error.

## How Has This Been Tested?

Ingest run on data files containing both populated and null "Properties" fields.

## Types of changes

## Checklist:


packages/go/ein/ad.go
```diff
@@ -35,9 +35,14 @@ func ConvertSessionObject(session Session) IngestibleSession {
...
 func ConvertObjectToNode(item IngestBase, itemType graph.Kind) IngestibleNode {
+	itemProps := item.Properties
+	if itemProps == nil {
+		itemProps = make(map[string]any)
+	}
+
 	return IngestibleNode{
 		ObjectID:    item.ObjectIdentifier,
-		PropertyMap: item.Properties,
+		PropertyMap: itemProps,
```
definitelynotagoblin, line 45: Do we want to support ingest of nodes with null "Properties"?  Or should we instead raise an error?

rvazarkar: Its fine as is, this lets us support third party ingest a bit easier and loosens up our requirements

superlinkx (approved):

PR #423, opened 2024-02-16
Database Management 

brandonshearin:
## Description

This changeset adds a new page underneath "administration" that enables a user to obliterate their data. 

**3/1 UPDATE** 
the api contract has changed slightly to accommodate more flexibility in deleting asset group selectors.  the request body used to have a boolean field named `deleteHighValueSelectors`, which has been changed to the name `deleteAssetGroupSelectors` and it accepts a list of ints (corresponding to asset group id's)

the ui has been updated with 5 checkboxes instead of 4:
Screenshot 2024-03-01 at 10 49 25 AM

api changes: 
- added a new "WipeDB" permission to the "administrator" role 
- added a new endpoint `POST api/v2/clear-database` 
- added 3 new methods to the bloodhound db interface to delete data: **DeleteAllAssetGroupSelectorsForAssetGroups()** , **DeleteAllDataQuality()**, and **DeleteAllFileUploads**
- add swagger docs for new endpoint (this is going to get deleted after cody's PR so i'll keep track of that)

ui changes:
- new `<DatabaseManagement />` and `<ConfirmationDialog/>` components

## Motivation and Context

give a user the ability to light fire to their data

## How Has This Been Tested?

unit tests have been added to cover UI changes, and unit and integration tests have been added for the api handler

## Types of changes

## Checklist:


cmd/api/src/auth/permission.go
```diff
@@ -95,5 +98,7 @@ func Permissions() PermissionSet {
 
 		SavedQueriesRead:  model.NewPermission("saved_queries", "Read"),
 		SavedQueriesWrite: model.NewPermission("saved_queries", "Write"),
+
+		WipeDB: model.NewPermission("wipedb", "wipedb"),
```
mistahj67, line 102: Style nit here but it seems like we follow a patter of lowercase for authority and then uppercase for name and i'm not sure if we wanted to make the authority generalized such as:
`db` and then for name `Wipe` perhaps?

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -0,0 +1,307 @@
...
+    const handleCheckbox = (event: React.ChangeEvent<HTMLInputElement>) => {
+        dispatch({
+            type: 'selection',
+            targetName: event.target.name,
+            checked: event.target.checked,
+        });
+    };
+
+    return (
+        <ContentPage title='Clear BloodHound data'>
```
urangel, line 201: We should capitalize _data_ here to align with the casing in the other page titles.

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -0,0 +1,307 @@
...
+                        ) : null}
+
+                        {state.mutationError ? (
+                            <Box color={theme.palette.error.main} display='flex' alignItems='center' gap='0.3rem'>
+                                <FontAwesomeIcon icon={faCircleXmark} />
+                                <FormHelperText sx={{ marginTop: '1.5px', color: theme.palette.error.main }}>
+                                    There was an error processing your request.
+                                </FormHelperText>
+                            </Box>
+                        ) : null}
```
urangel, line 231: We use the MUI Alert components in other places in the app. This looks like a good place to use an <Alert/> so that these kinds of messages look consistent.

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -0,0 +1,307 @@
...
+        dispatch({
+            type: 'selection',
+            targetName: event.target.name,
+            checked: event.target.checked,
+        });
+    };
+
+    return (
+        <ContentPage title='Clear BloodHound data'>
+            <div>
```
urangel, line 202: It would be nice to swap this out for an MUI \<Box/> so that it applies the generated MUI css classes. It is effectively the same right now but if we ever start introducing any theming (like dark mode 👀 ) it will be easier to have MUI handle sweeping style changes rather than having to update this div manually. Wherever possible, try to use MUI components.

cmd/api/src/database/agi.go
```diff
@@ -41,9 +41,11 @@ func (s *BloodhoundDB) CreateAssetGroup(ctx context.Context, name, tag string, s
 		}
 	)
 
-	return assetGroup, s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+	err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
 		return CheckError(tx.Create(&assetGroup))
 	})
+
+	return assetGroup, err
```
urangel, line 48: This is a nice clean up for readability 👍 What do you think about also moving the err declaration into the var block above? The same can be done for the change on line 180 if you decide to do so.

urangel (commented):
Good job making both the API and UI changes on this! Also, thanks for all the tests you included. This feature will be a big quality of life improvement. Most, if not all, of the comments I left were minor consistency nits but the logic checks out. 

On a final note, I understand that we want to make it difficult to accidentally delete your data and that is why we ask for typing the phrase in but it felt a bit too cumbersome to me. If nothing else, maybe we can make the phrase matching case insensitive? I enjoyed the way legacy BloodHound would swap the confirm delete button with the cancel button in case one was spamming clicks. Just my two cents 🤷

elikmiller:
I don't see our swagger docs updated with the new endpoint, those updates need to be included.

elikmiller:
The ticket calls out that we should "initiate a new analysis" after deleting data - I think this piece is missing.

mistahj67 (changes requested):
We discussed in DM but the permissions addition requires a migration to be added to create the permission as well as to add the permission to the admin role.

cmd/api/src/api/v2/database_management_integration_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+			graphdb, ok := lab.Unpack(harness, fixtures.GraphDBFixture)
+			assert.True(ok)
+
+			graphdb.ReadTransaction(context.Background(),
+				func(tx graph.Transaction) error {
+
+					return nil
+				})
+
```
mistahj67, line 71: we should assert / check the read transaction has no error right? also might be good to verify that the data is first in the db and then the data is no longer in the db by doing a get around the delete?

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -0,0 +1,307 @@
...
+            return {
+                ...state,
+                [targetName]: checked,
+                noSelectionError: false,
+            };
+        }
+        case 'open_dialog': {
+            const { collectedGraphData, dataQualityHistory, fileIngestHistory, highValueSelectors } = state;
+            if (
+                [collectedGraphData, dataQualityHistory, fileIngestHistory, highValueSelectors].filter(Boolean)
```
benwaples, line 120: Loving the separation of concerns here in this file!

nit: Could we pull this out to a named variable to help give context as to what we are checking? Its not very obvious at first and takes a couple reads to see whats going on!

Maybe something like 
```js
const noSelection = collectedGraphData, dataQualityHistory, fileIngestHistory, highValueSelectors].filter(Boolean).length === 0

if (noSelection) {
  ...
}

```

cmd/ui/src/views/DatabaseManagement/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+                                <>
+                                    <FontAwesomeIcon icon={faCircleXmark} /> Please input the phrase prior to clicking
+                                    confirm.
+                                </>
+                            ) : null
+                        }></TextField>
+                </Box>
+            </DialogContent>
+            <DialogActions>
+                <Button onClick={handleClose}>Cancel</Button>
```
benwaples, line 90: minor ux thing: IMO it would be nice to clear the error and input state on cancel as well!

cmd/ui/src/views/DatabaseManagement/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+                    <Typography variant='body1'>Continuing will delete all data from your environment.</Typography>
+                    <Typography
+                        variant='body1'
+                        color={theme.palette.error.main}
+                        fontWeight={theme.typography.fontWeightMedium}>
+                        This change is irreversible.
+                    </Typography>
+                    <Typography variant='body1'>Please input the phrase prior to click confirm.</Typography>
+                    <TextField
+                        placeholder='Please delete my data'
```
benwaples, line 72: We could reuse the `confirmationText ` variable here so if we need to change it in the future then we only need to change it in one place!!

cmd/api/src/api/registration/v2.go
```diff
@@ -131,6 +131,8 @@ func NewV2API(cfg config.Configuration, resources v2.Resources, routerInst *rout
 		routerInst.GET("/api/v2/features", resources.GetFlags).RequirePermissions(permissions.AppReadApplicationConfiguration),
 		routerInst.PUT("/api/v2/features/{feature_id}/toggle", resources.ToggleFlag).RequirePermissions(permissions.AppWriteApplicationConfiguration),
 
+		routerInst.POST("/api/v2/database-management", resources.HandleDatabaseManagement).RequirePermissions(permissions.WipeDB),
```
elikmiller, line 134: My opinion is that this endpoint should be renamed. To me "database management" implies a lot more beyond clearing data. Maybe "/api/v2/clear-database"?

cmd/api/src/api/v2/database_management_integration_test.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2023 Specter Ops, Inc.
```
elikmiller, line 1: Should be 2024 for new files

cmd/api/src/api/v2/database_management.go
```diff
@@ -0,0 +1,155 @@
...
+				api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("%s: %s", "error deleting all nodes", err.Error()), request),
+				response,
+			)
+			return
+		}
+
+	}
+
+	// delete custom high value selectors
+	if payload.HighValueSelectors {
```
elikmiller, line 90: What is the expectation if both `payload.CollectedGraphData` and `payload.HighValueSelectors` are true but an error occurs inside this block? Do we end up with only some data being deleted instead of all requested data deleted?

cmd/api/src/api/v2/database_management.go
```diff
@@ -0,0 +1,155 @@
...
+			api.WriteErrorResponse(
+				request.Context(),
+				api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("%s: %s", "there was an error deleting data quality history", err.Error()), request),
+				response,
+			)
+			return
+		}
+	}
+
+	if err := s.DB.AppendAuditLog(request.Context(), model.AuditEntry{
```
elikmiller, line 139: Flagging @juggernot325 since he is up-to-date with all things audit logging. Can you take a look here and validate that this is the appropriate way to create audit logs in this scenario?

cmd/api/src/auth/permission.go
elikmiller: I agree that the naming here could be improved

cmd/ui/src/views/DatabaseManagement/ConfirmationDialog.tsx
elikmiller: The Cancel and Confirm buttons should have two different colors

cmd/ui/src/views/DatabaseManagement/ConfirmationDialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+                        variant='standard'
+                        value={input}
+                        onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+                            setInput(event.target.value);
+                        }}
+                        error={error}
+                        helperText={
+                            error ? (
+                                <>
+                                    <FontAwesomeIcon icon={faCircleXmark} /> Please input the phrase prior to clicking
```
elikmiller, line 82: We should clear this message once the user types the appropriate phrase. It should also re-appear if the phrase is subsequently modified and is no longer matching.

cmd/api/src/api/v2/database_management.go
elikmiller: Also, it looks like the custom selector is deleted but the nodes themselves remained tagged as part of the asset group. Is that expected behavior @StephenHinck ?

elikmiller (changes requested):

cmd/api/src/api/v2/database_management.go
juggernot325: I know we discussed this while you were working on it and decided on generating the audit logs in this method. After looking at the PR as a whole I think we may want to go a different route. One thing that I think is big concern is that if there are any errors in the above operations there will not be an audit log generated at all, so that should be addressed.

I actually think this is going to need a novel implementation of audit logging given the complexity of this operation. Let's meet up and discuss some options.

cmd/api/src/database/db.go
```diff
@@ -138,22 +138,46 @@ type Database interface {
...
 	GetFileUploadJobsWithStatus(status model.JobStatus) ([]model.FileUploadJob, error)
+	DeleteAllFileUploads() error
 	ListSavedQueries(userID uuid.UUID, order string, filter model.SQLFilter, skip, limit int) (model.SavedQueries, int, error)
 	CreateSavedQuery(userID uuid.UUID, name string, query string) (model.SavedQuery, error)
 	DeleteSavedQuery(id int) error
 	SavedQueryBelongsToUser(userID uuid.UUID, savedQueryID int) (bool, error)
+	DeleteAssetGroupSelectors(ctx context.Context, assetGroupId int) error
+	Transaction(fn func(db *BloodhoundDB) error) error
+	Begin() *BloodhoundDB
+	Commit() error
```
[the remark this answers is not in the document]
brandonshearin: i removed this code. it was leftover from something i was trying that was testing related, but i ended up not going that direction

cmd/api/src/api/v2/database_management.go
brandonshearin: to your first question, if a user requests to delet emultiple types of data, and gets a partial failure, maybe we should surface that to the UI?

so if i delete graph data and file ingest history, and file ingest history fails, the user would see on the UI something like "we encountered an error deleting file ingest history. please try again."

brandonshearin:
> We discussed in DM but the permissions addition requires a migration to be added to create the permission as well as to add the permission to the admin role.

appreciate the guidance on this one.  i got this working now-- i added a new migration file titled v5.7.0.sql and added the sql to create the db.Wipe permission and then associated that permission with the Admin role

brandonshearin:
> The ticket calls out that we should "initiate a new analysis" after deleting data - I think this piece is missing.

added this to the code block for deleting graph data and for deleting custom high value selectors

mistahj67:
> I don't see our swagger docs updated with the new endpoint, those updates need to be included.

Might be worth coordinating with @codydbentley since he's knee deep in refactoring swagger currently

cmd/api/src/api/v2/database_management.go
juggernot325: The latest iteration of this looks good to me! It should successfully capture the intent audit log and capture actual results.

cmd/api/src/api/v2/database_wipe.go
```diff
@@ -0,0 +1,161 @@
...
+			api.WriteErrorResponse(
+				request.Context(),
+				api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("%s: %s", "error deleting all nodes", err.Error()), request),
+				response,
+			)
+			return
+		}
+
+		// if succesful, kick off analysis
+		s.TaskNotifier.RequestAnalysis()
```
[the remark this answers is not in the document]
brandonshearin: got it!

cmd/api/src/api/v2/database_wipe_integration_test.go
```diff
@@ -0,0 +1,89 @@
...
+		computer3     = fixtures.NewComputerFixture(computer3UUID, "computer 3", fixtures.BasicDomainFixture)
+	)
+
+	lab.Pack(harness, graphdb)
+	lab.Pack(harness, computer1)
+	lab.Pack(harness, computer2)
+	lab.Pack(harness, computer3)
+
+	lab.NewSpec(t, harness).Run(
+		lab.TestCase("the endpoint deletes graph data", func(assert *require.Assertions, harness *lab.Harness) {
```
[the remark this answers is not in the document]
brandonshearin: i added a lot of unit test coverage to this function to cover more bases.  our integration test suite like lab doesn't support writing a test like "create some high value selectors in the db, call the wipe database endpoint, and then assert that there are no high value selectors in the db."

i tried to follow this approach, but what i was running into was a race condition in the test.  i would setup some fixtures in lab to create some data, i.e. some high value selectors, but asset group selectors depend on a FK relationship to an asset group, so I made a fixture to create the asset group first, and then a selector.

some really wonky behavior happened where the asset group record would be created in memory in gorm but wouldn't be committed to the db, so there was a race condition when trying to create the selector.  creating the selector usually threw a FK constraint violation because the asset group wasn't actually committed to the db yet.

it was too burdensome to try to get something to work for the integration test, so i just unit tested instead!

cmd/api/src/database/dataquality.go
```diff
@@ -182,3 +183,30 @@ func (s *BloodhoundDB) GetAzureDataQualityAggregations(start time.Time, end time
 
 	return azureDataQualityAggregations, int(count), nil
 }
+
+func (s *BloodhoundDB) DeleteAllDataQuality() error {
+
+	errADQualityAggregations := CheckError(
+		s.db.Exec("DELETE FROM ad_data_quality_aggregations"),
```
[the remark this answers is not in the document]
brandonshearin: sure can! done

cmd/api/src/database/dataquality.go
```diff
@@ -182,3 +183,11 @@ func (s *BloodhoundDB) GetAzureDataQualityAggregations(start time.Time, end time
 
 	return azureDataQualityAggregations, int(count), nil
 }
+
+func (s *BloodhoundDB) DeleteAllDataQuality(ctx context.Context) error {
+	err := CheckError(
+		s.db.WithContext(ctx).Exec("DELETE FROM ad_data_quality_aggregations; DELETE FROM ad_data_quality_stats; DELETE FROM azure_data_quality_aggregations; DELETE FROM azure_data_quality_stats;"),
+	)
+
+	return err
```
mistahj67, line 192:
```suggestion
	return CheckError(
		s.db.WithContext(ctx).Exec("DELETE FROM ad_data_quality_aggregations; DELETE FROM ad_data_quality_stats; DELETE FROM azure_data_quality_aggregations; DELETE FROM azure_data_quality_stats;"),
	)
```

cmd/api/src/go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/unrolled/secure v1.13.0
 	github.com/zenazn/goji v1.0.1
 	go.uber.org/mock v0.2.0
+	go.uber.org/zap v1.13.0
```
mistahj67, line 50: What are these go.uber.org/zap for?

cmd/ui/src/views/Administration/Administration.tsx
```diff
@@ -48,6 +50,11 @@ const Administration: React.FC = () => {
                     path: ROUTE_ADMINISTRATION_DATA_QUALITY,
                     component: QA,
                 },
+                {
+                    label: 'Database Management',
```
mistahj67, line 54: Did we want to keep this as Database Management ? 🤔 Probably should check with product

mistahj67 (commented):
Backend logic LGTM! Thanks for your patience with all of my feedback!

mistahj67 (approved):
Approving to remove my requested changes blocker but definitely get a more seasoned front end approval before merging!

cmd/ui/src/views/Administration/Administration.tsx
brandonshearin: yep! this is the name specified in the acceptance criteria

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -0,0 +1,293 @@
+// Copyright 2024 Specter Ops, Inc.
```
maffkipp, line 1: I think it might be helpful to break the reducer, hook, and component out into separate files here

cmd/api/src/database/agi.go
```diff
@@ -204,6 +213,13 @@ func (s *BloodhoundDB) DeleteAssetGroupSelector(ctx context.Context, selector mo
 	})
 }
 
+func (s *BloodhoundDB) DeleteAssetGroupSelectorsForAssetGroup(ctx context.Context, assetGroupId int) error {
+	return CheckError(
+		s.db.Where("asset_group_id = ?", assetGroupId).
```
mistahj67, line 218: We need the `WithContext(ctx)` here on this db call

brandonshearin: gotcha

cmd/api/src/api/v2/database_wipe.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/specterops/bloodhound/src/api"
+	"github.com/specterops/bloodhound/src/model"
+)
+
+type DatabaseWipe struct {
+	DeleteCollectedGraphData bool `json:"deleteCollectedGraphData"`
+	DeleteHighValueSelectors bool `json:"deleteHighValueSelectors"`
+	DeleteFileIngestHistory  bool `json:"deleteFileIngestHistory"`
+	DeleteDataQualityHistory bool `json:"deleteDataQualityHistory"`
+	AssetGroupId             int  `json:"assetGroupId"`
```
elikmiller, line 38: If the purpose of this endpoint is to "delete high value selectors" - why is the onus on the requestor to provide the correct asset group ID? Can the service not derive the asset group ID for the appropriate asset group?

cmd/api/src/api/v2/database_wipe.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// return `BadRequest` if request specifies `high value selectors` without an `asset group ID`
+	if payload.DeleteHighValueSelectors && payload.AssetGroupId == 0 {
+		api.WriteErrorResponse(
+			request.Context(),
+			api.BuildErrorResponse(http.StatusBadRequest, "please provide an assetGroupId to delete", request),
+			response,
+		)
+		return
+	}
```
elikmiller, line 68: Assuming the answer to my previous question is "the service should not need the requestor to provide an asset group ID" then we can remove this validation

elikmiller (changes requested):

cmd/api/src/api/v2/database_wipe.go
StephenHinck: I forgot to consider in the AC for this ticket that multiple groups exist (and will become more prevalent in the future).

Would it make more sense to leave it this way, change `DeleteHighValueSelectors` to `DeleteGroupSelectors` and then move `AssetGroupId` to an array/list for improved long-term supportability (with an 'all' option)? I recognize this would be a change to AC and scope, and I am happy to discuss it.

cmd/api/src/api/v2/database_wipe.go
brandonshearin: removed!

cmd/api/src/api/v2/database_wipe.go
elikmiller: @StephenHinck it is your feature/call to make. I think that what @brandonshearin has implemented satisfies your original AC but if that is no longer desired then we will need time to discuss and make adjustments. Consider that once this changeset is in and released it will be harder to make changes to the API contract.

elikmiller (approved):

PR #425, opened 2024-02-20
feat: handle rendering of edges from node to itself

urangel:
## Description
This updates our rendering programs to identify and handle the rendering of "self-edges" (an edge originating and terminating at the same node). A radial offset is applied when there are multiple self-edges present. 

## Motivation and Context
These edges are currently not being rendered and they are becoming salient with the ESC3 edge logic that can possibly hinge on an EnrollOnBehalfOf edge that may be a self-edge.

image

image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/ui/src/rendering/programs/edge.self.ts
+++ b/cmd/ui/src/rendering/programs/edge.self.ts
@@ -14,16 +14,6 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-/**
- * Sigma.js WebGL Renderer Edge Program
- * =====================================
- *
- * Program rendering edges as thick lines but with a twist: the end of edge
- * does not sit in the middle of target node but instead stays by some margin.
- *
- * This is useful when combined with arrows to draw directed edges.
- * @module
- */
 import CurvedEdgeProgram from './edge.curved';
 import { Coordinates, NodeDisplayData } from 'sigma/types';
 import { Attributes } from 'graphology-types';
@@ -79,10 +69,6 @@ export const getControlPointsFromGroupSize = (
 };
 
 export default class SelfEdgeProgram extends CurvedEdgeProgram {
-    constructor(gl: WebGLRenderingContext) {
-        super(gl);
-    }
-
     process(
         sourceData: NodeDisplayData,
         targetData: NodeDisplayData,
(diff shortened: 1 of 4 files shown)


zinic:
I didn't notice any issue in trying this out locally. The code reviews well. No nits. Lots of good refactors to control the complexity involved.

zinic (approved):

cmd/ui/src/rendering/programs/edge.self.ts
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+        const control2 = { x: center.x + x1Offset, y: center.y + y1Offset };
+        const control3 = { x: center.x + x2Offset, y: center.y + y2Offset };
+
+        return { control2: control2, control3: control3 };
+    }
+};
+
+export default class SelfEdgeProgram extends CurvedEdgeProgram {
+    constructor(gl: WebGLRenderingContext) {
```
maffkipp, line 82: I get an eslint warning here -- JS classes already create an empty constructor by default (with the `super` call for derived classes) so it is not necessary to declare your own

cmd/ui/src/rendering/programs/edge.selfArrowHead.ts
```diff
@@ -0,0 +1,118 @@
...
+import { bezier } from 'src/rendering/utils/bezier';
+import { getNodeRadius } from 'src/rendering/utils/utils';
+import { getControlPointsFromGroupSize } from './edge.self';
+
+const POINTS = 3,
+    ATTRIBUTES = 9,
+    STRIDE = POINTS * ATTRIBUTES;
+
+export default class SelfEdgeArrowHeadProgram extends CurvedEdgeArrowHeadProgram {
+    constructor(gl: WebGLRenderingContext) {
```
maffkipp, line 37: same thing here

cmd/ui/src/rendering/programs/edge.selfArrowHead.ts
```diff
@@ -0,0 +1,118 @@
...
+            false,
+            false
+        );
+
+        const curveCircleIntersectionApproximation = bezier.getCoordinatesAlongCubicBezier(
+            start,
+            control2,
+            control3,
+            start,
+            0.8975
```
maffkipp, line 71: It might be good to pull this number into a constant to give a little more context to how we are doing this approximation

cmd/ui/src/rendering/programs/edge.selfArrowHead.ts
urangel: Thanks for pointing this out! I removed these constructors.

PR #426, opened 2024-02-20
remove default request timeouts, obey custom timeouts via prefer header

irshadaj:
## Description
* Remove the default timeouts of 30 seconds for all API requests
* Continue to support the `prefer` header to specify custom timeouts for individual requests

## How Has This Been Tested?

* Updated unit tests
* Manual testing to verify that the prefer header is processed as supplied

Hitting any endpoint without a `Prefer` request header returns a response without the `Preference-Applied` header in the response:
no wait

Adding the `Prefer` header to the request, we see the `Preference-Applied` header in the response:
wait

## Types of changes

## Checklist:

### DIFF
--- a/tools/docker-compose/neo4j.Dockerfile
+++ b/tools/docker-compose/neo4j.Dockerfile
@@ -16,4 +16,6 @@
 
 FROM docker.io/library/neo4j:4.4.0 as neo4j
 RUN echo "dbms.security.auth_enabled=false" >> /var/lib/neo4j/conf/neo4j.conf
+RUN echo "dbms.security.procedures.unrestricted=apoc.*" >> /var/lib/neo4j/conf/neo4j.conf
 RUN neo4j-admin memrec >> /var/lib/neo4j/conf/neo4j.conf
+RUN wget https://github.com/neo4j-contrib/neo4j-apoc-procedures/releases/download/4.4.0.0/apoc-4.4.0.0-core.jar -P /var/lib/neo4j/plugins/


cmd/api/src/api/middleware/middleware.go
```diff
@@ -195,21 +199,16 @@ func ParseHeaderValues(values string) map[string]string {
...
-	if hasWaitSpecified {
+func parsePreferHeaderWait(value string) (time.Duration, error) {
+	if waitStr, hasWaitSpecified := ParseHeaderValues(value)["wait"]; hasWaitSpecified {
 		if parsedNumSeconds, err := strconv.Atoi(waitStr); err != nil {
 			return 0, err
 		} else {
 			return time.Second * time.Duration(parsedNumSeconds), nil
 		}
+	} else {
+		return time.Duration(0), nil
```
mistahj67, line 210: I don't think this assertion is needed 🤷

cmd/api/src/api/v2/version.go
mistahj67, line 1: are these changes intended?

cmd/api/src/queries/graph_test.go
```diff
@@ -97,21 +93,21 @@ func TestGraphQuery_RawCypherSearch(t *testing.T) {
...
+		require.Equal(t, time.Minute*10, config.Timeout)
 
 		return nil
 	}).Times(2)
 
 	// Unset the user-set timeout in the BH context to validate QC runtime reduction of a complex query
-	outerBHCtxInst.Timeout.UserSet = false
-	outerBHCtxInst.Timeout.Value = time.Minute
+	// This will be set to a default of 30 min, with a reduction factor of 3, so we should have a 10 min config timeout
+	outerBHCtxInst.Timeout = 0
```
mistahj67, line 103: I'm confused here, the comment says it should be 30 minutes but you have 0 here

cmd/api/src/api/middleware/middleware.go
mistahj67: Was referring to the type assertion `time.Duration(0)` vs just 0

rvazarkar:
Tests look good and nothing stands out logic wise that would be a problem. Are there existing tests to validate that our timeouts actually work as intended? Maybe something that just sleeps for 10 seconds with a 5 second wait header to ensure we get a timeout

PR #428, opened 2024-02-21
Improved accuracy on saved queries for tags

StephenHinck:
## Description

Many of the saved cypher queries used `system_tags = 'admin_tier_0'` which will break when the system_tags field contains multiple tags. Updating to use `CONTAINS` instead for accuracy.

## Motivation and Context
Improve query accuracy

## How Has This Been Tested?

Tested in local env

## Types of changes

-   [X] Chore (a change that does not modify the application functionality)

## Checklist:

-   [X] All new and existing tests passed.


superlinkx (approved):

elikmiller:
Technically a user could create a custom asset group with the tag "admin_tier_0_fake" and these new queries would capture those nodes. Better solution might be something like `WHERE "admin_tier_0" IN split(n.system_tags, ' ')`

PR #433, opened 2024-02-22
fix: cypher pattern range parsing

zinic:
## Description
This change works around some minor deficiencies in the ANTLR generated parser for openCypher to correctly parse pattern ranges.

## Motivation and Context
As written, the openCypher parser would incorrectly set the starting range index of a pattern when only an ending range index is specified.

This was discovered while attempting to perform ranged pattern matching: `match p = (:User)<-[*1..3]-(t:Base) where t.name = 'example' return p`.

The above query was being parsed and emitted as: `match p = (:User)<-[*3..]-(t:Base) where t.name = 'example' return p` which would then result in an unbounded traversal and crash Neo4j.

## How Has This Been Tested?
An additional test case was added to cover this.

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/cypher/frontend/pattern.go
+++ b/packages/go/cypher/frontend/pattern.go
@@ -17,9 +17,12 @@
 package frontend
 
 import (
+	"fmt"
+	"github.com/antlr4-go/antlr/v4"
 	"github.com/specterops/bloodhound/cypher/model"
 	"github.com/specterops/bloodhound/cypher/parser"
 	"github.com/specterops/bloodhound/dawgs/graph"
+	"strconv"
 )
 
 type WhereVisitor struct {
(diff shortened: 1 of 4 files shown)


packages/go/cypher/frontend/pattern.go
```diff
@@ -145,13 +148,52 @@ func (s *RelationshipPatternVisitor) ExitOC_RightArrowHead(ctx *parser.OC_RightA
 }
 
 func (s *RelationshipPatternVisitor) EnterOC_RangeLiteral(ctx *parser.OC_RangeLiteralContext) {
-	s.ctx.Enter(&RangeLiteralVisitor{
-		PatternRange: &model.PatternRange{},
-	})
+	const (
```
zinic, line 151: Not the biggest fan of having to write a little state machine for this but I couldn't think of a better way to correctly negotiate pattern ranges.

superlinkx (approved):
Not seeing issues on my side, appreciate the regression test for sure. Probably wouldn't hurt for someone else more familiar with the parser code to double check it, but seems pretty sane.

PR #434, opened 2024-02-23
fix: plumb ctx into asset-groups db methods

mistahj67:
## Description

Plumb context into gorm asset-group db methods

I also noticed there was a number of unused db methods related to the asset-groups so I removed those to keep things nice and clean. (I didn't check the rest of the methods, but I do think it would be beneficial to give them all a quick audit)

## Motivation and Context

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/error.go
+++ b/cmd/api/src/api/error.go
@@ -17,6 +17,7 @@
 package api
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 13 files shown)


cmd/api/src/go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/unrolled/secure v1.13.0
 	github.com/zenazn/goji v1.0.1
 	go.uber.org/mock v0.2.0
+	go.uber.org/zap v1.13.0
```
juggernot325, line 50: It's curious how this new dependency got added. It doesn't look like it's used in any of the files in this PR.

zinic: Let's `zap` this dependency if we aren't using it. Pun intended. Suffer.

zinic (changes requested):
Requesting dependency clean up, otherwise changeset looks good.

cmd/api/src/go.mod
mistahj67: image

PR #437, opened 2024-02-26
asset group tags to not contain whitespaces

irshadaj:
## Description

Whitespaces in asset group tags conflict with delimiters in the "system_group" properties. This MR adds a check to disallow spaces in the tags field when creating an asset group. It also adds a database migration to update any existing asset group tags and trim the whitespace out.

## How Has This Been Tested?
Added unit test for validation
Manual test for database migration

Set up the database with one asset group containing a space in the tag
before

Ran the update to remove spaces
update

Verified that the spaces are removed from id=3
after

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/database/migration/migrations/v5.7.1.sql
+++ b/cmd/api/src/database/migration/migrations/v5.7.1.sql
@@ -14,7 +14,5 @@
 --
 -- SPDX-License-Identifier: Apache-2.0
 
--- Add new columns for audit_logs
-
 UPDATE asset_groups
 SET tag = REPLACE(tag, ' ', '');


cmd/api/src/api/v2/agi.go
```diff
@@ -200,6 +200,8 @@ func (s Resources) CreateAssetGroup(response http.ResponseWriter, request *http.
 
 	if err := api.ReadJSONRequestPayloadLimited(&createRequest, request); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if strings.Contains(createRequest.Tag, " ") {
```
zinic, line 203: Knee-jerk was just to strip and trim the string but this feels more correct.

zinic (approved):
Changes look good to me!

cmd/api/src/api/v2/agi.go
superlinkx, line 203: Quick question: should we disallow all forms of whitespace? It seems like it may be a good idea just to be safe. We'd probably use a regex to test for any whitespace character, rather than enumerate whitespace checks manually.

cmd/api/src/database/migration/migrations/v5.7.1.sql
superlinkx, line 1: We may want to make this v5.8.0 due to the next version definitely being a feature update

PR #439, opened 2024-02-27
ESC4: post processing

brandonshearin:
## Description

esc4 edge added to post processing
- Added new handler to the analysis package to perform post processing for ESC4 edges
- updates to the adcs cache:
   - renamed `CertTemplateControllers` -> `CertTemplateEnrollers`
   - introduced a new cached field: `CertTemplateControllers`, which are principals that have privlieges on a cert template through owns, generic all, write dacl, or wrtie owner edges

## How Has This Been Tested?

a staggering amount of scenarios have been integration tested

Screenshot 2024-03-01 at 1 27 31 PM

Screenshot 2024-03-01 at 1 53 40 PM

## Types of changes

## Checklist:


cmd/api/src/analysis/ad/adcs_integration_test.go
```diff
@@ -662,6 +662,352 @@ func TestADCSESC3(t *testing.T) {
...
+		require.Nil(t, err)
+
+		certTemplates, err := ad2.FetchNodesByKind(context.Background(), db, ad.CertTemplate)
+		require.Nil(t, err)
+
+		domains, err := ad2.FetchNodesByKind(context.Background(), db, ad.Domain)
+		require.Nil(t, err)
+
+		cache := ad2.NewADCSCache()
+		cache.BuildCache(context.Background(), db, enterpriseCertAuthorities, certTemplates)
```
urangel, line 687: There is a `FetchADCSPrereqs` method that can be utilized here and in the following similar blocks for conciseness and noise reduction.

cmd/api/src/test/integration/harnesses.go
```diff
@@ -5504,6 +5504,745 @@ func (s *RootADHarness) Setup(graphTestContext *GraphTestContext) {
...
+		EnrolleeSuppliesSubject: true,
+		NoSecurityExtension:     false,
+		RequiresManagerApproval: false,
+		SchemaVersion:           2,
+		SubjectAltRequireSPN:    false,
+		SubjectAltRequireUPN:    false,
+	})
+
+	s.Domain = graphTestContext.NewActiveDirectoryDomain("Domain", domainSid, false, true)
+	s.EnterpriseCA = graphTestContext.NewActiveDirectoryEnterpriseCA("EnterpriseCA", domainSid)
```
urangel, line 5594: I looked over the design document and didn't see it mentioned so it may not matter but it looks like the diagrams show a property of `IsUserSpecifiesSanEnabled=true` for the enterpriseCA which is not setup in the code for this harness. If the property doesn't have any effects on whether or not an ESC4 edge is created it would be good to remove it from the diagrams or if it does have effects than we should make sure to set it up here in the code and not just in the harness diagram.

packages/go/analysis/ad/esc4.go
```diff
@@ -0,0 +1,208 @@
...
+				),
+			)
+
+			// 2c. kick out early if cert template does meet conditions for ESC4
+			if valid, err := isCertTemplateValidForESC4(certTemplate); err != nil {
+				log.Warnf("error validating cert template %d: %v", certTemplate.ID, err)
+				continue
+			} else if !valid {
+				continue
+			}
```
urangel, line 77: Moving this to the start of the if/else chain after line 38 would save us from querying the db unnecessarily in case the certTemplate is not valid.

urangel: Oh actually this looks to follow the order that was specified in the design doc so it tracks even though it seems a bit different than the logic for other escalations.

urangel (approved):
The tests can be updated to have a smaller footprint but the work looks good to me so I'll leave my approval

PR #440, opened 2024-02-27
Audit log tests

juggernot325:
## Description

This PR adds a number of automated tests to verify the creation of audit logs is protected from regressions. There are a mix of unit and integration tests based on what was being tested. Tests were specifically added for each action that generates an audit log throughout the app. Some actions didn't have integration tests created yet, so those were added, including checks for the audit log generation.

## Motivation and Context

* Testing was skipped as part of #345 in an effort to move quickly. This PR fills those gaps.

## How Has This Been Tested?

* Verified all of the tests pass locally

## Types of changes

## Checklist:


cmd/api/src/auth/model.go
```diff
@@ -149,14 +149,20 @@ func (s authorizer) AllowsAtLeastOnePermission(ctx Context, requiredPermissions
 }
 
 func (s authorizer) AuditLogUnauthorizedAccess(request *http.Request) {
-	// Ignore read logs as they are less likely to occur from malicious access
+	commitId, err := uuid.NewV4()
+	if err != nil {
+		log.Errorf("error generating commit ID for unauthorized access: %s", err.Error())
```
superlinkx, line 154: This will log the error, but if we don't have a valid uuid, do we really want to try to commit it without one?

cmd/api/src/model/audit.go
```diff
@@ -156,6 +157,27 @@ type AuditEntry struct {
 	ErrorMsg string
 }
 
+// Necessary function for testing. Ensures all fields except CommitID match so mocking checks pass.
```
superlinkx, line 160: If this is only used in tests, we should probably define it with the tests (can't easily add it as a method, but a method is just a function where the first parameter is the struct you're operating on)

cmd/api/src/model/audit.go
```diff
@@ -156,6 +157,27 @@ type AuditEntry struct {
...
+	} else if s.Status != expected.Status {
+		return false
+	} else if !reflect.DeepEqual(s.Model, expected.Model) {
+		return false
+	}
+
+	return true
+}
+
+func (s AuditEntry) String() string {
```
superlinkx, line 177: Is this Stringer actually getting any use? I couldn't find any issues by deleting it, so I'm not sure why we need it in this form right now

superlinkx (changes requested):
One logical concern and two nits. Testing strategy looks good to me  and it looks like we're passing. Thank you for your hard work on this effort!

cmd/api/src/model/audit.go
juggernot325: This function and the `String()` function are needed to satisfy the `Matcher` interface in gomock. It needs to exist on the type to override the default match behavior, which fails when comparing the UUID. I used [this tutorial](https://medium.com/modanisa-engineering/writing-a-custom-matcher-for-testing-with-gomock-d3ef5f13db82) as a reference.

cmd/api/src/model/audit.go
juggernot325: Same as [this comment](https://github.com/SpecterOps/BloodHound/pull/440#discussion_r1504981753).

cmd/api/src/auth/model.go
superlinkx: That's a good point. I'm not sure what the best course of action would be, so leaving it as a pass through to commit without an ID seems sane to me

cmd/api/src/model/audit.go
superlinkx: Ah, that makes sense. Yeah, since we don't have a way to know what the UUID is, since it's generated on the fly, I get why you'd go this route. I'm still not a big fan of having code specific to mocks in our application code, but I'm not sure what I'd recommend instead.

Side note: if you wouldn't mind changing `interface{}` to `any`, it will meet our styleguide (`any` is functionally `interface{}` so it won't break the gomock caller

juggernot325: Will do! Also FWIW, I do agree on the co-mingling of test code + functional code, but this does solve the testing problem.

superlinkx: Agreed, this is a pragmatic solution for now. It sounds like changing our testing to avoid it would be more work, and this is already a significant amount of work as is. We can clean it up in the future when a better idea comes along.

superlinkx (approved):

PR #441, opened 2024-02-27
remove all whitespace from ag tags, revise migration version

irshadaj:
This is a follow up MR for review comments missed in https://github.com/SpecterOps/BloodHound/pull/437

Updated migration version to 5.8.0

Updated the migration to replace all whitespaces (tabs etc) as opposed just single spaces in the previous MR.

Manually tested the migration again by adding a new asset group with a tag containing single space, double space, and a tab:
before

Ran the migration SQL
update

Verified that all whitespace is removed
after

### DIFF
--- a/cmd/api/src/api/v2/agi.go
+++ b/cmd/api/src/api/v2/agi.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


superlinkx (approved):
Looks great! Appreciate getting this extended and tested

PR #442, opened 2024-02-27
fix: not including AZContains in inbound / outbound control traversal

mistahj67:
## Description

Added AZContains to the traversal strategy for both inbound and outbound control entities
Removed redundant `direction` argument from `FetchXEntityObjectControllers`
Removed unused `sortBy`  function in testing file

## Motivation and Context

Currently AZContains is ignored by the inbound / outbound controllers making it appear as though some Azure entities have 0 inbound control entities when that is not the case. This fixes that bug.

Removal of the direction from the FetchXEntityObjectControllers functions IMO helps prevent an issue where a developer may call an OutboundController with a direction.Inbound resulting in a bugged response. I didn't see any of those instances but felt it best to prevent that going forward.

## How Has This Been Tested?

Tested locally as well as added to integration tests.

image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/analysis/azure/azure_integration_test.go
+++ b/cmd/api/src/analysis/azure/azure_integration_test.go
@@ -21,7 +21,6 @@ package azure_test
 import (
 	"context"
 	schema "github.com/specterops/bloodhound/graphschema"
-	"sort"
 	"testing"
 
 	"github.com/specterops/bloodhound/graphschema/azure"
@@ -36,14 +35,6 @@ import (
 	"github.com/specterops/bloodhound/dawgs/query"
 	"github.com/specterops/bloodhound/src/test/integration"
 )
-
-func SortIDs(ids []graph.ID) []graph.ID {
-	sort.Slice(ids, func(i, j int) bool {
-		return ids[i].Int64() > ids[j].Int64()
-	})
-
-	return ids
-}
 
 func TestFetchEntityByObjectID(t *testing.T) {
 	testContext := integration.NewGraphTestContext(t, schema.DefaultGraphSchema())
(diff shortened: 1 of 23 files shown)


cmd/api/src/analysis/azure/azure_integration_test.go
mistahj67, line 1: Changes here is for AZContains traversal and not part of cleanup

packages/go/analysis/azure/filters.go
```diff
@@ -49,7 +49,7 @@ func FilterSecretReaders() graph.Criteria {
 }
 
 func FilterControlsRelationships() graph.Criteria {
-	return query.KindIn(query.Relationship(), append(azure.ControlRelationships(), azure.MemberOf)...)
+	return query.KindIn(query.Relationship(), append(azure.ControlRelationships(), azure.MemberOf, azure.Contains)...)
```
mistahj67, line 52: Changes here is for AZContains traversal and not part of cleanup

packages/go/analysis/azure/queries.go
```diff
@@ -319,33 +317,33 @@ func OutboundControlDescentFilter(ctx *ops.TraversalContext, segment *graph.Path
 	return shouldDescend
 }
 
-func OutboundControlPathFilter(ctx *ops.TraversalContext, segment *graph.PathSegment) bool {
-	return !segment.Edge.Kind.Is(azure.MemberOf)
+func OutboundControlPathFilter(_ *ops.TraversalContext, segment *graph.PathSegment) bool {
+	return !(segment.Edge.Kind.Is(azure.MemberOf) || segment.Edge.Kind.Is(azure.Contains))
```
mistahj67, line 321: Changes here is for AZContains traversal and not part of cleanup

packages/go/analysis/azure/queries.go
```diff
@@ -281,18 +281,16 @@ func FetchAppRoleAssignmentsTransitPaths(tx graph.Transaction, root *graph.Node,
 	})
 }
 
-func InboundControlDescentFilter(ctx *ops.TraversalContext, segment *graph.PathSegment) bool {
-	// The first traversal must be a control relationship for expansion
+func InboundControlDescentFilter(_ *ops.TraversalContext, segment *graph.PathSegment) bool {
```
mistahj67, line 284: Changes here is for AZContains traversal and not part of cleanup

superlinkx (approved):
Everything checks out from what I can see. Someone with more graph experience should give the logical change an extra pair of eyes just to be sure I didn't miss something subtle

packages/go/analysis/azure/queries.go
zinic: Minor nit here but `Kind.Is(...)` is variadic and the `||` is not needed.

zinic (changes requested):
Two minor changes I'd like to see plumbed, otherwise changeset reads good

zinic (approved):
LGTM, pull

PR #443, opened 2024-02-28
File upload UI updates

maffkipp:
## Description
This PR includes a few changes to the File Ingest UI:
- Improved error handling for failed file uploads -- displays the API error message to the user (if, for example, the JSON payload is malformed)
- Queries the new `/accepted-types` endpoint and uses the returned values for client file validation against the file's MIME type

There are also a few cleanup tasks included:
- Moved the main `FileIngest` component into the shared UI
- Added response types for `/file-upload` endpoints
- Moved all file upload queries into a separate hook
- found and fixed a bug caused by a race condition between API requests and react state updates

## Motivation and Context
- Update File Ingest to match changes currently being made to how files are processed in the API

## How Has This Been Tested?
- Updated existing tests to match new behavior (mostly just mocking the new endpoint) along with lots of manual testing

Screenshot 2024-02-28 at 11 56 47 AM

## Types of changes

## Checklist:


packages/javascript/bh-shared-ui/src/hooks/useFileIngest.tsx
```diff
@@ -0,0 +1,61 @@
...
+    contentType?: string;
+}) => {
+    return apiClient.uploadFileToIngestJob(jobId, fileContents, contentType).then((res) => res.data);
+};
+
+export const endFileIngestJob = ({ jobId }: { jobId: string }) =>
+    apiClient.endFileIngest(jobId).then((res) => res.data);
+
+export const fileUploadKeys = {
+    all: ['file-upload'] as const,
```
benwaples, line 27: Is the idea around having this as an array so that we can add to it later and itll automatically be spread to the functions below? If we dont think we will need to add more keys, then I think we could set this as a string and not have to spread the array below.

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
```diff
@@ -15,68 +15,200 @@
...
+                    } catch (error) {
+                        errorCount += 1;
+                    }
+                    setNewFileStatus(ingestFile.file.name, FileStatus.DONE);
+                }
+            }
+            await finishUpload(currentIngestJobId);
+
+            logFinishedIngestJob(errorCount);
+        } catch (error) {
```
benwaples, line 131: I have a couple of questions around error handling:

1. Just a little clarification for me: whats going on with the start/finishUpload requests? Is it there a transaction type of functionality happening on the API where if one file errors, then no file gets uploaded? Or is that have to do with the file streaming?
2. if this function does error (specifically with start/finishUpload), should we also roll the `fileUploadStep` back to `CONFIRMATION` or `ADD_FILES` in the catch block? Each individual file upload has its own error handler, so Im thinking if the start/finishUpload request errors then nothing gets uploaded?

benwaples (commented):
Testing this and works like a charm! I do have a couple smol comments but those are dealers choice all the way.

I am curious about the possible UX problem if/when things error, however it might just be that I need clarification on the topic and there is no problem :)

cmd/ui/src/views/FileIngest/FileIngest.tsx
```diff
@@ -1,220 +0,0 @@
-// Copyright 2023 Specter Ops, Inc.
```
urangel, line 1: Nice to see this movement of components into the shared UI!

packages/javascript/bh-shared-ui/src/components/FileDrop/FileDrop.tsx
```diff
@@ -54,6 +56,8 @@ const FileDrop: React.FC<{
     const handleMouseEnter = (_e: MouseEvent) => setHoverActive(true);
     const handleMouseLeave = (_e: MouseEvent) => setHoverActive(false);
 
+    const formatAcceptList = () => (accept && accept.length ? accept.join(',') : undefined);
```
urangel, line 59: This had me wondering what passing `undefined` to the input tag's accept attribute does. Do we have any plans to stop supporting `application/json` in the future? It would reduce some mental overhead for me to set a default value for the accept prop on this component of `application/json` and then only have to call `accept.join(',')` without the need to possibly set the accept list as `undefined`.

urangel (approved):
This is sort of a random but since it's contained within this code I wanted to make a soft request: could we make it so that the FileDrop dialog closes when clicking outside of the dialog/modal? The UX would be a little nicer imo and matches what our other dialogs do.  

The changeset looks good otherwise and I appreciate the duplication cleanup!

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
```diff
@@ -15,68 +15,200 @@
...
+        const filesHaveErrors = filesForIngest.filter((file) => file.errors).length > 0;
+        const filesAreUploading = filesForIngest.filter((file) => file.status === FileStatus.UPLOADING).length > 0;
+
+        if (filesHaveErrors || !filesForIngest.length) {
+            setSubmitDialogDisabled(true);
+        } else {
+            setSubmitDialogDisabled(false);
+        }
+
+        if (filesAreUploading) {
```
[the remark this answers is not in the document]
maffkipp: Done!

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
```diff
@@ -15,68 +15,200 @@
...
+    const listFileTypesForIngest = useListFileTypesForIngest();
+    const startFileIngestJob = useStartFileIngestJob();
+    const uploadFileToIngestJob = useUploadFileToIngestJob();
+    const endFileIngestJob = useEndFileIngestJob();
+
+    useEffect(() => {
+        const filesHaveErrors = filesForIngest.filter((file) => file.errors).length > 0;
+        const filesAreUploading = filesForIngest.filter((file) => file.status === FileStatus.UPLOADING).length > 0;
+
+        if (filesHaveErrors || !filesForIngest.length) {
```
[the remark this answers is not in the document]
maffkipp: Good callout on both of these checks, it reads way better now

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
```diff
@@ -15,68 +15,200 @@
...
+                    } else {
+                        addNotification(`File upload failed for ${ingestFile.file.name}`, 'IngestFileUploadFail');
+                        setUploadFailureError(ingestFile.file.name, 'Upload Failed');
+                    }
+                },
+            }
+        );
+    };
+
+    const truncateStatusMessage = (message: string, length: number) => {
```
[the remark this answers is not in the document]
maffkipp: Done! Moved that into some styles in the `FileValidationStatus` component

packages/javascript/bh-shared-ui/src/hooks/useFileIngest.tsx
maffkipp: I don't think this is necessary, I've removed it

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
maffkipp: Clarification: we send one request to start a job, one request each for each file we want to add to the job, and one request to close the job. If the job doesn't get closed, the job will eventually get marked as stale and discarded, but if a job gets closed and one or more of the files failed to ingest, my understanding is that the job will still run as expected.

I think it would be interesting to talk about what we want to do if the start or finish jobs fail. My thought would be instead of rolling them back a step, display an error and let them immediately retry, since a failure on the start or end requests is almost certainly an issue on our end. But there may be a case for bringing them back further in the process!

packages/javascript/bh-shared-ui/src/components/FileDrop/FileDrop.tsx
maffkipp: I tested this out, passing undefined to an html prop in JSX essentially makes the output html not have that prop. I definitely can see the argument for having a default file type -- my thought process was that if we do not return a list of file types from the API endpoint, I'd rather allow the user to upload invalid types than prevent them from uploading valid types. Since we do additional validation once the file hits our API anyways

urangel: I understand passing undefined is valid since trying this out but it took a bit of time to recognize that so my comment is more about clarity than validity. Your explanation about prevention of uploading valid file types by having a default of application/json makes sense since then one may not be able to pass a zip if something goes awry with the response for accepted types. Setting a default to an empty string still sits a little better with me since it avoids the ternary involved with undefined and will still allow all file types. Either way works though, feel free to leave as is and thanks for sharing your reasoning.

benwaples: Heyyo! Tagging on to this convo! 

I like the idea that if we don't get a list of accepted files from the API, then we default to allowing invalid types because the API is validating the content. 

I wanted to point out that In the off chance the api fails or doesnt have the list of file types, the file input would still be blocked from uploading because we check the file type before appending files to `filesForIngest`:
```ts
 const handleFileDrop = (files: FileList | null) => {
        if (files && files.length > 0) {
            const validatedFiles: FileForIngest[] = [...files].map((file) => {
                if (listFileTypesForIngest.data?.data.includes(file.type)) { // this check fails if accepted-types request fails
                    return { file, status: FileStatus.READY };
                } else {
                    return { file, errors: ['invalid file type'], status: FileStatus.READY };
                }
            });
            handleAppendFiles(validatedFiles);
        }
    };
```

I think we would need to update logic in `handleFileDrop` to allow uploading invalid types when we are missing the white list from the API!

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
benwaples: Thank yee very much for the clarification! This definitely makes me think the current error handling should totally work.

What you said about start/finish failing makes sense. I think we should ship this as is and if we run into issues with start/finish then we can re-address the error flow with more real world use-cases.

Thanks again for the detailed response :)

PR #444, opened 2024-02-28
feat: enhance signed request validation performance

ddlees:
## Description

- Stream processes the request signature of signed requests
- Payloads larger than 50MiB have their byte stream redirected to disk while signature is being computed and validated.
- Refactors function for generating a request signature to accept any hash.Hash factory function
- Implementation of SelfDestructingTemporaryFile
  - Automatically removes temp file if Closed
  - Automatically removes temp file if read to EOF

## Motivation and Context

As of version v5.6.0 the API must store the entire contents of a request’s body in memory to correctly validate the request signature for API token signing authentication. This leads to situations where large file uploads may cause out-of-memory conditions or otherwise harm API performance due to resource constraints.

## How Has This Been Tested?

- Many unit tests

## Types of changes

## Checklist:


cmd/api/src/api/signature.go
```diff
@@ -65,21 +68,81 @@ func tee(reader io.Reader, outA, outB io.Writer) error {
...
+		}
+		s.destroyed = true
+	}
+	return nil
+}
+
+func (s *SelfDestructingTempFile) Read(b []byte) (int, error) {
+	n, err := s.file.Read(b)
+	if err == io.EOF {
+		s.SelfDestruct()
```
zinic, line 104: This might have some weird side-effects if the file is read to `EOF` but `Seek` is then called. Thoughts?

zinic (dismissed):
Would like to see my comment addressed before giving this approval. The code reads well enough but there's a use-case flow I have a concern about.

cmd/api/src/api/signature.go
ddlees: When the file is read to EOF or explicitly closed the file is intended to be closed and deleted. This way, if a payload is "large" and we have to stream the request body to disk for validation, we can clean up after ourselves without having to coordinate any additional steps with the subsequent middleware/handler beyond what it "should" do normally.

So, if the SelfDestructingTempFile is read to EOF the underlying file descriptor should be released and a subsequent call to Seek I would expect to fail in the same way as any other closed os.File.

Is there some side-effect I'm not seeing?

cmd/api/src/api/auth_internal_test.go
```diff
@@ -109,3 +125,316 @@ func TestAuditLogin_UserNotFound(t *testing.T) {
 	mockDB.EXPECT().CreateAuditLog(testCtx, expectedAuditLog)
 	a.auditLogin(testCtx, commitId, model.User{}, loginRequest, string(model.AuditStatusFailure), ErrInvalidAuth)
 }
+
+func TestValidateRequestSignature(t *testing.T) {
```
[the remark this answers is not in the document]
ddlees: I have to use assert in a specific case so that the test doesn't bail before leaving a file on disk.

PR #445, opened 2024-02-28
Remove unused agi code

juggernot325:
## Description

This PR removes some database function (and the corresponding mock functions) that are no longer used across the code base for tidiness.

I also updated the `CreateAssetGroupSelector` API client function to use a `PUT` instead of a `POST` as the latter has been deprecated. While fixing this I noticed that the response body didn't match the struct we were trying to parse into, so I updated that to match as well.

## Motivation and Context

* Clean code is happy code

## How Has This Been Tested?

* Existing tests still pass

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/v2/apiclient/agi.go
+++ b/cmd/api/src/api/v2/apiclient/agi.go
@@ -1,17 +1,17 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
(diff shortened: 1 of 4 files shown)


mistahj67 (approved):
LGTM
Thanks for cleaning these up! 🧹 🧼

PR #449, opened 2024-02-29
Extracted ApiExplorer view to bh-shared-ui

codydbentley:
## Description

- moved ApiExplorer and related components to bh-shared-ui
- removed swagger-ui-react from BHCE

## Motivation and Context

This component was duplicated in BHCE and BHE, preparing for a docs change in the near future

## How Has This Been Tested?

manual testing and passing tests

## Types of changes

## Checklist:


packages/javascript/bh-shared-ui/package.json
```diff
@@ -28,14 +28,16 @@
...
         "luxon": "^3.3.0",
         "memoize-one": "^6.0.0",
         "notistack": "^2.0.5",
         "react-hook-form": "^7.45.1",
         "react-markdown": "^6.0.3",
         "react-query": "^3.39.3",
         "react-window": "^1.8.9",
-        "react-window-infinite-loader": "^1.0.9"
+        "react-window-infinite-loader": "^1.0.9",
+        "swagger-ui-react": "^4.15.2"
```
benwaples, line 40: Im getting unresolved dependency warnings while building `bh-shared-ui` from some of the packages imported into `OperationsEditionPlugin`. It looks like those modules are getting imported via sub-dependency's which is a little sketchy (this was happening before your changes btw, they just weren't caught via rollup). I think we aren't seeing errors because those dependencies exist via sub-dependencies right now so they resolve at run time!

I know Swaggers life in our app is limited, but It would be nice to not depend on those sub-dependencies existing for these docs to work in the meantime!

Can we add these to bh-shared-ui `package.json` dependencies:
- `immutable`, 
- `prop-types`, 
- and `react-immutable-proptypes`?

And in `rollup.config.js`, can we add to the external list: 
- `swagger-ui-react`, 
- `swagger-ui-react/swagger-ui.css`, 
- `prop-types`, 
- `immutable`, 
- `react-immutable-proptypes`, 
- and `lodash/toString`? 

That should resolve the dependency warning and also not rely on the subdependencys of other modules for `OperationsEditionPlugin`!

benwaples (approved):
Thanks for the clean up and reducing duplicated code:) 

I added a comment about a problem that was existing prior to these changes, however it would resolve an unlikely (but possible) issue at runtime!  

Still Approved because this works as is and this code will be removed soon regardless.

codydbentley (commented):
lgtm to me, send it

PR #450, opened 2024-02-29
fix: plumb context into client db methods

mistahj67:
## Description

Plumb context into gorm client db methods

## Motivation and Context

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/database/auth.go
+++ b/cmd/api/src/database/auth.go
@@ -63,7 +63,7 @@ func NewClientAuthToken(ownerID uuid.UUID, hmacMethod string) (model.AuthToken, 
 }
 
 type AuthContextInitializer interface {
-	InitContextFromToken(authToken model.AuthToken) (auth.Context, error)
+	InitContextFromToken(ctx context.Context, authToken model.AuthToken) (auth.Context, error)
 }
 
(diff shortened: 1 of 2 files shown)


juggernot325 (approved):

cmd/api/src/database/auth.go
```diff
@@ -74,7 +74,7 @@ func NewContextInitializer(db Database) AuthContextInitializer {
 	return contextInitializer{db: db}
 }
 
-func (s contextInitializer) InitContextFromToken(authToken model.AuthToken) (auth.Context, error) {
+func (s contextInitializer) InitContextFromToken(_ context.Context, authToken model.AuthToken) (auth.Context, error) {
```
zinic, line 77: Why not just plumb this into the `GetUser` call below?

zinic (changes requested):
Have a comment around `InitContextFromToken` that I'd like addressed before approval.

cmd/api/src/database/auth.go
zinic: Got it. Solid.

zinic (approved):
LGTM. Pull!

PR #452, opened 2024-02-29
feat: add bezier curve circle intersection utility, improve edge arrowhead positioning in small graphs

urangel:
## Description
A utility has been added to more accurately find an intersection between a bezier curve and a circle. This utility is used for rendering arrowheads in small graphs. The rendering of multiple edges with edges that have both forward and backward directions has been updated so that each edge has an arrowhead that points to the correct target node.

## Motivation and Context
It was noted that rendering multiple edges with different directions between two nodes caused an issue where one edge might have its arrowhead on the wrong edge which resulted in some edges having no arrowheads and some edges have two arrowheads which did not accurately convey the node relationships.

In addition, an attempt was made at improving rendering arrowheads on curved edges so that the direction and position of the arrowhead was better aligned with the edge. Unfortunately the calculations for this are relatively expensive and when rendering larger graphs the performance is noticeably worse with the improved arrowhead rendering. This rendering is implemented but turned off when there are over 250 edges in the rendered graph.

## How Has This Been Tested?
Some unit tests were added for fine approximations of a curve circle intersection. The original graph that was pointed out to have erroneous edge-arrowhead rendering was checked by uploading the associated dataset and manually ensuring that the edges and arrows were drawn in the correct places. 

Before:  
image
  
image
  
After:  
image

  
Screenshot 2024-02-29 at 9 37 00 AM

## Types of changes

## Checklist:

### DIFF
--- a/cmd/ui/src/rendering/shaders/edge.curved.vert.ts
+++ b/cmd/ui/src/rendering/shaders/edge.curved.vert.ts
@@ -50,7 +50,9 @@ void main() {
   vec2 compensationVector = vec2(-direction * unitNormal.y, direction * unitNormal.x) * (adaptedWebGLNodeRadius + adaptedWebGLArrowHeadLength);
 
   // Here is the proper position of the vertex
-  gl_Position = vec4((u_matrix * vec3(a_position + unitNormal * adaptedWebGLThickness + compensationVector, 1)).xy, 0, 1);
+  vec2 position = (u_matrix * vec3(a_position + unitNormal * adaptedWebGLThickness + compensationVector, 1)).xy;
+
+  gl_Position = vec4(position, 0, 1);
 
   v_thickness = webGLThickness / u_sqrtZoomRatio;
 
--- a/cmd/ui/src/rendering/utils/utils.ts
+++ b/cmd/ui/src/rendering/utils/utils.ts
@@ -13,6 +13,8 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
+import { Coordinates } from 'sigma/types';
+import { bezier } from './bezier';
 
 export const HIGHLIGHTED_LABEL_BACKGROUND_COLOR = '#FF6D66';
 export const HIGHLIGHTED_LABEL_FONT_COLOR = '#FFF';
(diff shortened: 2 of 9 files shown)


cmd/ui/src/views/Explore/utils.ts
```diff
@@ -182,11 +184,12 @@ export const initGraphEdges = (graph: MultiDirectedGraph, edges: GraphEdges) =>
                 size: 3,
                 type: 'arrow',
                 label: edge.label,
-                color: '#000000C0',
+                color: defaultEdgeColor,
                 groupPosition: 0,
                 groupSize: 1,
                 exploreGraphId: edge.exploreGraphId || key,
                 forceLabel: true,
+                needsPerformance: edges.length > 250,
```
elikmiller, line 192: This stands out to me - I'm curious how you came up with this heuristic?

urangel: This was a line in the sand relative to the max nodes we allow to render. It would be useful to do some benchmarks for determining some measurable acceptable threshold for rendering performance on varied hardware. This number would also ideally discriminate between the amount of edges drawn using bezier curves vs linear edges since the former requires more computation but that involves looping through the edges again to update the attribute which I chose to avoid.
 
The positions of the arrowheads have been fixed so that there are no edges without arrowheads and so that there are not two arrowheads on one edge that should only have one in both the "fine" and "fast" methods so pivoting here will only affect arrowhead positioning on an edge when node to node distance is small and edge group offset is large.

cmd/ui/src/rendering/shaders/edge.curved.vert.ts
```diff
@@ -50,7 +50,9 @@ void main() {
   vec2 compensationVector = vec2(-direction * unitNormal.y, direction * unitNormal.x) * (adaptedWebGLNodeRadius + adaptedWebGLArrowHeadLength);
 
   // Here is the proper position of the vertex
-  gl_Position = vec4((u_matrix * vec3(a_position + unitNormal * adaptedWebGLThickness + compensationVector, 1)).xy, 0, 1);
+  vec2 position = (u_matrix * vec3(a_position + unitNormal * adaptedWebGLThickness + compensationVector, 1)).xy;
+
+  gl_Position = vec4(position, 0, 1);
```
maffkipp, line 55: Why are we pulling out this `position` variable? It doesn't look like we are referencing it anywhere later

cmd/ui/src/rendering/utils/utils.ts
```diff
@@ -43,3 +45,78 @@ export const getNodeRadius = (highlighted: boolean, inverseSqrtZoomRatio: number
...
+    for (let index = 0; index < LUT.length; index++) {
+        const pointOnCurve = LUT[index];
+        const pointDistance = Math.abs(
+            bezier.getLineLength({ x: x, y: y }, { x: pointOnCurve.x, y: pointOnCurve.y }) - radius
+        );
+
+        // Realistically, there's only going to be an intersection if
+        // the distance to the circle center is already approximately
+        // the circle's radius.
+        if (prevDistance1 < distanceEpsilon && prevDistance2 > prevDistance1 && prevDistance1 < pointDistance) {
```
maffkipp, line 108: If nodes are past a certain distance from each other, the arrows are no longer drawn and the `curve circle intersection not found for drawing arrowhead` warning gets thrown in the console. It seems to be because at a certain distance, these conditions listed here are no longer being met and `findClosest` always returns `-1`.

My half-baked theory is that this is because our lookup tables are based on a constant resolution, and at a certain point there are no values in the lookup table close enough to the circles radius, but I am not 100% sure.

urangel: Yea I think you're right. Upping the resolution should work but needs more computation. I may be able to mitigate some of that by only evaluating in narrowed windows for `t` since we know that the nodes should be close to 0 or 1 values of `t` and will never be at 0.5 `t` for instance. I'll try this out

PR #467, opened 2024-03-06
fix: plumb context into audit logs db methods 

mistahj67:
## Description

Plumb context into gorm audit logs db methods

There is also some light cleanup in the listAuditLogs function

## Motivation and Context

## How Has This Been Tested?

## Types of changes

## Checklist:


cmd/api/src/utils/test/auditlogs.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package test
```
juggernot325, line 17: Thank you for moving this. I didn't like where it was, but also didn't want to spend time trying to find a better place to put it. This is a good place for it.

juggernot325 (approved):
Looks good! Thank you for the cleanup.

cmd/api/src/utils/test/auditlogs.go
mistahj67: Yep! I too wasn't sure where to put it so if you like the current directory, I like the current directory. 🙇

PR #470, opened 2024-03-06
Add config flag for disabling ingest

juggernot325:
## Description

This PR adds a new configuration flag for disabling ingest of files that have been uploaded. This is to fix some inconsistent behavior when multiple instances are sharing the same space.

## Motivation and Context

* Prevent a bug from occurring in development mode

## How Has This Been Tested?

* Set the flag to true and ensured that ingest does not complete
* Unset the flag and ensured that ingest functions as expected

## Types of changes

## Checklist:


cmd/api/src/daemons/datapipe/jobs.go
```diff
@@ -151,6 +151,11 @@ func (s *Daemon) processIngestTasks(ctx context.Context, ingestTasks model.Inges
 			return
 		}
 
+		if s.cfg.DisableIngest {
```
juggernot325, line 154: I'm looking for specific feedback on the placement of this check. I originally had it during the startup process and would just not start the data pipe daemon if the flag was set to `true`. One thing I didn't like about this implementation is that it wasn't clear why a file upload might not get past the ingesting step if the flag was unknowingly enabled. The benefit of setting it here is it will spam the logs with this message if `ingest_tasks` exist but the flag is enabled.

superlinkx: This seems like a perfectly reasonable compromise. This will only be an option devs will enable (it's not meant to be user consumable as there's no practical reason we know of for a user to want this behavior) so having some logging to remind the dev that ingest is disabled will prevent unnecessary digging for a non-existent bug

superlinkx (approved):
Appreciate getting this work in, will make dev a lot better for us

PR #471, opened 2024-03-06
fix: plumb ctx into ingest db methods 

mistahj67:
## Description

Plumb context into gorm ingest db methods

I also noticed that `GetUnfinishedIngestIDs` was an unused db methods related to the ingest so I removed it to keep things nice and clean.

## Motivation and Context

## How Has This Been Tested?

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/daemons/datapipe/datapipe.go
+++ b/cmd/api/src/daemons/datapipe/datapipe.go
@@ -133,8 +133,8 @@ func resetCache(cacher cache.Cache, cacheEnabled bool) {
 	}
 }
 
-func (s *Daemon) ingestAvailableTasks(ctx context.Context) {
-	if ingestTasks, err := s.db.GetAllIngestTasks(ctx); err != nil {
+func (s *Daemon) ingestAvailableTasks() {
+	if ingestTasks, err := s.db.GetAllIngestTasks(s.ctx); err != nil {
 		log.Errorf("Failed fetching available ingest tasks: %v", err)
 	} else {
 		s.processIngestTasks(s.ctx, ingestTasks)
@@ -150,16 +150,16 @@ func (s *Daemon) Start() {
 	defer datapipeLoopTimer.Stop()
 	defer pruningTicker.Stop()
 
-	s.clearOrphanedData(s.ctx)
+	s.clearOrphanedData()
 
 	for {
 		select {
 		case <-pruningTicker.C:
-			s.clearOrphanedData(s.ctx)
+			s.clearOrphanedData()
 
 		case <-datapipeLoopTimer.C:
 			// Ingest all available ingest tasks
-			s.ingestAvailableTasks(s.ctx)
+			s.ingestAvailableTasks()
 
 			// Manage time-out state progression for file upload jobs
 			fileupload.ProcessStaleFileUploadJobs(s.ctx, s.db)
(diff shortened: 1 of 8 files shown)


cmd/api/src/database/db.go
```diff
@@ -52,26 +54,25 @@ type Database interface {
...
-	GetUnfinishedIngestIDs() ([]int64, error)
+
+	// Ingest
+	ingest.IngestData
+	GetAllIngestTasks(ctx context.Context) (model.IngestTasks, error)
+	DeleteIngestTask(ctx context.Context, ingestTask model.IngestTask) error
+	GetIngestTasksForJob(ctx context.Context, jobID int64) (model.IngestTasks, error)
 
 	// Asset Groups
+	agi.AgiData
```
mistahj67, line 65: Missed cleaning this up in the asset-groups ctx PR but since these methods are defined elsewhere, having them on this interface is unnecessary.

cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -133,8 +133,8 @@ func resetCache(cacher cache.Cache, cacheEnabled bool) {
 	}
 }
 
-func (s *Daemon) ingestAvailableTasks() {
-	if ingestTasks, err := s.db.GetAllIngestTasks(); err != nil {
+func (s *Daemon) ingestAvailableTasks(ctx context.Context) {
+	if ingestTasks, err := s.db.GetAllIngestTasks(ctx); err != nil {
```
juggernot325, line 137: I'm probably just confused, but can you please provide details on why there's two different contexts in this function? This line uses the newly added parameter, but calls down below on 153, 158, and 162 are using the context on the receiver Daemon. Do we even need to pass in the parameter if the context is already available on the receiver?

cmd/api/src/daemons/datapipe/jobs.go
```diff
@@ -113,8 +113,8 @@ func ProcessIngestedFileUploadJobs(ctx context.Context, db database.Database) {
 }
 
 // clearFileTask removes a generic file upload task for ingested data.
-func (s *Daemon) clearFileTask(ingestTask model.IngestTask) {
-	if err := s.db.DeleteIngestTask(ingestTask); err != nil {
+func (s *Daemon) clearFileTask(ctx context.Context, ingestTask model.IngestTask) {
+	if err := s.db.DeleteIngestTask(ctx, ingestTask); err != nil {
```
juggernot325, line 117: Similar question here, why not use the context attached to the receiver?

cmd/api/src/database/db.go
juggernot325: I'm confused here, too, sorry. Your comment makes it sound like this line should be deleted, but this is showing as an addition on the diff.

juggernot325 (changes requested):
I have a few questions before marking this approved. I'm sure there's good reason, I just want to make sure I understand before signing off.

cmd/api/src/database/db.go
mistahj67: The 3 methods that are duplicated beneath it are encompassed within the agi.AgiData:
`GetAssetGroup`
`GetAllAssetGroups`
`CreateAssetGroupCollection`

cmd/api/src/daemons/datapipe/datapipe.go
mistahj67: No we do not and I completely missed that for these Daemon functions. Good catch, updating!

juggernot325 (approved):
Looks good! Thank you for making those changes.

PR #480, opened 2024-03-08
fix: plumb ctx into data quality db methods 

mistahj67:
## Description

Plumb context into gorm data quality db methods

**Removed circular dependency that leaked api structure into the data layer.
Added caching support for the count data type queries.**

## Motivation and Context

There existed a pretty gnarly circular dependency where we were were enforcing API response shape onto the data layer graph queries. This violates separation of concerns.

Further, there was no way for any count entity queries to be cached as the count data type was nested inside of the list data type query logic and exited early. 

## How Has This Been Tested?

## Types of changes

## Checklist:


cmd/api/src/queries/graph.go
```diff
@@ -773,67 +772,72 @@ func RunListQuery(ctx context.Context, db graph.Database, delegate any, node *gr
...
-		if fetchedResult, err := RunListQuery(ctx, s.Graph, params.ListDelegate, node, 0, 0); err != nil {
+		if fetchedResult, err := runEntityQuery(ctx, s.Graph, params.ListDelegate, node, 0, 0); err != nil {
 			return nil, err
 		} else {
 			result = fetchedResult
 		}
 	}
 
-	// Return early if this is just a count request
-	if params.RequestedType == model.DataTypeCount {
```
mistahj67, line 805: These requests were never cached previously because they always broke out early. By pulling out this to it's own function, we are able to now cache the DataTypeCount responses as well.

juggernot325 (approved):

PR #488, opened 2024-03-12
fix: edge arrowhead direction based on edge direction

urangel:
## Description
This changeset corrects edge arrowhead placement based on edge direction.

## Motivation and Context
Edge arrowheads were drawn targeting nodes incorrectly when there were different edge directions so this takes the edge direction into account in order to draw arrowheads to accurately display the relationships.

## Types of changes

## Checklist:


cmd/ui/src/rendering/programs/edge.curvedArrowHead.ts
```diff
@@ -150,9 +151,16 @@ export default class CurvedEdgeArrowHeadProgram extends AbstractEdgeProgram {
             return;
         }
 
+        let start, end: Coordinates;
+        if (data.direction === EdgeDirection.BACKWARDS) {
+            end = { x: sourceData.x, y: sourceData.y };
+            start = { x: targetData.x, y: targetData.y };
```
mistahj67, line 157: Small nit but by having both variables defined in the same order, I think it is a tad easier to see the difference here as well. 
```suggestion
            start = { x: targetData.x, y: targetData.y };
            end = { x: sourceData.x, y: sourceData.y };
```

urangel: ✔️

cmd/ui/src/rendering/programs/edge.curvedArrowHead.ts
```diff
@@ -150,9 +151,16 @@ export default class CurvedEdgeArrowHeadProgram extends AbstractEdgeProgram {
...
         }
 
+        let start, end: Coordinates;
+        if (data.direction === EdgeDirection.BACKWARDS) {
+            start = { x: targetData.x, y: targetData.y };
+            end = { x: sourceData.x, y: sourceData.y };
+        } else {
+            start = { x: sourceData.x, y: sourceData.y };
+            end = { x: targetData.x, y: targetData.y };
+        }
```
elikmiller, line 161: I'm curious why we have the notion of an edge direction? Isn't the direction of an edge simply based on its source and target nodes?

urangel: That is a good question and a good point. It could possibly be clearer but this is related to how the edges are first initially grouped for applying a different offset when there are multiple edges between two nodes as they are gathered together in a key value pair regardless of which the start or the end node are. It avoids creating overlapping edges. Here is where that happens: https://github.com/SpecterOps/BloodHound/blob/main/cmd/ui/src/views/Explore/utils.ts#L159

elikmiller:
Here's a cypher query I used to manually test this.

```
MATCH p=(n)-[r1]->(m)-[r2]->(n)
WHERE id(n) <> id(m)
RETURN p
LIMIT 5
```

cmd/ui/src/rendering/programs/edge.curvedArrowHead.ts
elikmiller: I see, regardless of the direction of an edge we want to be able to group them nicely. Makes sense!

PR #490, opened 2024-03-12
ESC4 Edge Composition

brandonshearin:
## Description

Composition for esc4!!

- added a new handler `PostADCSESC4Composition`
- the cypher statement that represents the composition was beefy, comprising 15 match statements, but there were some opportunities to reuse helpers and cut down on the beefiness
- mucho integration test coverage for all scenarios

graph rendering of the ecs4 edges present from the test file:
Screenshot 2024-03-13 at 6 50 12 AM

clicking on the edge from Alice reveals escalation 4 through her membership in SubDAs and Domain Admins with GenericAll on all templates
Screenshot 2024-03-13 at 6 51 48 AM

## Types of changes

-   [x ] New feature (non-breaking change which adds functionality)

## Checklist:

-   [x ] I have added and/or updated tests to cover my changes.
-   [ x] All new and existing tests passed.


JonasBK (approved):
Great work!

cmd/api/src/analysis/ad/adcs_integration_test.go
```diff
@@ -896,6 +896,225 @@ func TestADCSESC4(t *testing.T) {
...
+				require.Nil(t, err)
+
+				require.Equal(t, 7, len(composition.AllNodes()))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.Group11))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.Group0))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.CertTemplate1))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.EnterpriseCA))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.RootCA))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.NTAuthStore))
+				require.True(t, composition.AllNodes().Contains(harness.ESC4Template1.Domain))
```
urangel, line 962: Suggest assigning `composition.AllNodes()` to some variable so that we only call it once per result set

cmd/api/src/analysis/ad/adcs_integration_test.go
```diff
@@ -896,6 +896,235 @@ func TestADCSESC4(t *testing.T) {
...
 			}
 
+			return nil
+		})
+
+	})
+
+}
+
+func TestADCSESC4Composition(t *testing.T) {
```
urangel, line 906: Since we are setting up the same harnesses as the ESC4 post processing tests, could we instead move the test blocks that are in this function after the post processing assertions in the other? For example moving lines 941-967 to line 730. This could reduce testing overhead but wanted to ask in case there was some drawback to this.

packages/javascript/bh-shared-ui/src/components/HelpTexts/ADCSESC4/ADCSESC4.tsx
```diff
@@ -19,13 +19,15 @@ import WindowsAbuse from './WindowsAbuse';
 import LinuxAbuse from './LinuxAbuse';
 import Opsec from './Opsec';
 import References from './References';
+import Composition from '../ADCSESC6a/Composition';
```
urangel, line 22: This is neat. It seems like a good idea to make this component generally available instead of one for each edge. This seems good for now to avoid duplication as there is a lot to be desired in refactoring the edge info content which is out of scope for this changeset. I appreciate avoiding creating more cruft 🙌

cmd/api/src/analysis/ad/adcs_integration_test.go
brandonshearin: the reason i didn't do this is because the tests for PostADCSESC4() and for ESC4Composition() are several hundred lines, and even though they share some harness setup, this felt like the most natural seam between them.

btw all 5 of the composition tests use `ESC4Template1` harness, so theres only a one time setup cost/boilerplate. 
all of the composition tests use the template1 harness because template1 captures all of the different scenarios that esc4 manifests.

urangel: Yea it does make them more distinct and I had not noticed that they were all template1 in the composition tests. We aren't running into testing resource limits so this sounds good to me. Thank for the breakdown

packages/go/analysis/ad/esc4.go
```diff
@@ -206,3 +208,620 @@ func FetchPrincipalsWithWritePKIEnrollmentFlagOnCertTemplate(tx graph.Transactio
...
+			}),
+	}); err != nil {
+		return certTemplateSegments, certTemplates, err
+	}
+
+	return certTemplateSegments, certTemplates, nil
+}
+
+// composition: p2, p5, p8, p11, p15
+func findPathToDomainThroughEnterpriseCAsTrustedForNTAuth(
```
urangel, line 480: Both `enterpriseCATraversal` and  `ntAuthStoreToDomainTraversal` start with an outbound with no max and min depth from the start node which can be expensive. Breaking these into different steps might involve starting at the eca's that are gathered in `enterpriseCATraversal` and continuing from there for the `ntAuthStoreToDomainTraversal` so as not to repeat the same expensive traversal initially.

urangel (approved):
This took a while to look through so it is really impressive to see you implement this! I appreciate all the comments you left as it makes it easy to follow what is going on with all these moving parts

PR #494, opened 2024-03-13
fix: handle undefined input in explore search input

urangel:
## Description
This changeset adds handling for `undefined` input in the explore search. Additionally, the search reducer is moved to the shared-ui since it is used in both community and enterprise.

## Motivation and Context
We have a flow in BHE where search input state is filled in from information outside of the expected search results data format. The data that was passed in was incomplete and causing the app to reach the error boundary as there was no other explicit handling for the unexpectedly incomplete data. Empty defaults and checks for undefined input have been added for circumventing reaching an error state.

## How Has This Been Tested?
A unit test was added at the point where the runtime error was encountered.

## Types of changes

## Checklist:

### DIFF
--- a/packages/javascript/bh-shared-ui/src/hooks/useSearch/useSearch.tsx
+++ b/packages/javascript/bh-shared-ui/src/hooks/useSearch/useSearch.tsx
@@ -52,7 +52,7 @@ export const useSearch = (keyword: string, type: ActiveDirectoryNodeKind | Azure
 };
 
 export const getKeywordAndTypeValues = (
-    inputValue: string
+    inputValue = ''
 ): { keyword: string; type: ActiveDirectoryNodeKind | AzureNodeKind | undefined } => {
     const splitValue = inputValue.split(':');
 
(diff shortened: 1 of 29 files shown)


packages/javascript/bh-shared-ui/src/hooks/useSearch/useSearch.tsx
```diff
@@ -52,7 +52,7 @@ export const useSearch = (keyword: string, type: ActiveDirectoryNodeKind | Azure
 };
 
 export const getKeywordAndTypeValues = (
-    inputValue: string
+    inputValue = ''
```
urangel, line 55: This is the important part of the fix. Most everything else is noise from moving the search reducer to the shared ui

juggernot325: Thank you for pointing this out! This looks like a pretty straightforward fix after filtering out the package change noise.

juggernot325 (approved):
Looks good to me!

PR #496, opened 2024-03-15
Improved Status Messaging for File Ingest

irshadaj:
## Description

Zip upload can allow users to upload both JSON and non-JSON content, during this upload we do not communicate how many files were identified as JSON content and how many are non-JSON content. This results in a status complete with no context to what files were taken to the file ingest process. This will lead to confusion from the users following the upload if incomplete data exists.

We’d like to add Partially Complete which identifies when some of the files were identified as JSON and supply a count of files that were not ingested after zip unpack.

A full failure status must be shown when all files in an zip archive fail to ingest as json. 

## How Has This Been Tested?
Uploaded 3 different zip files that should return 3 different states

Success:
[success.zip](https://github.com/SpecterOps/BloodHound/files/14655466/success.zip)

Partial sucess:
[Archive.zip](https://github.com/SpecterOps/BloodHound/files/14655470/Archive.zip)

Failure:
[fail.zip](https://github.com/SpecterOps/BloodHound/files/14655471/fail.zip)

The expected end statuses were returned in the UI
ui

Also verified that the postgres columns are created as expected, and hold the right data as it pertains to the zip file and their expected return statuses

pg

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/daemons/datapipe/jobs.go
+++ b/cmd/api/src/daemons/datapipe/jobs.go
@@ -94,7 +94,7 @@ func CompleteAnalyzedFileUploadJobs(ctx context.Context, db database.Database) {
 			if job.FailedFiles > 0 {
 				if job.FailedFiles < job.TotalFiles {
 					status = model.JobStatusPartiallyComplete
-					message = fmt.Sprintf("%d Files failed to ingest as JSON Content", job.FailedFiles)
+					message = fmt.Sprintf("%d File(s) failed to ingest as JSON Content", job.FailedFiles)
 				} else {
 					status = model.JobStatusFailed
 					message = "All files failed to ingest as JSON Content"


cmd/api/src/daemons/datapipe/jobs.go
```diff
@@ -160,17 +176,25 @@ func (s *Daemon) preProcessIngestFile(path string, fileType model.FileType) ([]s
 	}
 }
 
-func (s *Daemon) processIngestFile(ctx context.Context, path string, fileType model.FileType) error {
+// processIngestFile reads the files at the path supplied, and returns the total number of files in the
+// archive, the number of files that failed to ingest as JSON, and an error
+func (s *Daemon) processIngestFile(ctx context.Context, path string, fileType model.FileType) (int, int, error) {
 	if paths, err := s.preProcessIngestFile(path, fileType); err != nil {
-		return err
+		return len(paths), 0, err
```
rvazarkar, line 183: Why return len(paths) here? If the zip file fails to process, do we even get anything back from this?

juggernot325 (approved):
Code looks good and behaved as expected when testing locally.

PR #509, opened 2024-03-19
fix: revert response shape for type=graph related entity queries

urangel:
## Description
The response shape for related entity queries with url param `type=graph` has been reverted in order to maintain the current api contract for this kind of request.

## Motivation and Context
This endpoint response changes shape based on the params that are passed in the request. While it is better and something that we should have in place to standardize these response shapes, normalizing this response resulted in a breaking change for viewing a node's relationships in the graph from the entity panel. The response was previously flat and was not sent back as pagination results and so the UI was looking in the wrong place for the data to render in the graph. Reverting this shape allows these graphs to render in the UI again.

## How Has This Been Tested?
Tests have been added asserting on the response shape for both graph type and list type queries. A local instance was spun up and graphs were able to be rendered from the entity panel drop downs in the UI.

https://github.com/SpecterOps/BloodHound/assets/16910931/11ae9d85-477e-4e55-af82-17af9b7f8d22

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/v2/ad_related_entity.go
+++ b/cmd/api/src/api/v2/ad_related_entity.go
@@ -26,6 +26,7 @@ import (
 	"github.com/specterops/bloodhound/dawgs/ops"
 	"github.com/specterops/bloodhound/graphschema/ad"
 	"github.com/specterops/bloodhound/src/api"
+	"github.com/specterops/bloodhound/src/model"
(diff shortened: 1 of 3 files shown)


elikmiller:
I've tested these changes locally in both BHCE and BHE. Looks like the fix works as expected - nice work!

elikmiller (approved):

PR #510, opened 2024-03-19
Resolve vulnerabilities in neo4j container

juggernot325:
## Description
This PR updates vulnerable libraries in our neo4j container used in development. I made the following fixes/changes to get down to only 1 "medium" vulnerability.

- Bump neo4j base image to 4.4.31
- Set the `allowlist` for procedures to only allow calls to `apoc.periodic.*` to mitigate CVE-2023-23926
- Updated `apoc` to 4.4.0.25
- Combined Docker commands to reduce the amount of layers needed for the container
- Added conditional command to add `memrec` to the neo4j config. This is configurable by setting the build arg `memconfig` to true. This is enabled by default in development docker compose deployments, but not testing.
- Updated the testing infrastructure to use BloodHound's custom neo4j container instead of the base image from Docker Hub

Note that this change means that we will now need to explicitly allow any procedures that we may want to run in the neo4j config file. 

## Motivation and Context
* Make our containers more secure

## How Has This Been Tested?
* Verified basic graph functionality in BloodHound still functions as expected
* Verified that calls to `apoc.periodic.*` procedures still function as expected
* Verified that calls to `apoc.import.graphml` do not function

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -71,6 +71,8 @@ services:
 
   graph-db:
     build:
+      args:
+        memconfig: true
       context: tools/docker-compose
       dockerfile: neo4j.Dockerfile
     environment:
--- a/docker-compose.testing.yml
+++ b/docker-compose.testing.yml
@@ -31,7 +31,9 @@ services:
 
   testgraph:
     restart: unless-stopped
-    image: neo4j:4.4.0
+    build:
+      context: tools/docker-compose
+      dockerfile: neo4j.Dockerfile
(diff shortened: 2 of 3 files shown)


juggernot325:
I noticed that the automated tests via GitHub actions do not use the same neo4j version as we do in development. I am going to do some research to see what all needs to be updated here and include it in this PR.

superlinkx (approved):

superlinkx (approved):

PR #511, opened 2024-03-19
fix: typo and wrong format for code snippit

benwaples:
## Description
Fix 2 typos and an incorrect formatting on code snippets

## Motivation and Context
BED-4271

## How Has This Been Tested?
Manually tested code snippet formatting and typos

## Types of changes

## Checklist:

### DIFF
--- a/packages/javascript/bh-shared-ui/src/components/HelpTexts/CodeController/CodeController.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/HelpTexts/CodeController/CodeController.tsx
@@ -93,7 +93,7 @@ function CodeController(props: PropsWithChildren<Props>) {
         const nextNonBlankLine = perLine.find((x, i) => i !== 0 && !!x.trim());
 
         const startingIndex = nextNonBlankLine?.split('').findIndex((x) => !!x.trim());
-        return perLine?.map((x, i) => (i === 0 ? x : x.slice(startingIndex))).join('\n');
+        return perLine?.map((x, i) => i === 0 ? x : x.slice(startingIndex)).join('\n');
     }, [children]);
 
     const handleCopy = async () => {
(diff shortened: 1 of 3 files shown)


packages/javascript/bh-shared-ui/src/components/HelpTexts/CodeController/CodeController.tsx
```diff
@@ -92,8 +92,8 @@ function CodeController(props: PropsWithChildren<Props>) {
         const perLine = (children?.toString() ?? '').split('\n');
         const nextNonBlankLine = perLine.find((x, i) => i !== 0 && !!x.trim());
 
-        const startingIndex = nextNonBlankLine?.split('').findIndex((x) => !!x.trim());
-        return perLine?.map((x) => x.slice(startingIndex)).join('\n');
+        const leftIndentationIndex = nextNonBlankLine?.split('').findIndex((x) => !!x.trim());
+        return perLine?.map((x, i) => i === 0 ? x : x.slice(leftIndentationIndex)).join('\n');
```
urangel, line 96: This is a good catch. Would this still function as we intend if the second line has some indentation? Something like this for example:
\`First line
------Indented second line
  
Non indented third line\`

elikmiller:
```
first line
    indented second line
non indented third line
```

urangel: ^Yes, like that. Thank you for the formatting help @elikmiller!

benwaples: I believe in this case the 3 and subsequent lines would be trimmed to the second line indentation.... I think a way to handle this could be by removing only whitespace up to that left indentation index. 

I can update with that!

urangel: This looks to provide enough guard rails and I appreciate the breakout from the inline map to a block of a few lines as it is easier to parse. There may be more edge cases with the formatting and the white space removal but since we want to move this content out of react components eventually this will hold us over.

Thanks for the update!

urangel (approved):

packages/javascript/bh-shared-ui/src/components/HelpTexts/CodeController/CodeController.tsx
benwaples: Thank you and great call out! I dont love all the edge cases this has to handle, but I agree it should suffice for now until we replace it.

PR #517, opened 2024-03-22
Move graph deletion to datapipe

rvazarkar:
## Description
Moves logic for graph deletion to the datapipe to prevent deadlocks

## Motivation and Context
The existing logic for clearing the DB was handled by the api handler in resources. However, if concurrent graph modifications were being performed in datapipe, this would lead to graph deadlocks. This MR moves the graph deletion logic into datapipe

https://specterops.atlassian.net/browse/BED-4286
## How Has This Been Tested?

Local testing

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/daemons/datapipe/datapipe.go
+++ b/cmd/api/src/daemons/datapipe/datapipe.go
@@ -40,6 +40,7 @@ const (
 
 type Tasker interface {
 	RequestAnalysis()
+	RequestDeletion()
 	GetStatus() model.DatapipeStatusWrapper
 }
 
@@ -49,6 +50,7 @@ type Daemon struct {
 	cache               cache.Cache
 	cfg                 config.Configuration
 	analysisRequested   *atomic.Bool
+	deletionRequested   *atomic.Bool
 	tickInterval        time.Duration
 	status              model.DatapipeStatusWrapper
 	ctx                 context.Context
@@ -66,6 +68,7 @@ func NewDaemon(ctx context.Context, cfg config.Configuration, connections bootst
 		cache:               cache,
 		cfg:                 cfg,
 		ctx:                 ctx,
+		deletionRequested:   &atomic.Bool{},
 		analysisRequested:   &atomic.Bool{},
 		orphanedFileSweeper: NewOrphanFileSweeper(NewOSFileOperations(), cfg.TempDirectory()),
 		tickInterval:        tickInterval,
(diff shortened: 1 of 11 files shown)


cmd/api/src/daemons/datapipe/delete.go
```diff
@@ -0,0 +1,36 @@
...
+			// deleting a node also deletes all of its edges due to a sql trigger
+			if err := batch.DeleteNode(nodeId); err != nil {
+				return err
+			}
+		}
+		return nil
+	}); err != nil {
+		return fmt.Errorf("error deleting all nodes: %w", err)
+	} else {
+		// if successful, handle audit log and kick off analysis
```
juggernot325, line 33: This comment is a little confusing. I would expect the code to perform these operations to live here based on the wording. Am I missing something?

justfile
```diff
@@ -99,7 +99,7 @@ update-favicon:
 
 # run go commands in the context of the api project
 go *ARGS:
-  @cd cmd/api/src && GODEBUG=cgocheck=2 go {{ARGS}}
+  @cd cmd/api/src && go {{ARGS}}
```
juggernot325, line 102: Was this change intentional? It seems out of scope from the original issue.

cmd/api/src/daemons/datapipe/delete.go
rvazarkar: leftover comment from copypaste

justfile
rvazarkar: It is out of scope, but that flag is deprecated and keeps just go * commands from running so I just fixed it while testing

juggernot325 (approved):
I did have a couple questions, but it's likely they won't require code changes. I'm preemptively approving under the assumption that nothing further will be needed. If that changes I'll revoke the approval.

cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -182,6 +206,25 @@ func (s *Daemon) Start(ctx context.Context) {
...
 }
 
+func (s *Daemon) deleteData() {
+	defer func() {
+		s.status.Update(model.DatapipeStatusIdle, false)
+		s.setDeletionRequested(false)
+		s.setAnalysisRequested(true)
+	}()
+	defer log.Measure(log.LevelInfo, "Purge Graph Data Completed")()
+	s.status.Update(model.DatapipeStatusPurging, false)
```
brandonshearin, line 216: should this be setting the purging state to true?

superlinkx: This method has a confusing signature if you're not looking at it in an IDE. The bool flag is to say whether to update analysis time, which we don't want to do because this isn't an analysis operation. Regardless of what you set that flag to, the status you give it will become the current status for the datapipe.

brandonshearin (approved):

PR #523, opened 2024-03-26
chore: gate graph clearing

urangel:
## Description
Puts graph clearing options behind a feature flag this enabled by default but not user updatable.

## Motivation and Context
Gives us levers to pull in the case of unexpected behavior.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/model/appcfg/flag.go
+++ b/cmd/api/src/model/appcfg/flag.go
@@ -90,8 +90,8 @@ func AvailableFlags() FeatureFlagSet {
 			Key:           FeatureClearGraphData,
 			Name:          "Clear Graph Data",
 			Description:   "Enables the ability to delete all nodes and edges from the graph database.",
-			Enabled:       false,
-			UserUpdatable: false,
+			Enabled:       true,
+			UserUpdatable: true,
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


juggernot325 (approved):

cmd/api/src/model/appcfg/flag.go
```diff
@@ -29,6 +30,7 @@ const (
 	FeatureReconciliation      = "reconciliation"
 	FeatureEntityPanelCaching  = "entity_panel_cache"
 	FeatureAdcs                = "adcs"
+	FeatureClearGraphData      = "clear-graph-data"
```
superlinkx, line 33: This should probably use underscores to match other flags

superlinkx (commented):
Should keep the flag name consistent with others, but otherwise this looks right to me. Let's just make sure we manually test enabling the flag to ensure it works as expected then get this in

cmd/api/src/model/appcfg/flag.go
urangel: You're right! This has been updated as suggested.

superlinkx (approved):

superlinkx (approved):

superlinkx (approved):

PR #525, opened 2024-03-26
feat: add migration for graph clear feature flag

urangel:
## Description
A migration was added for the clear graph data feature flag.

## Motivation and Context
Flags default to false when not in the database but we are enabling this by default so we need a migration for this flag.

## Types of changes

## Checklist:


cmd/api/src/database/migration/migrations/v5.8.0.sql
```diff
@@ -19,3 +19,5 @@ SET tag = REGEXP_REPLACE(tag, '\s', '', 'g');
 
 ALTER TABLE ingest_tasks
 ADD COLUMN IF NOT EXISTS file_type integer DEFAULT 0;
+
+UPDATE feature_flags SET enabled = true, user_updatable = false WHERE key = 'clear_graph_data';
```
superlinkx, line 23: This won't create the feature flag in the first place, we'll likely need to use a create if not exists here

PR #531, opened 2024-04-01
fix: add missing node types to data quality stats and display

urangel:
## Description
Adds some missing node kinds to the data quality page display for Azure.

## Motivation and Context
Some node kinds were missing from both our quality stats table in pg and also from the data quality page counts.

image

## Types of changes

## Checklist:


packages/javascript/bh-shared-ui/src/views/DataQuality/DomainInfo.tsx
```diff
@@ -35,6 +33,33 @@ const useStyles = makeStyles({
     },
 });
 
+export const DomainMap = {
```
maffkipp, line 36: you might be able to simplify this to a map of `displayText` values to `ActiveDirectoryNodeKind` keys if you move the `<NodeIcon />` component out into your `Object.map()` call

packages/javascript/bh-shared-ui/src/views/DataQuality/TenantInfo.tsx
```diff
@@ -35,6 +33,39 @@ const useStyles = makeStyles({
     },
 });
 
+export const TenantMap = {
```
maffkipp, line 36: possibly same refactor here as well

packages/javascript/bh-shared-ui/src/views/DataQuality/DomainInfo.tsx
urangel: This works well for AD kinds but there are a few differences in azure that would need special handling:

-  The az prefix on the kind (`users` column for azure stats is an `AZUser` kind)
-  Ensuring proper pluralization like for registry (`container_registries` azure stats column is an `AZContainerRegistry` but the plural of registry is registries instead of just appending an 's' like with the other kinds)
- Azure columns are snake cased while AD columns are not (`certtemplates` is not `cert_templates` while in azure we have `key_vaults` and not `keyvaults`). 

It seemed easier to have similar maps between Azure and AD than to handle azure mapping with special conditions.

maffkipp: this makes sense to me, thanks for the clarification!

maffkipp (approved):

PR #539, opened 2024-04-04
handle nil pointer for saml user session validation

irshadaj:
## Description

When a `User` is updated to be SAML, its `AuthSecret` is set to nil. Thereafter in `ValidateSessions`, we call a method on the nil `AuthSecret` leading to a panic in the logs. This MR adds logic to handle that case gracefully.

I also updated the Auth Middleware to return the actual error upon `ValidateSession` failure, as we were just discarding this information and returning a blanket failure statement before.

## How Has This Been Tested?
Manual testing

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/middleware/auth.go
+++ b/cmd/api/src/api/middleware/auth.go
@@ -67,7 +67,7 @@ func AuthMiddleware(authenticator api.Authenticator) mux.MiddlewareFunc {
 				switch authScheme {
 				case api.AuthorizationSchemeBearer:
 					if userAuth, err := authenticator.ValidateSession(request.Context(), schemeParameter); err != nil {
-						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusUnauthorized, err.Error(), request), response)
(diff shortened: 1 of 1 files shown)


cmd/api/src/api/middleware/auth.go
```diff
@@ -67,7 +67,7 @@ func AuthMiddleware(authenticator api.Authenticator) mux.MiddlewareFunc {
 				switch authScheme {
 				case api.AuthorizationSchemeBearer:
 					if userAuth, err := authenticator.ValidateSession(request.Context(), schemeParameter); err != nil {
-						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusUnauthorized, api.ErrorResponseDetailsAuthenticationInvalid, request), response)
+						api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusUnauthorized, err.Error(), request), response)
```
ddlees, line 70: We should revert this change. Any time we return a 401 Unauthorized we have to be careful so not to disclose any information that could be potentially leveraged by a bad actor.

ddlees (changes requested):

ddlees (approved):

PR #540, opened 2024-04-05
return errors during file ingest

irshadaj:
## Description

File ingest errors were initially being just logged (and not returned up the stack) because we didn't want a single file ingest failure to fail all of ingestion. However, that means that even when some files fail to ingest, the job will display as  having completed successfully, with no data showing up where expected and only the logs would reflect these failures. This is confusing to the user, and now that we have partial ingest job completion plumbed, we should be returning these errors so the status of ingest is more accurately set to partial completion or full failure.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/daemons/datapipe/ingest.go
+++ b/cmd/api/src/daemons/datapipe/ingest.go
@@ -64,9 +64,13 @@ func IngestGroupData(batch graph.Batch, converted ConvertedGroupData) error {
 
 	if err := IngestNodes(batch, ad.Entity, converted.NodeProps); err != nil {
 		errs.Add(err)
-	} else if err = IngestRelationships(batch, ad.Entity, converted.RelProps); err != nil {
+	}
+
+	if err := IngestRelationships(batch, ad.Entity, converted.RelProps); err != nil {
 		errs.Add(err)
-	} else if err = IngestDNRelationships(batch, converted.DistinguishedNameProps); err != nil {
+	}
+
+	if err := IngestDNRelationships(batch, converted.DistinguishedNameProps); err != nil {
 		errs.Add(err)
 	}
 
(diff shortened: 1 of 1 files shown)


rvazarkar:
Code wise, everything looks good, but could we get a test or two added pretty please?

cmd/api/src/daemons/datapipe/decoders.go
```diff
@@ -56,17 +57,22 @@ func decodeBasicData[T any](batch graph.Batch, reader io.ReadSeeker, conversionF
...
 
 	if count > 0 {
-		IngestBasicData(batch, convertedData)
+		if err = IngestBasicData(batch, convertedData); err != nil {
+			errs = append(errs, err)
+		}
 	}
 
-	return nil
+	return errors.Join(errs...)
```
computator, line 75: `errors.Join()` is great conceptually, but I think we have been avoiding using it because it does not play nice with the logging due to newlines? We have a custom compound error structure we've been using instead sometimes

cmd/api/src/daemons/datapipe/ingest.go
```diff
@@ -44,21 +45,44 @@ func ReadFileForIngest(batch graph.Batch, reader io.ReadSeeker) error {
...
-	IngestNodes(batch, ad.Entity, converted.NodeProps)
-	IngestRelationships(batch, ad.Entity, converted.RelProps)
+func IngestBasicData(batch graph.Batch, converted ConvertedData) error {
+	errs := []error{}
+
+	if err := IngestNodes(batch, ad.Entity, converted.NodeProps); err != nil {
+		errs = append(errs, err)
+	} else if err = IngestRelationships(batch, ad.Entity, converted.RelProps); err != nil {
+		errs = append(errs, err)
+	}
```
computator, line 55: Previously these would continue on to the next step if there was an error. Now they will not due to the else-if-tree. Is this intentional? If not, these would need to be split into individual if statements.

computator (commented):
Although I only commented on one instance of each, the potential issues noted apply to similar instances throughout the PR.

cmd/api/src/daemons/datapipe/ingest.go
```diff
@@ -44,21 +45,46 @@ func ReadFileForIngest(batch graph.Batch, reader io.ReadSeeker) error {
...
+func IngestGroupData(batch graph.Batch, converted ConvertedGroupData) error {
+	errs := util.NewErrorCollector()
+
+	if err := IngestNodes(batch, ad.Entity, converted.NodeProps); err != nil {
+		errs.Add(err)
+	} else if err = IngestRelationships(batch, ad.Entity, converted.RelProps); err != nil {
+		errs.Add(err)
+	} else if err = IngestDNRelationships(batch, converted.DistinguishedNameProps); err != nil {
+		errs.Add(err)
+	}
```
computator, line 71: I see you fixed one of the chained if blocks, but there's multiple others still 🙂

computator (changes requested):

computator (approved):

PR #541, opened 2024-04-05
Cypher to PgSQL Generic Translation

zinic:
## Description

openCypher to PgSQL translation machinery

## Motivation and Context

This changeset contains a syntax model for the PostgreSQL SQL dialect.

It also contains a translation implementation to take openCpyher input and output valid PostgreSQL SQL. This model is not intended to be a complete implementation of all available SQL dialect features but rather the subset of the dialect required to perform openCypher to pgsql translation.

## How Has This Been Tested?

* Unit testing
* 100+ translation cases
* Using the `pg` driver also results in integration test suite success

## Types of changes

## Checklist:


packages/go/cypher/backend/pgsql/format.go
```diff
@@ -240,207 +240,175 @@ func (s *Emitter) writeWhere(writer io.Writer, whereClause *model.Where) error {
...
+const (
+	defaultNodeBinding = "n"
+	defaultEdgeBinding = "r"
+)
 
-					if _, err := WriteStrings(writer, ".start_id or "); err != nil {
-						return err
-					}
+// ()-[]
+// ()<-[]
```
ddlees, line 274: Without much of an explanation these comments just look out of place. Could you elaborate on how this relates to the method below or remove it?

packages/go/cypher/backend/pgsql/format.go
```diff
@@ -240,207 +240,175 @@ func (s *Emitter) writeWhere(writer io.Writer, whereClause *model.Where) error {
...
-						return err
-					}
+	return nil
+}
 
-					if _, err := WriteStrings(writer, ".id"); err != nil {
-						return err
-					}
+// []->()
+// []-()
```
ddlees, line 318: Without much of an explanation these comments just look out of place. Could you elaborate on how this relates to the method below or remove it?

packages/go/cypher/model/pgsql/fold/extraction.go
```diff
@@ -0,0 +1,120 @@
...
+		conjoinedConstraintsByKey: map[string]*pgsql.Tree{},
+	}
+
+	if err := walk.PgSQL(expression, extractor); err != nil {
+		return nil, err
+	}
+
+	//for key, tree := range extractor.conjoinedConstraintsByKey {
+	//	visualization.MustWritePUML(tree.Root(), fmt.Sprintf("/home/zinic/digraphs/%s.puml", key))
+	//}
```
ddlees, line 117: Please remove or add meaningful comment as to why this is omitted from the function.

packages/go/cypher/model/pgsql/fold/extraction_test.go
```diff
@@ -0,0 +1,70 @@
...
+			continue
+		}
+
+		regularQuery, err := frontend.ParseCypher(frontend.NewContext(), testCase.Cypher)
+		require.Nil(t, err)
+
+		sqlAST, err := translate.TranslateCypherExpression(regularQuery.SingleQuery.SinglePartQuery.ReadingClauses[0].Match.Where.Expressions[0])
+		require.Nil(t, err)
+
+		visualization.MustWritePUML(sqlAST, "/home/zinic/digraphs/graph.puml")
```
ddlees, line 48: Boo... don't hardcode paths.

packages/go/cypher/model/pgsql/slices.go
```diff
@@ -0,0 +1,34 @@
...
+		if tTyped, isTType := any(fs[idx]).(T); !isTType {
+			var emptyT T
+			return nil, fmt.Errorf("slice type %T does not convert to %T", fs[idx], emptyT)
+		} else {
+			ts[idx] = tTyped
+		}
+	}
+
+	return ts, nil
+}
```
ddlees, line 18: We could probably simplify wherever this is being used by instead using `slicesext.Map[F, T any](fn func(F) T, fs []F) []T` where `fn` is a transformation function that leverages the compiler for type coercion.

ddlees (changes requested):

packages/go/cypher/model/pgsql/nodes.go
```diff
@@ -0,0 +1,39 @@
+package pgsql
+
+type StringLike interface {
+	String() string
+}
```
ddlees, line 5: This is functional equivalent to `fmt.Stringer`. Consider using that instead.

zinic: Accepted, will reauthor in next push

packages/go/cypher/model/pgsql/fold/extraction_test.go
zinic: Cheaper than taking time away from the problem I was focused on at the time. The visualizer needs some love in making it into a proper tool anyway but will address this in more detail once translation is in a better place.

packages/go/cypher/backend/pgsql/format.go
zinic: These are just old cruft and part of code paths that will eventually be deprecated so expect removal in a future changeset

packages/go/cypher/model/pgsql/fold/extraction.go
zinic: Removed these code paths

packages/go/cypher/model/pgsql/translate/translation.go
```diff
@@ -0,0 +1,794 @@
...
+		identifierTracker:   NewIdentifierTracker(),
+		identifierGenerator: NewIdentifierGenerator(),
+	}
+}
+
+func (s *Translator) ConstraintTracker() *ConstraintTracker {
+	return s.whereTranslation.constraints
+}
+
+func cypherVariableSymbol(expression cypher.Expression) (string, bool, error) {
```
brandonshearin, line 189: this function signature is a little opaque.  it looks like the functions purpose is to extract a binding from a pattern part or projection item?  maybe changing the name to "extractBindingFromExpression" would be clearer?

similar thoughts for the return signature.  it would be helpful to name the return parameters or just leave a comment on the method

packages/go/cypher/model/pgsql/translate/translation_cases/node.sql
```diff
@@ -0,0 +1,45 @@
...
+from n1;
+
+-- case: match (s), (e) where s.name = '1234' return s
+with n1 as (select n1.* from node n1 where n1.properties -> 'name' = '1234'),
+     n2 as (select n2.* from node n2)
+select (n1.id, n1.kind_ids, n1.properties)::nodecomposite as s
+from n1;
+
+-- case: match (s:A), (e:B) where s.name = e.name return s, e
+with n1 as (select n1.* from node n1 where n1.kind_ids operator (pg_catalog.&&) array []::int2[]),
```
brandonshearin, line 18: i haven't looked into the translator implementation for node labels yet, but wanted to make sure that you expect the binary expression in this where clause:

`n1.kind_ids operator (pg_catalog.&&) array []::int2[]`

to be an empty array?  logically this seems like its probably fine, because the node label "A" isn't an actual BH kind and im guessing whatever translates this clause under the hood is returning an empty array accordingly.

packages/go/cypher/model/pgsql/slices.go
zinic: Implemented with some minor tweaks

packages/go/cypher/model/pgsql/translate/translation.go
zinic: Good call out, will iterate

packages/go/cypher/model/pgsql/translate/translation_cases/node.sql
zinic: Kind mappings require a live database so I haven't gotten around to plugging that particular interface into the formatter. Leaving this unresolved to review once the KindMapper type is hoisted into the formatter function logic.

packages/go/cypher/model/pgsql/nodes.go
zinic: Refactored out

packages/go/cypher/model/pgsql/translate/tracking.go
```diff
@@ -0,0 +1,304 @@
...
+	default:
+		return "", fmt.Errorf("identifier with data type %s does not have a prefix case", dataType)
+	}
+}
+
+func NewIdentifierGenerator() IdentifierGenerator {
+	return IdentifierGenerator{}
+}
+
+type Constraint struct {
```
brandonshearin, line 39: trying to wrap my mind around these wonderful ancillary structures. What is a constraint?  where can a constraint appear in a cypher AST?  what gets stuffed into the dependencies of a constraint?  please leave some inline some comments in the code that illuminate some of these questions!

packages/go/cypher/model/pgsql/translate/tracking.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+	}
+
+	s.Constraints = append(s.Constraints, &Constraint{
+		Dependencies: dependencies,
+		Expression:   constraintExpression,
+	})
+}
+
+type TrackedIdentifier struct {
```
brandonshearin, line 104: i have the same request for this structure as i made for the Constraint structure.  it looks like TrackedIdentifier is a wrapper around a good ol pgsql.Identifier AST.  some comments around the usage of this structure would be grand.  when does an identifier become tracked? what does it mean to be tracked?  what parts of the translation depend on this structure?

packages/go/cypher/model/pgsql/translate/tracking.go
```diff
@@ -0,0 +1,304 @@
...
+
+	for _, constraint := range s.Constraints {
+		topLevelExpression = OptionalAnd(topLevelExpression, constraint.Expression)
+	}
+
+	s.Constraints = nil
+	return topLevelExpression
+}
+
+func (s *ConstraintTracker) Consume(available pgsql.IdentifierSet) (pgsql.IdentifierSet, pgsql.Expression) {
```
brandonshearin, line 63: what does it mean to consume()/constrain()?

packages/go/cypher/model/pgsql/translate/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Intent
+
+The GQIR is a simplification of the openCypher query model. The model is designed to allow for a single-walk translation
+of a given openCypher query. The GQIR is used to normalize and reorganize query components to serialize them for
+encoding into different end-query models.
+
+### Pattern Rewriting
+
+The example below demonstrates reading out label constraints and reorganizing them to simplify translation to PgSQL:
```
brandonshearin, line 13: its gonna be really nice when this readme is fleshed out.  is it possible to mention here which components of the Translator manage each of these responsibilities?  ie. what structure manages pattern rewriting,identifier rebinding, etc.

packages/go/cypher/model/pgsql/translate/expression.go
```diff
@@ -0,0 +1,230 @@
...
+		return map[pgsql.Identifier]struct{}{
+			typedOperand.Root(): {},
+		}
+
+	default:
+		return map[pgsql.Identifier]struct{}{}
+	}
+}
+
+type ExpressionTreeTranslator struct {
```
brandonshearin, line 80: could you provide a minimal example of how to use this structure?  perhaps as a test case in a new expression_test.go file?  i can extend off of your initial example with more coverage, but im curious how this machinery is intended to be used.  

seeing it work in isolation would be illuminating, but also not sure if its possible to reason about in isolation. what do you think?

superlinkx:
Waiting on panic removals, but otherwise the changeset seems good

superlinkx (commented):
LGTM

superlinkx (approved):

PR #556, opened 2024-04-16
Don't filter T0 assets from DCSync edges

juggernot325:
## Description

This PR prevents T0 assets from being incorrectly excluded from creating `DCSync` edges. The filtering was unintended and has been removed.

I also fixed an unrelated log message that I noticed had the incorrect log level. This should help prevent it from getting lost when parsing logs.

## Motivation and Context

* Ensure `DCSync` edges are present for T0 assets

## How Has This Been Tested?

* Updated the associated test for the new logic and ensured it was passing

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/analysis/analysis.go
+++ b/packages/go/analysis/analysis.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"slices"
-	"strings"
 
 	"github.com/specterops/bloodhound/dawgs/graph"
 	"github.com/specterops/bloodhound/dawgs/ops"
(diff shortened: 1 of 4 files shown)


packages/go/analysis/analysis.go
```diff
@@ -240,10 +239,10 @@ func GetLAPSSyncers(tx graph.Transaction, domain *graph.Node) ([]*graph.Node, er
 	}
 }
 
-func GetDCSyncers(tx graph.Transaction, domain *graph.Node, filterTierZero bool) ([]*graph.Node, error) {
```
juggernot325, line 243: This function is only called in one place so I opted for removing the parameter and filter logic instead of just passing false. I can add it back in and use the parameter if that is preferred, but I figured removing the unused code was cleaner.

packages/go/analysis/analysis.go
```diff
@@ -240,10 +239,10 @@ func GetLAPSSyncers(tx graph.Transaction, domain *graph.Node) ([]*graph.Node, er
...
+func GetDCSyncers(tx graph.Transaction, domain *graph.Node) ([]*graph.Node, error) {
 	var (
-		getChangesQuery    = fromEntityToEntityWithRelationshipKind(tx, domain, ad.GetChanges, filterTierZero)
```
mistahj67, line 245: It looks like the `fromEntityToEntityWithRelationshipKind` also has no need for filterTierZero as every use case after your change issues `false`. What do you think of cleaning that up as well?

packages/go/analysis/analysis.go
mistahj67: My vote is always to clean! 🧼

packages/go/analysis/analysis.go
juggernot325: Good catch! I removed that parameter and logic in the latest push.

mistahj67 (approved):
Tested locally and code LGTM!

Thanks for cleanup!

PR #571, opened 2024-04-22
BED-4331: Support Configuration of Cypher Controls

superlinkx:
## Description

Plumbs environment variables (and makes some naming updates to existing controls) to support the upcoming Cypher quality controls. This changes `disable_cypher_qc` to `disable_cypher_complexity_limit` (as it was a confusing name), changes `traversal_memory_limit` to `graph_query_memory_limit` (as it will cover more than just traversals), and adds `enable_cypher_mutation` to allow enabling mutation-type cypher queries.

This also includes a new user permission called "GraphDBMutate", available on Admins and Power Users.

Some work was done to begin plumbing actual mutation support already, but is incomplete. The environment variable can be enabled and it should allow for graph mutations if you have the new permission, but it has no guardrails and is still under development.

## Motivation and Context

Allow users to opt-out of cypher guardrails with environment variables

## How Has This Been Tested?

- Tests were added to RawCypherQuery, PrepareCypherQuery, and e2e test enhancements for CypherSearch
- Some manual testing was done to validate our initial mutation plumbing, allowing us to see if mutations would actually show up in the graph

## Types of changes

## Checklist:


examples/docker-compose/.env.example
```diff
@@ -28,8 +28,14 @@ NEO4J_WEB_PORT=7474
 
 # BloodHound Configuration Parameters
 
-# Uncomment if you want to disable cypher quality control
-#bhe_disable_cypher_qc=false
+# Uncomment if you want to disable cypher complexity limit
+#bhe_disable_cypher_complexity_limit=true
```
brandonshearin, line 32: can the comment here be more descriptive?  as a user im not sure what im disabling when i set this flag to true. 

i also noticed that there is another .env.example file at the root of the bhce project.  do these flags need to be specified there as well?

packages/go/cypher/frontend/query.go
```diff
@@ -363,6 +363,7 @@ func (s *SinglePartQueryVisitor) EnterOC_UpdatingClause(ctx *parser.OC_UpdatingC
 }
 
 func (s *SinglePartQueryVisitor) ExitOC_UpdatingClause(ctx *parser.OC_UpdatingClauseContext) {
+	s.ctx.HasMutation = true
```
brandonshearin, line 366: just curious, is there a reason why hasMutation is set to true on exit as opposed to on enter?

mistahj67: Exit was where the clause is actually added to the query struct so it felt more appropriate.

Also the filtering system relies on overriding the Enter variants and that caused some pain points so this ended up being the spot we tossed the flag. Not opposed to another area if one is a better fit / anyone has strong opinions though!

examples/docker-compose/.env.example
superlinkx: That .env.example is available for dev use, but I'm not sure if anyone typically uses it since you can modify your own config file instead. Probably not necessary to update with these flags as a result.

As for being more descriptive, I'm fairly confident that the users that want to disable this will know what is meant. They would have gotten errors saying the complexity was too high, and this gives them a flag to disable that if they want to.

We're happy to hear suggestions for a better comment, but we'll likely want to refine documentation of our environment variables at some point separately for more clarity

cmd/api/src/api/v2/cypher_search.go
```diff
@@ -30,14 +33,28 @@ type CypherSearch struct {
...
+			request.Context(),
+			api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response,
+		)
+	} else if preparedQuery.HasMutation && !s.Authorizer.AllowsPermission(authCtx, auth.Permissions().GraphDBMutate) {
+		s.Authorizer.AuditLogUnauthorizedAccess(request)
+		api.WriteErrorResponse(
+			request.Context(),
+			api.BuildErrorResponse(http.StatusForbidden, "Permission denied: User may not modify the graph.", request), response,
+		)
+	} else if graphResponse, err := s.GraphQuery.RawCypherSearch(request.Context(), preparedQuery, payload.IncludeProperties); err != nil {
```
definitelynotagoblin, line 57: Is RawCypherQuery not just opening a read transaction by default?  If not, is that something we want to address?

superlinkx: The definition of RawCypherSearch determines whether to use a Read or Write transaction based on preparedQuery.HasMutation. Previously, it would just pass directly to a Read transaction, but we now want to support both non-mutation and mutation queries

cmd/api/src/api/v2/cypher_search_integration_test.go
```diff
@@ -85,3 +86,47 @@ func Test_CypherSearch(t *testing.T) {
...
+	harness := lab.NewHarness()
+	customCfgFixture := fixtures.NewCustomConfigFixture(customCfg)
+	lab.Pack(harness, customCfgFixture)
+	customApiFixture := fixtures.NewCustomApiFixture(customCfgFixture)
+	lab.Pack(harness, customApiFixture)
+	adminApiClientFixture := fixtures.NewAdminApiClientFixture(customCfgFixture, customApiFixture)
+	lab.Pack(harness, adminApiClientFixture)
+
+	lab.NewSpec(t, harness).Run(
+		lab.TestCase("errors on mutations with enable cypher_mutations true", func(assert *require.Assertions, harness *lab.Harness) {
```
urangel, line 124: It looks like this description should say something like, "does not throw errors with cypher_mutations enabled" since currently it reads like it is supposed to error even though we are asserting a nil err.

cmd/api/src/auth/model.go
```diff
@@ -90,43 +90,52 @@ type AuditLogger interface {
...
+	auditLogger    AuditLogger
+	getPermissions GetPermissionsFunc
 }
 
-type authorizer struct {
-	auditLogger AuditLogger
+func NewCustomAuthorizer(auditLogger AuditLogger, getPermissionsFn GetPermissionsFunc) Authorizer {
+	if getPermissionsFn == nil {
+		getPermissionsFn = getPermissions
+	}
```
urangel, line 101: Is this an expected scenario? I only see it used with a non nil func passed in so was curious

packages/go/dawgs/ops/ops.go
```diff
@@ -195,11 +195,12 @@ func FetchPathSetByQuery(tx graph.Transaction, query string) (graph.PathSet, err
...
-				pathSetSize := size.Of(pathSet)
-
-				if currentPathSize > tx.TraversalMemoryLimit() || pathSetSize > tx.TraversalMemoryLimit() {
-					return pathSet, fmt.Errorf("%s - Limit: %.2f MB", "query required more memory than allowed", tx.TraversalMemoryLimit().Mebibytes())
+				if tx.GraphQueryMemoryLimit() > 0 {
+					currentPathSize := size.OfSlice(currentPath.Edges) + size.OfSlice(currentPath.Nodes)
+					pathSetSize := size.Of(pathSet)
+					if currentPathSize > tx.GraphQueryMemoryLimit() || pathSetSize > tx.GraphQueryMemoryLimit() {
+						return pathSet, fmt.Errorf("%s - Limit: %.2f MB", "query required more memory than allowed", tx.GraphQueryMemoryLimit().Mebibytes())
+					}
```
urangel, line 203: Nice refactor

cmd/api/src/auth/model.go
mistahj67: Because it's accepting a fn, just covering our bases with the nil check.

cmd/api/src/api/v2/cypher_search_integration_test.go
superlinkx: I'd even go shorter: "Mutations allowed with enable_cypher_mutations true"

urangel (approved):
Looking forward to having this in main!

PR #574, opened 2024-04-23
update cypher query timeouts, additional logging

irshadaj:
## Description

* More conservative default query timeouts for cypher
* Additional logging for better debugging
* Introduced a maximum timeout that will override abnormally high user specified timeouts

## How Has This Been Tested?
Unit tests 

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/queries/graph.go
+++ b/cmd/api/src/queries/graph.go
@@ -385,9 +385,9 @@ func (s *GraphQuery) prepareGraphQuery(rawCypher string, disableCypherQC bool) (
 		return graphQuery, newQueryError(err)
 	} else if !disableCypherQC && complexityMeasure.Weight > MaxQueryComplexityWeightAllowed {
 		// log query details if it is rejected due to high complexity
-		timeoutEvent := log.WithLevel(log.LevelError)
-		timeoutEvent.Str("query", graphQuery.strippedQuery)
-		timeoutEvent.Msg(fmt.Sprintf("Query rejected. Query weight: %.2f. Maximum allowed weight: %d", graphQuery.complexity.Weight, MaxQueryComplexityWeightAllowed))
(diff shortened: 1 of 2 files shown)


cmd/api/src/queries/graph.go
```diff
@@ -383,6 +384,11 @@ func (s *GraphQuery) prepareGraphQuery(rawCypher string, disableCypherQC bool) (
 	} else if complexityMeasure, err := analyzer.QueryComplexity(queryModel); err != nil {
 		return graphQuery, newQueryError(err)
 	} else if !disableCypherQC && complexityMeasure.Weight > MaxQueryComplexityWeightAllowed {
+		// log query details if it is rejected due to high complexity
+		timeoutEvent := log.WithLevel(log.LevelError)
+		timeoutEvent.Str("query", graphQuery.strippedQuery)
+		timeoutEvent.Msg(fmt.Sprintf("Query rejected. Query weight: %.2f. Maximum allowed weight: %d", graphQuery.complexity.Weight, MaxQueryComplexityWeightAllowed))
+
```
urangel, line 391: nit: timeoutEvent makes me think a timeout happened but this is bailing before actually executing the query. 

It might be helpful to add the constituent complexity values to the logged event so the user knows what levers they can pull to reduce complexity like number of patterns or projections.

cmd/api/src/queries/graph_test.go
```diff
@@ -101,12 +101,15 @@ func TestGraphQuery_RawCypherSearch(t *testing.T) {
 	// This will be set to a default of 30 min, with a reduction factor of 3, so we should have a 10 min config timeout
```
urangel, line 101: This comment should be updated to reflect the new default of 15 min as mentioned in the comment below.

urangel (approved):

PR #577, opened 2024-04-24
Open password reset dialog when updating user to use username/password based authentication

elikmiller:
## Description
In the Administration/Manage Users page, Administrators who update a user to use username/password based authentication when the user had previously used SAML are now presented with the password reset dialog to help ensure that the user has a valid secret before attempting their next login.

## Motivation and Context
Users lose their auth secret when not using username/password based authentication. If a user is updated to use username/password based authentication they can be left in a state where they cannot log in because there is no valid password for them to use. This change prompts admins to supply a password for these users to avoid this scenario.

## How Has This Been Tested?

New tests have been added to cover this behavior

## Types of changes

## Checklist:


urangel (approved):
Nice tests! Thanks for moving the UpdateUserRequest interface definition to the js client library for reuse 🙌

PR #578, opened 2024-04-24
chore: modify dockerfile to build with stbernard

superlinkx:
## Description

This PR updates the existing Dockerfile to build with `stbernard` instead of `beagle`. A few key changes were necessary to enable this functionality:

* Pulling the version from the environment variable if the git tag was unavailable. This was necessary to support building the edge release.
* Adding flags to override the `GOOS` and `GOARCH` environment variables. This enables `stbernard` to build against any environment via `go run`. Prior to this change the variables would conflict against the architecture that was calling `go run stbernard`.

Also included in this PR is the removal of Python from the Docker build as it was only used by `beagle.`

## Motivation and Context

* Consolidate on a single CLI tool for BloodHound

## How Has This Been Tested?

* Ran the builds locally and on GitHub actions and verified they built successfully

## Types of changes

## Checklist:

### DIFF
--- a/dockerfiles/bloodhound.Dockerfile
+++ b/dockerfiles/bloodhound.Dockerfile
@@ -41,7 +41,9 @@ RUN apk add --update --no-cache git go
 >>>>>>> 6b8694bb (chore: modify dockerfile to build with stbernard)
 
 COPY . /bloodhound
-RUN go run github.com/specterops/bloodhound/packages/go/stbernard deps
+
+RUN go build -o /stbernard github.com/specterops/bloodhound/packages/go/stbernard
+RUN /stbernard deps
 
 ########
 # Build
@@ -54,7 +56,7 @@ ENV GOARCH=${TARGETARCH}
 ENV CGO_ENABLED=0
 WORKDIR /bloodhound
 
-RUN go run github.com/specterops/bloodhound/packages/go/stbernard build
+RUN /stbernard build
 
 ########
 # Package other assets
--- a/dockerfiles/bloodhound.Dockerfile
+++ b/dockerfiles/bloodhound.Dockerfile
@@ -26,6 +26,10 @@ ARG AZUREHOUND_VERSION=v2.1.9
 FROM --platform=$BUILDPLATFORM docker.io/library/node:20-alpine AS deps
 ARG version=v999.999.999
 ARG checkout_hash=""
+ARG TARGETOS
+ARG TARGETARCH
+ENV GOOS=${TARGETOS}
+ENV GOARCH=${TARGETARCH}
 ENV SB_LOG_LEVEL=debug
 ENV VERSION=${version}
 ENV CHECKOUT_HASH=${checkout_hash}
--- a/dockerfiles/bloodhound.Dockerfile
+++ b/dockerfiles/bloodhound.Dockerfile
@@ -43,9 +43,10 @@ FROM deps AS builder
 ARG TARGETOS
 ARG TARGETARCH
 ENV CGO_ENABLED=0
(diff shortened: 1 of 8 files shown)


packages/go/stbernard/workspace/golang/build.go
```diff
@@ -33,15 +33,17 @@ import (
...
 		mu       sync.Mutex
+		version  semver.Version
 		buildDir = filepath.Join(workRoot, "dist") + string(filepath.Separator)
 	)
 
-	version, err := git.ParseLatestVersionFromTags(workRoot, env)
-	if err != nil {
-		return fmt.Errorf("parse latest version from git tags: %w", err)
+	if version, err = git.ParseLatestVersionFromTags(workRoot, env); err != nil {
+		log.Errorf("Failed to parse version from git tags, falling back to environment variable: %v", err)
```
superlinkx, line 45: I think this will be expected behavior in some cases. I recommend we move this to a Warnf

packages/go/stbernard/workspace/golang/build.go
```diff
@@ -33,15 +33,17 @@ import (
...
+		version = *semver.MustParse(env["VERSION"])
```
superlinkx, line 46: This will panic if we don't have a valid version in that env var, we may want to look at handling it as an error properly

packages/go/stbernard/command/builder/builder.go
```diff
@@ -69,6 +71,13 @@ func (s *command) Parse(cmdIndex int) error {
 	if err := cmd.Parse(os.Args[cmdIndex+1:]); err != nil {
 		cmd.Usage()
 		return fmt.Errorf("parsing %s command: %w", Name, err)
+	} else {
+		if *targetOs != "" {
+			s.env.Override("GOOS", *targetOs)
+		}
+		if *targetArch != "" {
+			s.env.Override("GOARCH", *targetArch)
+		}
```
superlinkx, line 80: These don't really need to be in the else clause, I'd just remove the else and let them be floating after the error check, since they have zero dependencies on that error check.

dockerfiles/bloodhound.Dockerfile
```diff
@@ -26,48 +26,27 @@ ARG AZUREHOUND_VERSION=v2.1.9
...
-# Build API
-################
-FROM deps AS api-builder
+FROM deps AS builder
 ARG TARGETOS
 ARG TARGETARCH
-ENV GOOS=${TARGETOS}
-ENV GOARCH=${TARGETARCH}
 ENV CGO_ENABLED=0
+ENV VERSION=${version}
```
superlinkx, line 46: We may want to use a more targeted ENV var just to prevent possible collisions in the future. This would also mean changing what St Bernard looks for. Perhaps something like SB_VERSION since SB-specific env vars are supposed to be prefixed.

packages/go/stbernard/workspace/golang/build.go
superlinkx, line 46: Additionally, it might be good to set this env to a constant somewhere. I took a shortcut on the SB_LOG_LEVEL one and threw it in main, but it might not hurt to do a proper refactor to keep all our environment related code together, perhaps pushing it all into the environment package (and having `main()` run NewEnvironment for us)

superlinkx (commented):
The only one of these suggestions I'm really worried about is the potential panic, otherwise these are minor suggestions that can be either changed or resolved as wontfix. Overall this looks really good and I appreciate getting this plumbed in properly.

zinic (approved):
LGTM, pull!

PR #581, opened 2024-04-25
fixed bug in logging high complexity query rejections

irshadaj:
## Description

A bug was introduced as part of https://github.com/SpecterOps/BloodHound/pull/574 where we were attempting to log some fields that hadn't yet been populated from the buffer. This would result in nil pointer errors, panics, and/or empty logged fields. This manifested as the following error in the console when a cypher query exceeded the maximum complexity threshold:

Screenshot 2024-04-25 at 12 51 00 PM

The changes in this MR will populate all fields to be logged, before the log event is set up.

## How Has This Been Tested?
Forced a query to exceed maximum complexity threshold by making the complexity analyzer (`QueryComplexity` function) return a high value

Screenshot 2024-04-25 at 1 00 51 PM

Ran a cypher query- this would trip the max complexity threshold and go into the high complexity log codepath. Error is shown at the bottom of the screen as expected.

Screenshot 2024-04-25 at 1 02 24 PM

Verified that the panic does not occur, and all expected fields are logged correctly

Screenshot 2024-04-25 at 12 57 53 PM

The other logging change sent out as part of the previous MR was related to neo4j timeouts. So I followed a similar strategy to force a neo4j timeout to occur, and verified that no nil pointers are thrown in that codepath.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/queries/graph.go
+++ b/cmd/api/src/queries/graph.go
@@ -374,20 +374,23 @@ type preparedQuery struct {
 
 func (s *GraphQuery) prepareGraphQuery(rawCypher string, disableCypherQC bool) (preparedQuery, error) {
 	var (
-		parseCtx   = frontend.DefaultCypherContext()
-		buffer     = &bytes.Buffer{}
-		graphQuery preparedQuery
+		parseCtx            = frontend.DefaultCypherContext()
+		queryBuffer         = &bytes.Buffer{}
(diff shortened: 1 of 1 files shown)


mistahj67 (approved):
I suspect there might be a way to have the single buffer but probably more work than it's worth.

Code LGTM and testing locally, no more panic, thanks for fixing!

PR #586, opened 2024-04-26
improve query complexity calculations

irshadaj:
## Description

* Improved code readability by using better suited data types in query complexity calculations
* Fixed missed loophole in query complexity calculation
* Refactored reduction factor into its own function to aid readability
* Added runtime measurements and logging for cypher queries to better inform the appropriateness of our reduction factor calculations in the future

## How Has This Been Tested?
Changed code paths are under extensive unit and integration tests, which had to be updated accordingly.

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/queries/graph_test.go
+++ b/cmd/api/src/queries/graph_test.go
@@ -168,7 +168,7 @@ func TestGraphQuery_RawCypherSearch(t *testing.T) {
 		// This will be set to a default of 15 min or 900 sec
 		// availableRuntime = 900 sec, query cost = 15
 		// reductionFactor = 1 + (15/5) = 4
-		// Therefore actual timeout = availableRuntime/reductionFactor : 900/3 = 225sec
+		// Therefore actual timeout = availableRuntime/reductionFactor : 900/4 = 225sec
 
 		outerBHCtxInst.Timeout = 0
 		preparedQuery, err := gq.PrepareCypherQuery("match ()-[:HasSession*..]->()-[:MemberOf*..]->() return n;")
--- a/cmd/api/src/queries/graph_internal_test.go
+++ b/cmd/api/src/queries/graph_internal_test.go
@@ -31,6 +31,41 @@ import (
 	"go.uber.org/mock/gomock"
 )
 
+func Test_ApplyTimeoutReduction(t *testing.T) {
+	// Query Weight			Reduction Factor 		  Runtime
+	// 	0-4						1						x
+	// 	5-9						2						x/2
+	//	10-14					3						x/3
+	//	15-19					4						x/4
+	//	20-24					5						x/5
+	//	25-29					6						x/6
+	//	30-34					7						x/7
+	//	35-39					8						x/8
+	//	40-44					9						x/9
+	//	45-49					10						x/10
+	// 	50						11						x/11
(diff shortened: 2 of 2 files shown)


cmd/api/src/queries/graph_test.go
```diff
@@ -141,28 +142,44 @@ func TestGraphQuery_RawCypherSearch(t *testing.T) {
 			txConfig := &graph.TransactionConfig{}
 			options[0](txConfig)
 
-			require.Equal(t, time.Minute*5, txConfig.Timeout)
+			require.Equal(t, time.Second*225, txConfig.Timeout)
```
superlinkx, line 145: I'm not sure I understand why this went from 5m to 3m45s, do we know what's causing this change?

superlinkx (commented):
Just one question on a test condition changing to better understand. Otherwise, this all looks as expected

cmd/api/src/queries/graph.go
```diff
@@ -507,12 +497,19 @@ func (s *GraphQuery) RawCypherSearch(ctx context.Context, pQuery PreparedQuery,
 		err = s.Graph.ReadTransaction(ctx, txDelegate, txOptions)
 	}
 
+	runtime := time.Now().Sub(start)
```
mistahj67, line 500:
```suggestion
	runtime := time.Since(start)
```

cmd/api/src/queries/graph.go
```diff
@@ -523,6 +520,24 @@ func (s *GraphQuery) RawCypherSearch(ctx context.Context, pQuery PreparedQuery,
...
+	// weights of 5-9 will get 1/2 the runtime duration
+	// weights of 10-15 will get 1/3 the runtime duration
+	// and so on until the max weight of 50 gets 1/11 the runtime duration
+
+	// NOTE reductionFactor will be the math.Floor() of the result of the division below
+	reductionFactor := 1 + (queryWeight / 5)
+	availableRuntimeInt /= reductionFactor
+	availableRuntime = time.Duration(availableRuntimeInt) * time.Second
+
+	log.Infof("Cypher query cost is: %d. Reduction factor for query is: %d. Available timeout for query is now set to: %d seconds", queryWeight, reductionFactor, availableRuntime)
```
mistahj67, line 536: Not ideal as it would require another parameter for just a log, but I wonder if we'd get a lot more out of this log if we included the stripped query to easily correlate

perhaps including it in the log event somehow 🤔

cmd/api/src/queries/graph.go
superlinkx: Nice catch, I think one of our analyzers can catch that too, it just might not be turned on yet

cmd/api/src/queries/graph.go
superlinkx: Actually yeah, we probably do need the query here. Correlation is otherwise virtually impossible

mistahj67 (approved):
LGTM thanks for adding test coverage!

cmd/api/src/queries/graph_test.go
superlinkx: Ah, that makes sense. It was hard to correlate just by looking at the test, but now I get it. A proper set of tests would be nice for sure

PR #590, opened 2024-04-30
feat: add audit logs to graph mutation attempts

mistahj67:
## Description

- Adds audit logs for intent, success, failure of all mutation attempts when `EnableCypherMutation` is true
- Cleans up audit log testing to use a single set of testing utils
- Removes unnecessary string type assertion for AuditLog statuses

## Motivation and Context

Any mutations done should be audit logged in order to keep track of who and how graph mutations occurred

## How Has This Been Tested?

Integration tests were added and it was tested locally and verified through pgadmin

## Types of changes

## Checklist:


cmd/api/src/api/v2/cypher_search.go
```diff
@@ -34,27 +36,68 @@ type CypherSearch struct {
...
+			model.AuditLogActionMutateGraph,
+			model.AuditLogStatusIntent,
+			model.AuditData{"query": preparedQuery.StrippedQuery}); err != nil {
+			api.WriteErrorResponse(
+				request.Context(),
+				api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request),
+				response,
+			)
+			return
+		}
```
juggernot325, line 83: This block is really difficult to read. Would you mind adjusting the indentation to make it more clear where the error handling code is vs the function call?

cmd/api/src/api/v2/cypher_search.go
```diff
@@ -34,27 +36,68 @@ type CypherSearch struct {
 
 func (s Resources) CypherSearch(response http.ResponseWriter, request *http.Request) {
```
juggernot325, line 37: Given the significant extra logic for mutation Cypher operations, I wonder if it would be worth a refactor to have two separate functions.

juggernot325 (changes requested):
Found one minor readability change that would be nice to fix up. I'm open to have a discussion on the suggested refactor.

cmd/api/src/api/registration/v2.go
```diff
@@ -159,8 +159,8 @@ func NewV2API(cfg config.Configuration, resources v2.Resources, routerInst *rout
 		// TODO discuss if this should be a post endpoint
 		routerInst.GET("/api/v2/graph-search", resources.GetSearchResult).RequirePermissions(permissions.GraphDBRead),
 
-		// Cypher Queries API
-		routerInst.POST("/api/v2/graphs/cypher", resources.CypherSearch).RequirePermissions(permissions.GraphDBRead),
+		// CypherQuery queries graph db
```
superlinkx, line 162: This seems unintentional

cmd/api/src/api/v2/cypherquery.go
```diff
@@ -60,69 +59,69 @@ func (s Resources) CypherSearch(response http.ResponseWriter, request *http.Requ
...
+		graphResponse, err = s.cypherMutation(request, preparedQuery, payload.IncludeProperties)
+	} else {
+		graphResponse, err = s.GraphQuery.RawCypherQuery(request.Context(), preparedQuery, payload.IncludeProperties)
+	}
 
-		// create an intent audit log
-		if err := s.DB.AppendAuditLog(request.Context(), auditLogEntry); err != nil {
+	if err != nil {
+		switch true {
+		case strings.Contains(err.Error(), "Permission denied: User may not modify the graph."):
```
superlinkx, line 69: I would not use a switch here, at least like this. We should definitely not look for a string in an error. Make the error a proper sentinel if we need it and use `errors.Is` to match the error we care about

cmd/api/src/api/v2/cypherquery.go
```diff
@@ -60,69 +59,69 @@ func (s Resources) CypherSearch(response http.ResponseWriter, request *http.Requ
...
-		}
-	}
-
-	if graphResponse, err := s.GraphQuery.RawCypherSearch(request.Context(), preparedQuery, payload.IncludeProperties); err != nil {
-		// audit log needs to be marked as failed if a mutation is present
-		if preparedQuery.HasMutation {
-			auditLogEntry.Status = model.AuditLogStatusFailure
-		}
-		if queries.IsQueryError(err) {
+		case queries.IsQueryError(err):
```
superlinkx, line 74: This could probably be updated to `errors.As(err, queries.QueryError)` but I'm not 100% certain. Not a big deal, but would be nice to eventually kill this function

cmd/api/src/api/v2/cypherquery.go
```diff
@@ -60,69 +59,69 @@ func (s Resources) CypherSearch(response http.ResponseWriter, request *http.Requ
...
 		api.WriteBasicResponse(request.Context(), graphResponse, http.StatusOK, response)
 	}
+}
 
-	if preparedQuery.HasMutation {
-		if err := s.DB.AppendAuditLog(request.Context(), auditLogEntry); err != nil {
-			// We don't want to update response as failed because having info on the mutation response trumps this error
-			log.Errorf("failure to create mutation audit log %s", err.Error())
-		}
+func (s Resources) cypherMutation(request *http.Request, preparedQuery queries.PreparedQuery, includeProperties bool) (model.UnifiedGraph, error) {
```
superlinkx, line 92: I think it might make sense to pull the authCtx off of request before sending it into this function. That way the function no longer needs to worry about request semantics, but just the AuthCtx for checking permissions.

cmd/api/src/api/v2/cypherquery.go
```diff
@@ -60,69 +59,69 @@ func (s Resources) CypherSearch(response http.ResponseWriter, request *http.Requ
...
+	var (
+		auditLogEntry model.AuditEntry
+		graphResponse model.UnifiedGraph
+		err           error
+	)
+
+	if !s.Authorizer.AllowsPermission(ctx.FromRequest(request).AuthCtx, auth.Permissions().GraphDBMutate) {
+		s.Authorizer.AuditLogUnauthorizedAccess(request)
+		return model.UnifiedGraph{}, errors.New("Permission denied: User may not modify the graph.")
```
superlinkx, line 101: Move this to a global variable to use as a ErrModificationDisallowed (or similar) sentinel

cmd/api/src/api/v2/cypherquery.go
superlinkx: I guess we'd also want to grab a context off of request and pass it down here too, but that's not a bad idea as it helps ensure the context requirement is obvious

mistahj67: I gave this a try actually but reverted because I then have to update the `AuditLogUnauthorizedAccess` to take in the 3 request parts it is concerned about and that ballooned out the arguments so I eventually embraced the request argument and nixed the authCtx entirely. I can add it back if it makes it more readable?

cmd/api/src/api/v2/cypherquery.go
mistahj67: I realized that we actually don't need the file whatsoever b/c we elevated the parse query function, it was the sole consumer of that error and those are all properly handled above

juggernot325 (approved):
I really like the `CypherSearch` > `CypherQuery` refactor. Thank you for doing that!

cmd/api/src/api/v2/cypherquery.go
superlinkx: Yeah, that's fine then. Was just maybe a way for us to clean up dependencies for later refactors, but not necessary

PR #599, opened 2024-05-06
BED-4358: Add Complexity Measurement For Graph Mutations

superlinkx:
## Description

This PR adds our initial complexity weights for different mutation statements during query analysis. As a byproduct, it also properly categorizes queries with no constraints (e.g. `match (n) return n`) as beyond the max weight allowed, so that when cypher complexity guardrails are enabled, these queries will be blocked appropriately. A decent number of our test cases needed to be updated to handle this change, and their complexities changed in response too.

Additionally, we fixed an identified bug with our multipart query parsing that caused WITH clauses to break (e.g. `WITH 'sid' AS sid MATCH (u:User {objectid: sid}) RETURN u`).

## Motivation and Context

BED-4358

## How Has This Been Tested?

Added additional tests for complexity and parsing
Updated a few breaking tests as appropriate
Spot checking by running some cypher we were concerned with

## Types of changes

## Checklist:


packages/go/cypher/analyzer/measure.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+func (s *ComplexityMeasure) onQuantifier(_ *model.WalkStack, _ *model.Quantifier) error {
+	// Quantifier expressions may increase the size of an inline projection to apply its contained filter and should
+	// be weighted
+	s.Weight += weight1
+	return nil
+}
+
+func (s *ComplexityMeasure) onRelationshipPattern(_ *model.WalkStack, node *model.RelationshipPattern) error {
```
urangel, line 199: I realize most of this code is just being moved from the analyzer.go file but I was wondering why we name the `*model.RelationshipPattern` parameter as `node`. Similarly for other param names in these methods. I might expect it to be named `relationshipPattern` like in `liftRelationshipPatternCriteria`. 

The reason I was curious was because I see that `match (n) return n` will be gated by setting it to be over the max complexity allowed. I often search for a specific edge in the cypher search, e.g., `match p= ()-[:DCSync]->() return p` which does not have any "node" constraints but does have a relationship constraint. Would DCSync then be included in the `ComplexityMeasure`'s `nodeLookupKinds`? Or will this query for returning these DCSync edges be disallowed because it has no "node" lookup kinds?

cmd/api/src/api/v2/cypherquery_integration_test.go
```diff
@@ -169,7 +168,7 @@ func Test_CypherSearch_WithCypherMutationsEnabled(t *testing.T) {
 			assert.True(ok)
 
 			var (
-				query         = "match (w) where w.name = 'harryp' delete w"
+				query         = "match (w: Wizard) where w.name = 'harryp' delete w"
```
urangel, line 171: yer a `Label` now Harry!

packages/go/cypher/frontend/query.go
```diff
@@ -363,43 +362,34 @@ func NewMultiPartQueryVisitor() *MultiPartQueryVisitor {
...
+	part := model.NewMultiPartQueryPart()
+	part.AddReadingClause(s.ctx.Exit().(*ReadingClauseVisitor).ReadingClause)
+	s.Query.Parts = append(s.Query.Parts, part)
 }
 
 func (s *MultiPartQueryVisitor) EnterOC_UpdatingClause(ctx *parser.OC_UpdatingClauseContext) {
 	s.ctx.Enter(NewUpdatingClauseVisitor())
 }
 
 func (s *MultiPartQueryVisitor) ExitOC_UpdatingClause(ctx *parser.OC_UpdatingClauseContext) {
```
urangel, line 378: What does "OC" stand for?

packages/go/cypher/analyzer/measure.go
superlinkx: I'm not entirely sure why node was chosen as the name for these parameters, other than maybe because they are nodes in the parser tree? Regardless, we did a decent amount of work to ensure that _any_ constraint would be sufficient to avoid that max complexity call onExit. That includes relationship labels, node labels, property filters, any where clause, or a limit.

packages/go/cypher/frontend/query.go
superlinkx: I'm not entirely sure. These are all overrides for functions that are defined by the antlr generated code

packages/go/cypher/analyzer/measure.go
urangel: Yea nodes in the parser tree was something I was considering. Okay, that gives me peace of mind to know that edge matching like I wrote would still be allowed. Thanks for the explanation!

packages/go/cypher/frontend/query.go
urangel: Cool was just wondering since I didn't see it spelled out anywhere at a glance.

urangel (approved):
Thanks for the new and updated tests!

packages/go/cypher/analyzer/measure.go
superlinkx: If you look at the JSON tests, you should see a few tests that validate that an edge label is sufficient as well. We tried to add any tests that seemed to be missing to ensure we'd catch any known edge cases

superlinkx:
```
{
    "name": "Query for the properties of an edge using keys()",
    "type": "string_match",
    "details": {
        "query": "match ()-[e:EDGE_OF_INTEREST]->() return keys(e)",
        "complexity": 6
    }
},
```

PR #602, opened 2024-05-08
feat: Update UI to better support cypher mutations

mistahj67:
## Description

- Moved empty response handling to rely on API error code 404
- If a mutation isn't present and nothing is found, throw 404 from API
- Update search placeholder to "Cypher Query"
- Update search button to more ambiguous "Run" 

## Motivation and Context
Now that mutations will be supported through the cypher tab, the UX needs to better reflect that

## How Has This Been Tested?

Tested locally

Before:
image
Sent when an empty return mutation was ran:
image

After:
image
image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/ui/src/ducks/explore/saga.ts
+++ b/cmd/ui/src/ducks/explore/saga.ts
@@ -195,15 +195,10 @@ function* runCypherSearchQuery(payload: CypherQueryRequest): SagaIterator {
 
         if (resultNodesAreEmpty && !resultEdgesAreEmpty) {
             yield put(putGraphData({}));
-            yield put(
-                addSnackbar(
-                    'The results are not rendered since only edges were returned',
-                    'cypherSearchOnlyContainsEdges'
-                )
-            );
+            yield put(addSnackbar('The results are not rendered since only edges were returned', 'cypherSearchOnlyContainsEdges'));
         } else if (resultNodesAreEmpty && resultEdgesAreEmpty) {
             yield put(putGraphData({}));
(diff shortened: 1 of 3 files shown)


cmd/ui/src/ducks/explore/saga.ts
```diff
@@ -195,15 +195,10 @@ function* runCypherSearchQuery(payload: CypherQueryRequest): SagaIterator {
...
-                addSnackbar(
-                    'The results are not rendered since only edges were returned',
-                    'cypherSearchOnlyContainsEdges'
-                )
-            );
+            yield put(addSnackbar('The results are not rendered since only edges were returned', 'cypherSearchOnlyContainsEdges'));
         } else if (resultNodesAreEmpty && resultEdgesAreEmpty) {
             yield put(putGraphData({}));
-            yield put(addSnackbar('No results match your criteria', 'cypherSearchEmptyResponse'));
+            yield put(addSnackbar('Command completed successfully', 'cypherSuccessResponse'));
```
benwaples, line 201: Should we also add this new message in the below else block? For queries that do not error and have results.

mistahj67: If they have results we just display the results with no snackbar like previously.

maffkipp (approved):

cmd/ui/src/ducks/explore/saga.ts
benwaples: Makes sense! Just was double checking on the AC :) Thank you!

PR #611, opened 2024-05-14
feat: auto tag T0 parent objects

JonasBK:
## Description

Parent OUs and containers of Tier Zero AD objects are automatically added to Tier Zero during analysis. Containers are only added if they have a Tier Zero child object with ACL inheritance enabled. 

A feature flag will allow the user to disable the feature.

Requires a migration.

## Motivation and Context

A Tier Zero node can be compromised in an attacker has control over the parent object.

Ticket: BED-3329.

This PR must be merged in and released in the same cycle as other PRs. See which in the BED-4430 ticket.

## How Has This Been Tested?

With local lab data.

image

## Types of changes

## Checklist:


cmd/api/src/model/appcfg/flag.go
```diff
@@ -125,6 +126,13 @@ func AvailableFlags() FeatureFlagSet {
 			Enabled:       false,
 			UserUpdatable: true,
 		},
+		FeatureAutoTagT0ParentObjects: {
```
urangel, line 129: This will need an explicit migration in a .sql file once this PR gets merged into main https://github.com/SpecterOps/BloodHound/pull/772

cmd/api/src/daemons/datapipe/agi.go
```diff
@@ -154,20 +155,26 @@ func ParallelTagAzureTierZero(ctx context.Context, db graph.Database) error {
...
 	defer log.Measure(log.LevelInfo, "Finished tagging Active Directory Tier Zero")()
 
-	if domains, err := adAnalysis.FetchAllDomains(ctx, db); err != nil {
+	autoTagT0ParentObjectsFlag, err := db.GetFlagByKey(ctx, appcfg.FeatureAutoTagT0ParentObjects)
+	if err != nil {
+		log.Errorf("error getting AutoTagT0ParentObjects feature flag: %w", err)
+		return err
+	}
+
+	if domains, err := adAnalysis.FetchAllDomains(ctx, graphDB); err != nil {
```
urangel, line 167:
```suggestion
	if autoTagT0ParentObjectsFlag, err := db.GetFlagByKey(ctx, appcfg.FeatureAutoTagT0ParentObjects); err != nil {
		return err
	} else if domains, err := adAnalysis.FetchAllDomains(ctx, graphDB); err != nil {
```

Suggestion for maintaining consistency in inlining initializer statements as well as handling error logging up the stack.

cmd/api/src/model/appcfg/flag.go
JonasBK: gotcha, so that means I should set the "My changes include a database migration" flag?

urangel: Correct. I believe we will be merging #772 for next release so this changeset should include a migration and have it be reflected in the PR description.

packages/go/analysis/ad/queries.go
```diff
@@ -130,6 +148,32 @@ func FetchActiveDirectoryTierZeroRoots(ctx context.Context, db graph.Database, d
 			attackPathRoots.AddSet(enforcedGPOs)
 		}
 
+		if (autoTagT0ParentObjectsFlag) {
```
urangel, line 151: Would a path with an attackPathRoot with ACL inheritance enabled and alternating containers/ous in the chain be valid? Something like: 
`(ou1:OU)-[:Contains]->(c1:Container)-[:Contains]->(ou2:OU)-[:Contains]->(c2:Container)-[:Contains]->(b:Base {system_tags: "admin_tier_0", isaclprotected: false})`

JonasBK: Great question - Yup, path would be valid. By disabling ACL inheritance you prevent ACL-based attacks, but it does not stop an attacker from linking a malicious GPO to the parent OU which enables a takeover of the child principals. GPOs can only be linked to OUs however.

urangel (approved):

PR #616, opened 2024-05-17
feat: featureflag for auto tagging T0 AD group members

JonasBK:
## Description

Add a feature flag to enable/disable auto-tagging of Tier Zero AD group members

Ticket: BP-702

## Motivation and Context

Some users (mostly BHE users) prefer no automatic Tier Zero tagging of Tier Zero AD group members as they would like to manually verify new members of Tier Zero groups. The feature flag allows them to disable the auto-tagging.

For BHE users, disabling the member auto-tagging will result in MemberOf findings, which comes with the BHE pull request 536. That PR should be merged at the same time as this PR.

## How Has This Been Tested?

Tested locally. Disabled and enabled it a couple of times to confirm it works.

We should perform a performance test with BHE to confirm that analysis does not increase significantly with this change. 

## Types of changes

## Checklist:


cmd/api/src/daemons/datapipe/agi.go
```diff
@@ -160,11 +160,13 @@ func TagActiveDirectoryTierZero(ctx context.Context, db database.Database, graph
 
 	if autoTagT0ParentObjectsFlag, err := db.GetFlagByKey(ctx, appcfg.FeatureAutoTagT0ParentObjects); err != nil {
 		return err
+	} else if autoTagMembersFlag, err := db.GetFlagByKey(ctx, appcfg.FeatureAutoTagT0ADMembers); err != nil {
```
juggernot325, line 163: I tried to test this locally and was unable to build the app. There's a change in main that breaks this line. The `db` variable is no longer present in the latest code. It looks like it was changed in #816.

juggernot325 (changes requested):

cmd/api/src/daemons/datapipe/agi.go
JonasBK: Fixed

JonasBK:
Issue identified. Will investigate.

JonasBK:
Closing this PR as we will implement the feature later in another way.

PR #621, opened 2024-05-21
feat: require current password to update password

mistahj67:
## Description
Closes Bed 4394

Adds the current password as a requirement to change password for your own password.
This also applies to the password reset form

## Motivation and Context

Currently password reset doesn't require the users current password. This is unsafe and we want to add the requirement that current password is required to change your own password.

## How Has This Been Tested?

U/I tests updated / added
API tests updated / added

image

image

## Types of changes

## Checklist:


cmd/api/src/api/v2/model.go
```diff
@@ -79,6 +79,7 @@ type DeleteSAMLProviderResponse struct {
 }
 
 type SetUserSecretRequest struct {
+	CurrentSecret      string `json:"current_secret"`
```
mistahj67, line 82: We don't want the validate tag as it's possible to set initial passwords via config, also we validate later in the handler against their pw anyways

cmd/ui/src/utils.ts
```diff
@@ -77,7 +77,8 @@ export const initializeBHEClient = () => {
                 if (
                     error?.response?.status === 401 &&
                     error?.response?.config.url !== '/api/v2/login' &&
-                    error?.response?.config.url !== '/api/v2/logout'
+                    error?.response?.config.url !== '/api/v2/logout' &&
+                    !error?.response?.config.url.match('/api/v2/bloodhound-users/[a-z0-9-]+/secret')
```
mistahj67, line 81: I tried a few different variants such as `e.preventDefault()` or `e.stopPropagation()` but no joy. Ended up opting for this which fixed the issue of immediate logout upon an invalid current password. Alternatively I could use a different code but `401` and `403` have built in default behavior and yet `401` felt the most appropriate error code to return from API when an invalid current password is sent. 

Of course, I'm open to suggestions if there's a cleaner way of reporting invalid current password!

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -634,39 +636,53 @@ func (s ManagementResource) PutUserAuthSecret(response http.ResponseWriter, requ
...
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
 	} else {
-		authSecret.UserID = targetUser.ID
-		authSecret.Digest = secretDigest.String()
-		authSecret.DigestMethod = s.secretDigester.Method()
-		authSecret.ExpiresAt = time.Now().Add(passwordExpiration).UTC()
-
-		if setUserSecretRequest.NeedsPasswordReset {
-			authSecret.ExpiresAt = time.Time{}
+		if loggedInUser.ID == targetUserID {
```
brandonshearin, line 656: how do we prevent a logged in user from changing the password of somebody else?

cmd/ui/src/views/Users/Users.tsx
```diff
@@ -159,8 +160,19 @@ const Users = () => {
...
+            needsPasswordReset,
+            currentSecret,
+        }: {
+            currentSecret?: string;
+            userId: string;
+            secret: string;
+            needsPasswordReset: boolean;
+        }) =>
             apiClient.putUserAuthSecret(userId, {
+                ...(currentSecret && { current_secret: currentSecret }),
```
brandonshearin, line 175: update the types.ts file in js-client-library to include the new current_secret field. the interface you are looking for is called `PutUserAuthSecretRequest`

cmd/api/src/api/v2/auth/auth.go
mistahj67: Admin permissions is checked in the auth middleware

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -634,39 +636,53 @@ func (s ManagementResource) PutUserAuthSecret(response http.ResponseWriter, requ
...
+			if targetUser.AuthSecret == nil {
+				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrNoUserSecret.Error(), request), response)
+			} else if err := s.authenticator.ValidateSecret(request.Context(), setUserSecretRequest.CurrentSecret, *targetUser.AuthSecret); err != nil {
+				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusUnauthorized, "Invalid current password", request), response)
```
ddlees, line 660: I have mixed feelings about returning a 401 here. If the current user has a session, then they are implicitly authenticated. However, if they input an incorrect password during verification does that mean they are no longer authenticated? If that's the case, then an unauthenticated user should not have a session and should be logged out. Conversely, if it means they are still authenticated and can retain their session then a 403 forbidden is probably more contextually appropriate.

cmd/ui/src/utils.ts
ddlees: See previous comment considering a 403 status instead.

packages/javascript/bh-shared-ui/src/components/PasswordDialog/PasswordDialog.tsx
```diff
@@ -77,13 +80,17 @@ const PasswordDialog: React.FC<{
...
 
-    const handleOnSave = (data: { password: string; confirmPassword: string; needsPasswordReset: boolean }) => {
-        return onSave({
-            userId: userId,
-            secret: data.password,
-            needsPasswordReset: Boolean(data.needsPasswordReset),
-        });
-    };
+    const handleOnSave = useCallback(
+        (data: ChangePasswordFormInputs) => {
```
brandonshearin, line 84: there seems to be a mixing of the words "secret" and "password" throughout this component. if there are two different names for the same thing, can we pick one and make it consistent

cmd/api/src/api/v2/auth/auth.go
mistahj67: Should be returned forbidden in that case correct. Should return `401 Not Authorized`. It's handled by `AuthorizeAuthManagementAccess` in `cmd/api/src/api/middleware/auth.go`

cmd/ui/src/views/Users/Users.tsx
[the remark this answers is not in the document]
mistahj67: Totally missed these, thanks for pointing them out! 🙇

packages/javascript/bh-shared-ui/src/components/PasswordDialog/PasswordDialog.tsx
mistahj67: We discussed offline, but it seems API related uses `secret` while UI related uses `password`. I'm open to refactoring but it might be out of scope for this particular task and I'd like to open the discussion to the wider engineering team / verify it's something we want to do.

packages/javascript/bh-shared-ui/src/views/UserProfile/UserProfile.tsx
```diff
@@ -46,8 +46,19 @@ const UserProfile = () => {
...
+            needsPasswordReset,
+            currentSecret,
+        }: {
+            userId: string;
+            currentSecret?: string;
+            secret: string;
+            needsPasswordReset: boolean;
+        }) =>
             apiClient.putUserAuthSecret(userId, {
+                ...(currentSecret && { current_secret: currentSecret }),
```
brandonshearin, line 61: this line can be simplified to:
`current_secret: currentSecret`.  

you can safely remove the && operator and the ... operator.  on mutations where currentSecret is undefined, the field is omitted from the post request body.

packages/javascript/bh-shared-ui/src/views/UserProfile/UserProfile.tsx
```diff
@@ -46,8 +46,19 @@ const UserProfile = () => {
...
 
     const updateUserPasswordMutation = useMutation(
-        ({ userId, secret, needsPasswordReset }: { userId: string; secret: string; needsPasswordReset: boolean }) =>
+        ({
+            userId,
+            secret,
+            needsPasswordReset,
+            currentSecret,
+        }: {
+            userId: string;
```
brandonshearin, line 55: it would improve readability if you extracted this type to somewhere above the component.  for example, you could put this before the <UserProfile/> function is defined:

```
type MutationArgs = {
    userId: string;
    currentSecret?: string;
    secret: string;
    needsPasswordReset: boolean;
};
```

then, you can update your declaration of this anonymous func that gets passed into useMutation to be of type MutationArgs.  something like this:

```
({ userId, secret, needsPasswordReset, currentSecret }: MutationArgs) =>
            apiClient.putUserAuthSecret(userId, {
                current_secret: currentSecret,
                needs_password_reset: needsPasswordReset,
                secret: secret,
            }),
```

mistahj67: I think i found an even cleaner way, lmk what you think!

brandonshearin (approved):
super slick update to the js-client code, really cleaned up the components!!

ddlees (approved):
Nice job!

PR #622, opened 2024-05-21
Bed 4345: removing API request timeouts

irshadaj:
## Description

Previous work removed much of the context middleware out of the BloodHound System.  Upon review there are some areas that are still having timeouts set by default and need to be cleaned up.

This MR removes the network timeouts, and adds context checking in signature verification so endpoints with large request bodies will obey timeouts and return immediately.

This has been marked as a breaking change because alteration of the network timeout within the configuration struct will not affect API timeouts any more.

## Motivation and Context

Context Middleware creates errors:"bloodhound-1 | {"level":"warn","time":"<timestamp>","message":"POST /api/v2/file-upload/6 took longer than the configured timeout of 70 seconds"} 

## How Has This Been Tested?
Set up a 70+ second sleep in the file uploads endpoint

Screenshot 2024-05-21 at 1 48 28 PM

When I uploaded a new file, the endpoint returns a success response after 70 seconds- this would have failed with a timeout error before

Screenshot 2024-05-21 at 1 55 26 PM

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/middleware/logging.go
+++ b/cmd/api/src/api/middleware/logging.go
@@ -27,9 +27,7 @@ import (
 	"github.com/specterops/bloodhound/log"
 	"github.com/specterops/bloodhound/src/api"
 	"github.com/specterops/bloodhound/src/auth"
-	"github.com/specterops/bloodhound/src/config"
 	"github.com/specterops/bloodhound/src/ctx"
-	"github.com/specterops/bloodhound/src/database"
 )
 
 // PanicHandler is a middleware func that sets up a defer-recovery trap to capture any unhandled panics that bubble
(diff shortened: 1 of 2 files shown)


cmd/api/src/api/middleware/logging.go
```diff
@@ -133,17 +133,23 @@ func LoggingMiddleware(cfg config.Configuration, idResolver auth.IdentityResolve
 				}
 			)
 
+			// assign a deadline, but only if a valid timeout has been supplied via the prefer header
+			timeout, err := RequestWaitDuration(request)
```
zinic, line 137: What if `err != nil`?

cmd/api/src/config/default.go
```diff
@@ -36,7 +36,6 @@ func NewDefaultConfiguration() (Configuration, error) {
 		return Configuration{
 			Version:                      0,
 			BindAddress:                  "127.0.0.1",
-			NetTimeoutSeconds:            70,  // Default timeout to avoid race conditions with 60 second gateway timeouts
```
zinic, line 39: Are we just getting rid of this configuration element? It looks like all of the potential wire-up for it is now removed as per this changeset.

If the intent is to remove this configuration element we should follow up and review relevant documentation or other mention of it.

zinic (changes requested):
I have to change requests for review for this changeset.

cmd/api/src/config/default.go
zinic: Think we found one last place this is mentioned but I'm going to go ahead and mark this resolved :+1:

zinic (approved):
Changes look good

PR #623, opened 2024-05-21
BED-4299  -- Display members of custom asset groups on Group Management

maffkipp:
## Description
`GET api/v2/asset-groups/{asset_group_id}/members` currently only queries the `system_tags` property to determine AG membership. This means any custom asset groups created through the API will not return any results via this endpoint, since membership in those groups is stored on the `user_tags` property instead.

This fix updates the query to check for the asset group tag in both the `system_tags` and `user_tags` field. I also added a check after the query to filter out nodes that do not have an exact match in the tag list, due to the potential for naming collisions with our `StringContains` query (for example, a custom asset group with tag `test_tag_1` also pulling in nodes with tag `test_tag`).

## Motivation and Context
If a user is creating custom asset groups via the API, we want them to be able to view and modify those asset groups via the Group Management page (or the `api/v2/asset-groups/{asset_group_id}/members` endpoint).

## How Has This Been Tested?
- Added integration tests to our graph query to make sure it returns custom groups
- Manual testing locally by creating a custom asset group, adding members, and checking the endpoint/group management page

Screenshot 2024-05-21 at 2 08 09 PM

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/queries/graph.go
+++ b/cmd/api/src/queries/graph.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


urangel (approved):
Good thinking filtering out substrings and thanks for adding a test!

PR #625, opened 2024-05-22
fix: errantly adding AZHasRole edge to non-tenant scoped principals

mistahj67:
Closes BED-4380
## Description

Stop `AZHasRole` edge from being created for appadmin / cloudappadmin for app scoped relationships
Also cleaned up some unused constants found

## Motivation and Context

These edges were errantly created and shouldn't exist.

## How Has This Been Tested?

Ran Azurehound and verified edge wasn't created.

Run the following in explore
```
match p = ()-[:AZAppAdmin|:AZCloudAppAdmin]->()
return p
```
Pick one and find the path from principal to the app service principal. The edge for `AZHasRole `should be removed

## Types of changes

## Checklist:

### DIFF
--- a/packages/go/analysis/azure/azure.go
+++ b/packages/go/analysis/azure/azure.go
@@ -104,13 +104,3 @@ func AzureNonDescentKinds() graph.Kinds {
 		azure.RunsAs,
 	}
 }
-
-func AzureIgnoredKinds() graph.Kinds {
-	return []graph.Kind{
-		azure.ScopedTo,
-		azure.Contains,
-		azure.GlobalAdmin,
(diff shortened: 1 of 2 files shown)


packages/go/analysis/azure/azure.go
```diff
@@ -104,13 +104,3 @@ func AzureNonDescentKinds() graph.Kinds {
 		azure.RunsAs,
 	}
 }
-
-func AzureIgnoredKinds() graph.Kinds {
```
mistahj67, line 108: Wasn't used so I deleted it. 🧼 🧹

PR #626, opened 2024-05-22
fix: compositions for multi-tier PKI

JonasBK:
## Description

Fix composition for ADCSESCx edges such that it also works when the enterprise CA is in a 2+ tier PKI hierarchy.

Various small issues fixed in composition logic as well.

## Motivation and Context

Ticket: BED-4427

Turned out the bug appeared when the affected enterprise CA is a tier 2+ CA, which means it has a longer IssuedSignedBy path to the RootCA. The fix makes sure we account for that.

## How Has This Been Tested?

Locally with a couple of data sets.

image
image

## Types of changes

## Checklist:


superlinkx (approved):
Sorry this one got lost for so long. Nothing here stands out, seems to follow guidance from the ticket. Pull when ready!

PR #630, opened 2024-05-24
fix: EnrollOnBehalfOf to be based on EffectiveEKUs

JonasBK:
## Description

Change the EnrollOnBehalfOf edge to created based on the EffectiveEKUs property instead of EKUs.

Ticket: BED-4433

## Motivation and Context

We create the EnrollOnBehalfOf edge based on the the EKUs property. It should be the EffectiveEKUs instead. These properties are the same most of the time but not always.

## Types of changes

## Checklist:


superlinkx (approved):
LGTM

PR #632, opened 2024-05-24
Correcting Logic and returning the ticket number

slokie-so:
## Description

Introduces Community PR Ticketing, allowing the workflow to confirm the organization prior to creating a ticket. Updates to Jira Issue Ticketing, resulting in a confirmation workflow after creation to ensure there is enough information collected prior to ticketing to Jira. Updates the incorrect link to the corrected HTML link

## Motivation and Context
Correcting Logic and returning the ticket number
Reduce incomplete information for Bug / Feature Requests, introduce Community PR tickets to allow Engineering to review and merge community PR with normalized workflow

## How Has This Been Tested?

YAML verified through validator plugin
Will need to test live, since self-hosted runner is the only permitted worker for this workflow

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/jira-issue-transfer.yml
+++ b/.github/workflows/jira-issue-transfer.yml
@@ -40,10 +40,10 @@ jobs:
             summary: ${{ github.event.issue.title }}
             description: "Github Issue Link: ${{ github.event.issue.html_url}} \r\n ${{ github.event.issue.body }}"
             fields: '{"labels":["GitHubReport"]}'
-          run: echo "Issue ${{ steps.bug.outputs.issue }} was created"
+            JIRA_TICKET: ${{ steps.bug.outputs.issue }}
         - name: Jira Create Enhancement
           id: enhancement
-          if: contains(github.event.issue.labels.*.name, 'ticketed', 'enhancement') && !contains(github.event.issue.labels.*.name, 'ticketed') 
+          if: contains(github.event.issue.labels.*.name, 'ticketed', 'enhancement')
           uses: atlassian/gajira-create@v3 
           with:
             project: BP
@@ -51,6 +51,9 @@ jobs:
             summary: ${{ github.event.issue.title }}
             description: "Github Issue Link: ${{ github.event.issue.html_url}} \r\n ${{ github.event.issue.body }}"
             fields: '{"labels":["GitHubReport"]}'
-          run: echo "Issue ${{ steps.enhancement.outputs.issue }} was created"
-     
-          
+            JIRA_TICKET: ${{ steps.enhancement.outputs.issue }}
+            
+        - name: Read Ticket Number
+          id: Ticket
+          run: echo "$JIRA_TICKET"
+    
--- a/.github/workflows/jira-issue-transfer.yml
+++ b/.github/workflows/jira-issue-transfer.yml
@@ -41,6 +41,7 @@ jobs:
             summary: ${{ github.event.issue.title }}
             description: "Github Issue Link: ${{ github.event.issue.html_url}} \r\n ${{ github.event.issue.body }}"
             fields: '{"labels":["GitHubReport"]}'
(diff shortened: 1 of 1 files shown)


.github/workflows/jira-issue-transfer.yml
```diff
@@ -31,20 +31,26 @@ jobs:
             JIRA_API_TOKEN: ${{ secrets.JIRA_API_TOKEN }}
             GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         - name: Jira Create Bug
-          if: github.event.label.name == 'ticketed' && !contains(github.event.issue.labels.*.name, 'ticketed') && contains(github.event.issue.labels.*.name, 'bug') 
+          id: bug
+          if: contains(github.event.issue.labels.*.name, 'ticketed', 'bug') && !contains(github.event.issue.labels.*.name, 'ticketed')
```
computator, line 35: This logic seems contradictory? The first checks if it contains `ticketed` and the second makes sure it does _not_ contain `ticketed`.

.github/workflows/jira-issue-transfer.yml
```diff
@@ -31,20 +31,26 @@ jobs:
...
           uses: atlassian/gajira-create@v3
           with:
             project: BP
             issuetype: 'Bug Report'
             summary: ${{ github.event.issue.title }}
             description: "Github Issue Link: ${{ github.event.issue.html_url}} \r\n ${{ github.event.issue.body }}"
             fields: '{"labels":["GitHubReport"]}'
-        - name: Jira Create Enhancement        
-          if: github.event.label.name == 'ticketed' && !contains(github.event.issue.labels.*.name, 'ticketed') && contains(github.event.issue.labels.*.name, 'enhancement')
+          run: echo "Issue ${{ steps.bug.outputs.issue }} was created"
```
computator, line 43: I don't believe you can have `with` and `run` in the same step. The action should have plenty of it's own logging that you can see in the logs section, but if you wanted to have this separate logging it probably either needs to be in it's own step with the appropriate `if` statements, or you can rearrange these to have 2 separate jobs each with their own job-level `if` statement and sequence of steps.

.github/workflows/jira-issue-transfer.yml
computator: In addition, `contains()` only supports 2 parameters

https://docs.github.com/en/actions/learn-github-actions/expressions#contains

computator (changes requested):
This reference page is super useful if you haven't looked at it yet

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions

.github/workflows/jira-issue-transfer.yml
```diff
@@ -31,20 +31,28 @@ jobs:
...
             issuetype: 'Product Feature'
             summary: ${{ github.event.issue.title }}
             description: "Github Issue Link: ${{ github.event.issue.html_url}} \r\n ${{ github.event.issue.body }}"
-            fields: '{"labels":["GitHubReport"]}'
\ No newline at end of file
+            fields: '{"labels":["GitHubReport"]}'
+            
+        - name: Read Ticket Number
+          id: Ticket
+          echo "Ticket Number: ${{ steps.enhancement.outputs.issue }} ${{ steps.bug.outputs.issue }}"
```
computator, line 57: Each of the steps is basically an array of options, so you can't put code directly in line with it. The text of the code would need to be passed as the value of the `run` key like this
```suggestion
          run: echo "Ticket Number: ${{ steps.enhancement.outputs.issue }} ${{ steps.bug.outputs.issue }}"
```
or in multiline/block syntax like this:
```suggestion
          run: |
            echo "Ticket Number: ${{ steps.enhancement.outputs.issue }} ${{ steps.bug.outputs.issue }}"
```

PR #643, opened 2024-06-04
Community PR Ticketing for Jira

slokie-so:
## Description

Introduces Community PR Ticketing, allowing the workflow to confirm the organization prior to creating a ticket. 

## Motivation and Context

Introduce Community PR tickets to allow Engineering to review and merge community PR with normalized workflow

## How Has This Been Tested?
YAML verified through validator plugin
Reviewed previously by BHE Engineering

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/pr-jira-issue-transfer.yml
+++ b/.github/workflows/pr-jira-issue-transfer.yml
@@ -1,0 +1,48 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+name: External Authorship Check
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  check-author:
+    runs-on: self-hosted
+    steps:
+      - name: Check if PR is authored by an external contributor
+        run: |
+          # Fetch the author association of the PR
+          AUTHOR_ASSOCIATION=$(jq --raw-output .pull_request.author_association $GITHUB_EVENT_PATH)
+          
+          # Define the list of allowed author associations for team members
+          TEAM_ASSOCIATIONS=("MEMBER")
+          
+          # Check if the author association is not in the team list
+          if [[ ! " ${TEAM_ASSOCIATIONS[@]} " =~ " ${AUTHOR_ASSOCIATION} " ]]; then
+            - name: Login
+              uses: atlassian/gajira-login@v3        
+    
+            - name: Jira Create
+              uses: atlassian/gajira-create@v3
+              with:
+                  project: BP
+                  issuetype: 'Product Feature'
+                  summary: ${{ github.event.pulls.title }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-jira-issue-transfer.yml
```diff
@@ -0,0 +1,48 @@
...
+              uses: atlassian/gajira-login@v3        
+    
+            - name: Jira Create
+              uses: atlassian/gajira-create@v3
+              with:
+                  project: BP
+                  issuetype: 'Product Feature'
+                  summary: ${{ github.event.pulls.title }}
+                  description: "Github Pull Request Link: ${{ github.event.pulls.html_url}} \r\n ${{ github.event.pulls.body }}"
+                  fields: '{"labels":["GitHubReport"]}'
```
computator, line 46: This is nested inside a bash `run:` block. You can't mix YAML and bash syntax or use bash as a templating language.

.github/workflows/pr-jira-issue-transfer.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Check if PR is authored by an external contributor
+        run: |
+          # Fetch the author association of the PR
+          AUTHOR_ASSOCIATION=$(jq --raw-output .pull_request.author_association $GITHUB_EVENT_PATH)
+          
+          # Define the list of allowed author associations for team members
+          TEAM_ASSOCIATIONS=("MEMBER")
+          
+          # Check if the author association is not in the team list
+          if [[ ! " ${TEAM_ASSOCIATIONS[@]} " =~ " ${AUTHOR_ASSOCIATION} " ]]; then
```
computator, line 35: I think these three checks should be able to be handled as a `if:` expression on the job. It should work with `contains()` in combination with the `fromJSON()` method similar to here https://docs.github.com/en/actions/learn-github-actions/expressions#example-matching-an-array-of-strings.

computator (changes requested):

.github/workflows/pr-jira-issue-transfer.yml
slokie-so: Uploaded the incorrect version,  corrected version and feel free to review

.github/workflows/pr-jira-issue-transfer.yml
```diff
@@ -23,26 +23,20 @@ jobs:
...
+    - name: checkUserMember
+      uses: marcocarvalho/team-membership@v3
+      with:
+        username: ${{ github.actor }}
+        team: 'bloodhound-enterprise'
+            
+    - name: Login
+      uses: atlassian/gajira-login@v3        
+    - name: Jira Create
+      if: steps.checkUserMember.outputs.isTeamMember == 'false'
```
computator, line 35: Despite the documentation included with the `team-membership` action, the `isTeamMember` output is actually a boolean not a string.
```suggestion
      if: steps.checkUserMember.outputs.isTeamMember == false
```

Also, you could include this same `if:` expression on the gajira-login step as well to prevent running it unnecessarily.

.github/workflows/pr-jira-issue-transfer.yml
```diff
@@ -23,26 +23,20 @@ jobs:
...
+      uses: atlassian/gajira-create@v3
+      with:
+        project: BP
+        issuetype: 'Product Feature'
```
computator, line 39: YAML is pretty flexible with strings, and as long as there are no special characters they can be unquoted. All your quoting is fine as is, but if you feel so inclined you can remove quotes on this line as well as on `team` in the team membership check.

`description` requires quotes due to containing a colon with a space after it (`: `) as well as the escape sequences. `fields` also requires quoting to pass it as a json string instead of interpreting it as a yaml expression since yaml is a superset of json.

computator (approved):
One minor type update is needed, but otherwise I believe this should work!

.github/workflows/pr-jira-issue-transfer.yml
computator: Oh, here's a handy tool that may be useful in the future:
https://eemeli.org/yaml-playground/

PR #649, opened 2024-06-07
BED-3671  -- add group shortcutting to SyncLAPSPassword Postprocessing

maffkipp:
## Description
Previously the logic for creating `SyncLAPSPassword` edges needed to query for group membership for all groups that had one of the two permissions. This change takes advantage of the group expansion we are already doing in a separate postprocessing step to avoid unnecessary db queries.

Additionally, the `SyncLAPSPassword` edge should no longer be created for members of groups that already have that edge, which should also reduce unnecessary noise in the graph; this difference is accounted for in the accompanying test harness.

Once this is merged, I can get another PR up to make the same change for `DCSync`.

## Motivation and Context
- Improve analysis performance by eliminating unnecessary group expansion
- avoid creating redundant `SyncLAPSPassword` edges in the graph

## How Has This Been Tested?
I've added a new integration test + harness to ensure these changes are doing what we expect

## Types of changes

## Checklist:


urangel (approved):
Looks good!

PR #653, opened 2024-06-12
draft: save datapipe status in postgres

brandonshearin:
## Types of changes

## Checklist:


cmd/api/src/database/datapipe_status.go
```diff
@@ -0,0 +1,54 @@
...
+
+		return tx.Error
+	}
+
+}
+
+func (s *BloodhoundDB) GetDatapipeStatus(ctx context.Context) (model.DatapipeStatusWrapper, error) {
+	var datapipeStatus model.DatapipeStatusWrapper
+
+	if tx := s.db.WithContext(ctx).Raw("select * from datapipe_status limit 1;").Scan(&datapipeStatus); tx.RowsAffected == 0 {
```
mistahj67, line 49: I think it might be a good thing to be explicit here
```suggestion
	if tx := s.db.WithContext(ctx).Raw("SELECT status, updated_at, last_updated_at FROM datapipe_status LIMIT 1;").Scan(&datapipeStatus); tx.RowsAffected == 0 {
```

cmd/api/src/database/datapipe_status.go
```diff
@@ -0,0 +1,54 @@
...
+
+		return tx.Error
+	} else {
+		updateSql += ";"
+		tx := s.db.WithContext(ctx).Exec(updateSql, status, now)
+
+		return tx.Error
+	}
+
+}
```
mistahj67, line 44: nit for possible better readability?
```suggestion
	if updateAnalysisTime {
		updateSql += ", last_complete_analysis_at = ?;"
		return s.db.WithContext(ctx).Exec(updateSql, status, now, now).Error
	} else {
		updateSql += ";"
		return s.db.WithContext(ctx).Exec(updateSql, status, now).Error
	}
}
```

cmd/api/src/database/datapipe_status_test.go
```diff
@@ -0,0 +1,53 @@
...
+	require.Equal(t, model.DatapipeStatusAnalyzing, status.Status)
+
+	// when `SetDatapipeStatus` is called with `true` for the `updateAnalysisTime` parameter, assert that the time is no longer null
+	require.Empty(t, status.LastCompleteAnalysisAt)
+	err = db.SetDatapipeStatus(testCtx, model.DatapipeStatusAnalyzing, true)
+	require.Nil(t, err)
+	status, err = db.GetDatapipeStatus(testCtx)
+	require.Nil(t, err)
+	require.NotEmpty(t, status.LastCompleteAnalysisAt)
+
```
mistahj67, line 52: Might I suggest a slightly more robust test
```suggestion
	// when `SetDatapipeStatus` is called with `true` for the `updateAnalysisTime` parameter, assert that the time is no longer null
	require.True(t, status.LastCompleteAnalysisAt.IsZero())
	err = db.SetDatapipeStatus(testCtx, model.DatapipeStatusAnalyzing, true)
	require.Nil(t, err)
	status, err = db.GetDatapipeStatus(testCtx)
	require.Nil(t, err)
	require.True(t, !status.LastCompleteAnalysisAt.IsZero())
```

cmd/api/src/daemons/datapipe/datapipe.go
mistahj67, line 1: I see there's a reference to `./mocks/tasker.go` in a go/generate comment at the top of this file.

I don't think that's needed anymore

mistahj67 (commented):
Mostly nits, otherwise looks solid!

mistahj67 (approved):
Code LGTM, great cleanup! 🧼

PR #661, opened 2024-06-21
BED-4366 -- parse negative numbers in Cypher queries

maffkipp:
## Description
Our arithmetic expression parsing for cypher queries was silently dropping the operator from negative numbers. This change adds a model for unary expressions that will carry the operator through and allow us to handle it as a separate case in the backend.

## Motivation and Context
Some of our saved queries use negative numbers for matching by property values, and we are currently returning unexpected results to the user when they use those queries.

## How Has This Been Tested?
Additional positive test cases have been added to confirm we now allow negative numbers, and that this does not otherwise impact the way we parse arithmetic expressions.

## Types of changes

## Checklist:


superlinkx (approved):
This looks clean, pull!

zinic (approved):

elikmiller:
I'm looking in the code for something that says `if(nextToken === "-") return currentResult * -1;` but my small brain cannot comprehend this changeset! 🥲

PR #664, opened 2024-06-24
BED-4485

stephanieslamb:
## Description

Added additional unit tests for existing functionality

## Motivation and Context

This change was to help get my feet wet in the codebase as well as adding some code coverage.  
https://specterops.atlassian.net/browse/BED-4485

## How Has This Been Tested?

No testing other than successful running of unit tests.

## Types of changes

-   [X] Chore (a change that does not modify the application functionality)

## Checklist:

-   [X] I have added and/or updated tests to cover my changes.
-   [X] All new and existing tests passed.


stephanieslamb:

cmd/api/src/model/audit_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	actionEntry2 := AuditEntry{
+		CommitID: uuid.Must(uuid.NewV4()),
+		Action:   "invalid action",
+		Model:    MockAuditable{AuditData{"key": "value"}},
+		Status:   "ok",
+		ErrorMsg: "err",
+	}
+
+	assert.False(t, actionEntry1.Matches(actionEntry2), "Expected actionEntry1 to match actionEntry2")
```
maffkipp, line 55: I think the error message here (and for the following tests that use `assert.False()`) might be flipped, since we are testing that the two entries do not match

stephanieslamb: i think this is a copy pasta error 🫠 my bad.  i think that this should be false, but the message should be "expected entry1 not to match entry2".  unless i am misunderstanding the function, i think it is checking that all fields match besides the commitid. so the first test case is checking that scenario and then the others are checking for mismatches on the other fields.  or thats how i thought about it in my head 😅  let me know if that is not the case.

maffkipp: That is my understanding too, the tests themselves all make sense to me!

maffkipp (approved):

PR #665, opened 2024-06-24
BED-4435: Enforce AG name tag uniqueness

iustinum:
## Description

Modified DB schema to prevent creation of asset groups with empty names/tags or duplicate tags.

## Motivation and Context

Previously, asset groups could be created with tags that duplicate existing tags or with empty names/tags. This could potentially cause widespread bugs in analysis and display given our extensive reliance on string matching for these tags.

## How Has This Been Tested?

- Updated application layer unit test function `TestResources_CreateAssetGroup`
- Added new test cases for:
  - Empty asset group name
  - Empty asset group tag
  - Whitespace in asset group tag
- Added new test file DB layer unit test function `migration/asset_group_constraints_test.go` to validate new schema constraints
- Verified compatibility with existing DB migration tests for asset groups
- Manually verified functionality through API Explorer UI

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/error.go
+++ b/cmd/api/src/api/error.go
@@ -63,6 +63,7 @@ const (
 	ErrorResponseAGNameTagEmpty						= "asset group name or tag must not be empty"
 	ErrorResponseAGDuplicateName					= "asset group name must be unique"
 	ErrorResponseAGDuplicateTag						= "asset group tag must be unique"
+	ErrorResponseDetailsUniqueViolation 			= "A unique constraint was violated"
 
 	FmtErrorResponseDetailsBadQueryParameters = "there are errors in the query parameters: %v"
 )
@@ -121,10 +122,6 @@ func HandleDatabaseError(request *http.Request, response http.ResponseWriter, er
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusNotFound, ErrorResponseDetailsResourceNotFound, request), response)
 	} else if errors.Is(err, context.DeadlineExceeded) {
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusInternalServerError, ErrorResponseRequestTimeout, request), response)
-	} else if err.Error() == "asset group with this name already exists" {
-		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusConflict, ErrorResponseAGDuplicateName, request), response)
-	} else if err.Error() == "asset group with this tag already exists" {
-		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusConflict, ErrorResponseAGDuplicateTag, request), response)
 	} else {
 		log.Errorf("Unexpected database error: %v", err)
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusInternalServerError, ErrorResponseDetailsInternalServerError, request), response)
--- a/cmd/api/src/api/error.go
+++ b/cmd/api/src/api/error.go
@@ -63,7 +63,7 @@ const (
 	ErrorResponseAGNameTagEmpty						= "asset group name or tag must not be empty"
 	ErrorResponseAGDuplicateName					= "asset group name must be unique"
 	ErrorResponseAGDuplicateTag						= "asset group tag must be unique"
-	ErrorResponseDetailsUniqueViolation 			= "A unique constraint was violated"
+	ErrorResponseDetailsUniqueViolation 			= "unique constraint was violated"
 
 	FmtErrorResponseDetailsBadQueryParameters = "there are errors in the query parameters: %v"
 )
--- a/cmd/api/src/api/error.go
+++ b/cmd/api/src/api/error.go
@@ -60,6 +60,9 @@ const (
 	ErrorResponseRequestTimeout                     = "request timed out"
 	ErrorResponseUserSelfDisable                    = "user attempted to disable themselves"
 	ErrorResponseAGTagWhiteSpace                    = "asset group tags must not contain whitespace"
+	ErrorResponseAGNameTagEmpty						= "asset group name or tag must not be empty"
+	ErrorResponseAGDuplicateName					= "asset group name must be unique"
+	ErrorResponseAGDuplicateTag						= "asset group tag must be unique"
 
 	FmtErrorResponseDetailsBadQueryParameters = "there are errors in the query parameters: %v"
 )
@@ -118,6 +121,10 @@ func HandleDatabaseError(request *http.Request, response http.ResponseWriter, er
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusNotFound, ErrorResponseDetailsResourceNotFound, request), response)
 	} else if errors.Is(err, context.DeadlineExceeded) {
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusInternalServerError, ErrorResponseRequestTimeout, request), response)
+	} else if err.Error() == "asset group with this name already exists" {
+		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusConflict, ErrorResponseAGDuplicateName, request), response)
+	} else if err.Error() == "asset group with this tag already exists" {
+		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusConflict, ErrorResponseAGDuplicateTag, request), response)
 	} else {
 		log.Errorf("Unexpected database error: %v", err)
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusInternalServerError, ErrorResponseDetailsInternalServerError, request), response)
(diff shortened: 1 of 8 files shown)


iustinum:
recheck

iustinum:

iustinum:
recheck

cmd/api/src/api/error.go
```diff
@@ -114,10 +118,15 @@ func BuildErrorResponse(httpStatus int, message string, request *http.Request) *
 // HandleDatabaseError writes an error (not found or other) depending on the database error encountered
 // Alternate: FormatDatabaseError()
 func HandleDatabaseError(request *http.Request, response http.ResponseWriter, err error) {
+	log.Infof("Handling database error: %v", err)
```
mistahj67, line 121: I'm not sure we want to log this as `Info` as it could get quite noisy

cmd/api/src/api/error.go
```diff
@@ -114,10 +118,15 @@ func BuildErrorResponse(httpStatus int, message string, request *http.Request) *
...
 	if errors.Is(err, database.ErrNotFound) {
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusNotFound, ErrorResponseDetailsResourceNotFound, request), response)
 	} else if errors.Is(err, context.DeadlineExceeded) {
 		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusInternalServerError, ErrorResponseRequestTimeout, request), response)
+	} else if strings.Contains(err.Error(), "duplicate key value violates unique constraint \"asset_groups_name_key\"") {
+		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusConflict, ErrorResponseAGDuplicateName, request), response)
+	} else if strings.Contains(err.Error(), "duplicate key value violates unique constraint \"asset_groups_tag_key\"") {
+		WriteErrorResponse(request.Context(), BuildErrorResponse(http.StatusConflict, ErrorResponseAGDuplicateTag, request), response)
```
mistahj67, line 129: This function is used in a number of areas outside of just asset groups, I think we either should generalize this further and / or find a home for these potentially higher up the stack. 

Futher, I think the `errors.Is()` is a must more elegant way to check these if it's possible to leverage that

cmd/api/src/database/migration/migrations/schema.sql
```diff
@@ -123,8 +123,8 @@ CREATE SEQUENCE IF NOT EXISTS asset_group_selectors_id_seq
 ALTER SEQUENCE asset_group_selectors_id_seq OWNED BY asset_group_selectors.id;
 
 CREATE TABLE IF NOT EXISTS asset_groups (
-    name text,
-    tag text,
+    name text UNIQUE NOT NULL,
+    tag text UNIQUE NOT NULL,
```
mistahj67, line 127: It's my understanding that this schema.sql is not run a second time and is more akin to "migration 0" so changes to this file won't be reflected and we should use another migration .sql file. But please get confirmation on that one! 🙇

mistahj67 (changes requested):

cmd/api/src/api/error.go
iustinum: That was for debugging - I will remove it!

cmd/api/src/database/migration/migrations/schema.sql
computator, line 127: I don't believe we would want to change this since it's the original migration which will have already been applied. We should only make changes in new migrations.

iustinum: Removed changes in `schema.sql` and applied to `v5.12.0.sql`

cmd/api/src/api/error.go
iustinum: I've placed the checks in `CreateAssetGroup` in `bhce/cmd/api/src/api/v2/agi.go` to keep `HandleDatabaseError` more general.

cmd/api/src/database/migration/migrations/schema.sql
iustinum: Resolved
https://github.com/SpecterOps/BloodHound/pull/665#discussion_r1657562087

cmd/api/src/api/v2/agi_test.go
```diff
@@ -25,6 +25,8 @@ import (
 	"net/url"
 	"testing"
 
+
```
mistahj67, line 28:
```suggestion
```

cmd/api/src/api/v2/agi_test.go
```diff
@@ -405,8 +407,34 @@ func TestResources_CreateAssetGroup(t *testing.T) {
...
+	require.Nil(t, err)
+
+	requestTemplate.
+		WithContext(&ctx.Context{
+			Host: &url.URL{},
+		}).
+		WithBody(jsonBody).
+		OnHandlerFunc(resources.CreateAssetGroup).
+		Require().
+		ResponseStatusCode(http.StatusBadRequest)
```
mistahj67, line 434: same as above

cmd/api/src/api/v2/agi_test.go
```diff
@@ -405,8 +407,34 @@ func TestResources_CreateAssetGroup(t *testing.T) {
...
```
mistahj67, line 421: I think you can also check that the response is the right description as well but if that ends up being to annoying with this testing framework, I'd not worry too much about it.

cmd/api/src/database/agi.go
```diff
@@ -43,9 +46,18 @@ func (s *BloodhoundDB) CreateAssetGroup(ctx context.Context, name, tag string, s
 		err error
 	)
 
-	err = s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
-		return CheckError(tx.Create(&assetGroup))
-	})
+    err = s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+        err := tx.Create(&assetGroup).Error
+        if err != nil {
+            log.Infof("Error creating asset group: %v", err)
```
mistahj67, line 52: Is this log intentional or just part of debugging?

cmd/api/src/database/migration/migrations/v5.12.0.sql
```diff
@@ -29,4 +29,10 @@ VALUES
 ON CONFLICT DO NOTHING;
 
 ALTER TABLE user_sessions
-    ADD COLUMN IF NOT EXISTS flags jsonb;
\ No newline at end of file
+    ADD COLUMN IF NOT EXISTS flags jsonb;
+
+ALTER TABLE asset_groups
```
mistahj67, line 34: We want to ensure these migrations are idempotent. 

There's not an `IF NOT EXISTS` option for `ADD CONSTRAINT` so you have to drop it with `DROP CONSTRAINT IF EXISTS` and then add it.

mistahj67 (changes requested):
Looks great so far, the blocking change request is the migration idempotency, the rest are just nits / suggestions

cmd/api/src/database/agi.go
iustinum: Debugging - removed.

cmd/api/src/database/migration/migrations/v5.12.0.sql
iustinum: This brings up another potential issue with running the migration script: handling existing null or duplicate entries to maintain idempotence. The script could fail when adding constraints if it encounters null or duplicated entries in the existing database.

Referencing the original Jira ticket about duplicate names/tags, which suggests appending a number to duplicate asset group tags: Should our script preemptively address existing null or duplicate entries before executing this migration? If the ALTER TABLE statement encounters NULL values in the name or tag column, it will fail.

I propose the following approach to ensure robustness:

1. Clean up any NULL values by assigning them default values.
2. Verify the absence of NULLs before enforcing the NOT NULL constraint.
3. Apply the NOT NULL constraint only if no NULL values are found.

Additionally, I believe we should integrate logic to rename existing duplicates. Your thoughts on this approach and any additional considerations?

iustinum:
Per https://github.com/SpecterOps/BloodHound/pull/665#discussion_r1658152670, this PR is marked as blocked as we work on a solution to safely migrate database to `v5.12.0` while preserving the integrity of the existing database.

cmd/api/src/database/migration/migrations/v5.12.0.sql
```diff
@@ -28,5 +28,23 @@ VALUES
     ('idle', NOW ())
 ON CONFLICT DO NOTHING;
 
-ALTER TABLE user_sessions
-    ADD COLUMN IF NOT EXISTS flags jsonb;
\ No newline at end of file
+ALTER TABLE IF EXISTS user_sessions
```
mistahj67, line 31: Nice catch! 🙇

mistahj67 (approved):
Code LGTM! 🚀

PR #667, opened 2024-06-25
fix: missing name property on azure apps spawned by service principals

mistahj67:
## Description
CLOSES BED-4429

Added a fallback utility fn to graph node properties.
Updated /search response nodes to fallback to display name and then object id if not name property does not exist.

## Motivation and Context

A number of applications are missing names b/c they were ingested via their respective service principal 

## How Has This Been Tested?

Ran locally

## Screenshots
Before:
image

After
image

## Types of changes

## Checklist:

### DIFF
--- a/cmd/api/src/api/bloodhoundgraph/conversions.go
+++ b/cmd/api/src/api/bloodhoundgraph/conversions.go
@@ -1,17 +1,17 @@
 // Copyright 2023 Specter Ops, Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-// 
+//
 //     http://www.apache.org/licenses/LICENSE-2.0
-// 
+//
 // Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


cmd/api/src/api/bloodhoundgraph/conversions.go
```diff
@@ -31,6 +32,7 @@ const (
 func NodeToBloodHoundGraph(node *graph.Node) BloodHoundGraphNode {
 	var (
 		nodeKindLabel       = analysis.GetNodeKindDisplayLabel(node)
+		name, _             = node.Properties.GetWithFallback(common.Name.String(), "NO NAME OR ID", common.ObjectID.String()).String()
```
rvazarkar, line 35: If we're making these changes, we should probably line up the behavior between the two. Right now, the search endpoint has an additional case this one doesn't. Otherwise, I like this solution a lot

mistahj67: Yes I intentionally kept it the same as it previously was to limit the change scope, but I'm all for consistency if we want to add display name as another fallback!

rvazarkar (approved):

PR #669, opened 2024-06-26
chore: added more base schemas to new openapi docs

sircodemane:
## Description

Added more schemas to the new openapi docs in preparation for paths

## Motivation and Context

This PR addresses: BED-4141

The last of the prep work to begin adding paths

## Types of changes

## Checklist:

  - Open API docs

### DIFF
--- a/packages/go/openapi/src/schemas/null.int64.yaml
+++ b/packages/go/openapi/src/schemas/null.int64.yaml
@@ -1,0 +1,24 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+type: object
+properties:
+  int64:
+    type: integer
+    format: int64
+  valid:
+    description: Valid is true if `int64` is not `null`.
+    type: boolean
--- a/packages/go/openapi/src/schemas/api.request.user.set-secret.yaml
+++ b/packages/go/openapi/src/schemas/api.request.user.set-secret.yaml
@@ -1,0 +1,22 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+type: object
+properties:
+  secret:
+    type: string
+  needs_password_reset:
+    type: boolean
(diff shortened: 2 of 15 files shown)


packages/go/openapi/src/schemas/null.int64.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+type: object
+properties:
+  int64:
+    type: integer
+    format: int64
+  valid:
+    description: Valid is true if `int64` is not `null`.
```
neumachen, line 23: What about `0`?

neumachen (commented):
Looks good to me. I do want to know if:

- Do the `int` types consider `0` valid?
- For the `uuid`, do we consider empty strings, or spaces valid?
- For `date-time`, do we have a specific format in mind? ISO? And is anything but null valid?
- For the following objects, should we be also defining required attributes, both for requests, and response? And I do think it's prudent to define the required values in responses to so we can convey the response object's expected structure.
    - [packages/go/openapi/src/schemas/api.request.user.set-secret.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-dd9c950243c1211750f1922f0e20eba93def086982533f398390151177180dcb)
    - [packages/go/openapi/src/schemas/api.request.user.update.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-f74142f6783a2ea8a8202e608b5c23c8fce805947c4cf23936c97026c5d3e294)
    - [packages/go/openapi/src/schemas/api.request.user.set-secret.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-dd9c950243c1211750f1922f0e20eba93def086982533f398390151177180dcb)
    - [packages/go/openapi/src/schemas/api.request.user.set-secret.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-c0e01ac432511153eea244499761dbb167af85e77740355ec15c25ee393101d0)
    - [packages/go/openapi/src/schemas/api.response.time-window.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-912e9678a658ecb3913bdc4c3aba7bfcc6cbe27c7ac47c4cf6c65948b766424d)

sircodemane:
> * Do the `int` types consider `0` valid?
> * For the `uuid`, do we consider empty strings, or spaces valid?
> * For `date-time`, do we have a specific format in mind? ISO? And is anything but null valid?

These are `null` type definitions, the valid flag is only for the purpose of identifying if a value is null or not.
A null value is `valid = false`, anything else is `valid = true`. example: https://pkg.go.dev/database/sql#NullInt32

For `date-time`, we use RFC-3339 and it's called out in most places where a description for `date-time` exists. This was a miss, i have updated it with a description.

> * For the following objects, should we be also defining required attributes, both for requests, and response? And I do think it's prudent to define the required values in responses to so we can convey the response object's expected structure.
>   
>   * [packages/go/openapi/src/schemas/api.request.user.set-secret.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-dd9c950243c1211750f1922f0e20eba93def086982533f398390151177180dcb)
>   * [packages/go/openapi/src/schemas/api.request.user.update.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-f74142f6783a2ea8a8202e608b5c23c8fce805947c4cf23936c97026c5d3e294)
>   * [packages/go/openapi/src/schemas/api.request.user.set-secret.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-dd9c950243c1211750f1922f0e20eba93def086982533f398390151177180dcb)
>   * [packages/go/openapi/src/schemas/api.request.user.set-secret.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-c0e01ac432511153eea244499761dbb167af85e77740355ec15c25ee393101d0)
>   * [packages/go/openapi/src/schemas/api.response.time-window.yaml](https://github.com/SpecterOps/BloodHound/pull/669/files#diff-912e9678a658ecb3913bdc4c3aba7bfcc6cbe27c7ac47c4cf6c65948b766424d)

This is a valid call out. For context, this is a fraction of an 18k line changeset for completely rewriting our openapi spec.
It is considered only the first iteration of a long term effort to improve our documentation. The main focus was improved source structure and accuracy of schema shapes. A good bit of nuance (like required fields) has been omitted for now and will be addressed in the future.

ddlees (approved):
😋

PR #670, opened 2024-06-26
chore: move type to api client

urangel:
## Description
Updates the type location for datapipestatus to be defined in the client.

## Motivation and Context

This PR addresses: BED-4490 indirectly

Cleaner and consolidated response typing that will be used in the accompanying BHE MR.
## How Has This Been Tested?
No material changes only typing. UI still builds so type checking is good.

## Types of changes

## Checklist:

  - Open API docs

### DIFF
--- a/cmd/ui/src/ducks/global/types.ts
+++ b/cmd/ui/src/ducks/global/types.ts
@@ -49,12 +49,6 @@ export interface Notification {
     options: any;
 }
 
-export interface DatapipeStatus {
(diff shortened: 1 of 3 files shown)


mistahj67 (approved):
Looks like there's an adjacent reference in `cmd/ui/src/mocks/factories.ts` as well that isn't used and can be cleaned up. But otherwise, code LGTM

elikmiller (approved):

PR #671, opened 2024-06-26
BED-4488 (round 2) - Code cleanup and additional UX improvements on helper texts

StephenHinck:
## Description

Perform additional code cleanup from BED-4488 and address feedback from UX

## Motivation and Context

This PR addresses: BED-4488

## How Has This Been Tested?

Validated in local environment plus prep-for-codereview

## Screenshots (optional):
image
image
image
image

## Types of changes

## Checklist:

  - Open API docs


cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -222,7 +222,7 @@ const DatabaseManagement = () => {
     return (
         <ContentPage title='Clear BloodHound Data'>
             <Box>
-                <Typography variant='body1'>
+                <Typography variant='body2'>
```
urangel, line 225: These helper texts seem like they are sub headers to the page. I would recommend using the component prop with the typography component to render these with elements that are more semantically appropriate like an h2. It might even make sense to update the ContentPage component to accept props for different headers and handle the rendering there similar to the `title` prop that it currently uses. 

Disregard if you feel that these helper texts should not be considered headers.

cmd/ui/src/views/QA/QA.tsx
```diff
@@ -60,10 +60,11 @@ const QualityAssurance: React.FC = () => {
 
     if (!contextType || (!contextId && (contextType === 'active-directory' || contextType === 'azure'))) {
         return (
-            <ContentPage
-                title='Data Quality'
-                data-testid='data-quality'
-                actionButton={
```
urangel, line 66: Since we are no longer displaying the data selector in a position inline with the header we should remove this actionButton prop from the ContentPage component as it is no longer being used. We would also want to remove the wrapping flex container from the ContentPage and instead have only the Typography h1 element and the children within the encapsulating div.

packages/javascript/bh-shared-ui/src/components/ContentPage/ContentPage.tsx
```diff
@@ -32,9 +32,6 @@ const ContentPage: React.FC<ContentPageProps> = ({ title, children, actionButton
                         <Typography variant='h1'>{title}</Typography>
                         {actionButton}
                     </Box>
-                    <Box mt={2} mb={4}>
-                        <Divider />
-                    </Box>
```
urangel, line 37: Removing this divider from the ContentPage component and also removing the actionButton prop as suggested in a different comment will make it essentially the same as the PageWithTitle component. We should probably scrap one or the other.

packages/javascript/bh-shared-ui/src/components/FileIngest/FileIngest.tsx
```diff
@@ -58,37 +58,33 @@ const FileIngest = () => {
...
-                    paginationProps={{
-                        page,
-                        rowsPerPage,
-                        count: totalCount,
-                        onPageChange: handlePageChange,
-                        onRowsPerPageChange: handleRowsPerPageChange,
-                    }}
-                />
-            </ContentPage>
+            <Box display='flex' justifyContent='flex-end' alignItems='center' minHeight='24px' mb={2}>
```
urangel, line 68: Would you please update the `mb` prop on this box to instead be an `my` prop? If one is viewing this at a smaller resolution, the help text is too close to the button.

only margin on bottom at smaller resolution:

Screenshot 2024-06-26 at 4 22 49 PM


after updating to `my`:

Screenshot 2024-06-26 at 4 26 54 PM

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
```diff
@@ -92,11 +92,10 @@ const DownloadCollectors = () => {
...
         <PageWithTitle title='Download Collectors' data-testid='download-collectors'>
-            <Typography variant='body1'>
-                <p>To get started, collect data using SharpHound or AzureHound.</p>
-                <p>
-                    BloodHound CE supports both {sharpHoundCELink} or {azureHoundCELink} collectors.
-                </p>
+            <Typography variant='body2' paragraph={true}>
+                To get started, collect data using SharpHound or AzureHound.
+                <br />
+                BloodHound CE supports both {sharpHoundCELink} or {azureHoundCELink} collectors.
```
urangel, line 98: Would it make more sense for this to read "BloodHound CE supports both {sharpHoundCELink} _**and**_ {azureHoundCELink} collectors." instead of using the `or` conjunction?

packages/javascript/bh-shared-ui/src/components/ContentPage/ContentPage.tsx
StephenHinck: The reason for the change was to make them stylistically consistent. The biggest remaining difference between them appears to be content offsets - ContentPage justifies all the way left, while PageWithTitle leaves some whitespace. It makes ContentPage appear nicer for standalone pages (Download Collectors, My Profile), while PageWithTitle is better for pages under Admin.

I'm happy to collapse the two. Is there a preference for which should remain?

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
StephenHinck: The intention is to provide a basic description and/or help links to guide users to what to expect. Usually, I expect an H2 to be a title for some specific section of content on the page. So while I'm not sure a sub-header is correct in this case, perhaps this is a new `pageDescription` prop that automatically puts the content in a box with a consistent variant instead of applying `body2` on each page?

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
StephenHinck: Yup, got it!

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
urangel: Yea I see what you're saying and agree. I may have been envisioning something like 
```html
<header>
  <h1>title</h1>
  <p>page description</p>
</header>
```
so what you are describing with `pageDescription` makes sense.

packages/javascript/bh-shared-ui/src/components/ContentPage/ContentPage.tsx
urangel: I would keep PageWithTitle and modify it like so:

```tsx
import { Box, Container, Typography, ContainerProps } from '@mui/material';
import React from 'react';

type PageWithTitleProps = ContainerProps<
    'div',
    {
        title?: string;
        pageDescription?: JSX.Element;
        children?: React.ReactNode;
    }
>;

const PageWithTitle: React.FC<PageWithTitleProps> = ({ title, pageDescription, children, ...rest }) => {
    return (
        <Container maxWidth='xl' {...rest}>
            <Box component={'header'} my={2}>
                {title && <Typography variant='h1'>{title}</Typography>}
                {pageDescription}
            </Box>
            {children}
        </Container>
    );
};

export default PageWithTitle;
```

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
```diff
@@ -67,38 +67,19 @@ const DownloadCollectors = () => {
...
-            </Typography>
-
+        <PageWithTitle
+            title='Download Collectors'
+            data-testid='download-collectors'
+            pageDescription={
+                <Typography variant='body2' paragraph>
+                    To get started, collect data using SharpHound or AzureHound.
+                    <br />
+                    BloodHound CE supports both {DocumentationLinks.sharpHoundCELink} and{' '}
```
mistahj67, line 79:
```suggestion
                    BloodHound CE supports both {DocumentationLinks.sharpHoundCELink} and
```

cmd/ui/src/views/SAMLConfiguration/SAMLConfiguration.tsx
```diff
@@ -71,25 +71,19 @@ const SAMLConfiguration: React.FC = () => {
...
-                <Typography variant='body1'>
-                    <p>
-                        BloodHound supports SAML for single sign-on (SSO). Learn how to deploy SAML {samlConfigDocLink}.
-                    </p>
-                </Typography>
+            <PageWithTitle
+                title='SAML Configuration'
+                data-testid='saml-configuration'
+                pageDescription={
+                    <Typography variant='body2' paragraph={true}>
```
mistahj67, line 82:
```suggestion
                    <Typography variant='body2' paragraph>
```

cmd/ui/src/views/SAMLConfiguration/SAMLConfiguration.tsx
```diff
@@ -71,25 +71,19 @@ const SAMLConfiguration: React.FC = () => {
...
+                        BloodHound supports SAML for single sign-on (SSO). Learn how to deploy SAML{' '}
```
mistahj67, line 83:
```suggestion
                        BloodHound supports SAML for single sign-on (SSO). Learn how to deploy SAML
```

packages/javascript/bh-shared-ui/src/components/FileIngest/FileIngest.tsx
```diff
@@ -58,37 +48,36 @@ const FileIngest = () => {
...
-                        Upload File(s)
-                    </Button>
-                </Box>
-            </ContentPage>
+            <PageWithTitle
+                title='File Ingest'
+                data-testid='manual-file-ingest'
+                pageDescription={
+                    <Typography variant='body2'>
+                        Upload data from SharpHound or AzureHound offline collectors. Check out our{' '}
```
mistahj67, line 56:
```suggestion
                        Upload data from SharpHound or AzureHound offline collectors. Check out our
```

mistahj67 (commented):
Just style nits here, I'd prefer a more seasoned UI stamp vs myself

urangel (approved):
Thank you!

Issue #672, opened 2024-06-26
Issues and Pull Requests README


sircodemane:
## Pull Requests
If a Pull Request is being planned, please ensure a GitHub issue has been opened first. All Pull Requests must be associated with a GitHub issue (or a Jira ticket for BHE team members), otherwise, it will be closed. The reasons for this are:
- The bug or feature may already be in progress internally.
- Validation is needed to ensure the contribution fits the needs of BHCE.
- The engineering team will assist in refining the requirements and setting clear expectations.

Once an issue is submitted, it will be tracked internally. It will start in a triage state, where discussions will take place to address the points mentioned above. After these discussions are completed, if it is indicated that you will be authoring the contribution, you will be informed that work can begin and a pull request can be opened when ready. Don't forget to check out our [Contributing Guide](https://github.com/SpecterOps/BloodHound/wiki/Contributing) to help you get started.

## Issues
For issues, please search our issue board to make sure the bug or feature hasn't already been brought to our attention. If an open issue already exists, this is a great opportunity to add more details and context. If opening a new issue, please fill out each section to the best of your ability. Providing as much information as possible will aid in the triage stage. If it is indicated that the feature or bug fix will not be implemented by you, the issue will be added to our internal work log and work will commence as soon as possible.

Thanks for taking the time to read and if you have any further questions, please reach out to us in our [community slack](https://ghst.ly/BHSlack).

PR #674, opened 2024-06-26
Added entity query responses and necessary bh-graph models to openapi

sircodemane:
## Description

Adds the response schemas for AD entity query endpoints and the BH Graph models

## Motivation and Context

This PR addresses: BED-4141

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/openapi/src/responses/entity-info-query-results.yaml
+++ b/packages/go/openapi/src/responses/entity-info-query-results.yaml
@@ -1,0 +1,38 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+description: |
+  **OK**
+
+  This response is polymorphic and depends on the the type of entity being queried and whether
+  the `count` param is true or not. All node types will return a `props` field with the graph node
+  properties. If `count=true` the response will also include additional fields with integer counts.
+content:
+  application/json:
+    schema:
+      type: object
+      properties:
+        data:
+          allOf:
+            - type: object
+              properties:
+                props:
+                  type: object
+                  additionalProperties:
+                    type: object
+            - type: object
+              additionalProperties:
+                type: integer
--- a/packages/go/openapi/src/responses/related-entity-query-results.yaml
+++ b/packages/go/openapi/src/responses/related-entity-query-results.yaml
@@ -1,0 +1,47 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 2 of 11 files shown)


packages/go/openapi/src/schemas/model.bh-graph.label.yaml
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+allOf:
+  - $ref: './model.bh-graph.font-icon.yaml'
```
juggernot325, line 18: We probably don't want to make this object dependent on properties from `font-icon`. While they happen to share the same 3 properties here, we don't necessarily want `label` to change if `font-icon` does.

packages/go/openapi/src/schemas/model.bh-graph.edge.yaml
juggernot325, line 1: The associated struct that this comes from appears to be `BloodHoundGraphLink` but I noticed the filename was for `edge`. Do we call these `edge` on the user-facing side?

packages/go/openapi/src/responses/related-entity-query-results.yaml
```diff
@@ -0,0 +1,47 @@
...
+          type: integer
+        - allOf:
+            - $ref: './../schemas/api.response.pagination.yaml'
+            - title: list
+              type: object
+              properties:
+                data:
+                  type: array
+                  items:
+                    title: paged-node-list-entry
```
juggernot325, line 39: This is probably only used in this one spot, but maybe we should move it out to its own file just in case we do need to reference it again. It least makes it consistent that every model has a file.

juggernot325 (commented):
I have a few suggestions for clean up. I'm not sure how deep you want to go on this pass, though.

packages/go/openapi/src/schemas/graph.props.last-seen.yaml
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+type: string
+format: date-time
+description: The RFC-3339 timestamp in which this object was last collected by BloodHound.
```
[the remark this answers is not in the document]
sircodemane: oof lol i can try to find them all and clean them up

packages/go/openapi/src/responses/entity-info-query-results.yaml
```diff
@@ -0,0 +1,38 @@
...
+      properties:
+        data:
+          allOf:
+            - type: object
+              properties:
+                props:
+                  type: object
+                  additionalProperties:
+                    type: object
+            - type: object
```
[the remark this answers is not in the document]
sircodemane: Good catch, you are correct! will fix this

packages/go/openapi/src/schemas/model.bh-graph.edge.yaml
sircodemane: Yep, links are edges, but we dont surface the naming anywhere. I'm not sure why we called them Links on the backend, but I chose edge here because it makes it clear what part of the graph it is. I'd favor changing the name of Link to Edge instead of changing edge to link

packages/go/openapi/src/responses/related-entity-query-results.yaml
sircodemane: The response model for this one is pretty unique and this is the only place it's used. If you feel strongly about it I can move it, but if it's "meh", my preference would be to leave it

packages/go/openapi/src/schemas/model.bh-graph.label.yaml
sircodemane: Yea you're right, these really should be separate. fixed

packages/go/openapi/src/responses/related-entity-query-results.yaml
juggernot325: My preference would be on moving it out for consistency, but I'm not going to make a big fuss over it. We can leave it as-is.

juggernot325 (approved):
Looks good! Thank you for making those updates.

PR #676, opened 2024-06-27
BED-4391: Make v5.12.0 migration idempotent

superlinkx:
## Description

Fixes the part of our v5.12.0 migration that is not idempotent

## Motivation and Context

Migrations should be written as idempotent whenever possible to ensure they can be reapplied safely

## How Has This Been Tested?

Deleted the migration record for v5.12.0 and restarted dev after running into a migration issue. This solves it.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/database/migration/migrations/v5.12.0.sql
+++ b/cmd/api/src/database/migration/migrations/v5.12.0.sql
@@ -25,7 +25,8 @@ CREATE TABLE IF NOT EXISTS datapipe_status (
 INSERT INTO
     datapipe_status (status, updated_at)
 VALUES
-    ('idle', NOW ());
+    ('idle', NOW ())
+ON CONFLICT DO NOTHING;
 
 ALTER TABLE user_sessions
-    ADD COLUMN flags jsonb;
+    ADD COLUMN IF NOT EXISTS flags jsonb;


maffkipp (approved):
looks good!

mistahj67 (changes requested):
We should also make the insert idempotent as well correct?

superlinkx:
> We should also make the insert idempotent as well correct?

It seems pretty safe as-is, do you have a suggestion on making it more correct?

mistahj67:
> It seems pretty safe as-is, do you have a suggestion on making it more correct?

It'll conflict on second attempt to insert because of the singleton
Adding `ON CONFLICT DO NOTHING` ought to fix that.

superlinkx:
> > It seems pretty safe as-is, do you have a suggestion on making it more correct?
> 
> It'll conflict on second attempt to insert because of the singleton Adding `ON CONFLICT DO NOTHING` ought to fix that.

Didn't run into it on local, but seems like a straightforward improvement, I'll get that added

mistahj67 (approved):
Looks good! Thanks for catching and fixing these 🙇

PR #681, opened 2024-07-01
BED-4445 & BED-4446: Hybrid Path Edge Post-Processing

superlinkx:
## Description

This adds the following new post-processed edges:

- SyncedToEntraUser
- SyncedToADUser

These edges are created for each Entra User node that has OnPremSyncEnabled and an OnPremID collected. The edges are matched to a corresponding AD User node when available, or it will create a new AD User node with just the objectid in the case where it is otherwise missing from the graph.

## Motivation and Context

This PR addresses:
- BED-4445: Hybrid Paths SyncedToEntraUser Edge
- BED-4446: Hybrid Paths SyncedToADUser Edge

These are the first pieces in supporting hybrid attack paths. For now we are just creating the new edges appropriately during post-processing. Proper support in BloodHound beyond just existing in the database will be addressed in future PRs.

## How Has This Been Tested?

- Several integration tests will be added before this leaves draft (one is done at time of writing)
- Local testing with several datasets of different sizes to ensure general usage and performance

Performance at this time is in-line with other parts of Azure Post-Processing. There is plenty of room for optimization from here, but the performance is acceptable for the scope of work.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


superlinkx:
We are still adding tests, but local testing with several datasets proved promising so far. Will be getting the PR description updated shortly.

packages/go/dawgs/drivers/neo4j/transaction.go
```diff
@@ -253,7 +253,7 @@ func (s *neo4jTransaction) createNode(properties *graph.Properties, kinds ...gra
 	} else if result := s.Raw(statement, queryBuilder.Parameters); result.Error() != nil {
 		return nil, result.Error()
 	} else if !result.Next() {
-		return nil, graph.ErrNoResultsFound
+		return nil, fmt.Errorf("%w: %w", graph.ErrNoResultsFound, result.Error())
```
superlinkx, line 256: This was frustrating during some of our debugging. Still not sold on returning a "not found" during creation, but for backwards compatibility, I'm wrapping the original error path as well as the potential additional error context from the result. Why `result.Error()` is nil until you call `result.Next()` is likely something I don't even want to know the answer to.

juggernot325:
Tests have been added! This should be good for review.

elikmiller:
I think we're grabbing some unnecessary updates from `main` from after the `stage/v5.12.0` branch was created.

PR #684, opened 2024-07-03
BED-4439: Broken collected filter

iustinum:
## Description

Fixes issue where domains without the `collected` property does not show up when filter is set to `collected=false`

## Motivation and Context

Improved functionality from a bugfix

## How Has This Been Tested?

- Manual testing with API Explorer UI
- Passes unit and integration tests

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/model/filter.go
+++ b/cmd/api/src/model/filter.go
@@ -105,6 +105,7 @@ func (s QueryParameterFilter) BuildGDBNodeFilter() graph.Criteria {
         value       = guessFilterValueType(s.Value)
     )
 
+	// TODO: Investigate whether we can set the collected property for domains that originate from trusts in ParseDomainTrusts
     switch {
     case s.Name == common.Collected.String() && s.Operator == Equals:
         switch s.Value {
--- a/cmd/api/src/model/filter.go
+++ b/cmd/api/src/model/filter.go
@@ -51,7 +51,6 @@ const (
 	FalseString    = "false"
 	IdString       = "id"
 	ObjectIdString = "objectid"
-	CollectedString = "collected"
 
 	ErrNotFiltered = errors.Error("parameter value is not filtered")
 )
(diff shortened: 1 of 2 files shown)


cmd/api/src/model/filter.go
```diff
@@ -50,6 +50,7 @@ const (
 	FalseString    = "false"
 	IdString       = "id"
 	ObjectIdString = "objectid"
+	CollectedString = "collected"
```
urangel, line 53: We have this defined as part of common properties so I would avoid defining this here. In line 107 you can instead use `common.Collected.String()`

cmd/api/src/model/filter.go
```diff
@@ -162,33 +201,15 @@ func guessFilterValueType(raw string) any {
...
-
-			case Equals:
-				criteria = append(criteria, query.Equals(query.NodeProperty(filter.Name), guessFilterValueType(filter.Value)))
-
-			case NotEquals:
-				criteria = append(criteria, query.Not(query.Equals(query.NodeProperty(filter.Name), guessFilterValueType(filter.Value))))
-			}
-		}
-	}
+    for _, filters := range s {
```
urangel, line 206: Nice refactor!

cmd/api/src/model/filter.go
```diff
@@ -96,8 +97,46 @@ type QueryParameterFilter struct {
 }
 
 type QueryParameterFilters []QueryParameterFilter
+
+
+func (f QueryParameterFilter) BuildGDBNodeFilter() graph.Criteria {
+    propertyRef := query.NodeProperty(f.Name)
+    value := guessFilterValueType(f.Value)
```
urangel, line 104: We try to use `s` for receiver names so please update `f` to `s` here. In addition, it would be good to wrap these declarations with a `var` so they format nicely.

urangel (approved):
Nice work! Thanks for updating to match our Go standards

PR #702, opened 2024-07-18
BED-4549: Simple Entity Paths To OAPI Spec

mvlipka:
## Description

Addresses feedback and splits the work into smaller commits that was initially done in [BED-4141](https://specterops.atlassian.net/browse/BED-4141)

## Motivation and Context

This PR addresses: [BED-4549](https://specterops.atlassian.net/browse/BED-4549)

Simplify and improve on our existing OpenAPI spec documentation

## How Has This Been Tested?
GoLand will only recognize valid OpenAPI specs when they are added as a path to the base document (openapi.yaml in this case), we can verify that there are no errors in the code through GoLand, but until all following tickets are complete we cannot test that the spec renders properly

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-4141]: https://specterops.atlassian.net/browse/BED-4141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BED-4549]: https://specterops.atlassian.net/browse/BED-4549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/go/openapi/src/paths/aiacas.aiaca.id.controllers.yaml
+++ b/packages/go/openapi/src/paths/aiacas.aiaca.id.controllers.yaml
@@ -1,0 +1,45 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+parameters:
+  - $ref: './../../parameters/header/prefer.yaml'
+  - $ref: './../../parameters/path/object-id.yaml'
+get:
+  operationId: GetAiaCaEntityControllers
+  summary: Get AIA CA entity controllers
+  description: Get a list, graph, or count of the principals that can control this AIA CA.
+  tags:
+    - AIA CAs
+    - Community
+    - Enterprise
+  parameters:
+    - $ref: './../../parameters/query/entity-query.skip.yaml'
+    - $ref: './../../parameters/query/entity-query.limit.yaml'
+    - $ref: './../../parameters/query/entity-query.type.yaml'
+    - $ref: './../../parameters/query/entity-query.sort-by.yaml'
+  responses:
+    200:
+      $ref: './../../responses/related-entity-query-results.yaml'
+    400:
+      $ref: './../../responses/bad-request.yaml'
+    401:
+      $ref: './../../responses/unauthorized.yaml'
+    403:
+      $ref: './../../responses/forbidden.yaml'
+    429:
+      $ref: './../../responses/too-many-requests.yaml'
+    500:
+      $ref: './../../responses/internal-server-error.yaml'
--- a/packages/go/openapi/src/paths/rootcas.rootcas.id.controllers.yaml
+++ b/packages/go/openapi/src/paths/rootcas.rootcas.id.controllers.yaml
@@ -1,0 +1,45 @@
+# Copyright 2024 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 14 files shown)


superlinkx (approved):
Thanks for this! Very excited to see this work continue to progress

packages/go/openapi/src/paths/aiacas.aiaca.id.controllers.yaml
```diff
@@ -0,0 +1,45 @@
...
+  tags:
+    - AIA CAs
+    - Community
+    - Enterprise
+  parameters:
+    - $ref: './../../parameters/query/entity-query.skip.yaml'
+    - $ref: './../../parameters/query/entity-query.limit.yaml'
+    - $ref: './../../parameters/query/entity-query.type.yaml'
+    - $ref: './../../parameters/query/entity-query.sort-by.yaml'
+  responses:
```
neumachen, line 33: You can use `default` for this, or reference the common responses so you don't have to repeat the same responses for each file you are changing.

Example:

```yaml
# ../../common-responses.yaml or default-responses.yaml, up to you.
responses:
  400:
    $ref: './../../responses/bad-request.yaml'
  401:
    $ref: './../../responses/unauthorized.yaml'
  403:
    $ref: './../../responses/forbidden.yaml'
  429:
    $ref: './../../responses/too-many-requests.yaml'
  500:
    $ref: './../../responses/internal-server-error.yaml'
```

```yaml
parameters:
  - $ref: './../../parameters/header/prefer.yaml'
  - $ref: './../../parameters/path/object-id.yaml'
get:
  operationId: GetAiaCaEntityControllers
  summary: Get AIA CA entity controllers
  description: Get a list, graph, or count of the principals that can control this AIA CA.
  tags:
    - AIA CAs
    - Community
    - Enterprise
  parameters:
    - $ref: './../../parameters/query/entity-query.skip.yaml'
    - $ref: './../../parameters/query/entity-query.limit.yaml'
    - $ref: './../../parameters/query/entity-query.type.yaml'
    - $ref: './../../parameters/query/entity-query.sort-by.yaml'
  responses:
    $ref: '../common-responses.yaml#/responses'
    200:
      $ref: '../responses/entity-info-query-results.yaml'
```

Also, notice that I've kept the 200 response separate, as these will likely be operation-specific.

packages/go/openapi/src/paths/rootcas.rootcas.id.controllers.yaml
```diff
@@ -0,0 +1,45 @@
...
+get:
+  operationId: GetRootCaEntityControllers
+  summary: Get Root CA entity controllers
+  description: Get a list, graph, or count of the principals that can control this Root CA.
+  tags:
+    - Root CAs
+    - Community
+    - Enterprise
+  parameters:
+    - $ref: '../parameters/query.entity.skip.yaml'
```
neumachen, line 29: The same applies here.

packages/go/openapi/src/paths/aiacas.aiaca.id.controllers.yaml
sircodemane: This doesn't work for a lot of tooling because even though some people do it, it's technically not valid according to OpenAPI spec. Many of the endpoints have different error response cases too

mvlipka: I also think this makes it harder to determine what responses and parameters are valid at a glance (without proper tooling).

neumachen: @sircodemane Thanks for the suggestion. You're right. As an alternative, we can use the [default field](https://swagger.io/specification/v3/#responses-object).

@mvlipka Regarding your concern:

1. Spec readability: It's reasonable to assume that developers will primarily refer to the rendered specification, where the file path reference would be apparent.

2. Maintenance: The referenced file path should provide sufficient context even for developers working directly with the source files.

3. Consistency and error prevention: My main concern is that repeating the same response objects across multiple status codes increases the risk of inconsistencies or oversights during updates.

Consolidating common responses could improve maintainability and reduce the likelihood of errors. However, I'm open to balancing these concerns with your other considerations.

mvlipka: Those are good points. I personally, think the explicit definition of the responses makes it more clear and easier to edit. I fear the trap of over obfuscating things which can lead to:
`CommonResponses`
`CommonResponsesWithNotFound`

Explicitly defining the routes doesn't take much more time, and we already have the reference which ensures we are sending valid responses and taking in valid parameters.

PR #709, opened 2024-07-19
Dark mode

urangel:
## Description
Dark mode makes its return to BloodHound! 

We are working on solidifying our design system and making styles more uniform throughout BloodHound. Updating our palettes and using them in a consistent way enables us to apply different themes to the application. As such, we begin this endeavor with introducing new styles and palettes for both light and dark themes. 

The dark mode feature will be hidden behind a feature flag that can be enabled via the Early Access Features page. Enabling this feature will allow one to toggle dark mode on and off from within the popover menu that opens when clicking on the gear icon in the header of the application. 

This setting is persisted and loaded in from local storage so that it remains between logging in and out as well as refreshing the page or closing it and returning to the site. 

Big thanks to @rvazarkar , @benwaples , and our UX Designer Ryan Robin for help with the implementation and design. 

Thanks as well to @chrismaddalena for help updating our logos for use within the dark mode setting.

## Motivation and Context

This PR addresses: BED-2566

## How Has This Been Tested?

Unit test was added for the logic of persisting the setting in between logins but otherwise most of the changes are style changes.

## Screenshots (optional):

image

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


sircodemane:
🎉 🍾

cmd/ui/src/App.tsx
```diff
@@ -14,47 +14,92 @@
...
+    components,
+} from 'bh-shared-ui';
+import { createBrowserHistory } from 'history';
 import React, { useEffect } from 'react';
+import { ErrorBoundary } from 'react-error-boundary';
 import { useQueryClient } from 'react-query';
-import { useLocation } from 'react-router-dom';
-import { AppNotifications } from 'bh-shared-ui';
-import Notifier from 'src/components/Notifier';
+import { unstable_HistoryRouter as BrowserRouter, useLocation } from 'react-router-dom';
```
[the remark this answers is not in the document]
elikmiller: Nevermind. I thought we got rid of it but it just moved from main.tsx to here.

So another question. Both App.tsx and main.tsx are sort of these massive wrapper components around the entire application. Why do we have both and what are the responsibilities of each?

cmd/ui/src/components/SettingsMenu.tsx
```diff
@@ -74,6 +81,12 @@ const SettingsMenu: React.FC<Props> = ({ anchorEl, handleClose }) => {
         dispatch(logout());
     };
 
+    const toggleDarkMode: React.MouseEventHandler<HTMLLIElement> = () => {
+        const theme = darkMode ? 'light' : 'dark';
+        body.setAttribute('class', theme);
+        dispatch(setDarkMode(!darkMode));
```
elikmiller, line 87: It looks like we're storing information about whether dark mode is enabled in a couple location - as a piece of state in redux, as well as an attribute on the top level <body> tag which isn't controlled by react.

Curious why both are required?

cmd/ui/src/App.tsx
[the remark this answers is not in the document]
urangel: The reason this was updated from what it was like before is so that we can use hooks to pull the dark mode state from the store so that it can inform the theme what palette to use. The theme was created in main.tsx previously and we can't use hooks there.

cmd/ui/src/components/SettingsMenu.tsx
```diff
@@ -139,11 +152,25 @@ const SettingsMenu: React.FC<Props> = ({ anchorEl, handleClose }) => {
...
+                    enabled={
+                        <MenuItem onClick={toggleDarkMode} data-testid={'global_header_settings-menu_nav-logout'}>
+                            <ListItemIcon>
+                                <FontAwesomeIcon icon={faCircleHalfStroke} />
+                            </ListItemIcon>
+                            <ListItemText primary={'Dark Mode'} />
+                            <Switch checked={darkMode}>Dark Mode</Switch>
+                        </MenuItem>
+                    }
+                    disabled={null}
```
elikmiller, line 171: Not required, the disabled prop can simply be omitted.

cmd/ui/src/App.tsx
```diff
@@ -14,47 +14,92 @@
...
+                    },
+                },
+            },
+        },
+        applicationHeader: {
+            flexGrow: 0,
+            zIndex: theme.zIndex.drawer + 1,
+        },
+        applicationContent: {
+            backgroundColor: theme.palette.neutral.primary,
```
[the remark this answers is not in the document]
urangel: I agree and feel like we could use one of the colors above the secondary defined neutral color so that there is still enough contrast between the background and the elements directly over the background. This was based on the given design but if there is a different consensus then I am happy to update this.

cmd/ui/src/components/SettingsMenu.tsx
urangel: We need to set the dark mode class in the body for the Doodle UI elements that we are starting to use. When we ultimately move from MUI, we should only need the class in the DOM to inform the theme. 

Setting the `dark` class somewhere like the element that has the id of `root` does not affect elements like dialog components and thus `tertiary` variant styles do not adapt accordingly in dark mode.

cmd/ui/src/store.ts
```diff
@@ -59,7 +62,12 @@ const loadState = () => {
     }
 };
 
-const saveState = (state: any) => {
+type PersistedState = {
+    auth: { sessionToken: string | null };
+    global: { view: { darkMode: boolean; notifications: string[] } };
```
elikmiller, line 67: I think we can just ignore `notifications` here unless we had a specific use case for wanting to persist those?

cmd/ui/src/store.ts
```diff
@@ -77,19 +85,30 @@ if (SAMLToken !== undefined) {
...
 store.subscribe(
     throttle(() => {
+        const state = store.getState();
         saveState({
-            auth: { sessionToken: store.getState().auth.sessionToken },
+            auth: { sessionToken: state.auth.sessionToken },
+            global: {
+                view: {
+                    darkMode: state.global.view.darkMode,
+                    notifications: [],
```
elikmiller, line 109: same here

cmd/ui/src/views/EarlyAccessFeatures/EarlyAccessFeatures.tsx
```diff
@@ -163,6 +159,12 @@ const EarlyAccessFeatures: React.FC = () => {
                                     <EarlyAccessFeatureToggle
                                         flag={flag}
                                         onClick={(flagId) => {
+                                            // TODO: Consider adding flag keys to cue files so we don't use random strings. Consider adding more flexibility/composability to side effects for toggling feature flags on and off
```
elikmiller, line 162: Can we author a ticket for this and remove this TODO?

cmd/ui/src/views/Explore/ExploreSearch/CypherSearch.tsx
```diff
@@ -63,6 +61,21 @@ const useStyles = makeStyles((theme) => ({
...
+        border: '1px solid',
+        borderColor: 'rgba(0,0,0,.23)',
+        borderRadius: theme.shape.borderRadius,
+        backgroundColor: '#002b36',
+        paddingTop: '5px',
+        minHeight: '120px',
+        // enables drag n drop resizing of editor
+        resize: 'vertical',
+        maxHeight: '500px',
+        overflow: 'auto',
```
elikmiller, line 77: I believe this `overflow` was just removed in BED-4365 - we remove that here (and for the light mode version too.)

cmd/ui/src/views/Explore/utils.ts
```diff
@@ -38,26 +40,30 @@ export const initGraphNodes = (graph: MultiDirectedGraph, nodes: GraphNodes, nod
 
         // Tier zero nodes should be marked with a gem glyph
         if (node.isTierZero) {
-            const glyph = GLYPHS[GlyphKind.TIER_ZERO];
+            const glyph =
+                theme.palette.color.primary === '#1D1B20'
```
elikmiller, line 44: Can't help but feel like this logic will break and go undetected for a while.

Maybe there is a more direct way to determine if the app is in dark mode or not (perhaps via another argument to this function?)

cmd/ui/src/store.ts
urangel: Not having this defined when loading from the persisted state would throw an undefined error. It would be better to untangle this a bit more and we aren't actually persisting notifications as we only save an empty array in the throttled save state but this avoids the app crashing without a bigger refactor.

cmd/ui/src/store.ts
urangel: See reasoning in previous comment. This was quick and easy while avoiding further refactors.

packages/javascript/bh-shared-ui/src/components/AssetGroupFilters/AssetGroupFilters.tsx
```diff
@@ -45,6 +45,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     },
     activeFilters: {
         '& button.expand-filters': {
+            color: `${theme.palette.color.primary} !important`,
```
elikmiller, line 48: Prefer a more specific selector over `!important`

cmd/ui/src/views/EarlyAccessFeatures/EarlyAccessFeatures.tsx
urangel: I'll author a ticket a for it 👍 Though I do like having the TODO here as hopefully it informs others in the area that this is a pattern that we do not want to continue without having to be aware that there is a ticket to make this change.

cmd/ui/src/views/Explore/ExploreSearch/CypherSearch.tsx
urangel: Good catch, thank you for pointing this out.

cmd/ui/src/views/Explore/utils.ts
urangel: Yea I agree and had forgotten to clean this up so thank you for pointing it out. I am going to pass in the dark mode state directly instead.

packages/javascript/bh-shared-ui/src/constants.ts
```diff
@@ -14,9 +14,248 @@
...
+        quinary: '#CACFD3',
+    },
+    background: {
+        paper: '#fafafa',
+        default: '#e4e9eb',
+    },
+    low: 'rgb(255, 195, 15)',
+    moderate: 'rgb(255, 97, 66)',
+    high: 'rgb(205, 0, 117)',
+    critical: 'rgb(76, 29, 143)',
```
elikmiller, line 59: nit: As a dev I don't find it immediately obvious why I would use `primary.main` over `color.primary` over `neutral.primary`

Similarly if this is the `lightPalette` why are there options like `primary.dark`

TLDR I think we can discuss some better naming conventions here

urangel: This is a result of trying to fit the way we defined our styles in the design system with how MUI has their styles defined while doing this for the hackathon. I agree things can and should be clearer but hopefully we move from MUI soon enough that this isn't in some weird in between state.

package.json
```diff
@@ -24,5 +24,8 @@
         "all-contributors-cli": "^6.26.1",
         "prettier": "^3.2.5",
         "typescript": "^5.1.6"
+    },
+    "dependencies": {
+        "@bloodhoundenterprise/doodleui": "^1.0.0-alpha.3"
```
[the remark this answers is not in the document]
urangel: 👍

cmd/ui/src/components/SettingsMenu.tsx
elikmiller: Interesting - I see the explanation in the tailwind docs now.

I suggest we find a top level div in `App.tsx` and control a className prop there (using react/redux patterns) rather than reach outside the react root node and manually manipulate the DOM. That way we keep the theme state in one place and don't have to do anything outside what react renders/handles for us.

cmd/ui/src/store.ts
elikmiller: My understanding is that by leaving stuff undefined redux will automatically overlay a default state.

For example, in the `auth` property we only care about persisting `sessionToken` meanwhile there are plenty of other pieces of state that we track.

```
    isInitialized: boolean;
    loginLoading: boolean;
    loginError: any;
    updateExpiredPasswordLoading: boolean;
    updateExpiredPasswordError: any;
    acceptEULALoading: boolean;
    acceptEULAError: any;
    sessionToken: string | null;
    user: getSelfResponse | null;
    fedramp_eula_accepted: boolean;
```

elikmiller: Aaaaand I was wrong. So we'll leave it as is!

cmd/ui/src/App.tsx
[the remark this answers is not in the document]
elikmiller: Makes sense!

cmd/ui/src/views/EarlyAccessFeatures/EarlyAccessFeatures.tsx
elikmiller: Okie dokie!

packages/javascript/bh-shared-ui/src/constants.ts
elikmiller: I'll add it to the list!

cmd/ui/src/views/EarlyAccessFeatures/EarlyAccessFeatures.tsx
benwaples: feature flags in cue: BED-4583

Comment is updated and moved in changeset: https://github.com/SpecterOps/BloodHound/pull/717/files.

cmd/ui/src/components/SettingsMenu.tsx
elikmiller: I took care of this in !721

elikmiller (approved):

elikmiller:
I'm scared but I am ready!

cmd/ui/src/views/Explore/ExploreSearch/CommonSearches.tsx
```diff
@@ -39,7 +39,10 @@ const useStyles = makeStyles(() => ({
     tab: {
         height: '35px',
         minHeight: '35px',
-        color: 'black',
+        color: theme.palette.color.primary,
+        '&.Mui-selected': {
```
benwaples, line 43: Should this be an override?

benwaples: confirmed that it targets tab color of MUI Tabs component and an override would better support any future uses of tabs

cmd/ui/src/components/SettingsMenu.tsx
```diff
@@ -139,11 +149,24 @@ const SettingsMenu: React.FC<Props> = ({ anchorEl, handleClose }) => {
                     onClick={() => openInNewTab('https://bloodhoundenterprise.io/our-solution/')}
                     data-testid='global_header_settings-menu_nav-checkout-BHE'>
                     <ListItemIcon>
-                        <EnterpriseIcon fill={'#000000'} width='1rem' height='1rem' />
+                        <EnterpriseIcon fill={theme.palette.color.primary} width='1rem' height='1rem' />
                     </ListItemIcon>
                     <ListItemText primary='BloodHound Enterprise' />
                 </MenuItem>
 
+                <FeatureFlag
```
maffkipp, line 157: Currently, if the `dark_mode` flag is not found, this menu item will display the string "Error". If possible, it would be preferable to hide this element entirely:
Screenshot 2024-07-22 at 1 04 55 PM

cmd/ui/src/views/Users/CreateUserForm.tsx
```diff
@@ -344,18 +344,14 @@ const CreateUserForm: React.FC<{
                             </FormHelperText>
                         )}
                         <Button
```
maffkipp, line 346: Pretty minor nit, now that these buttons pulled in from Doodle are fully outlined, they seem crammed against the edge in all our modal dialogs and might benefit from slightly larger bottom & right margins:
Screenshot 2024-07-22 at 1 14 26 PM

cmd/ui/src/components/SettingsMenu.tsx
benwaples: Great find! updated!

cmd/ui/src/views/Users/CreateUserForm.tsx
benwaples: Good eye and agreed, updated modal actions container to have spacing that matches content/header

PR #728, opened 2024-07-23
BED-4168: Fixed inaccurate AG member_count value 

iustinum:
## Description

Added a new `getAssetGroupMemberCount` helper function that queries the latest asset group collection

## Motivation and Context

Resolves the bug where asset group member counts were incorrectly reported as 0 in API responses and ensures accurate representation of group membership

## How Has This Been Tested?

Manually tested via API Explorer UI

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/database/agi.go
+++ b/cmd/api/src/database/agi.go
@@ -88,16 +88,16 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
 func (s *BloodhoundDB) GetAssetGroup(ctx context.Context, id int32) (model.AssetGroup, error) {
 	var assetGroup model.AssetGroup
 	if result := s.preload(model.AssetGroupAssociations()).WithContext(ctx).First(&assetGroup, id); errors.Is(result.Error, gorm.ErrRecordNotFound) {
-        return model.AssetGroup{}, ErrNotFound
+        return assetGroup, ErrNotFound
     } else if result.Error != nil {
-        return model.AssetGroup{}, result.Error
+        return assetGroup, result.Error
     }
 
     if latestCollection, collectionErr := s.GetLatestAssetGroupCollection(ctx, id); collectionErr != nil {
         if errors.Is(collectionErr, ErrNotFound) {
 			assetGroup.MemberCount = 0
         }
-		return model.AssetGroup{}, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
+		return assetGroup, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
     } else {
         assetGroup.MemberCount = len(latestCollection.Entries)
     }
@@ -129,7 +129,7 @@ func (s *BloodhoundDB) GetAllAssetGroups(ctx context.Context, order string, filt
     for idx, assetGroup := range assetGroups {
         if latestCollection, collectionErr := s.GetLatestAssetGroupCollection(ctx, assetGroup.ID); collectionErr != nil {
             if !errors.Is(collectionErr, ErrNotFound) {
-                return model.AssetGroups{}, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
+                return assetGroups, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
             }
             assetGroups[idx].MemberCount = 0
         } else {
--- a/cmd/api/src/database/agi.go
+++ b/cmd/api/src/database/agi.go
@@ -18,9 +18,9 @@ package database
 
 import (
 	"context"
-	"time"
-	"strings"
 	"fmt"
+	"strings"
+	"time"
 
 	"gorm.io/gorm"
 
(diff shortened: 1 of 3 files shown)


cmd/api/src/database/agi.go
```diff
@@ -86,47 +86,74 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
...
+        latestCollection model.AssetGroupCollection
+        result           = s.preload(model.AssetGroupCollectionAssociations()).
+                             WithContext(ctx).
+                             Where("asset_group_id = ?", assetGroupID).
+                             Order("created_at DESC").
+                             First(&latestCollection)
+    )
+
+    if result.Error != nil {
+        if errors.Is(result.Error, gorm.ErrRecordNotFound) {
```
zinic, line 100: The BloodHound `errors` package predates much of the added functionality in the [upstream errors package](https://pkg.go.dev/errors). We should prefer the stdlib `errors.Is(...)` instead of our implementation.

I'd ask for three changes here:
* Refactor this without a `nil` check since `errors.Is(...)` is nil-safe.
* Consume the [stdlib errors package](https://pkg.go.dev/errors) in the conditional.
* Add a deprecation notice to the BloodHound `errors` function to help guard others from exercising it.

cmd/api/src/database/agi.go
```diff
@@ -86,47 +86,74 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
...
-		} else {
-			assetGroups[idx].MemberCount = len(latestCollection.Entries)
-		}
-	}
+    var (
+        assetGroups model.AssetGroups
+        result      *gorm.DB
+    )
+
+    if order != "" && filter.SQLString == "" {
```
zinic, line 134: The following condition will see `order != ""` as the same result, meaning that invoking the comparison multiple times is unnecessary. Consider the following structure instead:
```golang
if order != "" {
   if filter.SQLString == "" {
      ...
   } else {
      ...
   }
}
```

zinic (changes requested):

cmd/api/src/database/agi.go
mistahj67: I think we can take it 1 step further since result is the *gorm.DB, it can built dynamically with result = result.XYZ and toss the preload / context part in the initialization step

```suggestion
	var (
		assetGroups model.AssetGroups
		result      = s.preload(model.AssetGroupAssociations()).WithContext(ctx)
	)

	if order != "" {
		result = result.Order(order)
	}
	
	if filter.SQLString != "" {
		result = result.Where(filter.SQLString, filter.Params)
	}
	
	if result = result.Find(&assetGroups); result.Error != nil {
		return assetGroups, CheckError(result)
	}
```

cmd/api/src/database/agi.go
```diff
@@ -86,47 +86,74 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
...
+                             Order("created_at DESC").
+                             First(&latestCollection)
+    )
+
+    if result.Error != nil {
+        if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+            return 0, nil
+        }
+        return 0, fmt.Errorf("error getting latest collection for asset group ID %d: %w", assetGroupID, result.Error)
+    }
```
superlinkx, line 104: This could be unnested as:

```go
if errors.Is(result.Error, gorm.ErrRecordNotFound) {
    return 0, nil
} else if result.Error != nil {
    return 0, fmt.Errorf("error getting latest collection for asset group ID %d: %w", assetGroupID, result.Error)
}
```

cmd/api/src/database/agi.go
```diff
@@ -86,47 +86,74 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
 	})
 }
 
+func (s *BloodhoundDB) getAssetGroupMemberCount(ctx context.Context, assetGroupID int32) (int, error) {
+    var (
+        latestCollection model.AssetGroupCollection
+        result           = s.preload(model.AssetGroupCollectionAssociations()).
+                             WithContext(ctx).
+                             Where("asset_group_id = ?", assetGroupID).
+                             Order("created_at DESC").
```
mistahj67, line 95: I noticed we have a bespoke db call here over the `GetLatestAssetGroupCollection`, is the original implementation not valid? If so we might want to consider moving this to that method and calling it here so the other callers get fixed as well. If both are valid, I think we might want to shy away from the bespoke db hit.

cmd/api/src/database/agi.go
superlinkx: See @zinic's comment, as I didn't see it before adding this. We want to use the stdlib version of `errors` here (may need to change other usages in this file though, so if it gets crazy feel free to ask for assistance), and this is roughly what @zinic was talking about with the unnecessary nil check.

mistahj67 (commented):
Do we have this covered by test(s)? If possible I'd appreciate if we can add at least 1.

cmd/api/src/database/agi.go
iustinum: The former might be the case as `GetLatestAssetGroupCollection` has been returning empty asset groups despite new custom members being added. I will adapt my fix into that function.

cmd/api/src/database/agi.go
iustinum: Given that `stdErrors.Is` is nil-safe, do I still need to add `result.Error != nil`, having made the switch?

superlinkx: You probably answered your own question judging by the edit, but just for anyone else wondering: 

Yes, you want to do your `errors.Is` branches first to check for specific errors you want to add special handling for. After that, you should always do a nil check to see if you need to handle general errors you're not specifically interested in.

zinic (approved):
Changeset looks good

cmd/api/src/database/agi.go
```diff
@@ -87,46 +86,58 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
...
+    if collectionErr != nil {
+        if errors.Is(collectionErr, ErrNotFound) {
+            assetGroup.MemberCount = 0
+            return assetGroup, nil
+        }
+        return assetGroup, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
+    }
+    assetGroup.MemberCount = len(latestCollection.Entries)
+
+    return assetGroup, nil
```
superlinkx, line 106: This could be rewritten to chain with the previous block:

```go
    else if latestCollection, collectionErr := s.GetLatestAssetGroupCollection(ctx, id); errors.Is(collectionErr, ErrNotFound) {
        assetGroup.MemberCount = 0
        return assetGroup, nil
    } else if collectionErr != nil {
        return assetGroup, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
    } else {
        assetGroup.MemberCount = len(latestCollection.Entries)
        return assetGroup, nil
    }
```

cmd/api/src/database/agi.go
```diff
@@ -87,46 +86,58 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
...
+    if filter.SQLString != "" {
+        result = result.Where(filter.SQLString, filter.Params)
+    }
+
+    if result = result.Find(&assetGroups); result.Error != nil {
+		if errors.Is(result.Error, gorm.ErrRecordNotFound) {
+			return assetGroups, ErrNotFound
 		}
+		return assetGroups, result.Error
 	}
```
superlinkx, line 128: This could also be unnested similar to our previous ones

cmd/api/src/database/agi.go
```diff
@@ -87,46 +86,58 @@ func (s *BloodhoundDB) DeleteAssetGroup(ctx context.Context, assetGroup model.As
...
 
-	return assetGroups, nil
+    for idx, assetGroup := range assetGroups {
+        if latestCollection, collectionErr := s.GetLatestAssetGroupCollection(ctx, assetGroup.ID); collectionErr != nil {
+            if !errors.Is(collectionErr, ErrNotFound) {
+                return assetGroups, fmt.Errorf("error getting latest collection for asset group %s: %w", assetGroup.Name, collectionErr)
+            }
+            assetGroups[idx].MemberCount = 0
```
superlinkx, line 135: This could be unnested by checking errors.Is first, then nil check

superlinkx (approved):
Just a couple of consistency comments for the updated code. Additionally, I noticed there were a few places where we ended up with spaces instead of tabs. I'd recommend running `goimports` on the changed files to ensure they get updated to consistent spacing before merge.

These are nits, the code as it exists is logically correct as far as I can see. Consistency would be nice to help with future maintenance, but it's not hard to see the refactor points either. Approving as a result, but feel free to make any suggested changes before the final merge.

superlinkx (approved):

PR #730, opened 2024-07-23
BED-4534: Saved Cypher Improvements & Description Field

mvlipka:
## Description

These changes add a new description field to the `saved_queries` model, which will allow users the ability to add short descriptions on what their cypher query does.
We are also adding a new model called `saved_queries_permissions` which will contain the relationships between users and the queries they have been shared - globally or directly.

## Motivation and Context

This PR addresses: BED-4535

Initial database implementation required for work on API to begin development. Ultimately will enable the ability to search and share cypher queries directly from the UI.

## How Has This Been Tested?

First I ensured that the UI was still able to save queries while sending up the new description field in the body:

image

We can see that the migration successfully ran on startup by having the description field exist, and we can see that the query was correctly saved with the `DEFAULT ''` value.

I then tested the API using Bruno to send a request to `/api/v2/saved-queries`
image

The query was saved with the description that was sent in the request body.

Since these changes were put in place before the API was made to enable sharing, I used DataGrip to test the `saved_queries_permissions` table

After clearing my postgres volume to ensure best testing results, I was able to create a globally shared cypher:
image

A cypher to a user:
image

And trying to share a cypher to the same user results in a unique constraint violation:
image

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/database/db.go
+++ b/cmd/api/src/database/db.go
@@ -153,7 +153,7 @@ type Database interface {
 	fileupload.FileUploadData
 
 	// Saved Queries
-	SavedQueries
+	SavedQueriesData
 
 	// Saved Queries Permissions
 	SavedQueriesPermissionsData
--- a/cmd/api/src/database/db.go
+++ b/cmd/api/src/database/db.go
@@ -44,8 +44,8 @@ const (
 )
 
 var (
-    ErrDuplicateAGName = errors.New("duplicate asset group name")
-    ErrDuplicateAGTag  = errors.New("duplicate asset group tag")
+	ErrDuplicateAGName = errors.New("duplicate asset group name")
+	ErrDuplicateAGTag  = errors.New("duplicate asset group tag")
 )
 
 func IsUnexpectedDatabaseError(err error) bool {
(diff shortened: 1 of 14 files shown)


cmd/api/src/database/db.go
```diff
@@ -81,6 +81,7 @@ type Database interface {
 	GetAssetGroupSelector(ctx context.Context, id int32) (model.AssetGroupSelector, error)
 	DeleteAssetGroupSelector(ctx context.Context, selector model.AssetGroupSelector) error
 	UpdateAssetGroupSelectors(ctx context.Context, assetGroup model.AssetGroup, selectorSpecs []model.AssetGroupSelectorSpec, systemSelector bool) (model.UpdatedAssetGroupSelectors, error)
+	DeleteAssetGroupSelectorsForAssetGroups(ctx context.Context, assetGroupIds []int) error
```
mvlipka, line 84: @ALCooper12 caught this in review, it was in the wrong place so we moved it to its proper home. Great catch!

cmd/api/src/database/saved_queries.go
```diff
@@ -74,3 +84,20 @@ func (s *BloodhoundDB) SavedQueryBelongsToUser(ctx context.Context, userID uuid.
...
+
+// GetSavedQueriesSharedWithUser returns all the saved queries that the given userID has access to, including global queries
+func (s *BloodhoundDB) GetSavedQueriesSharedWithUser(ctx context.Context, userID int64) (model.SavedQueries, error) {
+	savedQueries := model.SavedQueries{}
+
+	result := s.db.WithContext(ctx).Where("shared_to_user_id = ?", userID).Find(&savedQueries)
+
+	return savedQueries, CheckError(result)
+}
+
```
mvlipka, line 96: Moved these methods here. They pertain to the permissions but return a SavedQuery model, which made sense to put them here to me.
This also gives power to the end user implementing the endpoints something akin to:

```go
sharedQueries := GetSavedQueriesSharedWithUser(ctx, id)
globalQueries := GetSavedQueriesSharedGlobally(ctx)

return SavedQueriesSharedWithUser {
    GlobalQueries: globalQueries,
    SharedQueries: sharedQuereies
}
```

cmd/api/src/docs/json/definitions/v2.json
```diff
@@ -951,6 +951,9 @@
             },
             "name": {
                 "type": "string"
+            },
+            "description": {
```
mvlipka, line 955: The OAPI docs don't seem to be generating properly for any request that reference an external object. Those with embedded parameters are working, at least when testing locally.

Regardless, I updated docs here as well as in the new `.yaml` Redocly docs we will be moving to

PR #741, opened 2024-07-24
BED-4479 - ensure update key uniqueness by removing sort

zinic:
## Description

Batch updates are bucketed by a unique key that matches the generated openCypher query statement. This sort appears to damage the uniqueness of these buckets, resulting in mismatched kinds being applied to nodes during upsert.

This would manifest as nodes having more than one valid kind such as: `match (n) where n:Base and n:Group and n:User return n`

## Motivation and Context

This damages the validity of data stored in the graph as nodes are being incorrectly labeled.

## How Has This Been Tested?

Integration tests pass. Added an additional unit test with some explanation for the assertions.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/dawgs/drivers/neo4j/cypher.go
+++ b/packages/go/dawgs/drivers/neo4j/cypher.go
@@ -27,7 +27,7 @@ import (
 	"github.com/specterops/bloodhound/log"
 )
 
-func updateKey(identityKind graph.Kind, identityProperties []string, updateKinds graph.Kinds) string {
+func newUpdateKey(identityKind graph.Kind, identityProperties []string, updateKinds graph.Kinds) string {
 	keys := []string{
 		identityKind.String(),
 	}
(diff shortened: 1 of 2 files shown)


packages/go/dawgs/drivers/neo4j/cypher.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/specterops/bloodhound/log"
 )
 
-func updateKey(identityKind graph.Kind, identityProperties []string, updateKinds graph.Kinds) string {
+func newUpdateKey(identityKind graph.Kind, identityProperties []string, updateKinds graph.Kinds) string {
```
sircodemane, line 30: I have a suspicion that this `new` naming convention is maybe an artifact of testing a new implementation while preserving an old implementation. If that's not the case, I'm curious the reasoning for the naming convention

sircodemane (approved):

packages/go/dawgs/drivers/neo4j/cypher.go
zinic: Nope, this was a naming collision that was making me upset. `updateKey` was used both for function names and instantiation assignment. So now the function is `newUpdateKey` to differentiate.

zinic:
Handing this off to @neumachen

computator (approved):

PR #753, opened 2024-07-30
BED-4641 - Create Ingest Defect Graph Migration

zinic:
## Description

`Version_513_Migration` covers a bug discovered in ingest that would result in nodes containing more than one kind. For example, the following query should return no results: `(n:Base) where (n:User and n:Computer) return n` but, at time of writing, due to the ingest defect several environments will return results for this query.

For instances where nodes are discovered that contain more than one valid kind assignment, the node's kinds are reset to the matching base entity kind such that:

```
node.Kinds = Kinds{ad.Entity, ad.User, ad.Computer} must be reset to:
node.Kinds = Kinds{ad.Entity}
```

## Motivation and Context

This PR addresses: BED-4641

## How Has This Been Tested?

Tested locally with a 1M+ node dataset.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/migrations/graph.go
+++ b/cmd/api/src/migrations/graph.go
@@ -32,45 +32,10 @@ type Migration struct {
 	Execute func(db graph.Database) error
 }
 
-type GraphMigrator struct {
-	db graph.Database
-}
-
-func NewGraphMigrator(db graph.Database) *GraphMigrator {
-	return &GraphMigrator{db: db}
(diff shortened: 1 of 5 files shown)


urangel (approved):
`ParallelFetchNodes` is a cool function! The logic looks right to me after having dug through this.

PR #764, opened 2024-08-01
fix: correctly stream process json data normalization to utf8 for bot…

ddlees:
…h metatag validation and writing to disk

## Description

- correctly stream process json data normalization to utf8
- feed normalized data to metatag validator
- feed normalized data to disk writer

## Motivation and Context

This PR addresses: [GitHub issue or Jira ticket number]
BED-4463

The initial pass at normalizing json data that is prefixed with a byte order mark (BOM) was causing the `ValidateMetaTag` check to incorrectly fail when preparing to write the data to disk and when it was being read from disk for graph ingest. Additionally, we were reading in the whole payload into an uncompressed `[]byte` which is problematic for very large files. This PR addresses those issues by correcting data normalization via stream processing and feeding the data stream to the validator and disk writer.

## How Has This Been Tested?

Updated unit tests to reflect necessary interface changes.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/go/bomenc/normalize_test.go
```diff
@@ -288,59 +198,52 @@ func TestNormalizeToUTF8_LargeInput(t *testing.T) {
...
-		for _, c := range u16 {
-			utf16LE = append(utf16LE, byte(c), byte(c>>8))
-		}
-
-		// Expected UTF-8
-		buf := make([]byte, 4)
-		n := utf8.EncodeRune(buf, r)
-		expected = append(expected, buf[:n]...)
+	// Generate a large data set with 1000 Unicode code points
+	// NOTE: We don't want to begin the payload with a NULL character because it would then be impossible to discern between UTF16LE and
```
urangel, line 202:
```suggestion
	// NOTE: We don't want to begin the payload with a NULL character because it would then be impossible to discern between UTF16LE and UTF32LE
```
nit: seems the comment got cut off early

urangel (approved):
This looks to work for uploads of JSON files 👍 

It may be out of scope for this work but I believe it is still possible to zip JSON files with BOM's and upload the zip into BH's UI. Since there is no normalization done of the files within the zip, data that is uploaded this way will encounter a similar BOM issue and an error will display that the data is not valid JSON.

StephenHinck:
> This looks to work for uploads of JSON files 👍
> 
> It may be out of scope for this work but I believe it is still possible to zip JSON files with BOM's and upload the zip into BH's UI. Since there is no normalization done of the files within the zip, data that is uploaded this way will encounter a similar BOM issue and an error will display that the data is not valid JSON.

This is correct - we should also expect to strip encoding from zipped JSON files.

ddlees:
> > This looks to work for uploads of JSON files 👍
> > It may be out of scope for this work but I believe it is still possible to zip JSON files with BOM's and upload the zip into BH's UI. Since there is no normalization done of the files within the zip, data that is uploaded this way will encounter a similar BOM issue and an error will display that the data is not valid JSON.
> 
> This is correct - we should also expect to strip encoding from zipped JSON files.

I looked into what it would take to include normalizing decompressed json files and we would have to do some non-trivial work to accommodate the `ReadSeeker` interface that we expect when processing the file. SharpHound zips will not include BOM encoded json files but we shouldn't lose sight of this issue because I'm sure someone will bom encode some json files and zip them and try to upload that data to BloodHound at some point. Will attempt to deliver a solution in a separate PR.

PR #766, opened 2024-08-01
BED-4663: Fix Hybrid Paths Failure When Node Exists But Not As AD User

superlinkx:
## Description

Fixes an edge case where a node can exist with a given `objectid` but may not have a `User` label due to not being directly collected or some other case where the label is missing. This includes checking each node we think we need to create for existence, and if it does already exist, adding the `User` label to that node.

## Motivation and Context

This PR addresses: BED-4663

Fixes an important edge case we found in v5.13.0. This is being staged for hotfix v5.13.1

## How Has This Been Tested?

Used a debugger to pause on the first objectid that needed to be created because it wasn't found in our list of AD Users. During that pause, I did a `MERGE` for that objectid and ensured that the node had only `:Base`, no `:User`. I then resumed the debugger until the next item and checked Neo4j for the previous node. When not adding on the label, I saw the correct node show up as expected with only the `Base` label. When adding the additional logic to add the `User` label, I saw that it correctly identified the existing node and added the `User` label.

An integration test has also been added for this edge case, and was validated in a similar manual debugging effort while the test ran to validate replication of the edge case and correct outcome.

## Screenshots (optional):
image
Proof that we could properly work with existing nodes that lacked an AD User label, prior to ensuring that label is readded

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/analysis/hybrid/hybrid.go
+++ b/packages/go/analysis/hybrid/hybrid.go
@@ -174,24 +174,28 @@ func createMissingADUser(ctx context.Context, db graph.Database, objectID string
 		common.ObjectID.String(): objectID,
 	})
 
+	// Using a switch to make the complex error handling logic more clear
 	err = db.WriteTransaction(ctx, func(tx graph.Transaction) error {
 		newNode, err = analysis.FetchNodeByObjectID(tx, objectID)
-		if !errors.Is(err, graph.ErrNoResultsFound) && err != nil {
-			return fmt.Errorf("create missing ad user precheck: %w", err)
-		} else if err == nil {
+		switch {
+		// No node found, so it's safe to create a new AD User
+		case errors.Is(err, graph.ErrNoResultsFound):
+			if newNode, err = tx.CreateNode(properties, adSchema.Entity, adSchema.User); err != nil {
+				return fmt.Errorf("create missing ad user: %w", err)
+			} else {
+				return nil
+			}
+		// Node was found, so we need to update it to an AD User
+		case err == nil:
 			newNode.AddKinds(adSchema.User)
 			if err := tx.UpdateNode(newNode); err != nil {
 				return fmt.Errorf("update missing ad user label: %w", err)
 			} else {
 				return nil
 			}
-		}
-
-		newNode, err = tx.CreateNode(properties, adSchema.Entity, adSchema.User)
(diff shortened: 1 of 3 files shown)


packages/go/analysis/hybrid/hybrid.go
```diff
@@ -175,8 +175,24 @@ func createMissingADUser(ctx context.Context, db graph.Database, objectID string
...
+			}
+		}
+
 		newNode, err = tx.CreateNode(properties, adSchema.Entity, adSchema.User)
-		return err
+		if err != nil {
+			return fmt.Errorf("create missing ad user: %w", err)
+		} else {
+			return nil
+		}
```
mistahj67, line 195: Just a style nit per convention
```suggestion
		if newNode, err = tx.CreateNode(properties, adSchema.Entity, adSchema.User); err != nil {
			return fmt.Errorf("create missing ad user: %w", err)
		}
		
		return nil
```

mistahj67 (approved):
Approved but would love to see some integration tests if they are easily possible, but don't block as this is a high priority fix!

Last question for thought:

We are returning immediately hitting a single error here which breaks hybrid paths completely. Would it be better to collect the errors and keep processing instead of still halting on a single error?

superlinkx:
> Approved but would love to see some integration tests if they are easily possible, but don't block as this is a high priority fix!
> 
> Last question for thought:
> 
> We are returning immediately hitting a single error here which breaks hybrid paths completely. Would it be better to collect the errors and keep processing instead of still halting on a single error?

Integration test added after some thought on how to implement it. Required a minor harness revision, but it now successfully tests this edge case for us! Thanks for the suggestion. I debugged the test to make sure that each part worked as if we had hit the edge case, so I'm happy with the results and they can now be validated by others.

Collecting errors would be good, but we would need to think about how we appropriately remove the affected nodes out of our mappings for the edges, or else we could end up with undefined behavior pretty quickly. As this is meant to be a hotfix for our current issue, I'm requesting a bug ticket to track this as a future improvement.

zinic (approved):
Logic and tests look good to me

mistahj67 (approved):
Switch case is much cleaner 🧼 
Thanks for adding the test! 🙇

PR #771, opened 2024-08-05
BED-4581: added endpoint to update a saved query

irshadaj:
## Description

We don’t currently have a way to update a saved query, so we will be creating a new one

New update endpoint:PUT api/v2/saved-queries/{id} with body
```
{
    "name": string,
    "query": string,
    "description": string
}
```

Users should only be able to update queries owned by them

Admins should be able to update public queries

## Motivation and Context

This PR addresses: BED-4581

## How Has This Been Tested?

Added unit tests.

Also performed manual testing as follows:

Spun up the environment, created a saved query:

initial

Updated a single field, the description, and checked that it was done right:

desc

Proceeded to update all the fields to make sure that multiple changes are handled well:

all

Finally, verified via the ListSavedQueries endpoint that the updates are reflected in the database:

final

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/v2/saved_queries.go
+++ b/cmd/api/src/api/v2/saved_queries.go
@@ -189,7 +189,7 @@ func (s Resources) UpdateSavedQuery(response http.ResponseWriter, request *http.
 			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "query does not exist", request), response)
 			return
 		} else {
-			if isPublic, err := s.DB.IsSavedQueryPublic(request.Context(), savedQuery); err != nil {
+			if isPublic, err := s.DB.IsSavedQueryPublic(request.Context(), savedQuery.ID); err != nil {
 				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, err.Error(), request), response)
 				return
 			} else if !isPublic {
--- a/cmd/api/src/database/saved_queries.go
+++ b/cmd/api/src/database/saved_queries.go
@@ -32,7 +32,7 @@ type SavedQueriesData interface {
 	SavedQueryBelongsToUser(ctx context.Context, userID uuid.UUID, savedQueryID int) (bool, error)
 	GetSharedSavedQueries(ctx context.Context, userID uuid.UUID) (model.SavedQueries, error)
 	GetPublicSavedQueries(ctx context.Context) (model.SavedQueries, error)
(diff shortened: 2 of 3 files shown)


ALCooper12:
Hey @irshadaj, I think you forgot to add to the openapi yaml files!

cmd/api/src/database/saved_queries.go
```diff
@@ -71,6 +78,24 @@ func (s *BloodhoundDB) CreateSavedQuery(ctx context.Context, userID uuid.UUID, n
...
+	} else {
+		if name != "" {
+			savedQuery.Name = name
+		}
+		if query != "" {
+			savedQuery.Query = query
+		}
+		if description != "" {
+			savedQuery.Description = description
+		}
```
mvlipka, line 93: A user might want to remove a `description` from their saved query.
It probably doesn't matter as much for `query` and `name` but it might be feel weird to a user if they're intentionally deleting a saved query's cypher and when saving it is popping back up.

I'd think we could move the empty check logic outside of the database interaction and can just pass in `model.SavedQuery` that we expect to update.
https://gorm.io/docs/update.html

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -126,6 +126,27 @@ func (s Resources) CreateSavedQuery(response http.ResponseWriter, request *http.
...
+	)
+
+	if user, isUser := auth.GetUserFromAuthCtx(ctx2.FromRequest(request).AuthCtx); !isUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "No associated user found", request), response)
+	} else if err := api.ReadJSONRequestPayloadLimited(&updateRequest, request); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if savedQueryID, err := strconv.Atoi(rawSavedQueryID); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
+	} else if savedQuery, err := s.DB.GetSavedQuery(request.Context(), savedQueryID); err != nil || savedQuery.UserID != user.ID.String() {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "query does not exist", request), response)
```
mvlipka, line 142: Can use `api.HandleDatabaseError` for the db errors

mvlipka (changes requested):

PR #775, opened 2024-08-06
BED-4537: Create Share endpoint

ALCooper12:
## Description

- Created a new `ShareSavedQueries` endpoint
- Added unit tests
- Updated openapi docs

## Motivation and Context

- Ticket: BED-4537
- Subtask above is from BED-4534

- This newly created endpoint will allow users to have the ability to share a query with others, or set it as public

## How Has This Been Tested?
- 3 successfully expected api calls via Bruno + database changes
Screenshot 2024-08-05 at 5 56 33 PM
Screenshot 2024-08-05 at 5 57 53 PM
Screenshot 2024-08-05 at 5 58 36 PM
Screenshot 2024-08-05 at 5 59 55 PM
Screenshot 2024-08-05 at 6 00 38 PM

OpenAPI preview (but please check on how this looks like for you!)
Screenshot 2024-08-05 at 6 05 56 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


mvlipka (approved):
Looks good!
@irshadaj I'd still like another pair of eyes on it since I'm a bit bias here 😄

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -152,3 +153,76 @@ func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.
...
+
+			// Saving query permission to one or more users
+			for _, sharedUserID := range newlySharedUserIDs {
+				if savedPermission, err := s.DB.CreateSavedQueryPermissionToUser(request.Context(), permissionsForSavedQuery.QueryID, sharedUserID); err != nil {
+					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
+					return
+				} else {
+					apiResponse = append(apiResponse, savedPermission)
+				}
+			}
```
mistahj67, line 224: Since we are returning an array of responses anyway, why not append the errors and not return here instead?
That said, if we do want to return we have a helpful `HandleDatabaseError` util that would be perfect here
```suggestion
			// Saving query permission to one or more users
			for _, sharedUserID := range newlySharedUserIDs {
				if savedPermission, err := s.DB.CreateSavedQueryPermissionToUser(request.Context(), permissionsForSavedQuery.QueryID, sharedUserID); err != nil {
					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
					return
				} else {
					apiResponse = append(apiResponse, savedPermission)
				}
			}
```

mistahj67 (changes requested):
I had a number of questions as well that we DM'd and it looks like there's some additional work on the table still.

cmd/api/src/api/v2/saved_queries.go
mistahj67, line 224: It looks like you can pass a slice to gorm create and it'll automatically batch it and it has an upsert / on conflict where we can just do nothing

https://gorm.io/docs/create.html#Batch-Insert
https://gorm.io/docs/create.html#Upsert-x2F-On-Conflict

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -152,3 +153,76 @@ func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.
...
+	} else if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(request.Context(), user.ID, savedQueryID); errors.Is(err, database.ErrNotFound) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "Query does not exist", request), response)
+	} else if err := api.ReadJSONRequestPayloadLimited(&createRequest, request); errors.Is(err, database.ErrNotFound) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, api.ErrorResponseDetailsInternalServerError, request), response)
+	} else if !savedQueryBelongsToUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "Invalid saved_query_id supplied", request), response)
+	} else if permissionsForSavedQuery, err := s.DB.GetPermissionsForSavedQuery(request.Context(), int64(savedQueryID)); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "Error retrieving query permissions", request), response)
```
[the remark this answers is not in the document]
ALCooper12: I changed it to a `api.HandleDatabaseError(request, response, err)` instead!

packages/go/openapi/src/schemas/model.saved-queries-permissions-request.yaml
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+allOf:
+  - $ref: './model.components.int64.id.yaml'
+  - $ref: './model.components.timestamps.yaml'
+  - type: object
+    properties:
+      query_id:
```
sircodemane, line 22: This field doesnt appear to exist on the request Struct, as it seems to be handled by the URL parameter instead

cmd/api/src/database/saved_queries_permissions.go
```diff
@@ -52,16 +69,77 @@ func (s *BloodhoundDB) CreateSavedQueryPermissionToPublic(ctx context.Context, q
...
+	} else if sharedCount > 0 {
+		scopes[SavedQueryScopeShared] = true
+	}
+
+	return scopes, nil
+}
+
+// DeleteSavedQueryPermission deletes the saved query permission associated with the passed in permission id
+func (s *BloodhoundDB) DeleteSavedQueryPermission(ctx context.Context, permissionID int64) error {
+	return CheckError(s.db.WithContext(ctx).Table("saved_queries_permissions").Delete(&model.SavedQueriesPermissions{}, permissionID))
```
mistahj67, line 134: Is this missing a where clause? If we hit this with an empty id, does it wipe the entire table? Might be beneficial to toss a Limit(1) on this just in case.

cmd/api/src/database/saved_queries_permissions.go
```diff
@@ -52,16 +69,77 @@ func (s *BloodhoundDB) CreateSavedQueryPermissionToPublic(ctx context.Context, q
 	return permission, CheckError(s.db.WithContext(ctx).Create(&permission))
 }
 
+// CreateSavedQueryPermissionsBatch attempts to save the given saved query permissions in batches of 100 in a transaction
+func (s *BloodhoundDB) CreateSavedQueryPermissionsBatch(ctx context.Context, savedQueryPermissions []model.SavedQueriesPermissions) ([]model.SavedQueriesPermissions, error) {
+	result := s.db.WithContext(ctx).Clauses(clause.OnConflict{
+		DoNothing: true,
```
mistahj67, line 75: Since we are `on conflict, do nothing` do we actually even need to do the `hasAccess` db check above? 🤔 @irshadaj 

If we do keep the check, we might consider leveraging the `GetPermissionsForSavedQuery` and checking access via logic vs hitting the db for each shared user supplied.

packages/go/openapi/src/schemas/model.saved-queries-permissions.yaml
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+allOf:
+  - $ref: './model.components.int64.id.yaml'
+  - $ref: './model.components.timestamps.yaml'
+  - type: object
+    properties:
+      id: 
+        type: integer
+        format: int64
```
sircodemane, line 24: This property is actually covered by the model.components.int64.id part above. i suspect that this may have happened because the SavedQueryPermissions model has an ID field defined, though it would probably be better for that model to remove its ID field and replace the `Basic` embedded struct with a `BigSerial`, which has all of Basic + an int64 ID

.github/workflows/publish.yml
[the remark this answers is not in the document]
ALCooper12: No not intentional 😬

packages/go/openapi/src/schemas/model.saved-queries-permissions.yaml
```diff
@@ -0,0 +1,34 @@
...
+      shared_to_user_id:
+        type: array
+        items:
+          type: string
+          format: uuid
```
sircodemane, line 29: This field isn't an array, it's a `uuid.NullUUID`. you can include the `null.uuid.yaml` schema here

cmd/api/src/api/v2/savedqueriespermissions.go
```diff
@@ -0,0 +1,198 @@
...
+			isAdmin := user.Roles.Has(model.Role{Name: auth.RoleAdministrator})
+			if !isAdmin {
+				// If a user is not admin, then they need to own the query in order to unshare it
+				if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(request.Context(), user.ID, savedQueryID); err != nil {
+					api.HandleDatabaseError(request, response, err)
+					return
+				} else if !savedQueryBelongsToUser {
+					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusForbidden, "Query does not belong to the user", request), response)
+					return
+				}
```
ALCooper12, line 186: This was the only thing that was changed besides moving your function @mvlipka! 

It got switched from 401 to 403 since according to @mistahj67, there are some handlers on the frontend that will automatically logout a user whenever they receive a 401. And we don't want them to get logged out in this scenario

cmd/api/src/database/saved_queries_permissions.go
```diff
@@ -21,61 +21,119 @@ import (
...
+		model.SavedQueryScopeShared: false,
+	}
+
+	// Check if the query was shared with the user publicly
+	publicCount := int64(0)
+	if result := s.db.WithContext(ctx).Select("*").Table("saved_queries_permissions").Where("public = true AND query_id = ?", queryID).Count(&publicCount).Limit(1); result.Error != nil {
+		return scopes, CheckError(result)
+	} else if publicCount > 0 {
+		scopes[model.SavedQueryScopePublic] = true
+	}
```
mistahj67, line 104: Lets leverage the `IsSavedQueryPublic` method here

cmd/api/src/database/saved_queries_permissions.go
```diff
@@ -21,61 +21,119 @@ import (
...
+	} else if ownedCount > 0 {
+		scopes[model.SavedQueryScopeOwned] = true
+	}
+
+	// Check if the user has had the query shared to them
+	sharedCount := int64(0)
+	if result := s.db.WithContext(ctx).Select("*").Table("saved_queries_permissions").Where("query_id = ? AND shared_to_user_id = ?", queryID, userID).Count(&sharedCount).Limit(1); result.Error != nil {
+		return scopes, CheckError(result)
+	} else if sharedCount > 0 {
+		scopes[model.SavedQueryScopeShared] = true
```
mistahj67, line 119: Lets leverage the `IsSavedQuerySharedToUser` method here

cmd/api/src/database/saved_queries_permissions_test.go
```diff
@@ -129,36 +235,106 @@ func TestSavedQueriesPermissions_DeleteSavedQueryPermissionForUsers(t *testing.T
...
+	require.NoError(t, err)
+
+	require.Equal(t, database.SavedQueryScopeMap{
+		model.SavedQueryScopePublic: false,
+		model.SavedQueryScopeOwned:  true,
+		model.SavedQueryScopeShared: false,
+	}, scope)
+}
+
+func TestSavedQueriesPermissions_DeleteSavedQueryPermissionsForUsers(t *testing.T) {
```
mistahj67, line 254: Can we add the case we delete with no users and verify it deletes all

cmd/api/src/database/saved_queries_permissions_test.go
```diff
@@ -42,44 +42,120 @@ func TestSavedQueriesPermissions_SharingToUser(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	query, err := dbInst.CreateSavedQuery(testCtx, user.ID, "Test Query", "TESTING", "Example")
+	require.NoError(t, err)
+
```
mistahj67, line 47: Can we add sharing this to a second user and then below test the scope for that user to guarantee we delete the user shared row as well?

mistahj67 (approved):
Looks solid! I had a few small changes to shore up testing / cleaning but don't want to block merging 🚀 once you've tackled those. Great job and thanks for your patience while we iterated through the scenarios and logic, it came out clean! 🧼

mistahj67 (approved):
🚀

PR #780, opened 2024-08-06
Bed 4538

stephanieslamb:
## Description

Updated the existing GET ListSavedQueries endpoint to return public, shared and owned queries.  Added a query param, scope, which can return only public, shared or owned queries as well as any combination of scope types.  

## Motivation and Context

This PR addresses: BED-4538

This ticket is part of the on going initiative to improve saved queries.  

## How Has This Been Tested?

This has been tested via curl commands(see screenshots below) as well as unit tested. 

## Screenshots (optional):

Owned saved queries
Screenshot 2024-08-06 at 4 59 26 PM

Shared and public queries
Screenshot 2024-08-06 at 4 59 39 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/api/v2/saved_queries.go
```diff
@@ -89,9 +97,46 @@ func (s Resources) ListSavedQueries(response http.ResponseWriter, request *http.
...
+				case "public":
+					scopedQueries, err = s.DB.GetPublicSavedQueries(request.Context())
+					scopedCount = len(scopedQueries)
+				case "shared":
+					scopedQueries, err = s.DB.GetSharedSavedQueries(request.Context(), user.ID)
+					scopedCount = len(scopedQueries)
+				case "owned":
+					scopedQueries, scopedCount, err = s.DB.ListSavedQueries(request.Context(), user.ID, strings.Join(order, ", "), sqlFilter, skip, limit)
+				default:
+					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "invalid scope param", request), response)
```
superlinkx, line 123: Add return after to ensure we are closing the connection after error

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -75,12 +78,17 @@ func (s Resources) ListSavedQueries(response http.ResponseWriter, request *http.
...
 					}
-
 					queryFilters[name][i].IsStringData = savedQueries.IsString(filter.Name)
 				}
 			}
 		}
 
+		// needed to remove scope query param since we currently
+		// dont support query params that arent filterable
+		maps.DeleteFunc(queryFilters, func(s string, filters model.QueryParameterFilters) bool {
```
superlinkx, line 88: Move logic to ParseQueryParamFilter function as a carveout, like we do with pagination. This keeps our special logic in one place even if it's similarly hacky.

cmd/api/src/api/v2/saved_queries_test.go
```diff
@@ -262,6 +263,321 @@ func TestResources_ListSavedQueries(t *testing.T) {
...
+	mockDB.EXPECT().ListSavedQueries(gomock.Any(), userId, gomock.Any(), gomock.Any(), 1, 10).Return(model.SavedQueries{
+		{
+			UserID:      userId.String(),
+			Name:        "myQuery",
+			Query:       "Match(n) return n;",
+			Description: "Public query description",
+		},
+	}, 1, nil)
+
+	if req, err := http.NewRequestWithContext(createContextWithOwnerId(userId), "GET", endpoint, nil); err != nil {
```
superlinkx, line 287: If we change these to:

```go
req, err := http.NewRequestWithContext(createContextWithOwnerId(userId), "GET", endpoint, nil)
require.Nil(t, err)

q := url.Values{}
...
```

We can avoid indenting logic flows and better see what points are required to run the test

cmd/api/src/api/v2/saved_queries_test.go
```diff
@@ -262,6 +263,321 @@ func TestResources_ListSavedQueries(t *testing.T) {
...
+
+		req.Header.Set(headers.ContentType.String(), mediatypes.ApplicationJson.String())
+		req.URL.RawQuery = q.Encode()
+
+		router := mux.NewRouter()
+		router.HandleFunc(endpoint, resources.ListSavedQueries).Methods("GET")
+
+		response := httptest.NewRecorder()
+		router.ServeHTTP(response, req)
+		require.Equal(t, http.StatusOK, response.Code)
```
superlinkx, line 305: I always recommend using assert for your actual test criteria, as they will not fatally stop the test. In this case, the test has no logic after a single assert, but in the future there may be additional test cases or possible cleanup logic and changing these to asserts will help out the future engineer working on the test.

superlinkx (changes requested):

superlinkx:
Test changes aren't required, they're complete nits. The other two include a nice to have and a requirement to ensure we don't try to send more than one API response back. Once that's fixed, I'm happy to approve these changes as they're otherwise perfect!

cmd/api/src/api/v2/saved_queries_test.go
stephanieslamb: updated!

superlinkx (approved):

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -89,9 +89,47 @@ func (s Resources) ListSavedQueries(response http.ResponseWriter, request *http.
...
+
+				switch scope {
+				case "public":
+					scopedQueries, err = s.DB.GetPublicSavedQueries(request.Context())
+					scopedCount = len(scopedQueries)
+				case "shared":
+					scopedQueries, err = s.DB.GetSharedSavedQueries(request.Context(), user.ID)
+					scopedCount = len(scopedQueries)
+				case "owned":
+					scopedQueries, scopedCount, err = s.DB.ListSavedQueries(request.Context(), user.ID, strings.Join(order, ", "), sqlFilter, skip, limit)
```
mistahj67, line 113: If these are the same scopes as https://github.com/SpecterOps/BloodHound/blob/3a68a2240e716d4fd9a89fad77c014d7d03cee78/cmd/api/src/database/saved_queries_permissions.go

We should use the constants as those are Title case and these are lowercase

stephanieslamb: They are the same! I didn't know those had been added in Ari's PR.  Since her PR isn't merged yet, would you like me to go ahead and add them in mine? Or wait until her PR has been merged?

superlinkx: You may want to create separate constants in this file, as they are the same, but different (title vs lowercase, and different layers). I guess we could use the same constants if we were to `toLower` the constant before checking it against scope (may want to `toLower` the scope var in the switch too)

superlinkx: Actually, reviewing those constants, I'm not sure they should be title case. We could probably get those changed to lowercase and reuse them. It wouldn't hurt to make sure we're setting the incoming `scope` parameter to lowercase before checking against our strings/constants either way though

stephanieslamb: Updated this.  Let me know what you both think.  I did just add the consts that are in Ari's PR.

superlinkx: This is great. Ship it.

superlinkx (approved):

PR #793, opened 2024-08-14
BED-4707 added saved queries unshare endpoint

mvlipka:
## Description

This change will allow users that own a saved query as well as users that are admins to revoke saved query permissions granted to other users.

## Motivation and Context

This PR addresses: BED-4707

## How Has This Been Tested?

I created two shared query permissions shared to my user
image

I then tested that I could not unshare a query that I own:
image

Then I tested that I could unshare a query that was shared to me:
image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/api/registration/v2.go
```diff
@@ -164,6 +164,7 @@ func NewV2API(cfg config.Configuration, resources v2.Resources, routerInst *rout
 		routerInst.GET("/api/v2/saved-queries", resources.ListSavedQueries).RequirePermissions(permissions.SavedQueriesRead),
 		routerInst.POST("/api/v2/saved-queries", resources.CreateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.PUT(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.UpdateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
+		routerInst.PUT(fmt.Sprintf("/api/v2/saved-queries/{%s}/unshare", api.URIPathVariableSavedQueryID), resources.UnshareSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
```
mistahj67, line 167: Would a DELETE method on the `/share` endpoint be more apt? 🤔

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -242,3 +244,52 @@ func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.
...
+	} else if savedQueryID, err := strconv.ParseInt(rawSavedQueryID, 10, 64); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsFromMalformed, request), response)
+	} else if err := json.NewDecoder(request.Body).Decode(&unshareRequest); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else {
+		if unshareRequest.Self {
+			// User is trying to unshare from themselves
+			if err := s.DB.DeleteSavedQueryPermissionsForUser(request.Context(), savedQueryID, user.ID); err != nil {
+				api.HandleDatabaseError(request, response, err)
+			}
```
mistahj67, line 275: We need to verify that they aren't the query owner here correct? Otherwise that's a no-op. 

I can see why a user might want to unshare someone else sharing a query to them but you can't share your own query, it's filtered out in the `/share` handler?

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -242,3 +244,52 @@ func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.
...
+		} else {
+			// Unshare from users in user_ids
+			isAdmin := user.Roles.Has(model.Role{Name: auth.RoleAdministrator})
+			if savedQueryBelongsToUser, err := s.DB.SavedQueryBelongsToUser(request.Context(), user.ID, savedQueryID); err != nil {
+				api.HandleDatabaseError(request, response, err)
```
mistahj67, line 280: If they are admin we can skip this db hit right?

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -242,3 +244,52 @@ func (s Resources) DeleteSavedQuery(response http.ResponseWriter, request *http.
...
+			} else if isAdmin || savedQueryBelongsToUser {
+				for _, userID := range unshareRequest.UserIds {
+					if err := s.DB.DeleteSavedQueryPermissionsForUser(request.Context(), savedQueryID, userID); err != nil {
+						api.HandleDatabaseError(request, response, err)
+						return
+					}
+				}
```
mistahj67, line 287: can we batch this delete instead?

cmd/api/src/api/v2/saved_queries.go
mvlipka: Yeah, I can re-arrange that `if` block

cmd/api/src/api/v2/saved_queries.go
mvlipka: This would no-op in that situation. I can easily add a check to avoid the db call

cmd/api/src/api/registration/v2.go
mvlipka: That could work.
I was debating between DELETE and PUT on the current endpoint. This endpoint is _only_ deleting entries, but it's a modification to the `saved-queries` which is the route that it's attached to.

In hindsight it may have been good to breakaway from our hackathon URL scheme and put these endpoints on their own resource IE:
`/api/v2/saved-queries/{%s}/permissions/share`

mvlipka: Went ahead and changed the endpoint to be `DELETE api/v2/saved-queries/{%s}/permissions`

Also updated @ALCooper12 on the change to reflect the new routing in her PR

cmd/api/src/api/v2/saved_queries.go
mvlipka: Added a check for if a saved query is shared to the user when attempting to unshare from themselves.
When a user creates a saved query, do not have an associated saved_query_permission for their user id and the newly created saved query id.
We also prevent a user from sharing to themselves.

This means that a user will never have a sqp for a query they created. Simply checking if a user has a sqp to that saved query allows us to check for both scenarios:

1. Ensure a user does not own the query
2. Ensure a user has the query shared to them

mvlipka:
I re-tested using the same instructions in the description.
Only caveat is that the url is now:
`/api/v2/saved-queries/{saved_query_id}/permissions`

packages/go/openapi/src/paths/cypher.saved-queries.id.permissions.yaml
```diff
@@ -0,0 +1,62 @@
...
+delete:
+  operationId: DeleteSavedQueryPermissions
+  summary: Revokes permission of a saved query from users
+  description: Revokes permission of a saved query from a given set of users
+  tags:
+    - Cypher
+    - Community
+    - Enterprise
+  requestBody:
+    description: The request body for revoking permission a saved query from users
```
ALCooper12, line 35: Typo here. Is it supposed to say: `The request body for revoking the permission of a saved query from users` ?

packages/go/openapi/src/paths/cypher.saved-queries.id.permissions.yaml
```diff
@@ -0,0 +1,62 @@
...
+    required: true
+    content:
+      application/json:
+        schema:
+          type: object
+          properties:
+            user_ids:
+              type: array
+              items:
+                  $ref: './../schemas/model.components.uuid.yaml'
```
ALCooper12, line 45: I think the above should be 
```
type: array
items: 
  type: string
  format: uuid
```
instead. I say this because your `$ref: './../schemas/model.components.uuid.yaml'` has `readOnly: true` when your request body's field: `user_ids` isn't

Issue #417, opened 2024-02-14
Visual representation of Owned and all display system_tags in the property overview


olafhartong:
**Feature Description:**

I do like the new system_tag idea. However there is no visual representation of owned and it also does not show up in the Object Information, which makes it easy to miss. Is that something to still be added ?

**Current Behavior:**

Currently there is no clear way of knowing if an asset on a path is marked as owned, yet the High Value is represented with an icon.

Both tags are not visible on the property panel on the right, only in a separate menu.

**Desired Behavior:**

Owned should be visible like in the legacy version of BloodHound with some form of an indicator on the node. Additionally it would be great to also see the system_tags in the panel on the right.

**Use Case:**

Easier and more intuitive navigation

**Implementation Suggestions:**

Ideally there are more custom tags fields to be added to enrich the dataset

PR #795, opened 2024-08-15
Owned object visual representation

Palt:
## Description

Added a visual indicator to an owned object/node.

## Motivation and Context

This PR addresses: #417 
This change makes it easier to get an overview of owned object through a visual indicator.

## How Has This Been Tested?
Locally tested manually and using existing tests.

## Screenshots (optional):
bloodhound_visual_conf_owned_object

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/model/unified_graph.go
```diff
@@ -78,6 +79,7 @@ func FromDAWGSNode(node *graph.Node, includeProperties bool) UnifiedNode {
 		Kind:       analysis.GetNodeKind(node).String(),
 		ObjectId:   objectId,
 		IsTierZero: strings.Contains(systemTags, ad.AdminTierZero),
+		IsOwnedObject: strings.Contains(systemTags, "owned"),
```
urangel, line 82: There is a defined constant for "owned" that would be good to use here. Would you please replace the string here with the `OwnedAssetGroupTag` constant?

cmd/ui/src/ducks/explore/reducer.ts
```diff
@@ -51,10 +51,34 @@ const graphDataReducer = (state = initialGraphDataState, action: types.GraphActi
         } else if (action.type === types.TOGGLE_TIER_ZERO_NODE) {
             const systemTags = state.chartProps.items[action.nodeId].data.system_tags;
             // remove the tier zero tag from the node
-            if (systemTags === 'admin_tier_0') {
-                draft.chartProps.items[action.nodeId].data.system_tags = '';
+            if (systemTags.includes('admin_tier_0')) {
```
urangel, line 54: I appreciate that you considered the handling of both tags here in various situations. We can accomplish something similar by using the `replace` function which makes this logic more concise and clear. 

```typescript
        } else if (action.type === types.TOGGLE_TIER_ZERO_NODE) {
            const systemTags = state.chartProps.items[action.nodeId].data.system_tags;
            // remove the tier zero tag from the node
            systemTags.replace(TIER_ZERO_TAG, "").trim();
        } else if (action.type === types.TOGGLE_OWNED_OBJECT_NODE) {
            const systemTags = state.chartProps.items[action.nodeId].data.system_tags;
            // remove the owned object tag from the node
            systemTags.replace(OWNED_OBJECT_TAG, "").trim();
        }
```

Here is some example usage from trying this out in the browser console:

```
let dualTag = "admin_tier_0 owned"
> undefined
dualTag.replace("owned", "").trim()
> "admin_tier_0"
let singleTag = "admin_tier_0"
> undefined
singleTag.replace("owned", "").trim()
> "admin_tier_0"
let emptyTag = ""
> undefined
emptyTag.replace("owned", "").trim()
> ""
let flippedDualTag = "owned admin_tier_0"
> undefined
flippedDualTag.replace("owned", "").trim()
> "admin_tier_0" 
```

cmd/ui/src/utils.ts
```diff
@@ -209,7 +215,7 @@ export const transformToFlatGraphResponse = (graph: GraphResponse) => {
                 nodetype: value.kind,
                 name: value.label,
                 objectid: value.objectId,
-                system_tags: value.isTierZero ? 'admin_tier_0' : undefined,
+                system_tags: tags,
```
urangel, line 218: This value should be a string as that is the way the system_tags property is stored in the database. A quick fix to this can be made by joining the array elements, i.e., `tags.join(' ')`

cmd/ui/src/ducks/explore/reducer.ts
Palt: Great advice! Will look in to this. Do we need to consider multiple dangling spaces i.e. in the future if there is three or more system_tags? Ex:
```
let tripleTag = "admin_tier_0 some_tag owned"
> undefined
tripleTag.replace("some_tag", "").trim()
> "admin_tier_0  owned" // Two spaces
tripleTag.replace("some_tag", "").trim().replace(/ {2,}/g, " ")
> "admin_tier_0 owned" // Removed extra space in middle
```

urangel: Hmm, yea that is a good point! Handling that scenario would be nice in case we did have more tags like you mention. You are welcome to handle that here as you wish but if you are using regex pattern matching maybe leave a comment explaining what it is doing

cmd/ui/src/ducks/explore/reducer.ts
```diff
@@ -50,36 +51,16 @@ const graphDataReducer = (state = initialGraphDataState, action: types.GraphActi
...
-                    draft.chartProps.items[action.nodeId].data.system_tags = ''
-                }
-            } else {
-                // check if node is marked as owned or not
-                { systemTags.includes('owned') ?
-                    draft.chartProps.items[action.nodeId].data.system_tags = 'admin_tier_0 owned' :
-                    draft.chartProps.items[action.nodeId].data.system_tags = 'admin_tier_0'
-                }
-            }
+            if (systemTags.includes(TIER_ZERO_TAG)) {
```
Palt, line 54: Instead of working with regex would it be preferred to use `splice` and do something like the following:
```typescript
let systemTags = []
// Check if system_tags contains tags then split, else leave empty
{ state.chartProps.items[action.nodeId].data.system_tags ?
	systemTags = state.chartProps.items[action.nodeId].data.system_tags.split(" ") : null }

if (systemTags.includes(TIER_ZERO_TAG)) {
	// Remove tag
	systemTags.splice(systemTags.indexOf(TIER_ZERO_TAG), 1);
	draft.chartProps.items[action.nodeId].data.system_tags = systemTags.join(' ');
} else {
	// Add tag
	systemTags.push(TIER_ZERO_TAG);
	draft.chartProps.items[action.nodeId].data.system_tags = systemTags.join(' ');
}
```

urangel: Yea that looks nicer 👍

PR #808, opened 2024-08-22
BED-4726 added new ~eq predicate for filtering on strings

mvlipka:
## Description

This change adds the ability to query text columns in our database, if specified to be allowed, for approximately equals.

## Motivation and Context

This was created as part of the saved queries improvements as we required the ability to search for saved queries by their description. Previously it was only possible to search text columns for equals and not equals, this would be frustrating as you may only remember a portion of the description that you wrote.
We also added support for text searching the `saved_queries.name` column.

The `gin` index was added to the relevant columns in order to improve text search performance.
https://www.postgresql.org/docs/9.1/textsearch-indexes.html

This PR addresses: BED-4726

## How Has This Been Tested?
I created a couple of sample rows in the `saved_queries` table
image
NOTE: all of the queries for this must belong to the user that is performing the API request

I then sent a query to search for a description
image
We can see here that only one of the descriptions appears

Querying for a description that does not exist:
image

Querying for a name that returns multiple results:
image

Querying for a name returns a single result
image

Querying for a description with spaces in it:
image

Querying for a name with spaces that contains the term `test` in multiple rows, but only returns the case-insensitive match
image

Testing querying with `%20` to represent a space instead of `+`
image

Testing the existing `eq` predicate works with spaces
image

Testing the existing `neq` predicate works as it did previously
image

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/model/filter.go
```diff
@@ -344,6 +352,6 @@ func (s QueryParameterFilterParser) ParseQueryParameterFilters(request *http.Req
 
 func NewQueryParameterFilterParser() QueryParameterFilterParser {
 	return QueryParameterFilterParser{
-		re: regexp.MustCompile(`([\w]+):([\w\d\--_]+)`),
+		re: regexp.MustCompile(`([~\w]+):([\w\d\--_]+)`),
```
mvlipka, line 355: I created a test suite on regexr with both the existing predicates and the new one to verify it's working
regexr.com/8530l

cmd/api/src/model/filter.go
```diff
@@ -165,6 +170,9 @@ func (s QueryParameterFilterMap) BuildSQLFilter() (SQLFilter, error) {
 				predicate = EqualsSymbol
 			case NotEquals:
 				predicate = NotEqualsSymbol
+			case ApproximatelyEquals:
+				predicate = ApproximatelyEqualSymbol
+				filter.Value = fmt.Sprintf("%%%s%%", filter.Value)
```
mvlipka, line 175: I tested adding `%` in Bruno and adding `%` acts as a an empty character, see screenshots

cmd/api/src/model/filter.go
```diff
@@ -39,13 +39,15 @@ const (
...
 	NotEquals           FilterOperator = "neq"
+	ApproximatelyEquals FilterOperator = "~eq"
 
 	GreaterThanSymbol         string = ">"
 	GreaterThanOrEqualsSymbol string = ">="
 	LessThanSymbol            string = "<"
 	LessThanOrEqualsSymbol    string = "<="
 	EqualsSymbol              string = "="
 	NotEqualsSymbol           string = "<>"
+	ApproximatelyEqualSymbol  string = "ILIKE"
```
mvlipka, line 50: ILIKE allows for case insensitive search

cmd/api/src/model/filter.go
```diff
@@ -344,6 +352,6 @@ func (s QueryParameterFilterParser) ParseQueryParameterFilters(request *http.Req
 
 func NewQueryParameterFilterParser() QueryParameterFilterParser {
 	return QueryParameterFilterParser{
-		re: regexp.MustCompile(`([\w]+):([\w\d\--_]+)`),
+		re: regexp.MustCompile(`([~\w]+):([\w\--_ ]+)`),
```
mvlipka, line 355: I created a regexr suite of tests to validate the existing predicates and the new one
https://regexr.com/8583j

mvlipka: Golang correctly parses `+` and `%20` and replaces them with spaces, the empty character in the regex string is intentional so that we can support spaces in search terms

definitelynotagoblin (approved):

cmd/api/src/model/filter.go
mistahj67: I'm not sure how easy it is but if it's not too much more work, I think having a unit test around this regex would be a nice addition!

cmd/api/src/model/filter.go
mistahj67: ILIKE it!

cmd/api/src/model/filter.go
mvlipka: We already have a unit test for the function, but I can definitely add additional scenarios to it!
https://github.com/SpecterOps/BloodHound/blob/3a193574fad089963f433b6f2a3ba5ac2d0c41c1/cmd/api/src/model/filter_internal_test.go#L25

packages/go/openapi/doc/openapi.json
```diff
@@ -13929,7 +13929,7 @@
       },
       "api.params.predicate.filter.string": {
         "type": "string",
-        "description": "Filter results by column string value. Valid filter predicates are `eq`, `neq`.\n"
+        "description": "Filter results by column string value. Valid filter predicates are `eq`, `neq`, `~eq`.\n"
```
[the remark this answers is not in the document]
mvlipka: I can see that, it's related to `eq` so would be nice to have it read like a cascading
"equals, approximately equals, not equals"

cmd/api/src/model/saved_queries.go
```diff
@@ -53,9 +53,9 @@ func (s SavedQueries) IsSortable(column string) bool {
 func (s SavedQueries) ValidFilters() map[string][]FilterOperator {
 	return map[string][]FilterOperator{
 		"user_id":     {Equals, NotEquals},
-		"name":        {Equals, NotEquals},
+		"name":        {Equals, NotEquals, ApproximatelyEquals},
 		"query":       {Equals, NotEquals},
```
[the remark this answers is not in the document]
mvlipka: `query` may be updated frequently and adding the index on it could hit the write performance.
It could be something to think about later.

cmd/api/src/database/migration/migrations/v5.15.1.sql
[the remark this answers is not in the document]
mvlipka: Probably cutting it close to the next release since we're about to cut the RC.
I'll throw it in for v5.16.0

cmd/api/src/model/filter.go
mistahj67: Yes please! 🙇

mistahj67 (approved):
Thanks for your patience and adding the extra test coverage! 🚀

PR #841, opened 2024-09-05
refactor: adcs cache with RWMutex

mistahj67:
## Description

Added RWMutex to the adcs cache
Swapped to a getter / setter approach to help protect concurrent access to the inner properties with correct locking

## Motivation and Context

This PR addresses: BED-4796

Analysis was reporting concurrent writes to maps across goroutines./

## How Has This Been Tested?

Relied on testing but should be rigorously tested in larger environments

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/analysis/ad/adcscache.go
+++ b/packages/go/analysis/ad/adcscache.go
@@ -28,7 +28,7 @@ import (
 )
 
 type ADCSCache struct {
-	sync.RWMutex
+	mu *sync.RWMutex
 
 	// To discourage direct access without getting a read lock, these are private
 	authStoreForChainValid          map[graph.ID]cardinality.Duplex[uint32]
@@ -44,6 +44,7 @@ type ADCSCache struct {
 
 func NewADCSCache() ADCSCache {
 	return ADCSCache{
+		mu:                              &sync.RWMutex{},
 		authStoreForChainValid:          make(map[graph.ID]cardinality.Duplex[uint32]),
 		rootCAForChainValid:             make(map[graph.ID]cardinality.Duplex[uint32]),
 		expandedCertTemplateControllers: make(map[graph.ID]cardinality.Duplex[uint32]),
@@ -57,8 +58,8 @@ func NewADCSCache() ADCSCache {
 }
 
 func (s *ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterpriseCAs, certTemplates, domains []*graph.Node) {
-	s.Lock()
-	defer s.Unlock()
+	s.mu.Lock()
+	defer s.mu.Unlock()
 
 	err := db.ReadTransaction(ctx, func(tx graph.Transaction) error {
 		for _, ct := range certTemplates {
@@ -130,8 +131,8 @@ func (s *ADCSCache) DoesCAChainProperlyToDomain(enterpriseCA, domain *graph.Node
 	var domainID = domain.ID
 	var caID = enterpriseCA.ID.Uint32()
 
-	s.RLock()
-	defer s.RUnlock()
+	s.mu.RLock()
+	defer s.mu.RUnlock()
 	if _, ok := s.rootCAForChainValid[domainID]; !ok {
 		return false
 	} else if _, ok := s.authStoreForChainValid[domainID]; !ok {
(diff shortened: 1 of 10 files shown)


packages/go/analysis/ad/adcscache.go
```diff
@@ -27,46 +28,52 @@ import (
...
-	AuthStoreForChainValid          map[graph.ID]cardinality.Duplex[uint32]
-	RootCAForChainValid             map[graph.ID]cardinality.Duplex[uint32]
-	ExpandedCertTemplateControllers map[graph.ID][]uint32
-	CertTemplateEnrollers           map[graph.ID][]*graph.Node // principals that have enrollment on a cert template via `enroll`, `generic all`, `all extended rights` edges
-	CertTemplateControllers         map[graph.ID][]*graph.Node // principals that have privileges on a cert template via `owner`, `generic all`, `write dacl`, `write owner` edges
-	EnterpriseCAEnrollers           map[graph.ID][]*graph.Node // principals that have enrollment rights on an enterprise ca via `enroll` edge
-	PublishedTemplateCache          map[graph.ID][]*graph.Node // cert templates that are published to an enterprise ca
-	HasUPNCertMappingInForest       map[graph.ID]struct{}      // domains where at least one DC in the forest has Schannel UPN cert mapping enabled
-	HasWeakCertBindingInForest      map[graph.ID]struct{}      // domains where at least one DC in the forest has Kerberos weak cert binding enabled
+	sync.RWMutex
```
zinic, line 31: I have two concerns about this particular change:

This promotes an ambiguous call surface to a consumer of the `ADCSCache` type. The embedding will elevate all exported receiver functions defined in `sync.RWMutex` to the available call surface for `ADCSCache`.

This hides some important semantics. A mutex in golang comes with some usage caveats outlined in https://go.dev/src/sync/mutex.go namely: `A Mutex must not be copied after first use.`. While the current implementation correctly references the `ADCSCache` type by reference `s *ADCSCache` it is not immediately obvious to users of this type that it now has semantic restrictions on when it can be copied. 

My suggestion here is a refactor that sequesters the mutex as a non-exported struct member.

packages/go/analysis/ad/adcscache.go
```diff
@@ -91,51 +98,132 @@ func (s ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterprise
...
+	s.Lock()
+	defer s.Unlock()
+
+	if _, ok := s.expandedCertTemplateControllers[certId]; !ok {
+		s.expandedCertTemplateControllers[certId] = cardinality.NewBitmap32With(principalId)
+	} else {
+		s.expandedCertTemplateControllers[certId].Add(principalId)
+	}
+
+	return
```
zinic, line 163: Missed a bare `return` here that should be removed.

packages/go/analysis/ad/adcscache.go
```diff
@@ -91,51 +98,132 @@ func (s ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterprise
...
+	} else if _, ok := s.authStoreForChainValid[domainID]; !ok {
 		return false
 	} else {
-		return s.RootCAForChainValid[domainID].Contains(caID) && s.AuthStoreForChainValid[domainID].Contains(caID)
+		return s.rootCAForChainValid[domainID].Contains(caID) && s.authStoreForChainValid[domainID].Contains(caID)
 	}
 }
 
-func HasUPNCertMappingInForest(tx graph.Transaction, domain *graph.Node) (bool, error) {
+func (s *ADCSCache) GetExpandedCertTemplateControllers(id graph.ID) (cardinality.Duplex[uint32], bool) {
```
zinic, line 144: General question, not a fix suggestion: if all we do with the boolean return is assign it to the no-name variable `_` then should we just refactor these to return empty cardinality duplexes?

zinic: Same applies to all incarnations of this pattern in this changeset.

zinic (changes requested):
Locking logic looks sound. I appreciate the isolation of critical sections by using non-exported renames of the exported functions.

I have some concerns that I think warrant a small refactor and some thoughts around other code simplification tasks that might make this more ergonomic.

packages/go/analysis/ad/adcscache.go
mistahj67: 👍 Thanks for explaining, ❤️ the added protection this buys us.

packages/go/analysis/ad/adcscache.go
```diff
@@ -27,46 +28,53 @@ import (
 )
 
 type ADCSCache struct {
```
zinic, line 30: Struct and its members look far cleaner. Thanks for the refactor here.

packages/go/analysis/ad/adcscache.go
```diff
@@ -91,51 +99,125 @@ func (s ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterprise
...
+		return s.rootCAForChainValid[domainID].Contains(caID) && s.authStoreForChainValid[domainID].Contains(caID)
 	}
 }
 
-func HasUPNCertMappingInForest(tx graph.Transaction, domain *graph.Node) (bool, error) {
+func (s *ADCSCache) GetExpandedCertTemplateControllers(id graph.ID) cardinality.Duplex[uint32] {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	if expandedCertTemplateControllers, ok := s.expandedCertTemplateControllers[id]; !ok {
```
zinic, line 149: This is a great way to exercise the `value-not-found` return component of a map lookup while hiding it from consumers of the struct's exported contract.

packages/go/analysis/ad/adcscache.go
```diff
@@ -91,51 +99,125 @@ func (s ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterprise
...
+	defer s.mu.RUnlock()
+
+	return s.enterpriseCAEnrollers[id]
+}
+
+func (s *ADCSCache) GetPublishedTemplateCache(id graph.ID) []*graph.Node {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	published := s.publishedTemplateCache[id]
```
zinic, line 192: Looks like this was missed and probably could just be collapsed into the return statement like the functions above.

packages/go/analysis/ad/adcscache.go
```diff
@@ -91,51 +99,125 @@ func (s ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterprise
...
+	return published
+}
+
+func (s *ADCSCache) HasUPNCertMappingInForest(id uint32) bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	hasUPNCertMappingInForest := s.hasUPNCertMappingInForest.Contains(id)
```
zinic, line 200: Probably could just be collapsed into the return statement like the functions above.

packages/go/analysis/ad/adcscache.go
```diff
@@ -91,51 +99,125 @@ func (s ADCSCache) BuildCache(ctx context.Context, db graph.Database, enterprise
...
+
+	return hasUPNCertMappingInForest
+}
+
+func (s *ADCSCache) HasWeakCertBindingInForest(id uint32) bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	hasWeakCertBindingInForest := s.hasWeakCertBindingInForest.Contains(id)
```
zinic, line 209: Probably could just be collapsed into the return statement like the functions above.

zinic (approved):
Going to approve this with some minor nits in comment. Feel free to merge as-is.

PR #854, opened 2024-09-12
BED-4388: Properly handle paths passed to OrphanFileSweeper

computator:
## Description

This updates the OrphanFileSweeper to properly preserve intended files when cleaning up ingest files. Previously it did not handle being passed filenames with a path, but would only work for plain filenames. Now it will properly compare filenames with a path to see if they should be preserved.

## Motivation and Context

This PR addresses: BED-4388

If the API was restarted during ingest, multiple valid but not yet processed ingest files would be deleted. This would cause file not found errors when attempting to process those files, as well as requiring another collection to reacquire the deleted ingest data.

## How Has This Been Tested?

I ran AzureHound locally and triggered an on-demand collection. I then monitored the ingest directory during the collection process and restarted the API while there were multiple files yet to be processed. I verified that upon API restart most of the files were preserved, and that there were no missing file errors in the API logs.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/daemons/datapipe/cleanup.go
+++ b/cmd/api/src/daemons/datapipe/cleanup.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/specterops/bloodhound/log"
@@ -61,7 +62,7 @@ func NewOrphanFileSweeper(fileOps FileOperations, tempDirectoryRootPath string) 
 	return &OrphanFileSweeper{
 		lock:                  &sync.Mutex{},
 		fileOps:               fileOps,
-		tempDirectoryRootPath: tempDirectoryRootPath,
+		tempDirectoryRootPath: strings.TrimSuffix(tempDirectoryRootPath, string(filepath.Separator)),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ddlees (approved):

cmd/api/src/daemons/datapipe/cleanup.go
```diff
@@ -78,15 +79,27 @@ func (s *OrphanFileSweeper) Clear(ctx context.Context, expectedFileNames []strin
 	// Release the lock once finished
 	defer s.lock.Unlock()
 
+	log.Infof("Running OrphanFileSweeper for path %s", s.tempDirectoryRootPath)
```
mistahj67, line 82: Should this be info or debug?

ddlees: I think info is appropriate since were announcing the removal of data at a given path.

computator: I was debating on that as well but ended up choosing info since it would go along with the "x orphan files removed" message that is also info

mistahj67: I saw the lower one and was thinking it would be sufficient. But if this is gated to just on actual removal, that makes sense to me as well!

computator: It's not gated to removal, it runs every time. It's more just a "start of process" + location message, and the other one is the corresponding end of process message.

PR #869, opened 2024-09-17
Revert "BED-4776: Add .github/CODEOWNERS"

irshadaj:
Reverts SpecterOps/BloodHound#839

Reverting this since it is creating unwanted noise in emails by subscribing the entire team to every PR opened.


mistahj67:
General comment but I recall that @elikmiller had to adjust some things at the repository level when this change was implemented, did you or someone with the correct permissions also revert those changes as well outside of this change 🤔

zinic (approved):

PR #873, opened 2024-09-19
Bed 4752

stephanieslamb:
## Description

Updated specified log errors to warns in order to cut down noise.  

## Motivation and Context

This PR addresses: BED-4752

This change was requested to lessen confusion post releases or while investigating certain issues. 

## How Has This Been Tested?
No real code changes.  Unit tests still pass. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/go/analysis/ad/esc3.go
```diff
@@ -140,7 +141,9 @@ func PostEnrollOnBehalfOf(certTemplates []*graph.Node, operation analysis.StatTr
 	versionTwoTemplates := make([]*graph.Node, 0)
 
 	for _, node := range certTemplates {
-		if version, err := node.Properties.Get(ad.SchemaVersion.String()).Float64(); err != nil {
+		if version, err := node.Properties.Get(ad.SchemaVersion.String()).Float64(); errors.Is(err, graph.ErrPropertyNotFound) {
+			log.Warnf("Did not get schema version for cert template %d: %v", node.ID, err)
```
mvlipka, line 145: Nit: mixing `Did not` and `Didnt` can make it confusing to search logs

cmd/api/src/daemons/datapipe/jobs.go
```diff
@@ -214,7 +216,9 @@ func (s *Daemon) processIngestFile(ctx context.Context, path string, fileType mo
 
 				if err := file.Close(); err != nil {
 					log.Errorf("Error closing ingest file %s: %v", filePath, err)
-				} else if err := os.Remove(filePath); err != nil {
+				} else if err := os.Remove(filePath); errors.Is(err, fs.ErrNotExist) {
+					log.Warnf("Removing ingest file %s: %v", filePath, err)
```
mvlipka, line 220: Nit: `%w` probably makes more sense than `%v`.
While they are synonymous, `%w` enforces the `Error` interface to be implemented, which for a lot of the warnings we have, it seems like that's the case that we want.

I see `%v` was already being used, and we use it a lot elsewhere, so moreso just an observation

packages/go/ein/azure.go
```diff
@@ -41,7 +41,10 @@ const (
 	KeyVaultPermissionGet string = "Get"
 )
 
-var resourceGroupLevel = regexp.MustCompile(`^[\\w\\d\\-\\/]*/resourceGroups/[0-9a-zA-Z]+$`)
+var (
+	resourceGroupLevel = regexp.MustCompile(`^[\\w\\d\\-\\/]*/resourceGroups/[0-9a-zA-Z]+$`)
+	InvalidTypeErr     = errors.New("invalid type returned from directory object")
```
mvlipka, line 46: Great refactor!

mvlipka (approved):
LGTM
Couple of nits, but, it doesn't appear we really have as much of a standard when it comes to logging warnings.
Errors are generally prefixed with one of [`Error`, 'Failed', 'Unabled'], but we don't seem to have as much consensus when it comes to warn level messages.

cmd/api/src/daemons/datapipe/jobs.go
stephanieslamb: thanks for the explanation!  went ahead and updated.

mvlipka (approved):

cmd/api/src/queries/graph.go
```diff
@@ -21,6 +21,7 @@ package queries
 import (
 	"bytes"
 	"context"
+	e "errors"
```
zinic, line 24: This is a sign that we should refactor away from using our (admittedly) ancient internal `errors` package. Please update this file to remove the use of `github.com/specterops/bloodhound/errors`.

zinic (changes requested):
One request for refactor, changeset otherwise looks solid

cmd/api/src/queries/graph.go
stephanieslamb: updated file to use `errors`

zinic (approved):

PR #890, opened 2024-09-26
Bed-4852: SSO Provider Delete Endpoint

mvlipka:
## Description

Add a new endpoint to delete SSO Providers

## Motivation and Context

This PR addresses: BED-4852

Adds the ability for admins to delete SSO Providers and any associated references to them

## How Has This Been Tested?

**Basic Test**

I setup my database as followed:
image

Sent a request to
`/api/v2/sso-providers/{sso_provider_id}`
image

And received a 200 response back

Refreshing the tables in the database shows that the references were properly nulled out
image

**Attempting to delete a non-existent sso provider:**
image

**Attempting to delete a SAML provider after replacing DeleteSAMLProvider with the DeleteSSOProvider method**
image

After sending the request:
image

And the audit log with details:
image

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -10,4 +26,7 @@ CREATE TABLE IF NOT EXISTS oidc_providers
   created_at TIMESTAMP WITH TIME ZONE DEFAULT now(),
 
   UNIQUE (name)
-)
+);
+
+ALTER TABLE users ADD COLUMN oidc_provider_id BIGSERIAL;
```
mistahj67, line 31: Did we want to swap this back to serial to match the saml logic?

cmd/api/src/api/v2/auth/oidc.go
```diff
@@ -56,3 +62,22 @@ func (s ManagementResource) CreateOIDCProvider(response http.ResponseWriter, req
...
+	// Convert the incoming string url param to an int64
+	if oidcProviderID, err := strconv.ParseInt(rawOIDCProviderID, 10, 64); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if err = s.db.DeleteOIDCProvider(request.Context(), oidcProviderID); errors.Is(err, database.ErrNotFound) {
+		// Handle error if requested record could not be found
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, err.Error(), request), response)
+	} else if err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else {
+		api.WriteBasicResponse(request.Context(), "", http.StatusOK, response)
```
mistahj67, line 81: I think we can just write the statuscode and don't need to go through the full writeBasicResponse here

cmd/api/src/api/v2/auth/oidc.go
```diff
@@ -17,9 +17,15 @@
...
 import (
+	"errors"
 	"net/http"
+	"strconv"
 	"strings"
 
+	"github.com/gorilla/mux"
+
+	"github.com/specterops/bloodhound/src/database"
+
```
mistahj67, line 28: We want to combine all these into 1 block, same situation below

cmd/api/src/api/v2/auth/oidc_test.go
```diff
@@ -116,3 +121,57 @@ func TestManagementResource_CreateOIDCProvider(t *testing.T) {
...
 }
+
+func TestManagementResource_DeleteOIDCProvider(t *testing.T) {
+	var (
+		url               = "/api/v2/sso/providers/oidc/%s"
+		mockCtrl          = gomock.NewController(t)
+		resources, mockDB = apitest.NewAuthManagementResource(mockCtrl)
+	)
+
+	t.Run("successfully delete an OIDCProvider", func(t *testing.T) {
```
mistahj67, line 132: Can we include some users tied to an OIDC provider and also verify that the FK properly deletes them?

cmd/api/src/database/oidc.go
```diff
@@ -24,7 +24,17 @@ import (
 
 // OIDCProviderData defines the interface required to interact with the oidc_providers table
 type OIDCProviderData interface {
+	GetOIDCProvider(ctx context.Context, id int64) (model.OIDCProvider, error)
 	CreateOIDCProvider(ctx context.Context, name, issuer, clientID string) (model.OIDCProvider, error)
+	DeleteOIDCProvider(ctx context.Context, id int64) error
+}
+
+func (s *BloodhoundDB) GetOIDCProvider(ctx context.Context, id int64) (model.OIDCProvider, error) {
```
mistahj67, line 32: I believe we'll also need a get method for fetching via name / slug as well for the login flow.

cmd/api/src/database/oidc.go
```diff
@@ -37,3 +47,13 @@ func (s *BloodhoundDB) CreateOIDCProvider(ctx context.Context, name, issuer, cli
 
 	return provider, CheckError(s.db.WithContext(ctx).Table("oidc_providers").Create(&provider))
 }
+
+// DeleteOIDCProvider deletes a oidc_provider matching the given id
+func (s *BloodhoundDB) DeleteOIDCProvider(ctx context.Context, id int64) error {
```
mistahj67, line 52: Did we need auditlogs for all of these admin actions?

cmd/api/src/database/sso_providers_test.go
```diff
@@ -43,6 +44,30 @@ func TestBloodhoundDB_CreateSSOProvider(t *testing.T) {
...
+}
+
+func TestBloodhoundDB_DeleteSSOProvider(t *testing.T) {
+	var (
+		testCtx = context.Background()
+		dbInst  = integration.SetupDB(t)
+	)
+	defer dbInst.Close(testCtx)
+
+	t.Run("successfully delete an SSO provider", func(t *testing.T) {
```
mistahj67, line 61: Would it be possible to add a user to the saml provider and ensure that it properly sets it to null on both fields, ssoProviderId and samlProviderId?

And same with an OIDC provider / OIDC user?

cmd/api/src/api/constant.go
```diff
@@ -59,4 +59,6 @@ const (
 	URIPathVariableTokenID                           = "token_id"
 	URIPathVariableUserID                            = "user_id"
 	URIPathVariableSavedQueryID                      = "saved_query_id"
+	URIPathVariableOIDCProviderID                    = "oidc_provider_id"
```
mistahj67, line 62: This is from a previous iteration and not needed anymore right?
```suggestion
```

mvlipka: Yup! Unused, removing

cmd/api/src/database/auth.go
```diff
@@ -541,8 +541,13 @@ func (s *BloodhoundDB) DeleteSAMLProvider(ctx context.Context, provider model.SA
 		Model:  &provider, // Pointer is required to ensure success log contains updated fields after transaction
 	}
 
+	// Delete both the associated SSO Provider and the SAML Provider
 	return s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
-		return CheckError(tx.WithContext(ctx).Delete(&provider))
+		if err := s.DeleteSSOProvider(ctx, int(provider.ID)); err != nil {
```
mistahj67, line 546: isn't the provider.ID the saml provider id and not the sso provider id? 
```suggestion
		if err := s.DeleteSSOProvider(ctx, int(provider.SSOProviderID)); err != nil {
```

cmd/api/src/database/auth.go
```diff
@@ -541,8 +541,13 @@ func (s *BloodhoundDB) DeleteSAMLProvider(ctx context.Context, provider model.SA
...
+			return err
+		} else {
+			return CheckError(tx.WithContext(ctx).Delete(&provider))
+		}
```
mistahj67, line 550: Once we delete the SSOProvider id, the FK will cascade delete the saml provider record, making this always a no-op I think.

cmd/api/src/database/sso_providers.go
```diff
@@ -69,3 +67,40 @@ func (s *BloodhoundDB) GetAllSSOProviders(ctx context.Context, order string, sql
...
+
+	return provider, CheckError(s.db.WithContext(ctx).Table(ssoProviderTableName).Create(&provider))
+}
+
+// DeleteSSOProvider deletes a sso_provider entry with a matching id
+func (s *BloodhoundDB) DeleteSSOProvider(ctx context.Context, id int) error {
+	var (
+		ssoProvider, err = s.GetSSOProvider(ctx, id)
+		auditEntry       = model.AuditEntry{
+			Action: "DeleteSSOProvider",
```
mistahj67, line 88: Can we add this string to the constant definition in `bhce/cmd/api/src/model/audit.go` please?

cmd/api/src/database/sso_providers.go
```diff
@@ -69,3 +67,40 @@ func (s *BloodhoundDB) GetAllSSOProviders(ctx context.Context, order string, sql
...
+// CreateSSOProvider creates an entry in the sso_providers table
+// A slug will be created for the SSO Provider using the name argument as a base. The name will be lower cased and all spaces are replaced with `-`
+func (s *BloodhoundDB) CreateSSOProvider(ctx context.Context, name string, authProvider model.SessionAuthProvider) (model.SSOProvider, error) {
+	provider := model.SSOProvider{
+		Name: name,
+		Slug: strings.ToLower(strings.ReplaceAll(name, " ", "-")),
+		Type: authProvider,
+	}
+
+	return provider, CheckError(s.db.WithContext(ctx).Table(ssoProviderTableName).Create(&provider))
```
mistahj67, line 80: Did we want to have audit logs for create as well 🤔

cmd/api/src/model/sso_provider.go
```diff
@@ -30,6 +30,16 @@ type SSOProvider struct {
 	Serial
 }
 
+// AuditData returns the fields to log in the audit log
+func (s SSOProvider) AuditData() AuditData {
+	return AuditData{
```
mistahj67, line 35: You might be able to get fancy here and based on the type add in the details tied to the child provider info with a switch case like on our LIST endpoint 🤔 

You'll need to do the preload trick that @ddlees mentioned `db.Preload("OIDCProvider").Preload("SAMLProvider")`

mistahj67 (changes requested):

cmd/api/src/database/oidc.go
mistahj67: Disregard, added in a the login PR

cmd/api/src/database/auth.go
mvlipka: In this case it won't no-op.
Since we are deleting them in a transaction, the cascade delete won't actually trigger until the transaction is committed.

This gives us a slight benefit with the audit log:
image

When deleting the SAMLProvider, we also audit log the SSOProvider deletion.

mistahj67: I didn't realize we wanted to keep the individual child provider logs 🤔

mvlipka: Maybe not, but I could see this being beneficial to see which SAML provider was deleted. Sort of relating to the point made that we lose the context of what the SSO Provider was in the delete, this way we have a record of both entries being deleted

cmd/api/src/database/sso_providers.go
```diff
@@ -104,3 +116,12 @@ func (s *BloodhoundDB) DeleteSSOProvider(ctx context.Context, id int) error {
...
 }
+
+// GetSSOProviderUsers returns all the users associated with a given sso provider
+func (s *BloodhoundDB) GetSSOProviderUsers(ctx context.Context, id int) (model.Users, error) {
+	var (
+		users model.Users
+	)
+
+	return users, CheckError(s.db.WithContext(ctx).Table("users").Where("sso_provider_id = ?", id).Find(&users))
+}
```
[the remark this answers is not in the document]
mvlipka: Yeah, I had the same thinking.

Since it returns `model.Users` I was thinking it makes more sense to live alongside the queries pertaining to `Users`, but in this instance I followed what the SAML implementation did:
https://github.com/SpecterOps/BloodHound/blob/3681d66c3bff4c6bd0d4295f6af00b8c485dda53/cmd/api/src/database/db.go#L133

cmd/api/src/database/sso_providers.go
```diff
@@ -69,3 +68,60 @@ func (s *BloodhoundDB) GetAllSSOProviders(ctx context.Context, order string, sql
...
+		return err
+	}
+
+	err = s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+		result := tx.Table(ssoProviderTableName).Delete(ssoProvider)
+		if result.RowsAffected == 0 {
+			return ErrNotFound
+		}
+
+		return CheckError(result)
```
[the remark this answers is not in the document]
mvlipka: This was done to populate the AuditLog, since if we delete with just the ID then the AuditLog will only contain the entry's ID.

cmd/api/src/database/auth.go
mvlipka: Resolving, talked with @mistahj67 and came up with a cleaner way to produce the audit logs

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -191,7 +191,7 @@ func (s ManagementResource) DeleteSAMLProvider(response http.ResponseWriter, req
 		api.HandleDatabaseError(request, response, err)
 	} else if err := s.disassociateUsersFromSAMLProvider(request, providerUsers); err != nil {
 		api.HandleDatabaseError(request, response, err)
```
mistahj67, line 193: This isn't needed anymore as well right?

cmd/api/src/database/oidc_providers.go
```diff
@@ -29,9 +29,18 @@ const (
...
 }
 
+func (s *BloodhoundDB) GetOIDCProvider(ctx context.Context, id int) (model.OIDCProvider, error) {
+	var (
+		oidcProvider model.OIDCProvider
+		result       = s.db.WithContext(ctx).Table(oidcProvidersTableName).Where("id = ?", id).First(&oidcProvider)
+	)
+	return oidcProvider, CheckError(result)
+}
+
```
mistahj67, line 43: We were removing this one as well right?

cmd/api/src/database/auth.go
```diff
@@ -535,17 +535,6 @@ func (s *BloodhoundDB) GetSAMLProvider(ctx context.Context, id int32) (model.SAM
 	return samlProvider, CheckError(result)
 }
 
-func (s *BloodhoundDB) DeleteSAMLProvider(ctx context.Context, provider model.SAMLProvider) error {
-	auditEntry := model.AuditEntry{
-		Action: model.AuditLogActionDeleteSAMLIdentityProvider,
```
mistahj67, line 540: we should also delete this constant as it's no longer used

cmd/api/src/database/sso_providers.go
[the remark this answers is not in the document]
mistahj67: Fair enough 🤷

mistahj67 (commented):
I think there's just a smidge more cleanup and then we 🚀

cmd/api/src/api/v2/auth/auth.go
mvlipka: Just tested locally, and we indeed do not 😄

cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -81,17 +81,23 @@ WHERE saml_providers.sso_provider_id IS NULL;
...
                        FROM saml_providers saml
-                              JOIN sso_providers sso ON sso.id = saml.sso_provider_id
+                                JOIN sso_providers sso ON sso.id = saml.sso_provider_id
                        WHERE u.saml_provider_id = saml.id)
 WHERE sso_provider_id IS NULL
   AND saml_provider_id IS NOT NULL;
+
+-- Set the user's saml_provider_id to null when an sso_provider or saml_provider is deleted
+ALTER TABLE ONLY users
+    DROP CONSTRAINT fk_users_saml_provider;
```
mistahj67, line 101:
```suggestion
    DROP CONSTRAINT IF EXISTS fk_users_saml_provider;
```

cmd/api/src/api/v2/auth/sso.go
```diff
@@ -122,3 +129,33 @@ func (s ManagementResource) ListAuthProviders(response http.ResponseWriter, requ
...
+		requestContext   = ctx.FromRequest(request)
+	)
+
+	// Convert the incoming string url param to an int
+	if ssoProviderID, err := strconv.Atoi(rawSSOProviderID); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if user, isUser := auth.GetUserFromAuthCtx(requestContext.AuthCtx); isUser && user.SSOProviderID.Equal(null.Int32From(int32(ssoProviderID))) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusConflict, "user may not delete their own SSO auth provider", request), response)
+	} else if providerUsers, err := s.db.GetSSOProviderUsers(request.Context(), ssoProviderID); err != nil {
+		api.HandleDatabaseError(request, response, err)
```
mistahj67, line 151: I think this also needs the NotFound handling treatment as below.

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -189,9 +176,7 @@ func (s ManagementResource) DeleteSAMLProvider(response http.ResponseWriter, req
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusConflict, "user may not delete their own SAML auth provider", request), response)
 	} else if providerUsers, err := s.db.GetSAMLProviderUsers(request.Context(), identityProvider.ID); err != nil {
 		api.HandleDatabaseError(request, response, err)
-	} else if err := s.disassociateUsersFromSAMLProvider(request, providerUsers); err != nil {
-		api.HandleDatabaseError(request, response, err)
-	} else if err := s.db.DeleteSAMLProvider(request.Context(), identityProvider); err != nil {
```
mistahj67, line 194: There's a `v1` handler that needs to be swapped to `DeleteSSOProvider` as well

mistahj67 (commented):
Tested locally and hit a few small snags! Otherwise looks solid, love the cleanup ❤️

cmd/api/src/api/v2/auth/sso.go
mvlipka: This should be fine since a user may want to delete an SSO Provider that has not been assigned to any users yet.

mistahj67: Handled offline but the snag was in the db method preload, not here with users! Resolving

cmd/api/src/api/v2/auth/sso.go
```diff
@@ -122,3 +129,33 @@ func (s ManagementResource) ListAuthProviders(response http.ResponseWriter, requ
...
+	} else if err = s.db.DeleteSSOProvider(request.Context(), ssoProviderID); errors.Is(err, database.ErrNotFound) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, err.Error(), request), response)
+	} else if err != nil {
```
mistahj67, line 154:
```suggestion
	} else if err = s.db.DeleteSSOProvider(request.Context(), ssoProviderID); err != nil {
```

mistahj67 (approved):
Looks good, feels good, tests good; Send it! 🚀

PR #894, opened 2024-10-01
Bed-4851 OIDC API Provider Registration

mvlipka:
## Description
Modify the previously merged OIDC provider registration endpoint to support the new slug formatting, new sso_provider table, and move the URL to a new location to prevent collisions with existing endpoints.

## Motivation and Context

After our syncs on 09/27 and 09/30, we decided to make some changes to the last PR in order to better support an agnostic API for both oidc & saml providers.

This PR addresses: BED-4766

## How Has This Been Tested?
`just bh-dev` to start up BHCE and ensure the migrations ran
image

Verified a 201 response and that the data was inserted into the database with the correct formatting and referencing:
image

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -1,13 +1,50 @@
...
 (
-  id         BIGSERIAL PRIMARY KEY,
-  name       TEXT NOT NULL,
-  client_id  TEXT NOT NULL,
-  issuer     TEXT NOT NULL,
+    id              SERIAL PRIMARY KEY,
+    name            TEXT NOT NULL,
+    client_id       TEXT NOT NULL,
+    issuer          TEXT NOT NULL,
+    sso_provider_id INTEGER REFERENCES sso_providers (id) ON DELETE CASCADE NULL,
```
mvlipka, line 39: This supports auto-deleting the oidc_provider row when the associated sso_provider is deleted.

We can handle the scenario when an oidc_provider is deleted, the associated sso_provider is deleted in:
https://github.com/SpecterOps/BloodHound/pull/890

Alternatively we could use a pg trigger to accomplish this, but I'd personally want to handle it through our application

cmd/api/src/database/sso.go
```diff
@@ -0,0 +1,56 @@
...
+		Name: name,
+		Slug: slug,
+		Type: authType,
+	}
+
+	return provider, CheckError(s.db.WithContext(ctx).Table(ssoProviderTableName).Create(&provider))
+}
+
+// CreateSSOProviderWithTransaction uses a db transaction to create an entry in the sso_providers table
+func (s *BloodhoundDB) CreateSSOProviderWithTransaction(ctx context.Context, tx *gorm.DB, name, slug string, authType model.SessionAuthProvider) (model.SSOProvider, error) {
```
mvlipka, line 48: We don't do this elsewhere in the codebase, but I thought it would be helpful to move the responsibility of the transaction to the associated table's db interactor.

This allows us to centralize the database interaction into a single file rather than creating bespoke transactions in other table's db interactions.

cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -1,13 +1,49 @@
...
+    created_at      TIMESTAMP WITH TIME ZONE DEFAULT now()
+);
 
-  updated_at TIMESTAMP WITH TIME ZONE DEFAULT now(),
-  created_at TIMESTAMP WITH TIME ZONE DEFAULT now(),
+-- Create the reference from saml_providers to sso_providers
+ALTER TABLE ONLY saml_providers
+    ADD COLUMN IF NOT EXISTS sso_provider_id INTEGER NULL;
+ALTER TABLE ONLY saml_providers
+    ADD CONSTRAINT fk_saml_provider_sso_provider FOREIGN KEY (sso_provider_id) REFERENCES sso_providers (id) ON DELETE CASCADE;
```
ddlees, line 48: We may want to consider adding logic to hydrate the `sso_providers` table with the relevant information from the `saml_providers` table with this PR.

cmd/api/src/database/migration/migrations/v6.1.0.sql
ddlees: Agreed. This is a good use case for using `ON DELETE CASCADE` as opposed to a trigger.

cmd/api/src/database/migration/migrations/v6.1.0.sql
mvlipka: Agreed, I'll get started on that

cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -1,13 +1,49 @@
...
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- SSO Provider
+CREATE TABLE IF NOT EXISTS sso_providers
+(
+    id         SERIAL PRIMARY KEY,
+    name       TEXT NOT NULL,
+    slug       TEXT NOT NULL,
+    type       int  NOT NULL,
```
mvlipka, line 23: Might want a 3rd table for this, something like `sso_types` which would simply be
```
sso_types
id serial
name string
```

This does complicate our database a bit more than we previously had, but I've used this with success in other codebases and it seemed to be worth the effort

cmd/api/src/database/oidc.go
```diff
@@ -20,20 +20,35 @@ import (
...
+	CreateOIDCProvider(ctx context.Context, name, slug, issuer, clientID string) (model.OIDCProvider, error)
 }
 
-// CreateOIDCProvider creates a new entry for an OIDC provider
-func (s *BloodhoundDB) CreateOIDCProvider(ctx context.Context, name, issuer, clientID string) (model.OIDCProvider, error) {
-	provider := model.OIDCProvider{
-		Name:     name,
+// CreateOIDCProvider creates a new entry for an OIDC provider as well as the associated SSO provider
+func (s *BloodhoundDB) CreateOIDCProvider(ctx context.Context, name, slug, issuer, clientID string) (model.OIDCProvider, error) {
+	oidcProvider := model.OIDCProvider{
```
ddlees, line 37: Consider removing the `name` property from the OIDCProvider model since it not going to be in the table

mvlipka: Thought I caught all the instances of it 😭 
Good catch!

mvlipka: It's removed from the model, but I still have it passed into the method since it is required for the `CreateSSOProvider` method

cmd/api/src/database/sso.go
ddlees: This is fine. I want to say you can also do something like this if you want to avoid having an additional API to maintain:

``` go
	err := s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
	    bhdb := NewBloodhoundDB(tx, s.idResolver)
	    
		if ssoProvider, err := bhdb.CreateSSOProvider(ctx, name, slug, model.SessionAuthProviderOIDC); err != nil {
			return err
		} else {
		    ...
		}
	})
```

cmd/api/src/model/oidc_provider.go
```diff
@@ -16,10 +16,11 @@
...
+// OIDCProvider contains the data needed to initiate an OIDC secure login flow
 type OIDCProvider struct {
-	Name     string `json:"name"`
-	ClientID string `json:"client_id"`
-	Issuer   string `json:"issuer"`
+	ClientID      string `json:"client_id"`
+	Issuer        string `json:"issuer"`
+	SSOProviderID int    `json:"sso_provider_id"`
 
 	BigSerial
```
mistahj67, line 25: Can we swap this to Serial?

cmd/api/src/model/auth.go
```diff
@@ -565,6 +565,7 @@ type SessionAuthProvider int
 const (
 	SessionAuthProviderSecret SessionAuthProvider = 0
 	SessionAuthProviderSAML   SessionAuthProvider = 1
+	SessionAuthProviderOIDC   SessionAuthProvider = 2
```
mistahj67, line 568:
```suggestion
	SessionAuthProviderSecret SessionAuthProvider = iota
	SessionAuthProviderSAML
	SessionAuthProviderOIDC
```

cmd/api/src/api/v2/auth/oidc.go
```diff
@@ -42,17 +42,15 @@ func (s ManagementResource) CreateOIDCProvider(response http.ResponseWriter, req
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
 	} else if validated := validation.Validate(createRequest); validated != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, validated.Error(), request), response)
-	} else if strings.Contains(createRequest.Name, " ") {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "invalid name formatting, ensure there are no spaces in the provided name", request), response)
 	} else {
 		var (
-			formattedName = strings.ToLower(createRequest.Name)
+			formattedName = strings.ToLower(strings.ReplaceAll(createRequest.Name, " ", "-"))
```
mistahj67, line 47: What if we moved the slug generation into the SSO create db method instead? That way it's isolated to just what needs it and other callers don't need to know to implement it in a specific way.

cmd/api/src/database/sso.go
```diff
@@ -0,0 +1,56 @@
...
+type SSOProviderData interface {
+	CreateSSOProvider(ctx context.Context, name, slug string, authType model.SessionAuthProvider) (model.SSOProvider, error)
+	CreateSSOProviderWithTransaction(ctx context.Context, tx *gorm.DB, name, slug string, authType model.SessionAuthProvider) (model.SSOProvider, error)
+}
+
+// CreateSSOProvider creates an entry in the sso_providers table
+func (s *BloodhoundDB) CreateSSOProvider(ctx context.Context, name, slug string, authType model.SessionAuthProvider) (model.SSOProvider, error) {
+	provider := model.SSOProvider{
+		Name: name,
+		Slug: slug,
```
mistahj67, line 40: Here's where we can toss the slug generator from name I think that would eliminate the need to pass it all they way through.

cmd/api/src/model/auth.go
ddlees: The `SessionAuthProvider` value is stored in the DB. If for some reason the values listed after iota were to be reorganized the representation of the data stored would change. To avoid this potential mishap we should avoid using `iota` in this case.

cmd/api/src/database/sso.go
mvlipka: Ooh that's great! I'll get that in

cmd/api/src/model/auth.go
mistahj67: Fair enough!

cmd/api/src/database/sso_providers.go
```diff
@@ -0,0 +1,58 @@
...
+	"strings"
+
+	"github.com/specterops/bloodhound/src/model"
+)
+
+const (
+	ssoProviderTableName = "sso_providers"
+)
+
+var (
```
mvlipka, line 31: I did this since we had talked about using the existing SessionAuthProviders, but since we have the enum in our database, I wanted to have a few constants that can map to that enum as well

mistahj67: If we make the enum an int and match the below could we just continue to use the existing constant? 🤔

cmd/api/src/test/integration/database.go
```diff
@@ -59,6 +59,19 @@ func SetupDB(t *testing.T) database.Database {
 	return dbInst
 }
 
+// SetupTransaction returns a *gorm.DB which can be useful when writing integration tests that require a transaction
+func SetupTransaction(t *testing.T) *gorm.DB {
```
mistahj67, line 63: Is this being used still or is it from a previous iteration and we can clean it up?

cmd/api/src/database/oidc_providers.go
```diff
@@ -20,20 +20,37 @@ import (
...
-	return provider, CheckError(s.db.WithContext(ctx).Table("oidc_providers").Create(&provider))
+	// Create both the sso_providers and oidc_providers rows in a single transaction
+	// If one of these requests errors, both changes will be rolled back
+	err := s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		bhdb := NewBloodhoundDB(tx, s.idResolver)
+
+		if ssoProvider, err := bhdb.CreateSSOProvider(ctx, name, model.SessionAuthProviderOIDC); err != nil {
+			return err
+		} else {
+			oidcProvider.SSOProviderID = int(ssoProvider.ID)
```
mistahj67, line 50: Nit: With everything being serial, we shouldn't need the type assertion here.

mistahj67 (commented):
Did we want to also update the SAML create db op with a tx for creating SSO in this PR or wait for another one? 

We still need to update the users table as well I think. Might be able to swap to sso_id vs 2 separate columns 🤔 That's likely worth a full separate PR

cmd/api/src/database/sso_providers.go
mvlipka: I don't think there's a way to actually make an `int` enum in Postgres, unfortunately.
Open to ideas, though!

mvlipka:
> Did we want to also update the SAML create db op with a tx for creating SSO in this PR or wait for another one?
> 
> We still need to update the users table as well I think. Might be able to swap to sso_id vs 2 separate columns 🤔 That's likely worth a full separate PR

I was thinking another ticket, but there's no harm in throwing it in here as well while it's fresh in our minds

cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -14,29 +14,84 @@
...
+    ADD CONSTRAINT fk_users_sso_provider FOREIGN KEY (sso_provider_id) REFERENCES sso_providers (id) ON DELETE CASCADE;
+
+-- Backfill users with their proper sso_provider when they have a saml_provider_id
+UPDATE users u
+SET sso_provider_id = (SELECT sso.id
+                       FROM saml_providers saml
+                                JOIN sso_providers sso ON sso.id = saml.sso_provider_id
+                       WHERE u.saml_provider_id = saml.id)
+WHERE sso_provider_id IS NULL
+  AND saml_provider_id IS NOT NULL;
```
mistahj67, line 97: While it's likely blows up scope, we might want to put deprecated on the struct for the saml_provider_id. Idk how much extra work it would be to swap over to that in the logic or not but something to audit before merging this. If it's minimal, we might want to consider it as well as dropping the column from this table.... 🤷

cmd/api/src/model/oidc_provider.go
```diff
@@ -0,0 +1,34 @@
...
+	ClientID      string `json:"client_id"`
+	Issuer        string `json:"issuer"`
+	SSOProviderID int    `json:"sso_provider_id"`
+
+	Serial
+}
+
+func (s OIDCProvider) AuditData() AuditData {
+	return AuditData{
+		"id":              s.ID,
```
mistahj67, line 30: did we want to include the client id as well?

mistahj67 (commented):
Coming along solid! Sorry about the maybe scope creep question. Feel free to make a judgement call on that one.

cmd/api/src/database/migration/migrations/v6.1.0.sql
```diff
@@ -14,29 +14,84 @@
...
+SET sso_provider_id = (SELECT id FROM sso_providers WHERE name = saml_providers.name)
+WHERE saml_providers.sso_provider_id IS NULL;
+
+-- Add the sso_provider to the users table
+ALTER TABLE ONLY users
+    ADD COLUMN IF NOT EXISTS sso_provider_id INTEGER NULL;
+ALTER TABLE ONLY users
+    DROP CONSTRAINT IF EXISTS fk_users_sso_provider;
+ALTER TABLE ONLY users
+    ADD CONSTRAINT fk_users_sso_provider FOREIGN KEY (sso_provider_id) REFERENCES sso_providers (id) ON DELETE CASCADE;
```
ddlees, line 88: Does the existing logic delete users if their associated IDP is deleted or do we disable them and change them to basic auth?

cmd/api/src/database/migration/migrations/v6.1.0.sql
ddlees: There are already a ton of moving parts. I vote deal with it in a separate ticket.

cmd/api/src/database/migration/migrations/v6.1.0.sql
mvlipka: It looks like we currently disassociate them by setting their reference to null:
https://github.com/SpecterOps/BloodHound/blob/bf45ca32c028270dcba3428f7c0dd6628848f4fb/cmd/api/src/api/v2/auth/auth.go#L164

ddlees (approved):

PR #913, opened 2024-10-23
BED-4931: Add Audit Logging for SAML Authentication Attempts

iustinum:
## Description

This PR resolves an issue where successful SAML provider logins were not generating `LoginAttempt` entries in the `/api/v2/audit` endpoint, despite being logged in the Elastic logs.

## Motivation and Context

This change addresses BED-4931.

For secret-based authentication, audit logging is handled within the `authenticator.LoginWithSecret()` method via `auditLogin()`, which creates `LoginAttempt` entries visible in the `/api/v2/audit` response body. However, in the SAML authentication flow, login attempts were not being audited, as the handler for the `/api/v1/login/saml/{saml_provider_name}` endpoint did not create equivalent audit log entries.

This PR resolves the issue by introducing audit logging in the SAML authentication handler and ensuring that all login attempts, including those using SAML, are properly logged in the audit trail.

The solution aligns with the existing audit logging pattern and provides a consistent and complete audit trail for SAML authentication attempts.

## How Has This Been Tested?

The changes were tested by extending the existing test cases for the SAML login handler. A mock for `CreateAuditLog` was added to verify that audit log entries are created as expected. The mocking expectations were also updated to reflect the new code flow, including logging for both successful authentication attempts and failures.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/saml/saml.go
+++ b/cmd/api/src/api/saml/saml.go
@@ -26,17 +26,16 @@ import (
 	"sync"
 	"time"
 
-	"github.com/gofrs/uuid"
-	"github.com/specterops/bloodhound/src/database/types"
-
 	"github.com/crewjam/saml"
 	"github.com/crewjam/saml/samlsp"
+	"github.com/gofrs/uuid"
 	"github.com/gorilla/mux"
 	"github.com/specterops/bloodhound/errors"
 	"github.com/specterops/bloodhound/headers"
 	"github.com/specterops/bloodhound/log"
 	"github.com/specterops/bloodhound/mediatypes"
 	"github.com/specterops/bloodhound/src/api"
+	"github.com/specterops/bloodhound/src/auth"
 	"github.com/specterops/bloodhound/src/auth/bhsaml"
 	"github.com/specterops/bloodhound/src/config"
 	"github.com/specterops/bloodhound/src/ctx"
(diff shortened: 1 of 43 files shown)


cmd/api/src/api/saml/saml.go
```diff
@@ -266,29 +269,71 @@ func domainValue(host url.URL) string {
 	}
 }
 
+func (s ProviderResource) auditSAMLLogin(requestContext context.Context, user model.User, status model.AuditLogEntryStatus, err error) error {
```
mvlipka, line 272: It looks like we already have [auditLogin](https://github.com/SpecterOps/BloodHound/blob/96380ddb214a4034d1dbf26bbeb15223fd390fbf/cmd/api/src/api/auth.go#L92) that we could perhaps export and wrap since the functionality seems the same.

cmd/api/src/api/saml/saml.go
```diff
@@ -266,29 +269,71 @@ func domainValue(host url.URL) string {
...
+	if commitID, idErr := uuid.NewV4(); idErr != nil {
```
mvlipka, line 273: I think this makes sense in the current context since we can tie a single request to the request_id, but since the calls to `auditSAMLLogin` all happen within the same endpoint, I'm thinking we may want to pass in the `commitID` rather than generate it each call

iustinum: Resolved via reusing [auditLogin](url)

iustinum:
See #920 for latest change

iustinum:
#920

PR #918, opened 2024-10-28
fix: BED-4954 - Include SyncedToADUser and SyncedToEntraUser in Delet…

zinic:
## Description

Azure Post is missing SyncedToADUser: https://github.com/SpecterOps/BloodHound/blob/c347ed2bfa2d524494657269ab23828674f74376/packages/go/analysis/azure/post.go#L120

ActiveDirectory Post is missing SyncedToEntraUser: https://github.com/SpecterOps/BloodHound/blob/c347ed2bfa2d524494657269ab23828674f74376/packages/go/analysis/ad/post.go#L37

Lastly, as written the resulting query from DeleteTransitEdges(...) narrows the start and end node kinds to just one base type: (:Base)-[:SyncedToEntraUser]->(:Base). This presents a problem since AD/Entra Hybrid Paths, by definition, must cross boundaries: (:Base)-[:SyncedToEntraUser]->(:AZBase). As such, the DeleteTransitEdges(...) function must be rewritten to match on multiple start and end node kinds.

## Motivation and Context

This PR addresses: BED-4954

Calls to [DeleteTransitEdges(…)](https://github.com/SpecterOps/BloodHound/blob/c347ed2bfa2d524494657269ab23828674f74376/packages/go/analysis/post.go#L123) do not include the most recent AD/Entra Hybrid Path edges. Additionally, due to the way that the function is written, cross-domain edges will never be deleted.

## How Has This Been Tested?

Integration testing.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/analysis/post.go
+++ b/packages/go/analysis/post.go
@@ -120,7 +120,7 @@ type DeleteRelationshipJob struct {
 	ID   graph.ID
 }
 
-func DeleteTransitEdges(ctx context.Context, db graph.Database, fromKind, toKind graph.Kind, targetRelationships ...graph.Kind) (*AtomicPostProcessingStats, error) {
+func DeleteTransitEdges(ctx context.Context, db graph.Database, baseKinds graph.Kinds, targetRelationships ...graph.Kind) (*AtomicPostProcessingStats, error) {
 	defer log.Measure(log.LevelInfo, "Finished deleting transit edges")()
 
 	var (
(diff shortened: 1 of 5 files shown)


packages/go/analysis/post.go
```diff
@@ -134,9 +134,9 @@ func DeleteTransitEdges(ctx context.Context, db graph.Database, fromKind, toKind
 		if err := db.ReadTransaction(ctx, func(tx graph.Transaction) error {
 			fetchedRelationshipIDs, err := ops.FetchRelationshipIDs(tx.Relationships().Filterf(func() graph.Criteria {
 				return query.And(
-					query.Kind(query.Start(), fromKind),
+					query.KindIn(query.Start(), baseKinds...),
```
urangel, line 137: This is a bit of a tangent but is there reason we have both query.Kind and query.KindIn? Does the other serve as an alias? It looks like they do the same thing and I was curious what drove the update here

zinic: The AST that each function builds is different so I wanted distinction here, however, I agree with you this ends up feeling pretty obtuse. It might be worth just making `query.Kind()` variadic.

Unfortunately both `Kind` and `KindIn` are heavily used and would not be a simple refactor. I'll create a ticket to track it for continuity work.

urangel: Yea, similarly I've been wanting to update the definition of "admin_tier_0" to put into the common schema but it's used in a lot of places to update in a non related changeset without being noisy. FWIW I believe query.Kind and query.KindIn used to function differently in the past.

PR #896, opened 2024-10-04
BED-4766: API Endpoint to List OIDC Providers

iustinum:
## Description

This pull request introduces a unified SSO endpoint that provides detailed information on both OIDC and SAML providers. The goal is to offer a consolidated and accurate list of available authentication methods, ensuring that users are directed to the appropriate provider for successful authentication.

## Motivation and Context

This PR addresses the requirements outlined in ticket BED-4766.

Previously, SSO providers were listed using the `GET /api/v2/saml/sso endpoint`, which only supported SAML providers. With the addition of OIDC as an alternative authentication method to comply with FedRamp requirements, there is a need for a single, unified endpoint. This new endpoint will display information about both SAML and OIDC providers as a more flexible and agnostic approach. The unified endpoint will replace the existing SAML-only endpoint to ensure all authentication options are covered.

## How Has This Been Tested?

Unit Tests: Via `auth/sso_test` to verify endpoint behavior.
Integration Tests: Via `database/sso_providers_test` to verify interaction with the database.

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Associated an issue: https://github.com/SpecterOps/BloodHound/issues/672
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [X] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments (GoDocs / JSDocs)
- [X] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

PR #921, opened 2024-10-28
BED-4766: FF Backward Compatibility for ListAuthProviders Endpoint

iustinum:
## Description

This enables #896 to only return SAML providers when `FeatureOIDCSupport` is `false`

## Motivation and Context

This PR addresses: BED-4766

## How Has This Been Tested?

- Verified via DB test in `sso_providers_test.go`

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/database/sso_providers.go
+++ b/cmd/api/src/database/sso_providers.go
@@ -100,7 +100,7 @@ func (s *BloodhoundDB) GetAllSSOProviders(ctx context.Context, order string, sql
 		query = query.Where(sqlFilter.SQLString, sqlFilter.Params...)
 	}
 
-	// Backwards compatability when FF is disabled
+	// Backwards compatibility when FF is disabled
 	if oidcFeatureFlag, err := s.GetFlagByKey(ctx, appcfg.FeatureOIDCSupport); err != nil {
 		return providers, err
 	} else if !oidcFeatureFlag.Enabled {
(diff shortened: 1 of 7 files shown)


ALCooper12:
Idk if you're still working on your PR, but don't forget to list how this was tested!

cmd/api/src/database/sso_providers.go
```diff
@@ -99,6 +100,13 @@ func (s *BloodhoundDB) GetAllSSOProviders(ctx context.Context, order string, sql
 		query = query.Where(sqlFilter.SQLString, sqlFilter.Params...)
 	}
 
+	// Backwards compatibility when FF is disabled
+	if oidcFeatureFlag, err := s.GetFlagByKey(ctx, appcfg.FeatureOIDCSupport); err != nil {
```
mvlipka, line 104: Probably fine since this DB call is only used for our feature, but I'm thinking we'd probably want this in the endpoint rather than in the database call.

In my eyes the database call is doing what you're asking it, just grabbing SSOProviders, but it's the actual endpoint that implements the feature that is meant to be feature gated.

mistahj67: We started there but shoehorning in the FF check where it properly bubbled down to this layer got complicated with how the handlers parse filtering whereas this layer offered a quick and dirty workaround in the interim

mistahj67 (approved):
I agree with @mvlipka but seeing as this should only be for the short term, I think this is fine

PR #937, opened 2024-11-07
BED-4976: Fix EnrollOnBehalfOf logic issue

JonasBK:
## Description

Fixed a bug in the EnrollOnBehalfOf logic.

We now correctly check if the CA of the first CertTemplate is trusted for NTAuth.

Fixed harness tests to account for this as well.

## Motivation and Context

This PR addresses: BED-4976

## How Has This Been Tested?

Fixed harness tests for EnrollOnBehalfOf.

Checked that the edge is still generated and composition works for ESC3 with this dataset: 
[20231208161111_BloodHound.zip](https://github.com/user-attachments/files/17660073/20231208161111_BloodHound.zip)

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/go/analysis/ad/esc3.go
```diff
@@ -142,60 +142,71 @@ func PostADCSESC3(ctx context.Context, tx graph.Transaction, outC chan<- analysi
...
-	operation.Operation.SubmitReader(func(ctx context.Context, tx graph.Transaction, outC chan<- analysis.CreatePostRelationshipJob) error {
-		if results, err := EnrollOnBehalfOfVersionTwo(tx, versionTwoTemplates, certTemplates); err != nil {
-			return err
-		} else {
-			for _, result := range results {
-				if !channels.Submit(ctx, outC, result) {
-					return nil
-				}
-			}
+	for _, domain := range domains {
```
urangel, line 162: Might there be a way to combine this EnrollOnBehalfOf processing into the ESC3? I ask in case there is a way to avoid doing this double loop through domains and eca's twice (here and in the PostADCS func).

JonasBK: I like that idea! I think we could throw the EnrollOnBehalfOf logic under ESC3 and put the nodes that should get EnrollOnBehalfOf edge into variables, so we can perform lookups in those variables instead of checking for EnrollOnBehalfOf edges. At the end, we could create both the ESC3 and EnrollOnBehalfOf edges. That said, I don't have time to implement and test that before sometimes next year..

urangel: Makes sense. This would take some time to clean up!

urangel (approved):

PR #950, opened 2024-11-13
PostgreSQL Related Query Path Fixes

zinic:
## Description

This changeset addresses a few tickets parted out into independent commits. This also contains some additional `prepare-for-codereview` changes layered on top.

## Motivation and Context

This PR addresses the following tickets:
* BED-4997 - Post-Processing Constraint Violations
* BED-5003 - CyPgSQL Directionality Reversed for Expansion Query

## How Has This Been Tested?

Several new test cases have been added to cover changes.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/analysis/post.go
+++ b/packages/go/analysis/post.go
@@ -18,7 +18,9 @@ package analysis
 
 import (
 	"context"
+	"github.com/specterops/bloodhound/dawgs/cardinality"
 	"sort"
+	"sync"
 
 	"github.com/specterops/bloodhound/dawgs/graph"
 	"github.com/specterops/bloodhound/dawgs/ops"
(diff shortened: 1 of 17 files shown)


packages/go/analysis/post.go
```diff
@@ -120,6 +123,57 @@ type DeleteRelationshipJob struct {
...
+}
+
+// TrackCreatePostRelationshipJob decomposes a CreatePostRelationshipJob type and returns the result of tracking it.
+func (s EdgeConstraintMap) TrackCreatePostRelationshipJob(job CreatePostRelationshipJob) bool {
+	return s.Track(job.FromID, job.ToID, job.Kind)
+}
+
+// Track will attempt to track creation of the given relationship arguments. This function returns false
+// if the given relationship has already been tracked; true otherwise.
+func (s EdgeConstraintMap) Track(start, end graph.ID, edgeKind graph.Kind) bool {
```
mistahj67, line 148: Nit, this isn't used outside this file, can we swap this to a private function?

packages/go/cypher/models/pgsql/translate/expansion.go
```diff
@@ -712,6 +696,111 @@ func (s *Translator) buildExpansionPatternRoot(part *PatternPart, traversalStep
...
+	RightNodeConstraints *Constraint
+	EdgeConstraints      *Constraint
+	LeftNodeConstraints  *Constraint
+	PrimerVisible        *pgsql.IdentifierSet
+	RecursiveVisible     *pgsql.IdentifierSet
+	PrimerWhereClause    pgsql.Expression
+	RecursiveWhereClause pgsql.Expression
+}
+
+func PrepareExpansionRootComponents(part *PatternPart, traversalStep *PatternSegment, treeTranslator *ExpressionTreeTranslator) (ExpansionComponents, error) {
```
mistahj67, line 711: Similar nit feedback here re: private vs public

packages/go/cypher/models/pgsql/translate/expansion.go
```diff
@@ -712,6 +696,111 @@ func (s *Translator) buildExpansionPatternRoot(part *PatternPart, traversalStep
...
+	TerminalNode         *BoundIdentifier
+	RootNode             *BoundIdentifier
+	RightNodeConstraints *Constraint
+	EdgeConstraints      *Constraint
+	LeftNodeConstraints  *Constraint
+	PrimerVisible        *pgsql.IdentifierSet
+	RecursiveVisible     *pgsql.IdentifierSet
+	PrimerWhereClause    pgsql.Expression
+	RecursiveWhereClause pgsql.Expression
+}
```
superlinkx, line 709: It was mentioned during live review that this opens us up to a lot of nil pointers depending on use case. What mitigation strategies can we pursue until a more elegant structure presents itself? It seems like at a minimum, this type could be private as it's not used outside this package and given the volatile nature of it, probably shouldn't be used outside of this file

superlinkx (approved):
Slightly concerned by the nil pointers, but I don't think there's anything present as-is that would block this PR. Would like us to revisit that aspect in a follow up, so I'm approving as-is.

mistahj67 (commented):
Really appreciate the cleanup 🫶 🧼

Why wasn't the edge constraint safety net not applied to the 
`ExecuteCommand` or `PostHybrid`? 

That also made me wonder why we didn't opt to move it one layer up in post processing as well, though I suspect that might be to minimize memory

packages/go/cypher/models/pgsql/translate/expansion.go
zinic: Patching this up.

packages/go/cypher/models/pgsql/translate/expansion.go
zinic: Patching this up.

packages/go/analysis/post.go
zinic: I like this being a public contract for right now. It's the basic interaction of the datastructure and makes it more usable for little cost.

mistahj67 (approved):
Thanks for addressing nits! 🚀

superlinkx (approved):
Appreciate the due-diligence. This reads much safer and has my stamp of approval

PR #965, opened 2024-11-18
fix: set resolution for cross-spawn to a pinned fix for vuln finding

urangel:
## Description
This PR adds the cross-spawn package to the resolutions list of the package.json of the root of the project.

## Motivation and Context

This PR addresses: BED-5049

Our vulnerability scan in CI is blocking other PRs as it fails the vuln check: https://github.com/SpecterOps/BloodHound/actions/runs/11899005494/job/33157550242?pr=960

The resolution that is set was chosen from the fixed version listed in the failing CI step.

## How Has This Been Tested?

After setting the resolution, running and building the application locally works as expected. Specific attention was given to the API Explorer page as react-swagger-ui was one of the projects that included cross-spawn as a peer dependency. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


package.json
```diff
@@ -26,6 +26,7 @@
     "typescript": "^5.1.6"
   },
   "resolutions": {
-    "dompurify": "3.1.3"
+    "dompurify": "3.1.3",
+    "cross-spawn": "7.0.5"
```
benwaples, line 30: should we be using the ^ here so we can still use future versions that are compatible with 7.0.5? `^7.0.5`

benwaples (approved):
LGTM! Left one question, but if thats a larger discussion then we can merge as is because this will unblock CI

package.json
urangel: Yea that makes sense to me! I've updated to include this.

PR #978, opened 2024-11-23
BED-5067 feat: add edit sso provider support to ui

mistahj67:
## Description

- Renamed create provider dialogs to upsert
- Added edit support to provider dropdown

NOTE: https://github.com/SpecterOps/BloodHound/pull/975 should be merged first

## Motivation and Context

This PR addresses: BED-5067

Currently you cannot edit a SSO provider without deleting it first. This adds support for the UI to hit the new PATCH endpoints created by https://github.com/SpecterOps/BloodHound/pull/975

## How Has This Been Tested?
Locally and through unit tests

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/openapi/doc/openapi.json
+++ b/packages/go/openapi/doc/openapi.json
@@ -605,7 +605,7 @@
                   "type": "object",
                   "properties": {
                     "data": {
-                      "$ref": "#/components/schemas/model.auth-provider"
+                      "$ref": "#/components/schemas/model.oidc-provider"
                     }
                   }
                 }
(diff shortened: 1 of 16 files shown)


packages/javascript/bh-shared-ui/src/components/UpsertOIDCProviderForm.tsx
```diff
@@ -0,0 +1,143 @@
...
+    return (
+        <form autoComplete='off' onSubmit={handleSubmit(onSubmit)}>
+            <DialogContent>
+                <Grid container spacing={2}>
+                    <Grid item xs={12}>
+                        <Controller
+                            control={control}
+                            name='name'
+                            rules={{
+                                required: !oldSSOProvider && 'OIDC Provider Name is required',
```
benwaples, line 56: I see that all the required fields have this `!oldSSOProvider &&` check, does that mean that these fields are not required when updating a given provider record?

packages/javascript/bh-shared-ui/src/views/SSOConfiguration/SSOConfiguration.tsx
```diff
@@ -152,27 +170,48 @@ const SSOConfiguration: FC = () => {
...
-    const createOIDCProvider = async (oidcProvider: CreateOIDCProviderRequest) => {
-        setCreateProviderError('');
+    const upsertOIDCProvider = async (oidcProvider: UpsertOIDCProviderRequest) => {
+        setUpsertProviderError('');
         try {
-            await apiClient.createOIDCProvider(oidcProvider);
+            if (ssoProviderIdToDeleteOrUpdate) {
+                await apiClient.updateOIDCProvider(ssoProviderIdToDeleteOrUpdate, oidcProvider);
+            } else {
+                if (oidcProvider.name && oidcProvider.client_id && oidcProvider.issuer) {
```
benwaples, line 200: do we need this check in the case we are updating as well?

packages/javascript/bh-shared-ui/src/views/SSOConfiguration/SSOConfiguration.tsx
```diff
@@ -112,24 +115,39 @@ const SSOConfiguration: FC = () => {
...
+                switch (ssoProvider.type) {
+                    case 'SAML':
+                        openSAMLProviderDialog();
+                        break;
+                    case 'OIDC':
+                        openOIDCProviderDialog();
+                        break;
+                }
+                break;
+        }
```
benwaples, line 143: The compiler thanks you for your service

benwaples (changes requested):
Awesome work refactoring the create flow to an upsert flow! 

I left a couple questions that I think are related to an unexpected behavior: I can remove a providers name (or id or issuer for OIDC providers) which is required in the create flow, but Im able to "save" in the edit flow. The invalid changes don't actually take (which I assume is the API being well built and not accepting invalid inputs and only PATCHing the valid ones), but from a users perspective its a bit unexpected. In this case I think we should block the user from saving and show some errors if the changes are not going to be saved.

packages/go/openapi/doc/openapi.json
```diff
@@ -605,7 +605,7 @@
                   "type": "object",
                   "properties": {
                     "data": {
-                      "$ref": "#/components/schemas/model.auth-provider"
+                      "$ref": "#/components/schemas/model.oidc-provider"
```
sircodemane, line 608: the json is generated from the yaml and i'm not seeing any yaml file changes, did you update the json file directly?

packages/javascript/bh-shared-ui/src/components/UpsertOIDCProviderForm.tsx
mistahj67: They are not technically required no, the API will dynamically update only what's changed, that's correct. I suppose I could leave it required 🤷 wouldn't hurt

packages/javascript/bh-shared-ui/src/views/SSOConfiguration/SSOConfiguration.tsx
mistahj67: Nah the API doesn't need fields since it's a PATCH endpoint it accepts partial payloads

benwaples (approved):

PR #984, opened 2024-11-26
BED-5060: Prevent user from changing their own role/auth

wes-mil:
## Description

Prevent a user from updating their own role or auth method within user configuration.

Also addresses two bugs discovered while addressing ticket:
- An admin user could accidentally delete a user's password while adjusting their authentication method
- GetAuthSecret didn't return ERRNOTFOUND while attempting to get an auth secret by ID

## Motivation and Context

This PR addresses: BED_5060

An admin user could change their own role to a role that is not privileged or change their own authentication method to something else. While this could be intentional, it largely caused issues with misconfigured SSO providers or accidental role changing. Effectively preventing that user from authenticating or having privilege. In the worst case, this could occur have occured while there was only a single admin user in the account.

## How Has This Been Tested?

An admin user is created an then the user attempts to change their role through user configuration. On the UI these fields are now grayed out. Through the API, a PATCH request is made using an API token from an admin user attempting configure those fields. The API returns a 400 response now.

Test cases were added for the changes to database method UpdateUser.

## Screenshots (optional):

image

Screenshot from 2024-11-25 15-20-02

Screenshot from 2024-11-25 15-24-01

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/v2/auth/auth_test.go
+++ b/cmd/api/src/api/v2/auth/auth_test.go
@@ -165,7 +165,7 @@ func TestManagementResource_PutUserAuthSecret(t *testing.T) {
 
 func TestManagementResource_EnableUserSAML(t *testing.T) {
 	var (
-		adminUser         = model.User{AuthSecret: defaultDigestAuthSecret(t, "currentPassword"), Unique: model.Unique{ID: must.NewUUIDv4()}}
+		adminUser         = model.User{Unique: model.Unique{ID: must.NewUUIDv4()}}
 		goodRoles         = []int32{0}
 		goodUserID        = must.NewUUIDv4()
 		badUserID         = must.NewUUIDv4()


cmd/api/src/api/error.go
```diff
@@ -60,6 +60,8 @@ const (
 	ErrorResponsePayloadUnmarshalError              = "error unmarshalling JSON payload"
 	ErrorResponseRequestTimeout                     = "request timed out"
 	ErrorResponseUserSelfDisable                    = "user attempted to disable themselves"
+	ErrorResponseUserSelfRoleChange                 = "user attempted to change own role"
+	ErrorResponseUserSelfSSOChange                  = "user attempted to change own SSO"
```
mistahj67, line 64:
```suggestion
	ErrorResponseUserSelfSSOChange                  = "user attempted to change own SSO Provider"
```

cmd/api/src/database/auth.go
```diff
@@ -282,6 +282,15 @@ func (s *BloodhoundDB) UpdateUser(ctx context.Context, user model.User) error {
...
 
+		// AuthSecret must be manually deleted and then set to nil again to prevent recreation
+		if user.AuthSecret == nil {
+			if err := tx.Unscoped().Model(&user).WithContext(ctx).Association("AuthSecret").Unscoped().Clear(); err != nil {
+				return err
+			}
+
+			user.AuthSecret = nil
+		}
+
```
mistahj67, line 293: This feels a bit like we are swimming up stream. I'm hopeful we can find a smoother way if we pair!

mistahj67 (changes requested):
I think the disabled route is more apparent to the user but the ticket requests it not appear at all. Likely can verify with Ryan which we prefer!

cmd/api/src/database/auth.go
mistahj67: After pairing, we opted for the more verbose but single pathway of calling the DeleteAuthSecret method to ensure audit logs are maintained

mistahj67 (approved):
Tested locally, code LGTM. I'd still snag another ✅ since we ended up pairing on this, particularly the db method choice.

ALCooper12:
LGTM! Appreciate the explanations to my questions earlier!

ALCooper12 (approved):
LGTM! Appreciate the explanations to my questions earlier!

PR #1019, opened 2024-12-16
PQE-6 - Migrate E2E Disable Users Tests To BDD

ykaiboussiSO:
## Description

This PR introduces E2E Behavior Driven Development Framework using [Cucumber-JS](https://github.com/cucumber/cucumber-js/tree/main) and [Playwright API](https://playwright.dev). 

## Motivation and Context

We are currently exploring tools and best practices to automate Acceptance Criteria.

## How Has This Been Tested?

Tested E2E implementation against local BHCE instance.

```
5 scenarios (5 passed)
30 steps (30 passed)
0m07.044s (executing steps: 0m06.884s)
➜  BloodHound git:(PQE-6)
```

## Checklist:

  - Assigned myself to this PR
  - Open API docs


ykaiboussiSO:

packages/javascript/js-client-library/package-lock.json
benwaples, line 1: We should be able to remove this file. `package-lock.json` is created via `npm install` and we only use yarn in this project

cmd/e2e/package-lock.json
benwaples, line 1: We can also remove this package-lock.json as well.

cmd/e2e/src/tests/pageObjects/loginPage.ts
```diff
@@ -0,0 +1,41 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { type Page } from '@playwright/test';
+
+export default class LoginPage {
```
benwaples, line 19: I like this pattern! Is the idea that when we want to interact with a given page, we add a to `pageObject/` and can reuse those page interactions across different tests? 

A small organizational suggestion: we move `pageObjects/` under `helpers/` because these arent exactly tests but instead utils for tests.

cmd/e2e/README.md
```diff
@@ -0,0 +1,25 @@
...
+Setup Prisma Schema and Client 
+- `cp .env.example .env` & update DB URL
+- `just prisma-generate`
+
+Update Cross-Enviroment variables
+- `cp .env-cmdrc.example .env-cmdrc` 
+
+Scripts configured in `package.json` to support running different type of tests.
+
+- `just run-e2e-tests` runs all cucumber BDD tests using [just](https://github.com/casey/just)
```
benwaples, line 25: Do we need the bhce container running before running tests? I imagine we do so we have the DB running, but maybe Im missing something.

If so, can we add a step for running `just bh-dev` before running `just run run-e2e-tests`?

cmd/e2e/prisma/cleanup.ts
```diff
@@ -0,0 +1,36 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import prisma from './client.js';
+
+export class dbOPS {
+    // delete test data in dev environment only
+    async deleteUsers() {
+        const baseURL = `${process.env.BASEURL}`.toLowerCase();
+        if (
+            (`${process.env.ENV}` != 'production' || `${process.env.ENV}` != 'staging') &&
```
benwaples, line 24: Any particular reason were using `!=` here instead of `!==`? Its standard now to use `!==` to avoid any unexpected type coercion and the != is a bit of a code smell that theres something unexpected goin on here.

If we're not expecting the type coercion, then may we swap these out for `!==`, please?

cmd/e2e/prisma/cleanup.ts
```diff
@@ -0,0 +1,36 @@
...
+            (baseURL.toLowerCase().includes('localhost') || baseURL.toLowerCase().includes('127.0.0.1'))
```
benwaples, line 25: Looks like were running `.toLowerCase()` on `baseURL` twice, could we remove the added `toLowerCase()`'s on this line?

cmd/e2e/prisma/cleanup.ts
benwaples: Would `process.env.ENV === 'dev'`suffice in this check? Looks like this is a black list approach (meaning dont run this code in x envs), where as the suggested approach is a white list (only run in x envs). Difference being that if we are too add another environment, we wouldnt want to forget to add that env to this black list and end up deleting the users unintentionally

cmd/e2e/src/support/hooks.ts
```diff
@@ -0,0 +1,57 @@
...
+Before(async function (this: PlaywrightWorld) {
+    // create new instance of fixture for each scenario
+    await fx.openContext();
+    await fx.newPage();
+    this.fixture = fx.Fixture;
+});
+
+After(async function ({ result, pickle }) {
+    // capture screenshot for failed step
+    if (result?.status == Status.FAILED) {
```
benwaples, line 42: same comment here regarding equality operator and using `===` over `==`

benwaples (commented):
This is awesome! I was able to easily set up the project and run tests. I left a couple comments, mostly minor changes, but I think updating the readme will help for future people setting up the project as well. 

Ill do another review after any changes accepted/made!

cmd/e2e/prisma/cleanup.ts
ykaiboussiSO: Good call out. I will look into it.

cmd/e2e/prisma/seed.ts
```diff
@@ -0,0 +1,136 @@
...
+            (this.isExpired = isExpired),
+            (this.totpSecret = totpSecret),
+            (this.totpActivated = totpActivated);
+    }
+    async create() {
+        // sanity check when running against production environment
+        // production tagged tests should not include any seeding data.
+        if (process.env.ENV === 'production') {
+            console.error('exiting tests: no seeding data in production environment');
+            exit(1);
```
benwaples, line 77: Very cool and I havent seen process.exit used! Im curious if why we need `process.exit(1)` in this case instead of returning early? 

Looks like `exit()` will end the process entirely and potentially leave some async code hanging and unable to exit gracefully. I think an early return would suffice in this case because this check wants to prevent the below code from running and not necessarily exit the entire process. Unless Im missing something and we do want to exit the process entirely?

ykaiboussiSO: @benwaples, you are right. My initial thought is to `exit` the process to prevent seeding test data. But I did not factor `aysnc` calls while existing the process. I will look into it.

cmd/e2e/prisma/cleanup.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import prisma from './client.js';
+
+export class dbOPS {
+    // delete test data in dev environment only
+    async deleteUsers() {
+        const baseURL = `${process.env.BASEURL}`.toLowerCase();
+        if (`${process.env.ENV}` === 'dev' && (baseURL.includes('localhost') || baseURL.includes('127.0.0.1'))) {
+            const deleteUsers = prisma.users.deleteMany();
+            const deleteUserRoles = prisma.users_roles.deleteMany();
```
benwaples, line 25: Will this only delete the users inserted? Or do we need to pass an argument to target only the users/roles that were created from prisma? [Looking at their docks](https://www.prisma.io/docs/orm/reference/prisma-client-reference#deletemany) and it appears that it would delete all user records

cmd/e2e/prisma/cleanup.ts
```diff
@@ -0,0 +1,33 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import prisma from './client.js';
+
+export class dbOPS {
+    // delete test data in dev environment only
+    async deleteUsers() {
+        const baseURL = `${process.env.BASEURL}`.toLowerCase();
+        if (`${process.env.ENV}` === 'dev' && (baseURL.includes('localhost') || baseURL.includes('127.0.0.1'))) {
```
benwaples, line 23: Thanks for cleanin up this logic, I think the white list will work well! 

One thing regarding `process.env.<key>`, it returns `string | undefined`, so there is the possibility that we end up with a string of `'undefined'` instead of just `undefined` the value. Rather than wrapping `process.env` in the template literal, would it work to use the value from `process.env.<key>` directly? I think that means we would have to optionally chain `.toLowerCase()` on L22 in case `process.env.BASEURL` is undefined (which is unlikely because you have it by default in the config files).

cmd/e2e/src/helpers/init.ts
```diff
@@ -0,0 +1,44 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import * as fs from 'node:fs';
+
+function createsupportfolders(path: string): void {
```
benwaples, line 19: Nit:  looks like the camel casing is off on this one: `createSupportFolders`. Is that on purpose? I see that it's pulled directly from bhe so im guessing it was like that previously?

ykaiboussiSO: I will address it next commit.

cmd/e2e/prisma/cleanup.ts
ykaiboussiSO: I'm sorry, if I missed your feedback. we declared env types in `helpers/types` where `BASEURL` should return `string` only.

ykaiboussiSO:
Closing the PR. The project was an RnD around BDD tooling in Typescript. We gathered enough data point to revisit later if we see Playwright a good fit as end to end tool.

I would like to give a special shoutout to @urangel, @benwaples and @elikmiller for sharing their knowledge.

PR #1025, opened 2024-12-17
BED-5070: Add Auto-provisioning Support to SSO Auth Flow on Backend

ALCooper12:
## Description

- Created a new `SSOProviderConfig` struct to be used as a `jsonb` type/value inside the `sso_providers` table under a new column called `config`
- Added database migration in order to create the new `config` column and insert default values
- Updated and added new/existing database and model variables/functions within `sso_providers.go`,`samlprovider(s).go`, and `oidc_providers.go` files
- Added user creation auto-provisioning and attribute mapping for both SAML and OIDC
- Set up the enforcement of enable/disable auto-provisioning for the `SSOProviderConfig`

## Motivation and Context

- Ticket: BED-5070

- Currently, BloodHound users must be created manually for every new user. In larger customer environments, this can create a significant amount of work to create 30+ user accounts and this friction could reduce adoption. This PR solves the issue of auto-provisioning users who don't yet exist via SAML and OIDC into the BHCE/BHE platform. It also sets up default user role allocation and SSO Provider role modification

## How Has This Been Tested?

- Still working on this*****

## Screenshots (optional):
- Still working on this*****

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/v2/auth/oidc.go
+++ b/cmd/api/src/api/v2/auth/oidc.go
@@ -190,41 +190,41 @@ func (s ManagementResource) OIDCCallbackHandler(response http.ResponseWriter, re
 			} else if claims.DisplayName == "" {
 				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "Display Name claim is missing", request), response)
 				return
-			} else if user, err := s.db.LookupUser(request.Context(), claims.DisplayName); err != nil {
-				if errors.Is(err, database.ErrNotFound) {
-					user.EmailAddress = null.StringFrom(claims.Email)
-					user.PrincipalName = claims.Email
-					user.Roles = model.Roles{
-						{
-							Name:        "Read-Only",
-							Description: "Used for integrations",
-							Serial: model.Serial{
-								ID: 3,
+			} else if ssoProvider.Config.AutoProvision.Enabled {
+				if user, err := s.db.LookupUser(request.Context(), claims.DisplayName); err != nil {
+					if errors.Is(err, database.ErrNotFound) {
+						user.EmailAddress = null.StringFrom(claims.Email)
+						user.PrincipalName = claims.Email
+						user.Roles = model.Roles{
+							{
+								Serial: model.Serial{
+									ID: ssoProvider.Config.AutoProvision.DefaultRole,
+								},
 							},
-						},
+						}
+						user.SSOProviderID = null.Int32From(ssoProvider.ID)
+
+						// Need to find a work around since BHE cannot auto accept EULA as true
+						user.EULAAccepted = true
+
+						if claims.DisplayName == "" {
+							user.FirstName = null.StringFrom(claims.Name)
+						} else {
+							user.FirstName = null.StringFrom(claims.DisplayName)
+						}
+
+						if claims.FamilyName == "" {
+							user.LastName = null.StringFrom("Last name Not Found")
+						} else {
+							user.LastName = null.StringFrom(claims.FamilyName)
+						}
+
+						if _, err := s.db.CreateUser(request.Context(), user); err != nil {
+							api.HandleDatabaseError(request, response, err)
+						}
+
+						s.authenticator.CreateSSOSession(request, response, claims.Email, ssoProvider)
 					}
-					user.SSOProviderID = null.Int32From(ssoProvider.ID)
-
-					// Need to find a work around since BHE cannot auto accept EULA as true
-					user.EULAAccepted = true
-
-					if claims.DisplayName == "" {
-						user.FirstName = null.StringFrom(claims.Name)
-					} else {
-						user.FirstName = null.StringFrom(claims.DisplayName)
-					}
-
-					if claims.FamilyName == "" {
-						user.LastName = null.StringFrom("Last name Not Found")
-					} else {
-						user.LastName = null.StringFrom(claims.FamilyName)
-					}
-
-					if _, err := s.db.CreateUser(request.Context(), user); err != nil {
-						api.HandleDatabaseError(request, response, err)
-					}
-
-					s.authenticator.CreateSSOSession(request, response, claims.Email, ssoProvider)
 				}
 			} else {
 				s.authenticator.CreateSSOSession(request, response, claims.Email, ssoProvider)
--- a/cmd/api/src/api/v2/auth/oidc.go
+++ b/cmd/api/src/api/v2/auth/oidc.go
@@ -69,7 +69,7 @@ func (s ManagementResource) UpdateOIDCProviderRequest(response http.ResponseWrit
 		}
 
 		if upsertReq.Config.AutoProvision.Enabled {
-			if ssoProvider.Config.AutoProvision.DefaultRole > 5 || ssoProvider.Config.AutoProvision.DefaultRole < 0 {
+			if ssoProvider.Config.AutoProvision.DefaultRole > 5 || ssoProvider.Config.AutoProvision.DefaultRole < 1 {
 				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "role id is invalid", request), response)
 				return
 			}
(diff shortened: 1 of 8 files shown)


cmd/api/src/api/v2/auth/oidc.go
```diff
@@ -207,7 +230,44 @@ func (s ManagementResource) OIDCCallbackHandler(response http.ResponseWriter, re
 				// Technical or credentials issue
 				// Not explicitly covered; treat as a technical issue
 				redirectToLoginPage(response, request, "We’re having trouble connecting. Please check your internet and try again.")
+			} else if claims.Email == "" {
+				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "email claim is missing", request), response)
+				return
```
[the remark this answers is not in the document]
ALCooper12: Not too sure about preferred name, but I do believe that the upn could def be an email! I read that it's more likely in academic settings like @edu, but it happens. Like you mentioned, we'll tackle this area during the entra stuff

cmd/api/src/api/v2/auth/oidc.go
```diff
@@ -207,7 +231,44 @@ func (s ManagementResource) OIDCCallbackHandler(response http.ResponseWriter, re
...
 			} else {
+				if ssoProvider.Config.AutoProvision.Enabled {
+					if user, err := s.db.LookupUser(request.Context(), claims.Email); err != nil {
+						if errors.Is(err, database.ErrNotFound) {
```
superlinkx, line 240: The `errors.Is` can be hoisted up as the first check instead of `err != nil`, as it's safe against nil. This would also make it clearer that we missed the default error handling, which otherwise would go unnoticed here. The final form would be something like:

```go
if user, err := s.db.LookupUser(request.Context(), claims.Email); errors.Is(err, database.ErrNotFound) {
// Role stuff goes here
} else if err != nil {
// missing error handling goes here
}
```

There may be other chances to refactor to avoid so much nested checking, and I'd encourage further tinkering to avoid unnecessarily nested code to keep your happy path left aligned as much as possible

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -140,6 +141,20 @@ func (s ManagementResource) CreateSAMLProviderMultipart(response http.ResponseWr
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "expected only one \"metadata\" parameter", request), response)
 	} else if metadataXMLReader, err := metadataXMLFileHandles[0].Open(); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if autoProvisionEnabled, hasAutoProvisionEnabled := request.MultipartForm.Value["config.auto_provision.enabled"]; !hasAutoProvisionEnabled || len(autoProvisionEnabled) > 1 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "form is missing \"config.auto_provision.enabled\" parameter or \"config.auto_provision.enabled\" parameter has more than one value", request), response)
```
superlinkx, line 145: This could be split into two checks to allow for a more useful error message that gives exactly whether the parameter is missing or if it's too many values, rather than being ambiguous

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -140,6 +141,20 @@ func (s ManagementResource) CreateSAMLProviderMultipart(response http.ResponseWr
...
+	} else if isAutoProvisionEnabled, err := strconv.ParseBool(autoProvisionEnabled[0]); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "\"config.auto_provision.enabled\" parameter could not be converted to bool", request), response)
+	} else if defaultRoleId, hasDefaultRoleId := request.MultipartForm.Value["config.auto_provision.default_role_id"]; !hasDefaultRoleId || len(defaultRoleId) > 1 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "form is missing \"config.auto_provision.default_role_id\" parameter or \"config.auto_provision.default_role_id\" has more than one value", request), response)
```
superlinkx, line 149: Same as before

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -140,6 +141,20 @@ func (s ManagementResource) CreateSAMLProviderMultipart(response http.ResponseWr
...
+	} else if defaultRoleIdInt, err := strconv.Atoi(defaultRoleId[0]); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "\"config.auto_provision.default_role_id\" parameter could not be converted to int", request), response)
+	} else if defaultRole, err := s.db.GetRole(request.Context(), int32(defaultRoleIdInt)); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "\"config.auto_provision.default_role_id\" parameter is invalid", request), response)
+	} else if roleProvision, hasRoleProvisioned := request.MultipartForm.Value["config.auto_provision.role_provision"]; !hasRoleProvisioned || len(roleProvision) > 1 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "form is missing \"config.auto_provision.role_provision\" parameter or \"config.auto_provision.role_provision\" has more than one value", request), response)
```
superlinkx, line 155: Same as other two

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -203,7 +230,41 @@ func (s ManagementResource) UpdateSAMLProviderRequest(response http.ResponseWrit
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "expected only one \"name\" parameter", request), response)
 	} else if metadataXMLFileHandles, hasMetadataXML := request.MultipartForm.File["metadata"]; len(metadataXMLFileHandles) > 1 {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "expected only one \"metadata\" parameter", request), response)
+	} else if autoProvisionEnabled, hasAutoProvisionEnabled := request.MultipartForm.Value["config.auto_provision.enabled"]; len(autoProvisionEnabled) > 1 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "expected only one \"config.auto_provision.enabled\" parameter ", request), response)
 	} else {
+		if hasAutoProvisionEnabled {
```
superlinkx, line 236: This could be collapsed into an else-if to prevent unnecessary nesting

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -203,7 +230,41 @@ func (s ManagementResource) UpdateSAMLProviderRequest(response http.ResponseWrit
...
+			if isAutoProvisionEnabled, err := strconv.ParseBool(autoProvisionEnabled[0]); err != nil {
+				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "\"config.auto_provision.enabled\" parameter could not be converted to bool", request), response)
+				return
+			} else {
+				if isAutoProvisionEnabled {
```
superlinkx, line 241: This should also be collapsed

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -396,6 +457,40 @@ func (s ManagementResource) SAMLCallbackHandler(response http.ResponseWriter, re
 				// SAML credentials issue scenario again
 				redirectToLoginPage(response, request, "Your SSO was unable to authenticate your user, please contact your Administrator")
 			} else {
+				if ssoProvider.Config.AutoProvision.Enabled {
```
superlinkx, line 460: This should be collapsed

cmd/api/src/api/v2/auth/saml.go
```diff
@@ -396,6 +457,40 @@ func (s ManagementResource) SAMLCallbackHandler(response http.ResponseWriter, re
...
+					if user, err := s.db.LookupUser(request.Context(), principalName); err != nil {
+						if errors.Is(err, database.ErrNotFound) {
```
superlinkx, line 462: Check errors.Is first, then check err nil after, but not inside. Again, this led to an unchecked error case

cmd/api/src/model/sso_provider.go
```diff
@@ -27,9 +42,33 @@ type SSOProvider struct {
...
+func (cfg *SSOProviderConfig) Scan(value interface{}) error {
+	// Handle null values from the database
+	if value == nil {
+		*cfg = SSOProviderConfig{}
+		return nil
+	}
+
+	// Convert the database value to []byte
+	if bytes, ok := value.([]byte); !ok {
+		return errors.New("type assertion to []byte failed for SSOProviderConfig")
```
superlinkx, line 60: Should this be a sentinel error for us to check further up the stack?

superlinkx (changes requested):

superlinkx (approved):
Everything checks out, I think the refactor helped quite a bit in understanding blocks of code better. Always room for improvement, but we've definitely ended up leaving this better than we found it.

PR #1029, opened 2024-12-18
BED-4841: fix: fall back to ObjectID when DomainSID is missing while loading foreign members

computator:
## Description

Update the `FetchForeign*()` and `CreateForeignEntityMembership*Delegate()` methods to fall back to trying the `ObjectID` property when the `DomainSID` property is missing on the domain node.

## Motivation and Context

This PR addresses: BED-4841

The Foreign Members list fails to load on a domain missing `DomainSID` because the endpoints return a 500 error.

## How Has This Been Tested?

1. Loaded the AD example data
2. Queried for `match (n:Domain) return n`
3. Verified that none of the domains show an error for the Foreign Members section

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/analysis/ad/queries.go
+++ b/packages/go/analysis/ad/queries.go
@@ -653,17 +653,9 @@ func FetchDCSyncerPaths(tx graph.Transaction, node *graph.Node) (graph.PathSet, 
 }
 
 func FetchForeignGPOControllers(tx graph.Transaction, node *graph.Node, skip, limit int) (graph.NodeSet, error) {
-	var domainSID string
-	if sid, err := node.Properties.Get(ad.DomainSID.String()).String(); err != nil {
-		if sid, err := node.Properties.Get(common.ObjectID.String()).String(); err != nil {
-			return nil, err
-		} else {
-			domainSID = sid
-		}
-	} else {
-		domainSID = sid
-	}
-	if gpoIDs, err := ops.FetchNodeIDs(tx.Nodes().Filterf(func() graph.Criteria {
+	if domainSID, err := getNodeDomainSIDOrObjectID(node); err != nil {
+		return nil, err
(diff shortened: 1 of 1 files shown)


packages/go/analysis/ad/queries.go
```diff
@@ -653,9 +653,17 @@ func FetchDCSyncerPaths(tx graph.Transaction, node *graph.Node) (graph.PathSet,
...
+	var domainSID string
+	if sid, err := node.Properties.Get(ad.DomainSID.String()).String(); err != nil {
+		if sid, err := node.Properties.Get(common.ObjectID.String()).String(); err != nil {
+			return nil, err
+		} else {
+			domainSID = sid
+		}
+	} else {
+		domainSID = sid
+	}
```
urangel, line 665: Falling back to the object id is clever! 

It looks like we are probably even getting the domain sid upon ingest but then throw it away because of the narrow options for providing details about incoming relationships. Ideally we would fix the root cause but that may be out of scope for this work. 

Since this pattern is used multiple times in all of these foreign 'x' functions I would suggest creating a small function to abstract this logic so there is less code repetition. Then it would also be easily unit testable :D

Let me know if you have any questions. Happy to clarify anything.

AD7ZJ (approved):
It looks good to me! I verified it works on my environment and it fixes the issue of the API returning 500 errors.

packages/go/analysis/ad/queries.go
computator: I had been considering splitting it out, but I decided against it since it was only for a few functions and was pretty "generic" code. However, since you think it's a good idea as well I went ahead and split it out! I kinda wish we could just do something like `domainSID := Node.Properties.DomainSID || Node.Properties.ObjectID` 😆

urangel (approved):

rvazarkar (approved):

PR #1126, opened 2025-02-04
BED-5377: Enable recreating the default admin user with environment variable

superlinkx:
## Description

Adds new environment variable to allow recreating the default admin user

## Motivation and Context

This PR addresses: BED-5377

This supports an upstream need, but also means we'll have a supported way for users to reset their admin credentials if they lose access to their account without having to reset the entire database.

## How Has This Been Tested?

Worked through some scenarios locally with a dataset, seems to handle cases where the admin doesn't yet exist, where it already exists, and won't run by default, only if the environment variable gets intentionally set to true

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/bootstrap/server.go
+++ b/cmd/api/src/bootstrap/server.go
@@ -18,6 +18,7 @@ package bootstrap
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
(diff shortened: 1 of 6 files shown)


AD7ZJ (approved):

zinic (approved):
Nice!

The config is getting a little hairy but that's what a config does. I'm wondering if we shouldn't spend some time giving config, in general, some love. For another time however.

Pull!

PR #1142, opened 2025-02-11
BED-5129: Add top level mux middleware to ensure request body is closed

wes-mil:
## Description

The [SelfDestructingTempFile](https://github.com/SpecterOps/BloodHound/blob/e02041a3d6f693b3f2b7cc135bc1de2df3bf2084/cmd/api/src/api/signature.go#L79-L135) expects the io.ReadCloser to either be read to EOF or have .Close() explicitly called. The only real way to guarantee that one of these scenarios is met is to add a highest level middleware to the mux router that calls close at the very end of the middleware chain.

## Motivation and Context

This PR addresses: BED-5129

Previously, large numbers of temp files were getting created but never deleted. I traced the problem down to when content length was unknown, it would create a temp file even if it was a request with no body. Since the request had no body, it would never get cleaned up later.

## How Has This Been Tested?

Tested locally by removing the middleware and sending a GET request with a header of "transfer-encoding: chunked". This causes a temp file to be created in the api docker container. Then adding the middleware back, the files do not exist past the lifetime of the request.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/middleware/middleware.go
+++ b/cmd/api/src/api/middleware/middleware.go
@@ -284,6 +284,9 @@ func EnsureRequestBodyClosed() mux.MiddlewareFunc {
 		return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
 			next.ServeHTTP(response, request)
 
+			// This type cast is required because of the way that Go interfaces work. It is possible to have an
+			// interface pointed at a pointer that points at nil. This would result in the interface not being nil
+			// but still cause a panic while acting on the interface. https://go.dev/doc/faq#nil_error
 			switch b := request.Body.(type) {
 			case *gzip.Reader:
 				if b != nil {


cmd/api/src/api/signature.go
```diff
@@ -34,48 +34,6 @@ import (
 
 const ErrTemplateHMACSignature string = "unable to compute hmac signature: %w"
 
-// tee takes a source reader and two writers. The function reads from the source until exhaustion. Each read is written
```
superlinkx, line 37: Good catch on the tee not being used anywhere beyond old tests. This is excellent cleanup.

cmd/api/src/api/middleware/middleware.go
```diff
@@ -277,3 +278,22 @@ func FeatureFlagMiddleware(db database.Database, flagKey string) mux.MiddlewareF
 		})
 	}
 }
+
+func EnsureRequestBodyClosed() mux.MiddlewareFunc {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
+			next.ServeHTTP(response, request)
+
+			switch b := request.Body.(type) {
```
superlinkx, line 287: Recommending that we capture the reason this odd code is here in a comment. This is running into the nil != nil problem in Go, and the underlying cause is in stdlib interactions, so this seems like our only real option. Code-wise, the solution is solid we just want historical information available for the next poor engineer that needs to maintain this in the future.

superlinkx (approved):

PR #1151, opened 2025-02-14
BED-4887: AZResetPassword edge false positive on a role-assignable group.

AD7ZJ:
## Description

Before creating AZResetPassword from an AZRole to an AZUser, make sure the user is not a member of a role assignable group. 

## Motivation and Context

This PR addresses: https://specterops.atlassian.net/browse/BED-4887

## How Has This Been Tested?

Ingest SpecterDev.

Run pathfinding between `PARTNER TIER1 SUPPORT@SPECTEROPS DEVELOPMENT` and `ACHILES@SPECTERDEV.ONMICROSOFT.COM`

You should get the following results (important part is that there is no AZPasswordReset edge directly between the two): 
Screenshot from 2025-02-19 16-58-16

Prior to this fix, you would see the following:
Screenshot from 2025-02-19 16-53-39

This is not correct because ACHILES@SPECTERDEV.ONMICROSOFT.COM is a member of a role assignable group (ALL SPECTERDEV USERS@SPECTEROPS DEVELOPMENT) 
image

Since it's a member of a role assignable group, only Global Administrator Role, Privileged Authentication Administrator Role, or Partner Tier2 Support Role can perform a reset password operation. 

You can quickly disable my change by commenting out these lines I added in TenantRoleAssignments()
image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/analysis/azure/filters.go
+++ b/packages/go/analysis/azure/filters.go
@@ -73,10 +73,11 @@ func FilterGroupMembers() graph.Criteria {
 	)
 }
 
-func FilterGroupMembersUsers() graph.Criteria {
+func FilterRoleAssignableGroupMembersUsers() graph.Criteria {
 	return query.And(
 		query.Kind(query.Relationship(), azure.MemberOf),
 		query.Kind(query.Start(), azure.User),
+		query.Equals(query.EndProperty(azure.IsAssignableToRole.String()), "true"),
 	)
 }
 
--- a/packages/go/analysis/azure/role.go
+++ b/packages/go/analysis/azure/role.go
@@ -88,9 +88,9 @@ func (s RoleAssignmentMap) HasRole(id graph.ID, roleTemplateIDs ...string) bool 
 }
 
 type RoleAssignments struct {
-	Principals      graph.NodeKindSet
-	RoleMap         map[string]cardinality.Duplex[uint64]
-	GroupMembership map[graph.ID]cardinality.Duplex[uint64]
+	Principals                    graph.NodeKindSet
+	RoleMap                       map[string]cardinality.Duplex[uint64]
+	RoleAssignableGroupMembership cardinality.Duplex[uint64]
 }
 
 func (s RoleAssignments) GetNodeKindSet(bm cardinality.Duplex[uint64]) graph.NodeKindSet {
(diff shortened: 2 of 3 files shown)


packages/go/analysis/azure/filters.go
```diff
@@ -73,6 +73,13 @@ func FilterGroupMembers() graph.Criteria {
 	)
 }
 
+func FilterGroupMembersUsers() graph.Criteria {
+	return query.And(
+		query.Kind(query.Relationship(), azure.MemberOf),
+		query.Kind(query.Start(), azure.User),
+	)
```
definitelynotagoblin, line 80: I think this might also get Users as members of OUs
Might be intended but worth noting

AD7ZJ: Even when it's invoked from the function FetchGroupMembersUsers()? As long as FetchGroupMembersUsers() is called with an AZGroup as the node I think it should be something like (n:AZGroup)<-MemberOf-(u:AZUser)  If you called FetchGroupMembersUsers() with an OU instead of an AZGroup then I think you are right but I promise not to call it that way :D Actually I copied this from the pattern in FetchGroupMembers(), there is no protection in there either to ensure it gets called with an AZGroup type node.

StephenHinck: Users and OUs are part of the AD schema and connected via the `Contains` edge. The query specifies AZUsers (`azure.user`) and the AZMemberOf (`azure.memberof`), so I think you should be good.

packages/go/analysis/azure/role.go
```diff
@@ -142,6 +143,21 @@ func (s RoleAssignments) UsersWithRolesExclusive(roleTemplateIDs ...string) card
...
 }
 
+func (s RoleAssignments) UsersWithRoleAssignableGroupMembership() cardinality.Duplex[uint64] {
+	members := cardinality.NewBitmap64()
+
+	// loop through all the groups
+	for groupID, groupMemberIDBitmap := range s.GroupMembership {
+		group := s.Principals.Get(azure.Group)[groupID]
+		// if that group is role assignable, set the bits
+		if isRoleAssignable, err := group.Properties.Get(azure.IsAssignableToRole.String()).Bool(); err == nil && isRoleAssignable {
```
superlinkx, line 153: Do we think it would be a good idea to handle the error here? I know we can't extract information, but it may be good to at least log a warning that we couldn't convert to a Boolean value

superlinkx (commented):
Other than this missing error handling, this seems pretty sane to me. The loop doesn't appear to be a significant security concern on its face. If we ingested specterdev without noticing an impact, we're likely fine.

superlinkx (approved):
Code-wise this looks good, let's review the results of performance regression testing, but I'm approving to remove blockers once that's complete

PR #1163, opened 2025-02-21
BED-5301/5298: PenTest Result: Internal IP Address Disclosure AND Rate limit Bypass

ALCooper12:
## Description

- Replaced the tollbooth library/module with [https://github.com/ulule/limiter](url) in order to implement our current rate limiting middleware logic
- I also implemented a new middleware specifically for the `/api/v2/login` endpoint in order to address part of the requirements for BED-5300

## Motivation and Context

This PR addresses:
- BED-5301
- BED-5298
- A little bit of BED-5300

- This enables BHCE to follow stronger security practices, mitigate risks, and contribute to an overall improved security posture

## How Has This Been Tested?
- Local testing within the UI and terminal

## Screenshots (optional):
UI 
Screenshot 2025-02-27 at 3 44 11 PM

Terminal
Screenshot 2025-02-21 at 2 40 26 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/api/middleware/rate_limit.go
```diff
@@ -17,66 +17,42 @@
...
+
+				return rightMostIP
+			}
+		},
+	)
+
+	middleware := stdlib.NewMiddleware(limiter, ipGetter)
+
+	return middleware.Handler
+
```
superlinkx, line 55: I was trying to figure out why we couldn't use the reference X Forwarded For implementation from `limiter` but stumbled on the fact that you'd expect the _leftmost_ to be the IP, but because we don't have multiple reverse proxies _and_ the loadbalancer doesn't strip incoming X Forwarded For (in our case, almost always spoofed IPs), we can't rely on it. However, this puts us in a weird position because if someone is self hosting BloodHound, this will make it so anyone hosting with  multiple reverse proxies/load balancers in front of them will end up detecting internal IPs for rate limiting.

I don't know if there's a one size fits all solution here, but we should probably consider the consequences before moving forward.

ddlees: We can't solely rely on the left-most IP because of spoofing but if we take the rightmost IP we're basically rate limiting all traffic that comes in via the ALB (not good).

In a production deployment we only expect there to be 1 proxy in the `X-Forwarded-For` chain. It's not a perfect solution but I think it's sufficient for rate limiting at the service level:

- If there is no `X-Forwarded-For` header then limit by `request.RemoteAddr`
- If there are exactly 2 IPs in the `X-Forwarded-For` chain then we can limit by the left-most IP since the only 2 IPs should be the real client ID and the ALB proxy IP.
- If there are more than 2 IPs in the `X-Forwarded-For` chain then there are more proxies in the chain than we expect (likely spoofing) and we can limit on the second to last IP because it is the untrusted proxy right before it hits the ALB.

Mind you, this isn't perfect but anything more sophisticated than this requires additional configuration on the ALB and/or external services.

cmd/api/src/api/middleware/rate_limit.go
```diff
@@ -17,75 +17,94 @@
...
+			} else {
+				ips := strings.Split(xff, ",")
+
+				idxIP := len(ips) - cfg.TrustedProxies
+				if idxIP < 0 {
+					slog.WarnContext(r.Context(), "Not enough IPs in X-Forwarded-For, defaulting to first IP")
+					idxIP = 0
+				}
+
+				return strings.TrimSpace(ips[idxIP])
```
ddlees, line 66: In the case of just the ALB in the chain this will limit on the ip of the ALB and not the client ip

cmd/api/src/config/config.go
```diff
@@ -166,6 +166,7 @@ type Configuration struct {
 	FedRAMPEULAText              string                    `json:"fedramp_eula_text"` // Enterprise only
 	EnableTextLogger             bool                      `json:"enable_text_logger"`
 	RecreateDefaultAdmin         bool                      `json:"recreate_default_admin"`
+	TrustedProxies               int                       `json:"trusted_proxies"`
```
ddlees, line 169: I don't think we need this configuration value and can assume only 1 reverse proxy for the majority of cases. Otherwise, we might consider defaulting to 1 instead because coordinating configuration changes for production deployments is painful.

ddlees (changes requested):
Mostly good but we need to revise IP selection a bit.

cmd/api/src/api/middleware/rate_limit.go
```diff
@@ -17,75 +17,94 @@
...
+	rate := limiter.Rate{
+		Period: 1 * time.Second,
+		Limit:  DefaultRateLimit,
+	}
+
+	store := memory.NewStore()
+
+	instance := limiter.New(store, rate)
+
+	cfg, err := config.NewDefaultConfiguration()
```
ddlees, line 104: Change this to use the actual config not the default

cmd/api/src/api/middleware/rate_limit.go
computator: This was based on a misunderstanding and we resolved it in a discussion. The existing logic retrieves the correct IP.

cmd/api/src/api/middleware/rate_limit.go
computator: This was also based on a misunderstanding. The logic is correct.

cmd/api/src/api/middleware/rate_limit.go
```diff
@@ -17,75 +17,88 @@
...
 
+// RemoveRateLimitHeadersMiddleware removes rate limit headers that we do not want appearing in our responses
+func RemoveRateLimitHeadersMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
+		response.Header().Del("X-RateLimit-Limit")
+		response.Header().Del("X-RateLimit-Remaining")
+		response.Header().Del("X-RateLimit-Reset")
+		next.ServeHTTP(response, request)
+	})
+}
```
ddlees, line 85: Is there a reason we're removing these headers? It's a pretty common practice to return the required information for clients to exercise appropriate back-off logic.

ALCooper12: This was already discussed and resolved

cmd/api/src/api/registration/v2.go
```diff
@@ -37,9 +41,27 @@ func registerV2Auth(resources v2.Resources, routerInst *router.Router, permissio
...
+			Limit:  1,
+		}
 
-	router.With(middleware.DefaultRateLimitMiddleware,
+		store := memory.NewStore()
+
+		instance := limiter.New(store, rate)
+
+		return middleware.RateLimitMiddleware(resources.DB, instance)
+	},
```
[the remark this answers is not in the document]
ALCooper12: I was ganna wrap this logic like the default one too. But I thought since this scenario was ONLY specific to the login endpoint, it would be fine

cmd/api/src/api/middleware/rate_limit.go
```diff
@@ -17,75 +17,96 @@
...
+				return remoteIP
+			} else if xff := r.Header.Get("X-Forwarded-For"); xff == "" {
+				slog.DebugContext(r.Context(), "Expected X-Forwarded-For header for rate limiting but none found. Defaulted to remote IP Address", "IP Address", remoteIP)
+				return remoteIP
+			} else {
+				ips := strings.Split(xff, ",")
+
+				idxIP := len(ips) - trustedProxies
+				if idxIP < 0 {
+					slog.WarnContext(r.Context(), "Not enough IPs in X-Forwarded-For, defaulting to first IP")
```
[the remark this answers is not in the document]
computator: We can just log the XFF header like we do with some of the other logging statements

cmd/api/src/api/registration/v2.go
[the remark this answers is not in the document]
ALCooper12: Already did this so far!

computator (approved):

PR #1182, opened 2025-02-27
BED-5445: Deep linking node search

maffkipp:
## Description
Adds single node search functionality to our deep linking implementation on the Explore page. This feature can be tested by enabling the `back_button_support` feature flag. Some other aspects of the Explore page (particularly pathfinding search) may appear temporarily broken when this flag is enabled.

## Motivation and Context
First search type implemented in the larger deep linking initiative.

This PR addresses: [BED-5455](https://specterops.atlassian.net/browse/BED-5445)

## How Has This Been Tested?
Tests have been added to cover the `useNodeSearch` hook that wraps the new query param interactions.

## Types of changes

## Checklist:
  - Assigned myself to this PR
  - Open API docs

[BED-5455]: https://specterops.atlassian.net/browse/BED-5455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useNodeSearch/useNodeSearch.ts
```diff
@@ -0,0 +1,60 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { useEffect, useState } from 'react';
+import { SearchValue } from '../../../store';
+import { useExploreParams } from '../../useExploreParams';
+
+/* Reusable logic for syncing up a single node search field with browser query params on the Explore page. The value of the search field is tracked internally, and is only pushed to query params once the event handler is called by the consumer component. Direct changes to the associated query params will be synced back to the search field. */
+export const useNodeSearch = () => {
+    const [searchTerm, setSearchTerm] = useState<string>('');
+    const [selectedItem, setSelectedItem] = useState<SearchValue | undefined>(undefined);
```
benwaples, line 24: I see what youre saying about the adornment missing on initial render... definitely not the worst thing but I have an idea: 

I think the `selectedItem` could be derived from the `searchTerm`. From what I can tell, its intention in `<ExploreSearchCombobox />` appears to be to display the type adornment and thats it (current behavior). If we were using it to display the `selectedItem` in the list of nodes, I don think it would be possible to derive from `searchTerm`.

So if all we need is the node type which matches the inputs value then I think its safe to say that if a search term returns a single node, then we can display that nodes type?? In which case we could remove the selectedItem prop and say:
```js
// selectedItem could be renamed to startAdornment or somethin
const selectedItem = data?.length === 1 ? data[0] : null;
```

what are your thoughts on thus?

benwaples (commented):
This looks great!! I like the organization and how you did `useNodeSearch`

I left one comment on a potential solution for the missing adornment! But if we determine its not feasible/doesnt make sense im happy to stamp this as is

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useNodeSearch/useNodeSearch.ts
```diff
@@ -0,0 +1,79 @@
...
+internally, and is only pushed to query params once the event handler is called by the consumer component. Direct changes to the associated query
+params will be synced back to the search field. */
+export const useNodeSearch = () => {
+    const [searchTerm, setSearchTerm] = useState<string>('');
+    const [selectedItem, setSelectedItem] = useState<SearchValue | undefined>(undefined);
+
+    const { primarySearch, searchType, setExploreParams } = useExploreParams();
+
+    const graphQuery = nodeSearchGraphQuery({ primarySearch, searchType });
+    const { data: graphData } = useQuery(graphQuery);
```
benwaples, line 33: good call keeping this! I realized that if we did use the `useExploreGraph` directly here then the `matchedNode` below may not return the node that was searched for in the case where the current searched for node doesnt return in the graph results.

benwaples (approved):
Couple comments about build/tests, but functionality is great! Thank you!

cmd/ui/src/views/Explore/ExploreSearch/NodeSearch.tsx
```diff
@@ -14,28 +14,19 @@
...
-    const dispatch = useAppDispatch();
-
-    const primary = useAppSelector((state) => state.search.primary);
-    const { searchTerm, value: selectedItem } = primary;
-
-    const handleNodeEdited = (edit: string): SourceNodeEditedAction =>
-        dispatch(searchbarActions.sourceNodeEdited(edit));
-    const handleNodeSelected = (selected?: SearchValue): SourceNodeSelectedAction =>
-        dispatch(searchbarActions.sourceNodeSelected(selected));
+    const { searchTerm, selectedItem, editSourceNode, selectSourceNode } = useNodeSearchSwitch();
```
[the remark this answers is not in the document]
maffkipp: added a mock endpoint, going to continue running these tests with the feature flag disabled for the time being!

PR #1190, opened 2025-03-04
fix(BED-5436): Constraint violations in LinkWellKnownGroups during ingestion

neumachen:
## Description

This PR fixes an issue where the LinkWellKnownGroups function encountered graph DB constraint violations during ingestion. The function failed when attempting to create well-known group nodes with SIDs existing in the database but with different node types.

The fix modifies the getOrCreateWellKnownGroup function to:
1. Only filter by ObjectID when looking up nodes, not by kind
2. Add the Group kind to existing nodes if it's missing
3. Update the node to ensure it has the correct properties

## Motivation and Context

This PR addresses: [BED-5436](https://specterops.atlassian.net/browse/BED-5436)

During ingestion, users encountered errors when the LinkWellKnownGroups function attempted to run. The specific error was a Neo4j constraint violation (Neo.ClientError.Schema.ConstraintValidationFailed) because nodes with the same SID but different types already existed in the database. This could happen due to previous incomplete or incorrect ingestions.

The solution ensures that SID linking will work correctly even when nodes with the same SID but different types already exist, preventing constraint violations and allowing ingestion to complete successfully.

## How Has This Been Tested?

Added comprehensive tests that verify the behavior with various node scenarios:
- Normal scenario with no pre-existing nodes
- Pre-existing node without Group kind
- Pre-existing node with different kind (constraint violation scenario)
- Pre-existing node with Group kind already

Tests are run against Neo4j and PostgreSQL databases to ensure compatibility across supported database types.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Associated an issue: https://specterops.atlassian.net/browse/BED-5436
  - Open API docs

[BED-5436]: https://specterops.atlassian.net/browse/BED-5436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/go/analysis/ad/ad_testing_tools_test.go
```diff
@@ -0,0 +1,186 @@
...
+	nodeName, err := expectedNode.Properties.Get(common.Name.String()).String()
+	require.NoError(t, err)
+
+	defaultQueryCriterias := []graph.Criteria{
+		query.Equals(query.NodeProperty(ad.DomainSID.String()), domainSID),
+		query.Equals(query.NodeProperty(ad.DomainFQDN.String()), domainFQDN),
+		query.Equals(query.NodeProperty(common.Name.String()), nodeName),
+	}
+
+	queryCriterias = append(queryCriterias, defaultQueryCriterias)
```
zinic, line 69: This may need a var expand in the append since `defaultQueryCriterias` is a slice which means this will result in the slice being added instead of its elements.

superlinkx (changes requested):
We should really not use underscores and long names for Go files. This isn't currently a hard rule, but it's best practice in the Go world. There are a number of things with the way you're trying to set up tests that will cause some issues with the way we do testing in general, and specifically will be incompatible with the work we're doing for the integration testing initiative. If you have some time in the morning, I'd like to review with you and see if we can bring the tests up to spec which might also help fix some of your issues with testing both drivers.

packages/go/analysis/go.mod
```diff
@@ -21,6 +21,8 @@ go 1.23
 require (
 	github.com/RoaringBitmap/roaring v1.9.4
 	github.com/bloodhoundad/azurehound/v2 v2.0.1
+	github.com/go-faker/faker/v4 v4.6.0
+	github.com/rs/xid v1.6.0
```
superlinkx, line 25: I missed the new dependencies earlier. I don't think this one is necessary, for Domain SIDs, we usually just generate one and use it in the test as a constant. This also comes with the upside of it matching the exact format of a microsoft domain SID

packages/go/analysis/go.mod
```diff
@@ -21,6 +21,8 @@ go 1.23
 require (
 	github.com/RoaringBitmap/roaring v1.9.4
 	github.com/bloodhoundad/azurehound/v2 v2.0.1
+	github.com/go-faker/faker/v4 v4.6.0
```
superlinkx, line 24: I'd like to get more input on whether we want faker in the project. Ultimately, it's a somewhat useful library, but we haven't used anything like it before so I'd like to vet it's worth adding a dependency here

packages/go/analysis/ad/ad_test.go
```diff
@@ -0,0 +1,339 @@
...
+			ad.Domain,
+		},
+		Properties: graph.AsProperties(graph.PropertyMap{
+			common.Collected: true,
+			common.Name:      faker.DomainName(),
+			ad.DomainSID:     xid.New().String(),
+			ad.DomainFQDN:    faker.DomainName(),
+		}),
+	}
+}
```
superlinkx, line 303: I'd recommend at a minimum removing the faker/xid calls and using constants here. Ideally, I think the call stack for generating the graph is a little too tall and should be flattened a bit more (i.e.  don't use a function to create the node, just format it inline upstream).

After further discussions, I think it would be wise to drop the faker and xid deps, as they are not providing sufficient value here and we don't currently have a need for them in other testing initiatives right now. It's not that they're bad dependencies, it's just that we want to avoid unnecessary dependencies and the bar for usefulness needs to be higher

superlinkx (changes requested):
Let's rip out the dependencies and use constants for our tests for now

packages/go/analysis/ad/ad_test.go
```diff
@@ -0,0 +1,339 @@
...
+) (
+	*graph.Node,
+	error,
+) {
+	t.Helper()
+
+	var fetchedNode *graph.Node
+	err := graphDB.ReadTransaction(ctx, func(tx graph.Transaction) error {
+		filteredNode, err := tx.Nodes().Filterf(func() graph.Criteria {
+			criterias := make([]graph.Criteria, len(queryCriterias))
```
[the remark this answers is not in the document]
neumachen: I made it criterias since it's a collection, and appropriate as plural.

packages/go/analysis/ad/internal/nodeprops/reader.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/specterops/bloodhound/dawgs/graph"
+	"github.com/specterops/bloodhound/graphschema/ad"
+	"github.com/specterops/bloodhound/graphschema/common"
+)
+
+func errorReadDomainSIDandNameAsString(message string) error {
+	return fmt.Errorf("failed to read domain SID and name: %s", message)
+}
+
+func ReadDomainIDandNameAsString(nodeToRead *graph.Node) (string, string, error) {
```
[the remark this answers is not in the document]
neumachen: @irshadaj this is used in the ad.go file. That change was not committed.

packages/go/analysis/ad/ad_test.go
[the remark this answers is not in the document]
superlinkx: Criteria is already the plural form of criterion. Recommend changing to criteria for consistency with the rest of the codebase

PR #1166, opened 2025-02-24
Bed 5520: searchTab and searchType on tab change

benwaples:
## Description
Ties current tab to the `searchTab` value and sets `searchType` and `searchTab` on tab change of explore search tab. 

## Motivation and Context

This PR addresses: BED-5522

Adds deep linking support for the currently selected search tab. And matches current functionality on tab change where we refetch the search of that tab.

## How Has This Been Tested?
added unit tests and manually tested

Enable `deepLinking` browser mocks to test manually

## Types of changes
- New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Associated an issue: https://github.com/SpecterOps/BloodHound/issues/672
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [x] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments (GoDocs / JSDocs)
- [x] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

PR #1205, opened 2025-03-08
BED-5446:  Deep linking pathfinding search

maffkipp:
## Description
Adds pathfinding search functionality to our deep linking implementation on the Explore page. This feature can be tested by enabling the back_button_support feature flag.

There are some broken behaviors around tab switching that we can address after merging the changes for [BED-5520](https://github.com/SpecterOps/BloodHound/pull/1166) as well as some skipped tests that we can fix up with the `history` package from [BED-5453](https://github.com/SpecterOps/BloodHound/pull/1170).

## Motivation and Context

This PR addresses: [BED-5446](https://specterops.atlassian.net/browse/BED-5446)

This is another one of the required search types we are implementing for the deep linking initiative.

## How Has This Been Tested?
Updated a number of tests to account for our refactors. Added some tests that are currently being skipped for our custom url state hooks, as mentioned above.

## Types of changes

## Checklist:
  - Assigned myself to this PR
  - Open API docs

[BED-5520]: https://specterops.atlassian.net/browse/BED-5520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BED-5453]: https://specterops.atlassian.net/browse/BED-5453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BED-5446]: https://specterops.atlassian.net/browse/BED-5446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+export const mapParamsToFilters = (params: string[], initial: EdgeCheckboxType[]): EdgeCheckboxType[] => {
+    return initial.map((edge) => ({
+        ...edge,
+        checked: !!params.includes(edge.edgeType),
+    }));
+};
+
+export const compareEdgeTypes = (initial: string[], comparison: string[]): boolean => {
+    const a = initial.slice(0).sort();
+    const b = comparison.slice(0).sort();
```
benwaples, line 32: ahhh a great use case for `toSorted` but its not "widely available" yet! Looks like it should be widely available this year or next though.

cmd/ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -17,59 +17,31 @@
...
-        dispatch(searchbarActions.destinationNodeEdited(edit));
-
-    const handleSourceNodeSelected = (selected: SearchValue): SourceNodeSelectedAction => {
-        const doPathfindSearch = !!destinationSelectedItem;
-        return dispatch(searchbarActions.sourceNodeSelected(selected, doPathfindSearch));
-    };
-
-    const handleDestinationNodeSelected = (selected: SearchValue): DestinationNodeSelectedAction =>
-        dispatch(searchbarActions.destinationNodeSelected(selected));
+    const pathfinding = usePathfindingSearchSwitch();
```
benwaples, line 26: I think we will need this state switch also in the PathfindingSwapButton (opposite arrow button next to the filters button), its currently only using redux state!

benwaples (commented):
This is looking fantastic, thank you! 

I left one comment on regarding the swap button, seems we just need to reuse the state switch in there as well! 

There is one behavior I wanted to bring up -- when i perform a pathfinding search, then remove the destination node, and attempt a new single node search (from pathfinding tab) it is still attempting a pathfinding search.... I think it boils down to we dont have a good trigger for when to remove the secondarySearch or primarySearch values (like onBlur or something). This also comes up when I remove a value, toggle searchTabs, and the secondarySearch input is repopulated with what I removed (which would also be the case for primaryValue i think). Ive attached a video to demo!

I feel like we need some mechanism to remove those values, but that is a bit tricky....  if the value ends up at "" onEdit, do we pop off the primarySearch/secondarySearch? Or we add an onBlur handler? Or some event listener.... 


https://github.com/user-attachments/assets/b5722bab-0ffb-40af-9611-f47243a47469

cmd/ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
maffkipp: good catch!! I've plumbed my changes through to this button as well, and i swapped out the button styles for tailwind while i was in there since it was an easy change.

maffkipp:
@benwaples I've updated my logic to clear out the opposing term's query param in the case that no node is selected whenever a single node search is triggered from the pathfinding search box.

Some of the behavior noted around changing tabs I also have a reasonable idea for how to fix, but I think we will need to pull in your search tab changes first so we can make decisions based on the `searchTab` param.

benwaples (approved):
Thanks for the changes!! LGTM!

PR #1210, opened 2025-03-11
BED-5448: entityRelationshipEndpoints

benwaples:
## Description
instead of relationship endpoints mapped to individual entity sections, this pulls out those sections so they can be indexed by a `queryType`.

queryType is unique by `<entity_kind> - <section_label[]>` 

## Motivation and Context

This PR addresses: BED-5448

To support relationship searches with deep links, we need to be able to store what relationship query executed for graph data.

## How Has This Been Tested?
Existing tests pass

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/javascript/bh-shared-ui/src/utils/content.ts
```diff
@@ -292,313 +233,102 @@ export const allSections: Partial<Record<EntityKinds, (id: string) => EntityInfo
...
-                    endpoint: ({ counts, skip, limit, type }: EntitySectionEndpointParams) =>
-                        apiClient
-                            .getAZEntityInfoV2('key-vaults', id, 'key-readers', counts, skip, limit, type, {
-                                signal: controller.signal,
-                            })
-                            .then((res) => {
-                                if (type !== 'graph') res.data.countLabel = 'Key Readers';
-                                return res.data;
-                            }),
+                    queryType: 'azkeyvault-vault_readers-key_readers',
```
elikmiller, line 236: I think the plan is to perhaps split this string on `-` and use that to map these query types to nested accordion sections.

But since the query itself does not care about the `vault_readers` part of this type does it make sense to include it? Or is it preferable to use a type like `azkeyvault-key_readers`?

packages/javascript/bh-shared-ui/src/utils/content.ts
```diff
@@ -292,313 +233,102 @@ export const allSections: Partial<Record<EntityKinds, (id: string) => EntityInfo
...
-                                counts,
-                                skip,
-                                limit,
-                                type,
-                                {
-                                    signal: controller.signal,
-                                }
-                            )
-                            .then((res) => res.data),
+                    queryType: 'azmanagementgroup-descendant_objects-descendant_automation_accounts',
```
elikmiller, line 269: Might be `azmanagementgroup-descendant_objects-descendant_management-groups` instead?

packages/javascript/bh-shared-ui/src/utils/content.ts
```diff
@@ -292,313 +233,102 @@ export const allSections: Partial<Record<EntityKinds, (id: string) => EntityInfo
...
```
elikmiller, line 309: See above comment!

packages/javascript/bh-shared-ui/src/utils/content.ts
benwaples: Im not sure the use case of splitting on `-`, can you explain mapping these query types to nested accordions? These `queryTypes` are available for each section/accordion that has an relationship endpoint, including nested accordions.

I included the parent label (if there was one) so we didnt have to make each child of a section unique from children of another parent of the same entity type... this currently isnt an issue, but figured it was a better safe than sorry. However, depending on the intention of splitting the string by `-`, we may want these not to include the parent label!

packages/javascript/bh-shared-ui/src/utils/content.ts
benwaples: fixed, thank you!

packages/javascript/bh-shared-ui/src/utils/content.ts
benwaples: updated to remove parent labels from queryType

packages/javascript/bh-shared-ui/src/utils/content.ts
benwaples: updated, thank you!

specter-flq (approved):

PR #1211, opened 2025-03-12
BED-5268: Tiering create selector

AD7ZJ:
## Description

Implement the `api/v2/asset-group-labels/{%s}/selectors` endpoint. 

## Motivation and Context

This PR addresses: https://specterops.atlassian.net/browse/BED-5268

## How Has This Been Tested?

Added unit tests for resource and database methods. Tested that the endpoint works using bruno. 

## Types of changes

## Checklist:

  - Assigned myself to this PR


cmd/api/src/api/v2/assetgrouplabels.go
```diff
@@ -0,0 +1,74 @@
...
+	} else if errs := validation.Validate(sel); len(errs) > 0 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
+	} else if idStr, hasLabelID := mux.Vars(request)[api.URIPathVariableAssetGroupLabelID]; !hasLabelID {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrNoAssetGroupLabelId, request), response)
+	} else if id, err := strconv.Atoi(idStr); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupLabelId, request), response)
+	} else if _, err := s.DB.GetAssetGroupLabel(request.Context(), id); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupLabelId, request), response)
+	} else if actor, isUser := auth.GetUserFromAuthCtx(ctx.FromRequest(request).AuthCtx); !isUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusUnauthorized, "unknown user", request), response)
```
computator, line 66: Perhaps we should check for a valid user earlier in the sequence so it's harder to get information back from error messages if you are unauthorized.

cmd/api/src/database/migration/migrations/v7.2.0.sql
```diff
@@ -0,0 +1,107 @@
...
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+CREATE TABLE IF NOT EXISTS kind
+(
+  id   SMALLSERIAL,
+  name varchar(256) not null,
+  primary key (id),
+  unique (name)
+);
```
computator, line 23: I thought this table already exists?

Edit:
Ahh I see below the comment about it being duplicated.

cmd/api/src/api/v2/assetgrouplabels.go
AD7ZJ: Since the v2 handler requires `permissions.GraphDBWrite` I don't think we can even get into this function without being authenticated? This line is just to get the user ID to log in the created_by column.

cmd/api/src/database/migration/migrations/v7.2.0.sql
AD7ZJ: I'll add a comment about that here too though :)

cmd/api/src/api/v2/assetgrouplabels.go
AD7ZJ: Good idea, I fixed that!

cmd/api/src/database/migration/migrations/v7.2.0.sql
```diff
@@ -0,0 +1,110 @@
...
+    created_by text,
+    updated_at timestamp with time zone,
+    updated_by text,
+    disabled_at timestamp with time zone,
+    disabled_by text,
+    name text NOT NULL,
+    description text,
+    is_default boolean,
+    allow_disable boolean,
+    auto_certify boolean,
```
computator, line 98: should these bools be `NOT NULL` and should they have a default?

cmd/api/src/model/assetgrouplabels.go
```diff
@@ -0,0 +1,66 @@
...
+
+type AssetGroupLabelSelector struct {
+	ID                int       `json:"id"`
+	AssetGroupLabelId int       `json:"asset_group_label_id"`
+	CreatedAt         time.Time `json:"created_at"`
+	CreatedBy         string    `json:"created_by"`
+	UpdatedAt         time.Time `json:"updated_at"`
+	UpdatedBy         string    `json:"updated_by"`
+	DisabledAt        null.Time `json:"disabled_at"`
+	DisabledBy        string    `json:"disabled_by"`
```
computator, line 59: This should be a `null.String` as well

cmd/api/src/database/migration/migrations/v7.2.0.sql
```diff
@@ -0,0 +1,110 @@
...
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- This table is normally created by dawgs, as defined in schema_up.sql
+-- We add it here to aid in unit testing since the asset group management feature 
+-- also depends on the kind table.
```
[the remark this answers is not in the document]
AD7ZJ: Fixed!

cmd/api/src/api/registration/v2.go
```diff
@@ -301,5 +301,8 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		//TODO: Update the permission on this once we get something more concrete
 		routerInst.GET("/api/v2/analysis/status", resources.GetAnalysisRequest).RequirePermissions(permissions.GraphDBRead),
 		routerInst.PUT("/api/v2/analysis", resources.RequestAnalysis).RequirePermissions(permissions.GraphDBWrite),
+
+		// Asset group API
```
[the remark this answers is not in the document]
AD7ZJ: Good idea, I'll move it up next to that and fix the comment.

packages/go/dawgs/drivers/pg/query/sql/schema_up.sql
```diff
@@ -78,7 +78,9 @@ create table if not exists graph
 );
 
 -- The kind table contains name to ID mappings for graph kinds. Storage of these types is necessary to maintain search
--- capability of a database without the origin application that generated it.
+-- capability of a database without the origin application that generated it. 
+-- To support FK in asset_group_labels table, the kind table is now maintained by the stepwise migration files.
```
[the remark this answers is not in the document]
AD7ZJ: good finds!

cmd/api/src/api/v2/assetgrouplabels.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+	}
+	return true, nil
+}
+
+func (s *Resources) CreateAssetGroupLabelSelector(response http.ResponseWriter, request *http.Request) {
+	var (
+		err        error
+		sel        model.AssetGroupLabelSelector
+		actorIdStr = mux.Vars(request)[api.URIPathVariableAssetGroupLabelID]
```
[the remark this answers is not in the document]
computator: This still needs to be fixed

cmd/api/src/api/v2/assetgrouplabels.go
```diff
@@ -0,0 +1,74 @@
...
+	)
+	defer measure.ContextMeasure(request.Context(), slog.LevelDebug, "Asset Group Label Selector Create")()
+
+	if err = json.NewDecoder(request.Body).Decode(&sel); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if errs := validation.Validate(sel); len(errs) > 0 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
+	} else if id, err := strconv.Atoi(actorIdStr); err != nil {
```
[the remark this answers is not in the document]
computator: This too

cmd/api/src/api/v2/attack_path.go
```diff
@@ -21,13 +21,14 @@ import (
...
-	ErrorInvalidRFC3339     = "invalid RFC-3339 datetime format: %v"
-	ErrorNoDataType         = "no data type specified in url"
+	ErrorParseParams          = "unable to parse request parameters"
+	ErrorDecodeParams         = "unable to decode request parameters"
+	ErrNoDomainId             = "no domain id specified in url"
+	ErrorNoFindingType        = "no finding type specified"
+	ErrorInvalidFindingType   = "invalid finding type specified: %v"
+	ErrorInvalidRFC3339       = "invalid RFC-3339 datetime format: %v"
+	ErrorNoDataType           = "no data type specified in url"
+	ErrInvalidAssetGroupTagId = "invalid asset group tag id specified in url"
```
stephanieslamb, line 31: I feel like this error should live somewhere else and not in a file related to attack paths? I'm not sure if `error.go` would be a better spot?

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -0,0 +1,74 @@
...
+		tagIdStr = mux.Vars(request)[api.URIPathVariableAssetGroupTagID]
+	)
+	defer measure.ContextMeasure(request.Context(), slog.LevelDebug, "Asset Group Tag Selector Create")()
+
+	if err = json.NewDecoder(request.Body).Decode(&sel); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if errs := validation.Validate(sel); len(errs) > 0 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
+	} else if id, err := strconv.Atoi(tagIdStr); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupTagId, request), response)
```
computator, line 61: I was thinking about it more, maybe this check should be first? What's the point of parsing/validating the request body if the URL isn't even valid.

Also since this is validating the URL not the request body, I think it should return a 404 for an invalid ID (and thus invalid URL). This also corresponds to how the `api.HandleDatabaseError` call below will return a 404 when it doesn't find the ID.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -0,0 +1,74 @@
...
+func areCypherSelectorSeedsValid(graph queries.Graph, seeds []model.SelectorSeed) (bool, error) {
+	for _, seed := range seeds {
+		if seed.Type == model.SelectorTypeCypher {
+			if _, err := graph.PrepareCypherQuery(seed.Value, queries.QueryComplexityLimitSelector); err != nil {
+				return false, err
+			}
+		}
+	}
+	return true, nil
+}
```
computator, line 46: The bool return value for this is redundant. We can just check for the existence of an error. If the error return is `nil` then it was successful and they are all valid.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -0,0 +1,144 @@
...
+
+// AssetGroupTagData defines the methods required to interact with the asset_group_tags table
+type AssetGroupTagData interface {
+	CreateAssetGroupTag(ctx context.Context, tagType model.AssetGroupTagType, userId string, name string, description string, position null.Int32, requireCertify null.Bool) (model.AssetGroupTag, error)
+	GetAssetGroupTag(ctx context.Context, assetGroupTagId int) (model.AssetGroupTag, error)
+}
+
+// AssetGroupTagSelectorData defines the methods required to interact with the asset_group_tag_selectors and asset_group_tag_selector_seeds tables
+type AssetGroupTagSelectorData interface {
+	CreateAssetGroupTagSelector(ctx context.Context, assetGroupTagId int, userId string, name string, description string, isDefault bool, allowDisable bool, autoCertify bool, seeds []model.SelectorSeed) (model.AssetGroupTagSelector, error)
```
computator, line 40: Would `isDefault` and/or `allowDisable` ever be set programmatically, or just by default tags added via DB migrations? If not, we can remove both these params and the corresponding column inserts and just allow it to follow the DB defaults which would be correct for user-added tags.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -0,0 +1,144 @@
...
+	var (
+		selector = model.AssetGroupTagSelector{
+			AssetGroupTagId: assetGroupTagId,
+			CreatedBy:       userId,
+			UpdatedBy:       userId,
+			Name:            name,
+			Description:     description,
+			IsDefault:       isDefault,
+			AllowDisable:    allowDisable,
+			AutoCertify:     autoCertify,
```
computator, line 53: We probably shouldn't set any of these fields since they are just overwritten by the `.Scan()` lower down (and if they are not we want our tests to catch that).

cmd/api/src/database/assetgrouptags.go
```diff
@@ -0,0 +1,144 @@
...
+	}); err != nil {
+		return model.AssetGroupTagSelector{}, err
+	}
+
+	return selector, nil
+}
+
+func (s *BloodhoundDB) GetAssetGroupTag(ctx context.Context, assetGroupTagId int) (model.AssetGroupTag, error) {
+	var tag model.AssetGroupTag
+	if result := s.db.WithContext(ctx).Raw(fmt.Sprintf("SELECT id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, deleted_at, deleted_by, position, require_certify FROM %s WHERE id = ? AND deleted_at IS NULL", tag.TableName()), assetGroupTagId).First(&tag); result.Error != nil {
```
computator, line 93: I don't think we need to select `deleted_at` and `deleted_by` since we are already filtering it to ones with `deleted_at` set to null.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -0,0 +1,144 @@
...
+func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.AssetGroupTagType, userId string, name string, description string, position null.Int32, requireCertify null.Bool) (model.AssetGroupTag, error) {
+	var (
+		tag = model.AssetGroupTag{
+			Type:           tagType,
+			CreatedBy:      userId,
+			UpdatedBy:      userId,
+			Name:           name,
+			Description:    description,
+			Position:       position,
+			RequireCertify: requireCertify,
```
computator, line 109: Same situation here about not pre-filling the values.

cmd/api/src/api/v2/attack_path.go
AD7ZJ: I agree, how about just make it a constant at the top of the assetgrouptag.go resources file? It is specific to this file anyway...

cmd/api/src/database/assetgrouptags.go
```diff
@@ -0,0 +1,144 @@
...
+		auditEntry = model.AuditEntry{
+			Action: model.AuditLogActionCreateAssetGroupTag,
+			Model:  &tag, // Pointer is required to ensure success log contains updated fields after transaction
+		}
+	)
+
+	if tagType == model.AssetGroupTagTypeLabel {
+		position = null.Int32{}
+		requireCertify = null.Bool{}
+	}
```
computator, line 121: Rather than just setting these to zero values, we should prob return an error if they are not already zero-values so that the caller knows they have a logic error. This way would just mask any logic errors.

cmd/api/src/api/v2/assetgrouptags.go
AD7ZJ: That's fair, I'll re-order it

cmd/api/src/api/v2/attack_path.go
computator: If we just want to handle this particular one that seems fine, do we want to move/fix the other generic ones as well though?

cmd/api/src/model/assetgrouphistory.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"time"
+
+	"github.com/specterops/bloodhound/src/database/types/null"
+)
+
+type AssetGroupHistoryAction string
+
+const (
+	AssetGroupHistoryActorSystem = "SYSTEM"
```
computator, line 28: This is also used outside history for the default T0 tag too right?

cmd/api/src/database/assetgrouphistory.go
```diff
@@ -0,0 +1,45 @@
...
+
+// AssetGroupHistoryData defines the methods required to interact with the asset_group_history table
+type AssetGroupHistoryData interface {
+	CreateAssetGroupHistoryRecord(ctx context.Context, actor, target string, action model.AssetGroupHistoryAction, assetGroupTagId int, environmentId, note string) error
+	GetAssetGroupHistoryRecords(ctx context.Context) ([]model.AssetGroupHistory, error)
+}
+
+func (s *BloodhoundDB) CreateAssetGroupHistoryRecord(ctx context.Context, actor, target string, action model.AssetGroupHistoryAction, assetGroupTagId int, environmentId, note string) error {
+	return CheckError(s.db.WithContext(ctx).Exec(fmt.Sprintf("INSERT INTO %s (actor, target, action, asset_group_tag_id, environment_id, note) VALUES (?, ?, ?, ?, ?, ?)", assetGroupHistoryTable),
+		actor, target, action, assetGroupTagId, null.StringFrom(environmentId), null.StringFrom(note)))
```
computator, line 39: Since `environmentId` and `note` are passed as strings they will never be null, thus removing the benefit of allowing null for the columns in the db. For those to be effective we'll need to pass along the types to the callers as well.

computator (changes requested):

cmd/api/src/database/assetgrouptags.go
AD7ZJ: I am not sure, will the default tags be added programatically or only via db migrations?

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -0,0 +1,222 @@
...
+						},
+						IsDefault:   false,
+						AutoCertify: false,
+					})
+				},
+				Test: func(output apitest.Output) {
+					apitest.StatusCode(output, http.StatusBadRequest)
+					apitest.BodyContains(output, "invalid asset group tag id specified in url")
+				},
+			},
```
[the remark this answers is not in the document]
AD7ZJ: I'm not sure, since the code is there I thought I might as well have a test for it

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -0,0 +1,222 @@
...
+			{
+				Name: "NonExistantTagUrlId",
```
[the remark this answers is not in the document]
AD7ZJ: fixed

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+	apitest.
+		NewHarness(t, resourcesInst.CreateAssetGroupTagSelector).
+		Run([]apitest.Case{
+			{
+				Name: "BadRequest",
+				Input: func(input *apitest.Input) {
+					apitest.SetContext(input, userCtx)
+					apitest.SetURLVar(input, api.URIPathVariableAssetGroupTagID, "1")
+					apitest.BodyString(input, "{\"name\":[\"BadRequest\"]}")
```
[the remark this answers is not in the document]
AD7ZJ: that works, thanks!

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -0,0 +1,222 @@
...
+						Seeds: []model.SelectorSeed{
+							{Type: 0, Value: "this should be a string of cypher"},
+						},
+						IsDefault:   false,
+						AutoCertify: false,
+					})
+				},
+				Setup: func() {
+					mockDB.EXPECT().GetAssetGroupTag(gomock.Any(), gomock.Any()).
+						Return(model.AssetGroupTag{}, errors.New("entity not found")).Times(1)
```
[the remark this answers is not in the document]
AD7ZJ: nice catch, fixed

cmd/api/src/api/v2/assetgrouptags_test.go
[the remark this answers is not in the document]
AD7ZJ: I'll add that!

cmd/api/src/api/v2/assetgrouplabels.go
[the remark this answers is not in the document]
AD7ZJ: got it

cmd/api/src/database/assetgrouptags.go
AD7ZJ: well, I did that because they need to be filled out for the audit log, otherwise the "intent" entry has all the values empty.

computator: Ahh ok that makes sense. I thought it only used it at the end.

cmd/api/src/database/assetgrouphistory.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/specterops/bloodhound/src/database/types/null"
+	"github.com/specterops/bloodhound/src/model"
+)
+
+const (
+	assetGroupHistoryTable = "asset_group_history"
```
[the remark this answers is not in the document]
AD7ZJ: Fixed

cmd/api/src/database/assetgrouptags.go
AD7ZJ: that's true, fixed

cmd/api/src/database/assetgrouptags.go
AD7ZJ: It's for the audit log again

cmd/api/src/database/assetgrouptags.go
AD7ZJ: I added a check to return an error if either is not null.

cmd/api/src/model/assetgrouphistory.go
AD7ZJ: Ok I'll move it into assetgrouptags model.

cmd/api/src/database/assetgrouphistory.go
AD7ZJ: Good idea, fixed that!

cmd/api/src/database/assetgrouptags.go
```diff
@@ -0,0 +1,145 @@
...
+	if err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+		bhdb := NewBloodhoundDB(tx, s.idResolver)
+
+		var kindId int
+		if result := tx.Raw(fmt.Sprintf("INSERT INTO %s (name) VALUES (?) RETURNING id", kindTable), tag.ToKind()).Scan(&kindId); result.Error != nil {
+			return CheckError(result)
+		} else if result := tx.Raw(fmt.Sprintf(`
+			INSERT INTO %s (type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify) 
+			VALUES (?, ?, ?, ?, NOW(), ?, NOW(), ?, ?, ?) 
+			RETURNING id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, deleted_at, deleted_by, position, require_certify`,
```
computator, line 133: This could also skip `deleted_at` and `deleted_by` since we are just creating the record.

cmd/api/src/database/migration/migrations/v7.3.0.sql
```diff
@@ -0,0 +1,110 @@
...
+-- limitations under the License.
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Set `back_button_support` feature flag as user updatable
+UPDATE feature_flags SET user_updatable = true WHERE key = 'back_button_support';
+
+-- Specify the `back_button_support` feature flag is currently only for BHCE users
+UPDATE feature_flags SET description = 'Enable users to quickly navigate between views in a wider range of scenarios by utilizing the browser navigation buttons. Currently for BloodHound Community Edition users only.' WHERE key = 'back_button_support';
+
```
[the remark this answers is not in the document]
AD7ZJ: aah yes, oops. Fixed!

cmd/api/src/database/migration/migrations/v7.3.0.sql
```diff
@@ -0,0 +1,110 @@
...
+    PRIMARY KEY (id),
+    CONSTRAINT fk_asset_group_tags_asset_group_selectors FOREIGN KEY (asset_group_tag_id) REFERENCES asset_group_tags(id) ON DELETE CASCADE
+);
+
+-- Add asset_group_tag_selector_seeds table
+CREATE TABLE IF NOT EXISTS asset_group_tag_selector_seeds
+(
+    selector_id int,
+    type int,
+    value text,
```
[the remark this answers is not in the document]
AD7ZJ: ok done.

cmd/api/src/model/assetgrouptags.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/specterops/bloodhound/dawgs/graph"
+	"github.com/specterops/bloodhound/src/database/types/null"
+)
+
+const (
+	AssetGroupHistoryActorSystem = "SYSTEM"
```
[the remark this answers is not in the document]
AD7ZJ: yep

cmd/api/src/database/assetgrouptags.go
AD7ZJ: Ok sounds good.

computator (approved):
I think this all looks good!

PR #1213, opened 2025-03-12
BED-5249: feat: Get Tier/Label List Endpoint

computator:
## Description

- Adds `/api/v2/asset-group-tags`
  - Filter ability to select by `type` and other properties
  - `counts` param to include selector and member counts
- Resolves issues when `autoCertify` is unspecified for `CreateAssetGroupTagSelector`

## Motivation and Context

This PR addresses: https://specterops.atlassian.net/browse/BED-5249

Make it possible to retrieve a list of asset group tags.

## How Has This Been Tested?

Primarily tested by writing new tests.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -39,6 +41,69 @@ const (
...
+			if selectorCounts, err := s.DB.GetAssetGroupTagSelectorCounts(request.Context(), ids); err != nil {
+				api.HandleDatabaseError(request, response, err)
+				return
+			} else {
+				resp.Counts.Selectors = selectorCounts
+			}
+			memberCounts := make(map[int]int, len(tags))
+			for _, tag := range tags {
+				// TODO: use a more efficient query method
+				if nodelist, err := s.GraphQuery.GetNodesByKind(request.Context(), tag.ToKind()); err != nil {
```
computator, line 94: If we end up creating the method in the referenced comment we can use it here too instead of `GetNodesByKind`

https://github.com/SpecterOps/BloodHound/pull/1307/files#r2027599359

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -39,6 +41,69 @@ const (
 	ErrInvalidSelectorType    = "invalid selector type"
 )
 
+type GetAssetGroupTagsResponse struct {
```
stephanieslamb, line 44: this can probably be private?

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -39,6 +41,69 @@ const (
...
+		pnameTagType       = "type"
+		pnameIncludeCounts = "includeCounts"
+	)
+	var (
+		pvalsTagType = map[string]model.AssetGroupTagType{
+			strconv.Itoa(int(model.AssetGroupTagTypeLabel)): model.AssetGroupTagTypeLabel,
+			strconv.Itoa(int(model.AssetGroupTagTypeTier)):  model.AssetGroupTagTypeTier,
+			"": model.AssetGroupTagTypeAll, // default
+		}
+		pvalsIncludeCounts = map[string]bool{
```
stephanieslamb, line 63: what does the "p" stand for in these const and variable names?

computator: I had thought about that, but it's also used in the `v2_test` package so it would need to be exported right?

cmd/api/src/api/v2/assetgrouptags.go
computator: I had thought about that, but it's also used in the `v2_test` package so it would need to be exported right?

cmd/api/src/api/v2/assetgrouptags.go
computator: Parameter! 🙂

stephanieslamb: thats what i figured ☺️  but thought i would double check! i havent seen this pattern used yet.

computator: We haven't! I just made it 🙂

cmd/api/src/api/v2/assetgrouptags.go
zinic: The function is coming along for the ride in the above changeset - highly recommend rebasing on top of it and using it once the changeset drops.

computator: Awesome!

cmd/api/src/api/v2/ad_entity.go
```diff
@@ -63,7 +63,7 @@ func (s *Resources) PatchDomain(response http.ResponseWriter, request *http.Requ
 }
 
 func (s *Resources) handleAdEntityInfoQuery(response http.ResponseWriter, request *http.Request, entityType graph.Kind, countQueries map[string]any) {
-	if hydrateCounts, err := api.ParseOptionalBool(request.URL.Query().Get(api.QueryParameterHydrateCounts), true); err != nil {
+	if hydrateCounts, err := api.ParseOptionalBool(request.URL.Query().Get(api.QueryParameterIncludeCounts), true); err != nil {
```
urangel, line 66: It would line things up more if we also renamed `hydrateCounts` here to `includeCounts`
```suggestion
	if includeCounts, err := api.ParseOptionalBool(request.URL.Query().Get(api.QueryParameterIncludeCounts), true); err != nil {
```

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -29,12 +30,78 @@ import (
...
 	"github.com/specterops/bloodhound/src/utils/validation"
 )
 
+type GetAssetGroupTagsResponse struct {
+	Tags   model.AssetGroupTags `json:"tags"`
+	Counts struct {
+		Selectors map[int]int   `json:"selectors"`
+		Members   map[int]int64 `json:"members"`
+	} `json:"counts,omitempty"`
+}
```
urangel, line 46: Can we please embed this `Counts` struct inside of the `tags` object? 🙏 It would make it more ergonomic for us to use for displaying the Summary view that is planned for later since it would allow us to only loop through the tags without doing two other loops to line up the selectors and members counts for the associated tag

urangel: Here is what I'm thinking since I'm not sure I explained that well lol

```go
type AssetGroupTagCounts = struct {
	Selectors int   `json:"selectors"`
	Members   int64 `json:"members"`
}

type AssetGroupTagView struct {
	ID             int                 `json:"id"`
	Type           AssetGroupTagType   `json:"type"`
	KindId         int                 `json:"kind_id"`
	Name           string              `json:"name"`
	Description    string              `json:"description"`
	CreatedAt      time.Time           `json:"created_at"`
	CreatedBy      string              `json:"created_by"`
	UpdatedAt      time.Time           `json:"updated_at"`
	UpdatedBy      string              `json:"updated_by"`
	DeletedAt      null.Time           `json:"deleted_at"`
	DeletedBy      null.String         `json:"deleted_by"`
	Position       null.Int32          `json:"position"`
	RequireCertify null.Bool           `json:"require_certify"`
	Counts         AssetGroupTagCounts `json:"counts,omitempty"`
}

type GetAssetGroupTagsResponse = []AssetGroupTagView
```

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -29,12 +30,78 @@ import (
...
 	"github.com/specterops/bloodhound/src/queries"
 	"github.com/specterops/bloodhound/src/utils/validation"
 )
 
+type GetAssetGroupTagsResponse struct {
+	Tags   model.AssetGroupTags `json:"tags"`
+	Counts struct {
+		Selectors map[int]int   `json:"selectors"`
+		Members   map[int]int64 `json:"members"`
+	} `json:"counts,omitempty"`
```
AD7ZJ, line 45: I've noticed the `Counts` field always gets returned regardless of the counts URL parameter. Am I understanding omitempty right? Maybe it needs to be a struct pointer instead of a straight struct?

computator: Good catch. Most of my testing has been via tests, so I hadn't noticed that.

cmd/api/src/api/v2/assetgrouptags.go
computator: This should be fixed

cmd/api/src/database/assetgrouptags.go
```diff
@@ -167,6 +173,52 @@ func (s *BloodhoundDB) GetAssetGroupTag(ctx context.Context, assetGroupTagId int
...
+	}
+	return tags, nil
+}
+
+func (s *BloodhoundDB) GetAssetGroupTagSelectorCounts(ctx context.Context, tagIds []int) (map[int]int, error) {
+	var counts = make(map[int]int, len(tagIds))
+	// initalize values to 0 for any ids that end up with no rows
+	for _, i := range tagIds {
+		counts[i] = 0
+	}
```
computator, line 200: TODO: figure out why I wanted to fill out the map for ids without matching rows

cmd/api/src/api/v2/assetgrouptags.go
urangel: Thank you! E>

cmd/api/src/database/assetgrouptags.go
```diff
@@ -71,6 +73,10 @@ func (s *BloodhoundDB) CreateAssetGroupTagSelector(ctx context.Context, assetGro
 		}
 	)
 
+	if !autoCertify.Valid {
+		return model.AssetGroupTagSelector{}, fmt.Errorf("auto_certify must be set to true or false")
+	}
+
```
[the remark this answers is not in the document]
AD7ZJ: Good idea!

cmd/api/src/database/assetgrouptags.go
computator: So I had this for 2 reasons:
- So this pseudocode test passes:
  ```go
  idlist := []int{1,2,3,4}
  res := GetCounts(idlist)
  require.Equal(len(res), len(idlist))
  ```
- For the conceptual idea that we should expect to receive a result "row" for every ID passed to the method. If we asked for a count for id _x_, we should expect to receive an explicitly defined value for it, even if that value is 0. This would also be important if we were to ever use `maps.Keys()` or similar on the result.

computator: Thoughts @mistahj67?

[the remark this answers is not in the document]
computator: Resolving in preference of the other thread here https://github.com/SpecterOps/BloodHound/pull/1213#discussion_r2038570498

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - Asset Isolation
+    - Enterprise
+    - Community
+  parameters:
+  - name: type
+    description: Type of tag. Either Tier = 1 or Label = 2
+    in: query
+    required: false
+    schema:
+      type: string
```
computator, line 34: I'm not sure how to specify that `type` uses the filter syntax and that additional columns are also filterable

cmd/api/src/database/assetgrouptags.go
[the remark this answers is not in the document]
computator: Fixed!

AD7ZJ (approved):
Looks good to me.

PR #1214, opened 2025-03-13
BED-5452 + BED-5454: Support for `selectedItem` and `expandedPanelSections` URL query parameters

elikmiller:
## Description

When the `back_button_support` feature flag is enabled...
- The currently selected Explore page graph item (node or edge) is maintained as a URL query parameter `selectedItem`
  - The currently selected item determines which item in the graph is currently highlighted as well as which item appears in the right-hand item information panel
- The currently selected/open Panel (node or edge) is maintained as a URL query parameter `expandedPanelSections`. (At the same time we maintain the global state functionality for when flag is off)
- On load the Panel checks the url for the param and sets the open panels if any. 
- When clicking on an accordion it sets `expandedPanelSections`, `searchType`, `relationshipQueryType` and `relationshipQueryItemId`.
- It supports nested accordions so if a nested accordion is clicked the same as above will happen, but we will have two `expandedPanelSections`params.
- When clicking on a node inside the accordion it sets `primarySearch`, `searchType` and `exploreSearchTab`. 
- Object Information and Relationship Information section were also decoupled from the open/close functionality of the sections in the panel to make it possible to only have one open at a time.

## Motivation and Context

This PR addresses:
- https://specterops.atlassian.net/browse/BED-5452
- https://specterops.atlassian.net/browse/BED-5454

These changes are part of a larger effort which will allow users to use the browser forward and back buttons to more seamlessly navigate through various graph results on the Explore page. Additionally, users will be able to share links with the aforementioned query parameters and share their search results with other users.

## How Has This Been Tested?

Unit tests have been updated where necessary. Currently, while the feature is still in early access, most tests are manual.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/ui/src/views/Explore/EdgeInfo/EdgeInfoContent.tsx
```diff
@@ -78,27 +84,54 @@ const EdgeInfoContent: FC<{ selectedEdge: NonNullable<SelectedEdge> }> = ({ sele
...
+                            dispatch(edgeSectionToggle({ section: sectionKeyLabel, expanded: !isExpandedSection }));
+                        };
+
+                        const handleOnChange = () => {
+                            handleCurrentSectionToggle();
+                            if (backButtonFlag?.enabled) {
+                                setExpandedPanelSectionsParam();
+                            }
+                            if (sendOnChange && !backButtonFlag?.enabled) {
+                                getOnChange(
```
benwaples, line 117: Discussed offline but leaving a comment here for a placeholder -- looks like getOnChange returns a callback that will actually execute the composition query for V1 flows

cmd/ui/src/views/Explore/EntityInfo/EntityInfoPanel.tsx
```diff
@@ -32,14 +39,26 @@ interface EntityInfoPanelProps {
...
     const previousSelectedNode = usePreviousValue(selectedNode);
 
     useEffect(() => {
-        if (previousSelectedNode?.id !== selectedNode?.id) {
-            setExpandedSections({ 'Object Information': true });
+        if (backButtonFlag?.enabled && previousSelectedNode?.id !== selectedNode?.id && expandedPanelSections) {
+            const initialExpandedSections = { ...formatPanelSectionsParams(expandedPanelSections) };
+            setExpandedSections(initialExpandedSections);
+        } else if (!backButtonFlag?.enabled && previousSelectedNode?.id !== selectedNode?.id) {
+            setExpandedSections({});
```
benwaples, line 52: For V1 flows, looks like currently opened accordions stay open when selecting new entities of the same type so I think we still need that first `setExpandedSections({ 'Object Information': true });` still? Something like:

```tsx
  if (previousSelectedNode?.id !== selectedNode?.id) {
            if (backButtonFlag?.enabled && expandedPanelSections) {
                const initialExpandedSections = { ...formatPanelSectionsParams(expandedPanelSections) };
                setExpandedSections(initialExpandedSections);
            } else {
                setExpandedSections({ 'Object Information': true });
            }
        }
```

But take that with a grain of salt, im not sure how the section open by default, `Object Information`, works now

benwaples: disregard - something was off locally

packages/javascript/bh-shared-ui/src/hooks/useExploreSelectedItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const setSelectedItem = useCallback(
+        (itemId: string) => {
+            if (itemId !== selectedItem)
+                setExploreParams({
+                    selectedItem: itemId,
+                    expandedPanelSections: null,
+                });
+        },
+        [selectedItem, setExploreParams]
+    );
```
benwaples, line 36: I really like encoding resetting the `expandedPanelSections` here! One thing we may need to figure out in the future is how to batch update this with other query param updates so the user doesnt have multiple changes to their history from a single action. Looks like we dont have a use case like that currently, but something to note for the future!

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItemV2.tsx
```diff
@@ -0,0 +1,163 @@
...
+import { Button } from '@bloodhoundenterprise/doodleui';
+import { Dialog, DialogActions, DialogContent, DialogTitle, MenuItem } from '@mui/material';
+import { NodeResponse, apiClient, useExploreSelectedItem, useNotifications } from 'bh-shared-ui';
+import { FC, useState } from 'react';
+import { useMutation, useQuery } from 'react-query';
+import { selectOwnedAssetGroupId, selectTierZeroAssetGroupId } from 'src/ducks/assetgroups/reducer';
+import { toggleOwnedObjectNode, toggleTierZeroNode } from 'src/ducks/explore/actions';
+import { useAppDispatch, useAppSelector } from 'src/store';
+
+const AssetGroupMenuItem: FC<{ assetGroupId: number; assetGroupName: string }> = ({ assetGroupId, assetGroupName }) => {
```
benwaples, line 26: Nit: Mind adding the V2 tag on the component?

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItemV2.tsx
```diff
@@ -0,0 +1,163 @@
...
+                    object_id: `object_id=eq:${(selectedItemQuery.data as NodeResponse)?.objectId}`,
+                },
+            })
+            .then((res) => res.data.data?.members)
+    );
+
+    const handleAddToAssetGroup = () => {
+        if (selectedItemQuery.data) {
+            mutation.mutate({ nodeId: (selectedItemQuery.data as NodeResponse).objectId, action: 'add' });
+        }
```
benwaples, line 79: Casting here seems pretty safe because this option is only available for nodes! As an option to avoid casting `as NodeResponse` could be able to check for the `objectId in selectedItem.data` before accessing it:

```suggestion
        if (selectedItemQuery.data && 'objectId' in selectedItemQuery.data) {
            mutation.mutate({ nodeId: selectedItemQuery.data.objectId, action: 'add' });
        }
```

cmd/ui/src/components/GraphEdgeEventsV2.tsx
```diff
@@ -0,0 +1,196 @@
...
+    getEdgeDataFromKey,
+    getEdgeLabelTextLength,
+    getEdgeSourceAndTargetDisplayData,
+} from 'src/ducks/graph/utils';
+import { getBackgroundBoundInfo, getSelfEdgeStartingPoint } from 'src/rendering/programs/edge-label';
+import { getControlPointsFromGroupSize } from 'src/rendering/programs/edge.self';
+import { bezier } from 'src/rendering/utils/bezier';
+import { useAppDispatch, useAppSelector } from 'src/store';
+
+const GraphEdgeEvents: FC = () => {
```
benwaples, line 31: Nit: `GraphEdgeEventsV2`

cmd/ui/src/views/Explore/ContextMenu/CopyMenuItemV2.tsx
```diff
@@ -0,0 +1,82 @@
...
+        backgroundColor: theme.palette.background.paper,
+        padding: 0,
+        paddingTop: '0.5rem',
+        paddingBottom: '0.5rem',
+        boxShadow: theme.shadows[8],
+        marginLeft: '2px !important',
+    },
+}));
+
+const CopyMenuItemV2 = () => {
```
benwaples, line 36: Updating this component wasnt called out in any ticket and we wouldve missed it. TY TY TY for catching that!!

cmd/ui/src/views/Explore/EdgeInfo/EdgeInfoContent.tsx
specter-flq: all set

cmd/ui/src/views/Explore/EdgeInfo/EdgeInfoContent.tsx
```diff
@@ -78,27 +85,54 @@ const EdgeInfoContent: FC<{ selectedEdge: NonNullable<SelectedEdge> }> = ({ sele
...
                         const sendOnChange =
                             EdgeCompositionRelationships.includes(selectedEdge.name) && section[0] === 'composition';
 
+                        const setExpandedPanelSectionsParam = () => {
+                            setExploreParams({
+                                expandedPanelSections: [sectionKeyLabel],
+                                ...(sectionKeyLabel === 'composition'
+                                    ? { searchType: 'composition' }
+                                    : { searchType: null }),
+                                ...(sectionKeyLabel === 'composition' && { relationshipQueryItemId: selectedItem }),
```
benwaples, line 101: Is there an edge case im missing that requires 2 separate conditionals instead of one? By 1 conditional I mean:

```suggestion
                                ...(sectionKeyLabel === 'composition'
                                    ? { searchType: 'composition', relationshipQueryItemId: selectedItem }
                                    : { searchType: null }),
```

cmd/ui/src/views/Explore/EdgeInfo/EdgeInfoContent.tsx
benwaples: Appreciate the change! Looks great

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItemV2.tsx
specter-flq: Updated! Great suggestion imo, and something I want to save to my toolbox for type checking, thanks again!

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItemV2.tsx
specter-flq: Updated!

cmd/ui/src/components/GraphEdgeEventsV2.tsx
specter-flq: Updated!

cmd/ui/src/views/Explore/ContextMenu/ContextMenuV2.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Menu, MenuItem } from '@mui/material';
+
+import { NodeResponse, Permission, searchbarActions, useExploreSelectedItem, usePermissions } from 'bh-shared-ui';
+import { FC } from 'react';
+import { selectOwnedAssetGroupId, selectTierZeroAssetGroupId } from 'src/ducks/assetgroups/reducer';
+import { useAppDispatch, useAppSelector } from 'src/store';
+import AssetGroupMenuItem from './AssetGroupMenuItemV2';
+import CopyMenuItem from './CopyMenuItemV2';
+
+const ContextMenu: FC<{ contextMenu: { mouseX: number; mouseY: number } | null; handleClose: () => void }> = ({
```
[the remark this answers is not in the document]
specter-flq: Updated!

packages/javascript/bh-shared-ui/src/hooks/useExploreSelectedItem.tsx
elikmiller: action/reducer pattern for query params?

cmd/ui/src/views/Explore/ContextMenu/AssetGroupMenuItemV2.tsx
elikmiller: There is also an exported type guard to do just this - see `isNode` from `useGraphItem`

cmd/ui/src/views/Explore/EdgeInfo/EdgeInfoContent.tsx
specter-flq: good catch, I tried it but it was only one param for some reason, let me give it another try

cmd/ui/src/views/Explore/EntityInfo/EntityInfoDataTable.tsx
```diff
@@ -71,7 +111,22 @@ const EntityInfoDataTable: React.FC<EntityInfoDataTableProps> = ({ id, label, en
...
+    const handleCurrentSectionToggle = () => {
+        if (backButtonFlag?.enabled && allSectionsMap) {
+            collapseNonSelectedSections(expandedSections, allSectionsMap, label); // We want to keep only one open at a time
+        }
+        toggleSection(label);
+    };
+
+    const setNodeSearchParams = (item: SelectedNode) => {
+        setExploreParams({
+            primarySearch: item.id ?? item.name,
```
benwaples, line 123: I know I called this pattern out specifically in the ticket (item.id ?? item.name), however in other places we have left off the `item.name` nullish coalesce entirely to avoid any conflicts at all. Mind updating here to do the same? 

That way we are being consistent throughout and once we decide if falling back to the name is okay, we can do that in one change!

specter-flq: No worries sounds good, updated, let me know if that is what you meant. Thanks again.

cmd/ui/src/views/Explore/EntityInfo/EntityInfoDataTable.tsx
```diff
@@ -71,7 +111,22 @@ const EntityInfoDataTable: React.FC<EntityInfoDataTableProps> = ({ id, label, en
         }
     };
 
-    const handleOnClick = (item: any) => {
+    const handleCurrentSectionToggle = () => {
+        if (backButtonFlag?.enabled && allSectionsMap) {
+            collapseNonSelectedSections(expandedSections, allSectionsMap, label); // We want to keep only one open at a time
```
benwaples, line 116: Apologies in advance for the long message! Ill be on earlier tomorrow and perhaps we could call to review this comment together?

Id like to make a wee suggestion because it looks like `collapseNonSelectedSections` mutates context state and theres a nested loop in a nested loop in `collapseNonSelectedSections` (which probably would never be a problem because these lists are super small but O(n^3) isnt ideal).

Instead of adding the `allSectionsMap` prop, what about a prop such as `parentLabels?: string[]`  (with a default value of `[]`) which would equate all labels that appeared before the current one. The idea being that a child would always have its ancestors labels because EntityInfoDataTable is called recursively. So at the top in `EntityInfoDataTableList` the prop would be undefined, but in `EntityInfoDataTable`, we pass `parentLabels={[...parentLabels, label]}` 

after that, opening accordions could look like: 
```tsx
const labelList = [...parentLabels, label];
        setExploreParams({
            expandedPanelSections: labelList,
...
```

toggling: 
```tsx
const currentSectionIsOpen = expandedPanelSections?.includes(label);
        const toggledSections = currentSectionIsOpen ? parentLabels : [...parentLabels, label];

        if (backButtonFlag?.enabled) {
            setExploreParams({ expandedPanelSections: toggledSections });
...
```

and controlling the context state:
```tsx
const toggledSectionsMap = toggledSections.reduce((a: Record<string, boolean>, c) => {
            a[c] = true;
            return a;
        }, {});

        setExpandedSections(toggledSectionsMap);
...
```

cmd/ui/src/views/Explore/GraphItemInformationPanel.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (selectedItemQuery.isError) {
+        return (
+            <EntityInfoPanel
+                sx={infoPaneStyles}
+                selectedNode={{ graphId: selectedItem, id: '', name: 'Unknown', type: 'Unknown' as EntityKinds }}
+            />
+        );
+    }
+
+    if (isEdge(selectedItemQuery.data!)) {
```
benwaples, line 51: nit: if we need the non null assertion here and its required for some functionality, mind adding a comment as to why? Or if it's to get around `selectedItemQuery.data` being falsey, maybe we could add a check before this isEdge() to ensure `selectedItemQuery.data` is what we think it is:

```ts
if (selectedItemQuery.data && isEdge(selectedItemQuery.data))
```

cmd/ui/src/views/Explore/GraphItemInformationPanel.tsx
```diff
@@ -0,0 +1,83 @@
...
+                id: selectedItemQuery.data.target,
+                objectId: selectedItemQuery.data.targetNode.objectId,
+                name: selectedItemQuery.data.targetNode.label,
+                type: selectedItemQuery.data.targetNode.kind,
+            },
+        };
+        return <EdgeInfoPane sx={infoPaneStyles} selectedEdge={selectedEdge} />;
+    }
+
+    if (isNode(selectedItemQuery.data!)) {
```
benwaples, line 72: same thing about non null assertion

cmd/ui/src/views/Explore/GraphViewV2.tsx
```diff
@@ -14,20 +14,17 @@
 //
```
benwaples, line 14: I love how much this file has shrunk!

cmd/ui/src/views/GroupManagement/GroupManagement.tsx
```diff
@@ -61,10 +70,17 @@ const GroupManagement = () => {
...
                 ...openNode,
             };
-            dispatch(searchbarActions.sourceNodeSelected(searchNode));
-            dispatch(setSelectedNode(openNode));
 
-            navigate(ROUTE_EXPLORE);
+            if (backButtonFlagQuery.data?.enabled) {
+                navigate({
+                    pathname: ROUTE_EXPLORE,
+                    search: `?selectedItem=${getGraphNodeByObjectId.data?.id}&searchType=node&primarySearch=${openNode?.id}`,
```
benwaples, line 77: This makes me think we could use some type safety when building these search queries! No need to change anything here, but im going to think about how we could build this search string a little easier

packages/javascript/bh-shared-ui/src/hooks/useExploreSelectedItem.tsx
benwaples: yeah like a batch function with redux! I just want to avoid the user clicking one button but having to hit the back button multiple times to get to the previous state

packages/javascript/bh-shared-ui/src/hooks/useGraphItem.tsx
```diff
@@ -0,0 +1,114 @@
...
+            return apiClient
+                .cypherSearch(`MATCH (n) WHERE n.objectid = "${itemId}" RETURN n LIMIT 1`, undefined, true)
+                .then((res) => {
+                    if (!itemId) {
+                        return undefined;
+                    }
+
+                    const firstElement: any = Object.values(res.data?.data?.nodes)[0];
+
+                    const id = Object.keys(res.data?.data?.nodes)[0];
```
benwaples, line 99: very little nit: looks like we could use `Object.entries` here:

```suggestion
                    const [id, firstElement] = Object.entries(res.data?.data?.nodes[0]
```

packages/javascript/bh-shared-ui/src/utils/parseItemId.ts
```diff
@@ -0,0 +1,41 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+export const parseItemId = (itemId: string) => {
+    // Edge identifiers can be either `rel_<sourceNodeId>_<edgeKind>_<targetNodeId>`...
+    let match = itemId.match(/^(?:rel_)?(\d+)_(.+)_(\d+)$/);
```
benwaples, line 19: I thiiiiink there is a possibility of this being a vulnerability because we could have user input getting dropped directly into the regex, im not sure if we need to sanitize or escape anything. Id like to spend a little time researching this. 

I do really like this approach for tho and think its very handy!!

packages/javascript/bh-shared-ui/src/utils/parseItemId.ts
```diff
@@ -0,0 +1,41 @@
...
+    if (match) {
+        return {
+            itemType: 'edge',
+            cypherQuery: `MATCH p=(s)-[r:${match[2]}]->(t) WHERE ID(s) = ${match[1]} AND ID(t) = ${match[3]}  RETURN p LIMIT 1`,
```
benwaples, line 23: does the API sanitize these queries? Id think so, but also worth taking a moment to research!

cmd/ui/src/views/Explore/EdgeInfo/EdgeInfoContent.tsx
benwaples: Nice!! Im guessing there has been a lot of back and forth in these components so solid chance it was needed at one point!

Thanks for the update!!

cmd/ui/src/views/Explore/EntityInfo/EntityInfoDataTable.tsx
benwaples: spot on! Thank you!!

cmd/ui/src/views/Explore/EntityInfo/EntityInfoDataTable.tsx
specter-flq: Beautiful! Love it, thank you for that, I updated the code based on that logic, let me know!

packages/javascript/bh-shared-ui/src/utils/parseItemId.ts
benwaples: jk I was thinking about this wrong - my concern would be if we used the user supplied value in the regex, but not at all the case.

cmd/ui/src/views/Explore/EntityInfo/EntityInfoDataTable.tsx
benwaples: Absolutely nailed that! Thank you for the changes!

cmd/ui/src/views/GroupManagement/GroupManagement.tsx
benwaples: Can we add `exploreSearchTab=node` here as well please?

benwaples: It works as is because node tab is the default, but I think having it be explicit here would be good if the default tab ever changed!

packages/javascript/bh-shared-ui/src/hooks/useExploreSelectedItem.tsx
elikmiller: sounds like a big value add!

packages/javascript/bh-shared-ui/src/hooks/useGraphItem.tsx
elikmiller: Yeah, I think we've already briefly chatted about needing to use Object IDs as the "selectedItem" key and I'd expect we could clean up the logic here to account for that change.

Would you like me to make the change now or defer it?

cmd/ui/src/views/Explore/GraphItemInformationPanel.tsx
elikmiller: The react-query v3 docs have this to say:

"For most queries, it's usually sufficient to check for the isLoading state, then the isError state, then finally, assume that the data is available and render the successful state."

However, when using typescript, it is not inferred that the `data` property is not `undefined` even after checking the `isLoading` and `isError` properties. I think this is an improvement that comes with a newer version of react-query.


Would you prefer I add an in-code comment? I believe we use a similar pattern elsewhere in the app.

benwaples: ahh yeah okay thats ringing a bell. IIRC we may have ran into an issue where a query was in an idle state because it was disabled and so it bypassed the other checks. However, the underlying query here doesnt have a disabled flag set at all so we should be good. 

Thanks for taking a look, feel free to leave this as is!

packages/javascript/bh-shared-ui/src/hooks/useGraphItem.tsx
benwaples: defer if we are planning on changing this area anyway!

benwaples (approved):

PR #1258, opened 2025-03-20
BED-5266 feat: selector engine

mistahj67:
## Description

- Adds support for selecting and tagging nodes 
- Fixed DAWGs bug around deleting node kinds

## Motivation and Context

This PR addresses: BED-5266

## How Has This Been Tested?

Locally

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/database/migration/migrations/v7.4.0.sql
+++ b/cmd/api/src/database/migration/migrations/v7.4.0.sql
@@ -1,3 +1,19 @@
+-- Copyright 2025 Specter Ops, Inc.
+--
+-- Licensed under the Apache License, Version 2.0
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- SPDX-License-Identifier: Apache-2.0
+
 -- Add asset_group_tag_selector_nodes table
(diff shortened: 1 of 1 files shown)


cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -0,0 +1,567 @@
...
+		}
+	}
+
+	return nil
+}
+
+func fetchChildNodes(ctx context.Context, tx traversal.Traversal, node *graph.Node, result *graph.ThreadSafeNodeSet, resultSrcSet *sync.Map) error {
+	var pattern traversal.PatternContinuation
+
+	switch {
```
urangel, line 213: Why do some of these cases use the passed in node as the start node and some as the end? Should the cases where the passed in node is the end in the path pattern still be considered under the fetch child nodes umbrella?

It seems like there can be four different scenarios:
- node is end and inbound path pattern is used
- node is start and inbound path pattern is used
- node is end and outbound path pattern is used
- node is start and outbound path pattern is used

Maybe I am thinking of fetching child nodes too literally?

urangel (approved):
I left a couple of questions but don't want to block on them. This is great work. Thank you, @mistahj67

cmd/api/src/model/assetgrouptags.go
```diff
@@ -157,3 +210,15 @@ func (s AssetGroupTagSelector) ValidFilters() map[string][]FilterOperator {
...
 	Selectors AssetGroupTagSelectors `json:"selectors"`
 }
+
+type AssetGroupSelectorNode struct {
+	SelectorId  int                          `json:"selector_id"`
+	NodeId      graph.ID                     `json:"node_id"`
+	Certified   AssetGroupCertification      `json:"certified"`
+	CertifiedBy null.String                  `json:"certified_by"`
+	Source      AssetGroupSelectorNodeSource `json:"source"`
+}
```
urangel, line 220: Did we want to add the time related fields from the `Basic` struct to this model?

PR #1264, opened 2025-03-21
[BED-4716] Collector Downloads page refactor

definitelynotagoblin:
## Description
Desync *hound collector downloads from embedded binaries (only enterprise binaries tho)

## Motivation and Context

This PR addresses: https://specterops.atlassian.net/browse/BED-4716

## How Has This Been Tested?
Localhost

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
```diff
@@ -14,90 +14,113 @@
...
+
+    return (
+        <RiskBadge type='labeled' label={label} outlined={false} color={color} title={label} className={darkColor} />
+    );
 };
 
+// Sometimes you just need a dashed line connecting two underlined text elements together
+// Sometimes you need that dashed line to be slightly offset
+const DashedFlexConnector: React.FC<{ offset?: boolean }> = ({ offset }) => {
+    return <div className={cn("flex-1 min-w-0 decoration-dashed decoration-neutral-light-5 underline whitespace-nowrap text-clip overflow-hidden pointer-events-none select-none aria-hidden:", { 'underline-offset-4': offset })}>{"\u00A0".repeat(500)}</div>;
```
definitelynotagoblin, line 123: I need to find a solution besides this offset prop.
The offset only needs to change for instances of CollectorCard as they exist within the Older Versions drop down Accordion.
 I suspect the parent is applying some styling to its children that is messing this up but I need to dig into it

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
```diff
@@ -14,90 +14,113 @@
...
+}
+
+const CollectorLabel: React.FC<CollectorLabelProps> = ({ label, isPrerelease = false }) => {
+    // RiskBadge isn't set up out the gate to accomodate dark mode
+    // so we improvise
+    const color = isPrerelease ? 'rgba(243, 96, 54, 0.25)' : 'rgba(51, 49, 143, 0.15)';
+    const darkColor = isPrerelease ? 'dark:bg-[#02C577]' : 'dark:bg-[#33318F]';
+
+    return (
+        <RiskBadge type='labeled' label={label} outlined={false} color={color} title={label} className={darkColor} />
```
definitelynotagoblin, line 116: It works, but it is kinda like solving a problem with a surgical sledgehammer

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
```diff
@@ -19,7 +19,8 @@ import { CollectorCardList, DocumentationLinks, PageWithTitle, apiClient } from
 import fileDownload from 'js-file-download';
 import { useDispatch } from 'react-redux';
 import { addSnackbar } from 'src/ducks/global/actions';
-import { CollectorType, useGetCollectorsByType } from 'src/hooks/useCollectors';
+import { useGetCollectorsByType } from 'src/hooks/useCollectors';
+import { CollectorType } from 'js-client-library/src/types';
```
elikmiller, line 23: same here!

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
```diff
@@ -14,90 +14,113 @@
...
+}
+
+interface CollectorCardProps extends React.HTMLAttributes<HTMLDivElement> {
+    collectorType: CollectorType;
+    version: string;
+    downloadArtifacts: CollectorDownloadFile[];
+    timestamp?: number;
+    label?: LabelType;
+    isPrerelease?: boolean;
+    needsUnderlineOffset?: boolean;
```
elikmiller, line 46: This doesn't read like an expected prop for a `CollectorCard` - what is the intent here?

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
```diff
@@ -143,18 +146,22 @@ const DownloadCollectors = () => {
...
-                                .sort((a, b) => b.version.localeCompare(a.version))}
-                        />
+                                    noLabels
+                                    collectors={sharpHoundCollectorsQuery.data!.data.versions.map((collector) => ({
+                                        collectorType: 'sharphound' as const,
+                                        version: collector.version,
+                                        downloadArtifacts: [
+                                            {
+                                                os: 'windows',
+                                                arch: 'x86',
```
elikmiller, line 156: Are we expecting these values to be hard coded and not from the server?

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
definitelynotagoblin: Accordion element breaks styling on the dashed-underline-bridge of the CollectorCard.
I don't like this prop and I'd like to remove it, but I have to dig in to find what styling CollectorCard is inheriting from Accordion that seems to be causing the issue that necessitates this hack.

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
definitelynotagoblin: From the current FOSS manifest endpoint we don't get these values, so I'm hard-coding them here.  I don't think we can change the contract on this endpoint, but maybe there's somewhere else I should hard-code these values?

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
```diff
@@ -14,90 +14,113 @@
...
-}
+export type LabelType = 'latest' | 'prerelease';
 
-const COLLECTOR_TYPE = {
-    sharphound: 'SharpHound',
-    azurehound: 'AzureHound',
+export const COLLECTOR_TYPE_LABEL: { [key in CollectorType]: string } = {
+    sharphound_enterprise: 'Sharphound Enterprise',
+    sharphound: 'Sharphound Community',
+    azurehound: 'Azurehound',
```
urangel, line 28: nit: We should capitalize the 'h' in 'Hound' for these values.

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
urangel: It is likely easier and more appropriate to make a different component than try to reuse the RiskBadge here.

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
definitelynotagoblin: Yeah I'd just asked about that since the design sheet has these H's lowercased, but Business is in agreement - SharpHound

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
```diff
@@ -14,90 +14,113 @@
...
+    const darkColor = isPrerelease ? 'dark:bg-[#02C577]' : 'dark:bg-[#33318F]';
+
+    return (
+        <RiskBadge type='labeled' label={label} outlined={false} color={color} title={label} className={darkColor} />
+    );
 };
 
+// Sometimes you just need a dashed line connecting two underlined text elements together
+// Sometimes you need that dashed line to be slightly offset
+const DashedFlexConnector: React.FC<{ offset?: boolean }> = ({ offset }) => {
```
definitelynotagoblin, line 122: Ah I left the offset prop in there, will remove with next commit

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
maffkipp: In my testing, this appears to be extending off the edge of my browser in both chrome and firefox -- assuming to the full width of those 500 characters.

packages/javascript/bh-shared-ui/src/components/CollectorCardList/CollectorCardList.tsx
```diff
@@ -14,41 +14,121 @@
...
+const CollectorCardList: React.FC<CollectorCardListProps> = ({ collectors, noLabels = false }) => {
     const theme = useTheme();
 
+    // Few enough collectors that this isn't worth memoizing
+    // And the only stateful changes that should cause rerender
+    // are collectors and theme
+    const sortedCollectors = collectors.toSorted((a, b) =>
+        b.version.toLowerCase().localeCompare(a.version.toLowerCase())
+    );
+    const latestStable = sortedCollectors.filter((c) => !c.isPrerelease)[0];
```
maffkipp, line 60: You can use [Array.prototype.find()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) here to grab the first matching occurrence instead of needing to add the `[0]` at the end

maffkipp (changes requested):

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
```diff
@@ -32,13 +33,13 @@ const DownloadCollectors = () => {
     const downloadCollector = (collectorType: CollectorType, version: string) => {
         apiClient
             .downloadCollector(collectorType, version)
```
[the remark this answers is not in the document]
definitelynotagoblin: I don't think my linter is configured correctly.  Let me know if you're still seeing this, and I'll try to get the package import types to lint correctly.

packages/javascript/bh-shared-ui/src/components/CollectorCardList/CollectorCardList.tsx
```diff
@@ -14,41 +14,148 @@
...
+    onClickDownloadChecksum: () => void;
+}
+
+interface CollectorCardProps {
+    version: string;
+    downloadArtifacts: CollectorDownloadFile[];
+    timestamp?: number;
+    label?: LabelType;
+    isPrerelease?: boolean;
+}
```
elikmiller, line 43: I think these should be defined elsewhere - and actually I think some of these are duplicated already in other components. Do we need these types/consts defined here in `CollectorCardList`?

definitelynotagoblin: Hmm maybe not necessarily.  These aren't a 1:1 translation of the CollectorCard props, but maybe I can have these be `Extension`s of those prop interfaces 🤔 

The `OLLECTOR_SHORT_LABEL` const isn't being used elsewhere, but maybe there's value it putting it somewhere more generally accessible

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
definitelynotagoblin: Originally we'd intended for all collectors foss and enterprise to come through the new manifest api endpoint, but changes down the pipeline brought us back to using the old endpoint for the foss collectors.

packages/javascript/bh-shared-ui/src/components/CollectorCardList/CollectorCardList.tsx
definitelynotagoblin: Been consolidating these interfaces down, but am I barking up the wrong tree with this pattern:
```
interface CollectorCardListProps {
    collectorType: CollectorType;
    collectors: (Omit<CollectorCardProps, 'collectorType' & 'downloadArtifacts'> & {
        downloadArtifacts: Omit<CollectorDownloadFile, 'displayName'>;
    })[];
    noLabels?: boolean;
}
```

Where `CollectorCardProps` and `CollectorDownloadFile` are imported?
I'm trying to minimize the amount of unnecessary exposure from one component to the next up its composition stack, but I'm afraid I might be adding more complexity than I'm removing.

definitelynotagoblin: Rolling with it for now but do let me know if I'm running in the wrong direction.

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
definitelynotagoblin: We've moved to leaving BHCE DownloadCollectors as it had been with the components as they exist now, and using the new manifest endpoint for BHE, so this should get cleaned up.

packages/javascript/bh-shared-ui/src/components/CollectorCardList/CollectorCardList.tsx
definitelynotagoblin: Cross-linking
https://github.com/SpecterOps/bloodhound-enterprise/pull/204/files#r2013033650

packages/javascript/js-client-library/src/types.ts
```diff
@@ -264,6 +264,9 @@ export interface LoginResponse {
     };
 }
 
+export type CollectorType = 'sharphound_enterprise' | 'sharphound' | 'azurehound';
+export type CommunityCollectorType = Exclude<CollectorType, 'sharphound_enterprise'>;
```
definitelynotagoblin, line 268: I'm actually going to "flip" these definitions so `CollectorType` becomes `CommunityCollectorType | 'sharphound_enterprise'`

packages/javascript/bh-shared-ui/src/components/CollectorCard/CollectorCard.tsx
```diff
@@ -18,18 +18,19 @@ import { Button } from '@bloodhoundenterprise/doodleui';
...
-    onClickDownload?: (collectorType: 'sharphound' | 'azurehound', version: string) => void;
-    onClickDownloadChecksum?: (collectorType: 'sharphound' | 'azurehound', version: string) => void;
+    onClickDownload?: (collectorType: CommunityCollectorType, version: string) => void;
+    onClickDownloadChecksum?: (collectorType: CommunityCollectorType, version: string) => void;
     isLatest?: boolean;
     isDeprecated?: boolean;
 }
 
-const COLLECTOR_TYPE = {
+const COLLECTOR_TYPE: Record<CommunityCollectorType, string> = {
```
urangel, line 33: Nice refactor ^_^

packages/javascript/js-client-library/src/types.ts
```diff
@@ -360,6 +364,36 @@ export interface GetCollectorsResponse {
     };
 }
 
+export interface GetCommunityCollectorsResponse {
+    data: Record<CommunityCollectorType, CollectorManifest[]>;
+}
+
+export interface GetEnterpriseCollectorsResponse {
+    data: Record<EnterpriseCollectorType, CollectorManifest[]>;
+}
```
urangel, line 373: These would be good to have in the adjacent responses.ts file! I might also update the type declaration to use the `BasicResponse` type for consistency:

```suggestion
export type GetCommunityCollectorsResponse = BasicResponse<Record<CommunityCollectorType, CollectorManifest[]>>;

export type GetEnterpriseCollectorsResponse = BasicResponse<Record<EnterpriseCollectorType, CollectorManifest[]>>;
```

cmd/ui/src/views/DownloadCollectors/DownloadCollectors.tsx
urangel: This is no longer relevant it seems so am resolving!

packages/javascript/bh-shared-ui/src/components/CollectorCardList/CollectorCardList.tsx
urangel: Moving the conversation over to linked thread and resolving this!

urangel (approved):

maffkipp (approved):

PR #1307, opened 2025-04-03
Bed-5621 - feat: add AGT /member/counts endpoint

mistahj67:
## Description

- Added `/api/v2/asset-group-tags/{%s}/members/counts`
- Refactored primary kind utility for O(1) lookups
- Added `CountNodesByKind` helper method
- Bring malformed id checks in line with status code 404

## Motivation and Context

This PR addresses: Bed-5621

## How Has This Been Tested?
Locally and with unit test

## Screenshots (optional):
image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/v2/assetgrouptags.go
+++ b/cmd/api/src/api/v2/assetgrouptags.go
@@ -35,12 +35,6 @@ import (
 	"github.com/specterops/bloodhound/src/utils/validation"
 )
 
-const (
-	ErrInvalidAssetGroupTagId         = "invalid asset group tag id specified in url"
-	ErrInvalidAssetGroupTagSelectorId = "invalid asset group tag selector id specified in url"
-	ErrInvalidSelectorType            = "invalid selector type"
-)
-
 // Checks that the selector seeds are valid.
 func validateSelectorSeeds(graph queries.Graph, seeds []model.SelectorSeed) error {
 	// all seeds must be of the same type
@@ -71,7 +65,7 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 	defer measure.ContextMeasure(request.Context(), slog.LevelDebug, "Asset Group Tag Selector Create")()
 
 	if assetTagId, err := strconv.Atoi(assetTagIdStr); err != nil {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupTagId, request), response)
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, api.ErrorResponseDetailsIDMalformed, request), response)
 	} else if _, err := s.DB.GetAssetGroupTag(request.Context(), assetTagId); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else if err := json.NewDecoder(request.Body).Decode(&sel); err != nil {
@@ -102,11 +96,11 @@ func (s *Resources) UpdateAssetGroupTagSelector(response http.ResponseWriter, re
 		slog.Error("Unable to get user from auth context")
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, "unknown user", request), response)
 	} else if assetTagId, err := strconv.Atoi(assetTagIdStr); err != nil {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupTagId, request), response)
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, api.ErrorResponseDetailsIDMalformed, request), response)
 	} else if _, err := s.DB.GetAssetGroupTag(request.Context(), assetTagId); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else if selectorId, err := strconv.Atoi(rawSelectorID); err != nil {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupTagSelectorId, request), response)
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, api.ErrorResponseDetailsIDMalformed, request), response)
 	} else if selector, err := s.DB.GetAssetGroupTagSelectorBySelectorId(request.Context(), selectorId); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else if err := json.NewDecoder(request.Body).Decode(&selUpdateReq); err != nil {
@@ -214,7 +208,7 @@ func (s *Resources) GetAssetGroupTagSelectors(response http.ResponseWriter, requ
 		defer measure.ContextMeasure(request.Context(), slog.LevelDebug, "Asset Group Label Get Selector")()
 
 		if assetGroupTagID, err := strconv.Atoi(assetTagIdStr); err != nil {
-			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, ErrInvalidAssetGroupTagId, request), response)
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, api.ErrorResponseDetailsIDMalformed, request), response)
 		} else if selectorSqlFilter, err := selectorQueryFilter.BuildSQLFilter(); err != nil {
 			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "error building SQL for filter", request), response)
 		} else if selectorSeedSqlFilter, err := selectorSeedsQueryFilter.BuildSQLFilter(); err != nil {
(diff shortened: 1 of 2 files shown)


packages/go/analysis/analysis.go
```diff
@@ -57,6 +56,7 @@ func NewCompositionCounter() CompositionCounter {
 var (
 	metaKind       = graph.StringKind("Meta")
 	metaDetailKind = graph.StringKind("MetaDetail")
+	ValidKinds     = buildValidKinds()
```
kpowderly, line 59: probably can make var `validKinds` private, doesn't look like needed to be exported anywhere

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -158,3 +159,27 @@ func (s *Resources) GetAssetGroupTag(response http.ResponseWriter, request *http
...
+	TotalCount int            `json:"total_count"`
+	Counts     map[string]int `json:"counts"`
+}
+
+func (s *Resources) GetAssetGroupTagMemberCountsByKind(response http.ResponseWriter, request *http.Request) {
+	if tagId, err := strconv.ParseInt(mux.Vars(request)[api.URIPathVariableAssetGroupTagID], 10, 32); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
+	} else if tag, err := s.DB.GetAssetGroupTag(request.Context(), int(tagId)); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else if nodes, err := s.GraphQuery.GetNodesByKind(request.Context(), tag.ToKind()); err != nil {
```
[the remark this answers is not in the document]
computator: I could use that here too
https://github.com/SpecterOps/BloodHound/pull/1213/files#diff-0ee72f9b2f75848106e4b4eb037630493be520d2728305e1ca5b2c0adf9107c0R93-R94

cmd/api/src/queries/graph.go
```diff
@@ -664,6 +667,37 @@ func (s *GraphQuery) GetEntityCountResults(ctx context.Context, node *graph.Node
 	return results
 }
 
+func (s *GraphQuery) CountNodesByKind(ctx context.Context, kinds ...graph.Kind) (int64, error) {
+	var (
+		numNodes int64
+	)
```
computator, line 673: Since this is the only variable, would it be preferable to have it as a single var line?
```suggestion
	var numNodes int64
```

cmd/api/src/queries/graph.go
```diff
@@ -664,6 +667,37 @@ func (s *GraphQuery) GetEntityCountResults(ctx context.Context, node *graph.Node
...
+}
+
+func (s *GraphQuery) GetPrimaryNodeKindCounts(ctx context.Context, kinds ...graph.Kind) (map[string]int, error) {
+	results := map[string]int{}
+
+	return results, s.Graph.ReadTransaction(ctx, func(tx graph.Transaction) error {
+		return tx.Nodes().Filter(query.KindIn(query.Node(), kinds...)).FetchKinds(func(cursor graph.Cursor[graph.KindsResult]) error {
+			for next := range cursor.Chan() {
+				primaryKindStr := graphschema.PrimaryNodeKind(next.Kinds).String()
+				results[primaryKindStr] += 1
```
computator, line 693: Are these lines separate for clarity? Otherwise they could be combined and the assignment could be skipped.

cmd/api/src/queries/graph.go
```diff
@@ -664,6 +667,37 @@ func (s *GraphQuery) GetEntityCountResults(ctx context.Context, node *graph.Node
...
+
+	if err := s.Graph.ReadTransaction(ctx, func(tx graph.Transaction) error {
+		var err error
+		numNodes, err = tx.Nodes().Filter(query.KindIn(query.Node(), kinds...)).Count()
+		return err
+	}); err != nil {
+		return 0, err
+	}
+
+	return numNodes, nil
```
computator, line 683: It seems like this return could be combined similar to the one in the function below? Something like:
```go
return numNodes, s.Graph.ReadTransaction(...
```

cmd/api/src/queries/graph_integration_test.go
```diff
@@ -227,6 +227,23 @@ func TestGetEntityResults_QueryShorterThanSlowQueryThreshold(t *testing.T) {
...
+	testContext.SetupActiveDirectory()
+	testContext.DatabaseTest(func(harness integration.HarnessDetails, db graph.Database) {
+		graphQuery := queries.GraphQuery{
+			Graph: db,
+		}
+
+		results, err := graphQuery.GetPrimaryNodeKindCounts(context.Background(), ad.Entity, azure.Entity)
+		require.Nil(t, err)
+
+		// While this is a very thin test, any more specificity would require constant updates each time the harness added new kind
```
computator, line 242: You should be able to check if the nodes at least have the `Entity` type you queried for though, correct? And other kinds are also constants so perhaps I'm not seeing how that would require updates 🤔

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -158,3 +158,28 @@ func (s *Resources) GetAssetGroupTag(response http.ResponseWriter, request *http
 		api.WriteBasicResponse(request.Context(), getAssetGroupTagResponse{Tag: assetGroupTag}, http.StatusOK, response)
 	}
 }
+
+type getAssetGroupTagMemberCountsResponse struct {
```
computator, line 162: This appears to be different than the `ListAssetGroupMemberCountsResponse` in `assetgrouptags_test.go`?

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -158,3 +158,28 @@ func (s *Resources) GetAssetGroupTag(response http.ResponseWriter, request *http
...
+	TotalCount int            `json:"total_count"`
+	Counts     map[string]int `json:"counts"`
+}
+
+func (s *Resources) GetAssetGroupTagMemberCountsByKind(response http.ResponseWriter, request *http.Request) {
+	if tagId, err := strconv.ParseInt(mux.Vars(request)[api.URIPathVariableAssetGroupTagID], 10, 32); err != nil {
```
computator, line 168: This seems easier? No additional parameters, and it returns an int so you don't have to cast it in the next line. Also TIL that `Atoi` contains optimizations for common cases that make it faster than `ParseInt`.
```suggestion
	if tagId, err := strconv.Atoi(mux.Vars(request)[api.URIPathVariableAssetGroupTagID]); err != nil {
```

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -158,3 +158,28 @@ func (s *Resources) GetAssetGroupTag(response http.ResponseWriter, request *http
...
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
```
computator, line 169: This should be a 404. It's a path parameter, so if it's invalid that just means that URL doesn't happen to exist.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -158,3 +158,28 @@ func (s *Resources) GetAssetGroupTag(response http.ResponseWriter, request *http
...
+	} else if tag, err := s.DB.GetAssetGroupTag(request.Context(), int(tagId)); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else if primaryNodeKindsCounts, err := s.GraphQuery.GetPrimaryNodeKindCounts(request.Context(), tag.ToKind()); err != nil {
+		api.HandleDatabaseError(request, response, err)
```
computator, line 173: I'm not sure if GraphQueries work the same as regular DB queries, but if so we would want to check for ErrNotFound so that we can still return a list of 0 counts when there's no results instead of having `HandleDatabaseError` kick out a 404.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -158,3 +158,28 @@ func (s *Resources) GetAssetGroupTag(response http.ResponseWriter, request *http
...
+	} else {
+		data := getAssetGroupTagMemberCountsResponse{
+			Counts: primaryNodeKindsCounts,
+		}
+
+		for _, count := range primaryNodeKindsCounts {
+			data.TotalCount += count
```
computator, line 180: Wouldn't this inflate the `TotalCount` if any of the node kinds included nodes that apply to more than one category? I'm not sure if that's possible or not in this case.

cmd/api/src/api/v2/assetgrouptags.go
kpowderly: I think it's a little inconsistent how this is handled in the code so I wanted to respond based on my own ignorance... but as someone that doesn't know much about the codebase, I actually think a bad request makes more sense in this case because it indicates that nothing can be done with the request that was initiated. 

A 404 would indicate that we tried to do something with a bad path parameter and that the user should therefore _continue trying_ in the future because that path parameter might eventually exist wherever we are looking for it.

Hopefully that makes sense... but I'm not dying on any hill over it ;)

computator: Yes I understand that logic. It makes sense. That said, I don't happen to agree in this case. Yes someone can re-request a URL that returns a 404 because theoretically it may exist in the future, but by that argument a totally random made up URL like `/api/v2/blah-asdf` should also return a 400. Obviously that's incorrect, seeing as we don't want any random URL to say "you requested this incorrectly" but rather that there's nothing there (404). Also, for these new endpoints we (correctly) return a 404 when an id doesn't exist to say there's nothing there (f/e `/endpoint/1234`). From a URL perspective that's not any different than any other random string being in that segment of the URL. With `/endpoint/asdf` the segment with `asdf` also does not exist, the same as the string `1234` doesn't exist, so why should one return a 404 and the other 400? 🤷‍♂️

The other reason is that a 400 primarily refers to an invalid request body. Like POSTing an invalid JSON blob or something like that: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400

Thoughts? 🙂

kpowderly: That's fair! I think you're right in this case -- Ideally, it would be nice for some of these endpoints to do some sort of input validation because a missing parameter in a path is very different than a param that was fat-fingered. It might help make some of the responses a little more black/white vs the grey zone. 

Appreciate the discussion ❤️

zinic (approved):
Changes pass muster. Pull.

computator (approved):

PR #1310, opened 2025-04-03
BED-5673: ParseDomainTrusts float Parsing For TrustedAttributes

mvlipka:
## Description

When ingesting TrustAttributes for a Domain, the decoder will sometimes assign a `float64` as the type for TrustAttributes due to the usage of `any`
This adds the new type casing for `float32` and `float64` to the existing type switch and allows for the value to be set to `nil` when a value could not be parsed, rather than `0` that it currently displays in the Entity Info Panel

Also cleans up some string usage and unit tests

## Motivation and Context

This PR addresses: BED-5673

## How Has This Been Tested?

I ingested the data attached to the ticket, searched for the `EXTERNAL.LOCAL` node and inspected the `TrustedBy` edges to ensure the `Trusted Attributes` value is no longer appearing and that the API is returning `null` for `trustattributes`
See ticket comments for screenshots

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/go/ein/ad.go
```diff
@@ -700,12 +720,12 @@ func ParseDomainTrusts(domain Domain) ParsedDomainTrustData {
...
-						"tgtdelegationenabled": trust.TGTDelegationEnabled,
-						"trustattributes":      finalTrustAttributes,
-						"trusttype":            trust.TrustType,
-						"transitive":           trust.IsTransitive},
+						ad.IsACL.String():                false,
+						ad.SidFiltering.String():         trust.SidFilteringEnabled,
+						ad.TGTDelegationEnabled.String(): trust.TGTDelegationEnabled,
+						ad.TrustAttributes.String():      finalTrustAttributes,
+						ad.TrustType.String():            trust.TrustType,
+						"transitive":                     trust.IsTransitive},
```
mvlipka, line 728: This type doesn't exist as an ad constant

packages/go/ein/ad_test.go
```diff
@@ -74,64 +75,137 @@ func TestConvertObjectToNode_DomainInvalidProperties(t *testing.T) {
...
+			},
+		}
+
+		result := ein.ParseDomainTrusts(domainObject)
+		require.Len(t, result.TrustRelationships, 1)
+
+		rel := result.TrustRelationships[0]
+		assert.Contains(t, rel.RelProps, ad.TrustAttributes.String())
+		assert.Equal(t, nil, rel.RelProps[ad.TrustAttributes.String()])
+	})
```
mvlipka, line 208: These are the new tests for floats and invalid parsing

packages/go/ein/ad_test.go
```diff
@@ -22,18 +22,19 @@ import (
 	"github.com/specterops/bloodhound/ein"
```
mvlipka, line 22: Most of the changes here are just replacing strings with ad constants and cleaning up tests to be more concise
Actual testing of the new changes are added at the bottom of the diff

packages/go/ein/ad.go
```diff
@@ -662,23 +662,43 @@ func ParseGpLinks(links []GPLink, itemIdentifier string, itemType graph.Kind) []
...
-				finalTrustAttributes = i
+				convertedTrustAttributes = i
 			}
 		case int:
-			finalTrustAttributes = converted
+			convertedTrustAttributes = converted
+		case float32:
+			convertedTrustAttributes = int(converted)
+		case float64:
+			convertedTrustAttributes = int(converted)
```
mvlipka, line 692: There is a potential chance for this to overflow, but we'd have to have over 2.4 billion trusts for a domain:
https://learn.microsoft.com/en-us/windows-server/identity/ad-ds/plan/active-directory-domain-services-maximum-limits#trust-limitations

packages/go/ein/ad.go
rvazarkar: Can we add it to the schema?

mvlipka: Not a problem, added!

packages/go/ein/ad.go
```diff
@@ -662,23 +662,43 @@ func ParseGpLinks(links []GPLink, itemIdentifier string, itemType graph.Kind) []
...
+			invalidTrustAttribute    bool
+			finalTrustAttributes     any
+		)
+
+		// The type of the TrustAttributes is decided during decoding due to the `any` type usage
+		// We need to switch on the type and convert it to an int, or nil when there was an error parsing the value
 		switch converted := trust.TrustAttributes.(type) {
 		case string:
 			if i, err := strconv.Atoi(converted); err != nil {
 				slog.Error(fmt.Sprintf("Error converting trust attributes %s to int", converted))
```
zinic, line 682: I question how useful log outputs are here. I see sporadic outputs like this in the fleet application logs all the time but we've done nothing about them nor do they feel particularly actionable. No call for action but this section of the code can be pretty noisy and we don't really do much with the log output.

packages/go/ein/ad.go
zinic: Type switches like these are always ripe for missed `oops, I sent in an int16 this time and didn't catch it`. The default case has an error output that could benefit from logging the type encountered with `%T` and the converted reference:

```
fmt.Sprintf("expected a type X but got %T instead", converted)
```

zinic (approved):
Tow minor nits to consider regarding logging in the area. I'm fine with this changeset as-is but if you take a swing and cleaning up the logging tag me again.

packages/go/ein/ad.go
mvlipka: May be worth converting to a warning rather than an error since we continue processing when this occurs

packages/go/ein/ad.go
mvlipka: In its current state, the log is pretty similar:
`"level":"ERROR","message":"Error converting trust attributes %!s(float64=32) to int"`

I'll give it a little cleanup

packages/go/ein/ad.go
StephenHinck: I would agree as this has frequently caused confusion for folks on the community as well to understand whether, when these errors arise were they actually errors, etc. I would make sure the log levels are accurate, and are indicating actionable issues. Thank you for flagging!

zinic (approved):

PR #1323, opened 2025-04-04
BED-5266: chore: plumb tier enabled FF

mistahj67:
## Description

- Plumb a tiering toggle through analysis and various endpoints to rely on either tagging approach
- Cleanup tagging so it's more isolated
- General cleanup

## Motivation and Context

This PR addresses: BED-5266

## How Has This Been Tested?

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/daemons/datapipe/agt.go
+++ b/cmd/api/src/daemons/datapipe/agt.go
@@ -667,49 +667,49 @@ func clearAssetGroupTags(ctx context.Context, db database.Database, graphDb grap
 	return nil
 }
 
-func TagAssetGroupsAndTierZero(ctx context.Context, db database.Database, graphDb graph.Database, additionalFiltersToClear ...graph.Criteria) error {
+func TagAssetGroupsAndTierZero(ctx context.Context, db database.Database, graphDb graph.Database, additionalFiltersToClear ...graph.Criteria) []error {
+	var errors []error
+
 	if appcfg.GetTieringEnabled(ctx, db) {
 		// Tiering enabled, we don't want system tags present
 		if err := clearSystemTags(ctx, graphDb, additionalFiltersToClear...); err != nil {
 			slog.Error(fmt.Sprintf("AGT: wiping old system tags: %v", err))
-			return err
+			errors = append(errors, err)
 		}
 		if err := selectAssetGroupNodes(ctx, db, graphDb); err != nil {
 			slog.Error(fmt.Sprintf("AGT: selecting failed: %v", err))
-			return err
+			errors = append(errors, err)
 		}
 
 		if err := tagAssetGroupNodes(ctx, db, graphDb); err != nil {
 			slog.Error(fmt.Sprintf("AGT: tagging failed: %v", err))
-			return err
+			errors = append(errors, err)
 		}
 	} else {
 		// Tiering disabled, we don't want nodes with tagged kinds
 		if err := clearAssetGroupTags(ctx, db, graphDb); err != nil {
 			slog.Error(fmt.Sprintf("AGT: clearing tags failed: %v", err))
-			return err
+			errors = append(errors, err)
 		}
 
 		if err := clearSystemTags(ctx, graphDb, additionalFiltersToClear...); err != nil {
 			slog.Error(fmt.Sprintf("Failed clearing system tags: %v", err))
-			return err
-		}
-
-		if err := updateAssetGroupIsolationTags(ctx, db, graphDb); err != nil {
+			errors = append(errors, err)
+		} else if err := updateAssetGroupIsolationTags(ctx, db, graphDb); err != nil {
 			slog.Error(fmt.Sprintf("Failed updating asset group isolation tags: %v", err))
-			return err
+			errors = append(errors, err)
 		}
 
 		if err := tagActiveDirectoryTierZero(ctx, db, graphDb); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -560,3 +561,75 @@ func TagAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph
...
+	return nil
+}
+
+func TagAssetGroupsAndTierZero(ctx context.Context, db database.Database, graphDb graph.Database, additionalFiltersToClear ...graph.Criteria) error {
+	if appcfg.GetTieringEnabled(ctx, db) {
+		// Tiering enabled, we don't want system tags present
+		if err := clearSystemTags(ctx, graphDb, additionalFiltersToClear...); err != nil {
+			slog.Error(fmt.Sprintf("AGT: wiping old system tags: %v", err))
+			return err
+		}
```
urangel, line 596: Have we considered what if any downstream effects this might have? For instance, many of our prebuilt queries that are available in the UI depend on the `system_tags` property and look for the presence of the value `admin_tier_0` within that property string. "Locations of Tier Zero / High Value objects" is an example of one such query. Is it understood that many of these queries will not function as expected or do we need to take some time to figure out what all may be impacted by the absence of the system tags property and update it accordingly?

cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -560,3 +561,75 @@ func TagAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph
...
+
+		if err := tagActiveDirectoryTierZero(ctx, db, graphDb); err != nil {
+			slog.Error(fmt.Sprintf("Failed tagging Active Directory attack path roots: %v", err))
+			return err
+		}
+
+		if err := parallelTagAzureTierZero(ctx, graphDb); err != nil {
+			slog.Error(fmt.Sprintf("Failed tagging Azure attack path roots: %v", err))
+			return err
+		}
```
urangel, line 631: From what I understand, these processes will still happen when the tier flag is enabled except that they will be driven by selectors that mirror the criteria but live in pg, is that correct? I wonder if there is a way we can make that more explicit. Maybe it just feels more opaque to me because I am used to how things work today and I might not think to look in pg for that logic.

packages/go/dawgs/graph/node.go
```diff
@@ -58,6 +59,17 @@ type Node struct {
 	Properties   *Properties `json:"properties"`
 }
 
+// TODO Cleanup after Tiering GA
+func (s *Node) IsTierZero() bool {
```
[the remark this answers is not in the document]
urangel: straight to jail

jk lol

That's unfortunate that we have to use the strings instead of the constants but the TODO will hopefully suffice for cleaning this up in the future.

cmd/api/src/model/appcfg/flag.go
```diff
@@ -87,3 +88,13 @@ type GetFlagByKeyer interface {
 	// GetFlagByKey attempts to fetch a FeatureFlag by its key.
 	GetFlagByKey(context.Context, string) (FeatureFlag, error)
 }
+
+// TODO Cleanup after Tiering GA
+func GetTieringEnabled(ctx context.Context, service GetFlagByKeyer) bool {
```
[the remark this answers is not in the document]
urangel: This is nice; I like it 🙌

cmd/api/src/daemons/datapipe/agt.go
urangel: I've made a ticket to cover the UI portion of this concern and so am resolving this conversation

cmd/api/src/daemons/datapipe/agi.go
```diff
@@ -38,17 +38,82 @@ import (
...
+								node.Properties.Set(tagPropertyStr, assetGroup.Tag)
+							} else {
+								return err
+							}
+						} else {
+							node.Properties.Set(tagPropertyStr, tags+" "+assetGroup.Tag)
+						}
+
+						if err := tx.UpdateNode(node); err != nil {
+							return err
```
urangel, line 80: Maybe it is better to log the errors on lines 73 and 80 here to allow for the rest of the loop to continue. Should these be considered hard exits?

cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -560,3 +561,75 @@ func TagAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph
...
+				return nil
+			}); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func TagAssetGroupsAndTierZero(ctx context.Context, db database.Database, graphDb graph.Database, additionalFiltersToClear ...graph.Criteria) error {
```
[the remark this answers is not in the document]
urangel: The grouping is nice especially since it is duplicated where used currently. 

A side effect from grouping these and returning on the first error is that we would no longer allow for a later step to process if an earlier step fails since the error is returned immediately if there is one (for example, if tagging active directory fails for some reason, then tagging azure wouldn't happen). Unless we have evaluated whether or not it is okay to return early from an upstream error I think it would be better to follow a similar pattern of error aggregation so as not to block downstream processes.

urangel (approved):

PR #1335, opened 2025-04-09
feat: BED-5468 - added a graph fixture loader for graph json files

sircodemane:
## Description

This adds a function for importing a graph defined in JSON

## Motivation and Context

This PR addresses: BED-5468

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


superlinkx:
Please run `just prepare-for-codereview`

packages/go/lab/loader.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	return nil
+}
+
+// LoadGraphFixtureFile takes a graph.Database interface, a fs.FS interface,
+// and a path string. It will attempt to read the given path from the FS and
+// parse then file into a GraphFixture, and finally import the GraphFixture
+// Nodes and Edges into the graph.Database.
+func LoadGraphFixtureFile(db graph.Database, fSys fs.FS, path string) error {
+	if fh, err := fSys.Open(path); err != nil {
```
superlinkx, line 99: Let's avoid using `if err :=` when it becomes nested. I'd rather see a more traditional `if err != nil` for the outer in this case

packages/go/lab/loader.go
```diff
@@ -0,0 +1,110 @@
...
+	Type string `json:"type"`
+
+	// Properties is the key:value map used for storing extra information
+	// on the Edge.
+	Properties map[string]any `json:"properties"`
+}
+
+// ParseGraphFixtureJsonFile takes in a fs.File interface and parses its contents into a
+// GraphFixture struct.
+func ParseGraphFixtureJsonFile(fh fs.File) (GraphFixture, error) {
```
superlinkx, line 54: Is there a reason this isn't just part of the `LoadGraphFixtureFile` function and is exported right now? Seems like it would make sense to just fold it into the parent function until other usages are apparent

superlinkx (changes requested):
In addition to the code bits, has this been tested in any way to make sure it works?

sircodemane:
> Please run `just prepare-for-codereview`

i have become the thing i sought to fix :ded:

sircodemane:
> In addition to the code bits, has this been tested in any way to make sure it works?

@superlinkx the original PoC served as a manual function test. Since this is test tooling, tests would be contrived compared to use case testing. That was why I was asking yesterday if I should convert a couple harness tests. The failure points in the code are fairly limited since there's not really any business logic.

packages/go/lab/loader.go
sircodemane: Not sure I understand the reasoning behind this one. If it's to avoid shadowing the `err`, it still gets shadowed even if the outer `if` is collapsed

superlinkx:
Having a sample test or even just running this code in some way would be sufficient, I'd just like confirmation that the code has been run in some fashion before we merge, which I'm fairly sure I brought up during that discussion. Doesn't need official tests, just a confirmed code run for now

packages/go/lab/loader.go
superlinkx: Nope, it's because nested chaining is awful and I seek to remove it. We can take this offline if you'd like.

packages/go/lab/loader.go
superlinkx: I personally feel this is a premature abstraction, but feel free to override

local-harnesses/build.config.json.template
```diff
@@ -9,7 +9,6 @@
     "log_path": "bhapi.log",
     "enable_startup_wait_period": false,
     "datapipe_interval": 1,
-    "graph_driver": "neo4j",
```
superlinkx, line 12: This is incorrect, we want BHCE to continue to dev with neo4j by default. We need to remove the second "graph_driver": "pg" instance

local-harnesses/integration.config.json.template
```diff
@@ -9,7 +9,7 @@
     "log_path": "bhapi.log",
     "enable_startup_wait_period": false,
     "datapipe_interval": 1,
-    "graph_driver": "neo4j",
+    "graph_driver": "pg",
```
superlinkx, line 12: We do not want to use pg by default in BHCE at this time

cmd/api/src/analysis/ad/ad_integration_test.go
superlinkx, line 1: I absolutely love how this turned out, this already looks very ergonomic

packages/go/lab/loader.go
```diff
@@ -0,0 +1,161 @@
...
+	// on the Edge.
+	Properties map[string]string `json:"properties"`
+}
+
+// LoadGraphFixture requires a graph.Database interface and a GraphFixture struct.
+// It will import the nodes and edges from the GraphFixture and inserts them into
+// the graph database. It uses the `id` property of the nodes as the local
+// identifier and then maps them to database IDs, meaning that the `id` given in
+// the file will not be preserved.
+func LoadGraphFixture(db graph.Database, g *GraphFixture) error {
```
superlinkx, line 79: I still feel like we only need to expose one of these functions. For now, we have no need for LoadGraphFixture outside of the package, so I'd prefer to keep it internal

superlinkx (changes requested):
Overall looks good, we need to make sure to use the neo4j driver though

packages/go/lab/loader.go
```diff
@@ -0,0 +1,161 @@
...
+	var nodeMap = make(map[string]graph.ID)
+	if err := db.WriteTransaction(context.Background(), func(tx graph.Transaction) error {
```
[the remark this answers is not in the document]
superlinkx: We could do that, though how do you check by object id if two things were linked in post processing? I guess if no object id is given, we could autogen since it's better than not having anything, though some things will need valid sids instead of UUIDs to function correctly anyway. It's mostly a question of, if you need an objectid for a given test, you probably need to spell it out in the fixture so you can properly assert on it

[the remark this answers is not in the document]
sircodemane: we wouldn't want to autogen because some tests need them, and if a test needs an objectid, it should be explicitly set

packages/go/lab/loader.go
sircodemane: i'm realizing i did actually collapse one of the functions, but i think i do want to keep these two separated. i believe in the future that this will be simplified once generic ingest is introduced, but if the concern is public interface, then we should have a separate discussion around that.

PR #1358, opened 2025-04-15
Replace trustedby

JonasBK:
## Description

- Replace TrustedBy edge with four new edges: SameForestTrust, CrossForestTrust, AbuseTGTDelegation, and SpoofSIDHistory
- Implement LastCollected property for nodes
- Change LastSeen to be set at the time of the ingest task
- Add GroupScope node property

## Motivation and Context

The TrustedBy edge is only abusable in certain cases. Since edges must be either fully traversable or non-traversable, we have decided to replace TrustedBy with new, more precise edge types.

The LastCollected property is required for reconciliation of the trust edges.

The GroupScope property relevant for cross forest attacks, and is collected with this SHCommon PR: https://github.com/SpecterOps/SharpHoundCommon/pull/196

This PR addresses: BP-1253, BP-132

## How Has This Been Tested?

The following Domains.json file can be ingested to test the implementation: 

[20240923195225_domains (1).json](https://github.com/user-attachments/files/19754564/20240923195225_domains.1.json)

Upload of this data will results in the following edges:

```
MATCH p=(n:Domain)-[:SameForestTrust|CrossForestTrust|SpoofSIDHistory|AbuseTGTDelegation]-(m:Domain)
RETURN p
```
image

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/migrations/manifest.go
```diff
@@ -281,6 +281,55 @@ func Version_277_Migration(ctx context.Context, db graph.Database) error {
 	})
 }
 
+// Version_740_Migration is intended to split the TrustedBy edge to into SameForestTrust and CrossForestTrust edges
```
urangel, line 284: nit: Moving this block between the `RequiresMigration` func and the `Version_730_Migration` would maintain the existing migration order (more maintainable/clear; might not be expected here and missed) in this file as they start with most recent and get older down the file.

cmd/api/src/migrations/manifest.go
```diff
@@ -281,6 +281,55 @@ func Version_277_Migration(ctx context.Context, db graph.Database) error {
...
+				Kind:       newEdgeKind,
+				Properties: edgeProperties,
+			}); err != nil {
+				return err
+			}
+
+			// Delete old edge
+			if err := batch.DeleteRelationship(rel.ID); err != nil {
+				return err
+			}
```
urangel, line 326: Would it be possible to `UpdateRelationship` instead of doing both Create and Delete?

cmd/api/src/daemons/datapipe/azure_convertors.go
urangel, line 1: Since `nowUTC` is being plumbed all the way down into the *IngestX* funcs in ingest.go, we should be able to use that there within `NormalizeEinNodeProperties` to set the last collected property and not need to pass it into all of these convert funcs in both this file and in the convertors.go file. 

Doing that would make it apparent that the LastSeen property and the LastCollected property are getting set to the same value which brings me back to feeling that we should try and tighten up last seen without adding a new property.

I'm going to dig to see if there is a way we can accomplish the same thing without adding last collected but I think we should at least omit the nowUTC param passing into the convertor funcs and consolidate in the ingest.go file.

packages/javascript/bh-shared-ui/src/components/HelpTexts/SpoofSIDHistory/Abuse.tsx
```diff
@@ -0,0 +1,117 @@
+// Copyright 2023 Specter Ops, Inc.
```
urangel, line 1: Not sure how important this is but new files should have the year they were created in the copyright.

packages/javascript/bh-shared-ui/src/components/HelpTexts/SameForestTrust/Abuse.tsx
```diff
@@ -0,0 +1,170 @@
...
+                <br />
+                1) Obtain a SYSTEM session on a DC in the attacker-controlled domain
+                <br />
+                2) Create a certificate template allowing ESC1 abuse
+                <br />
+                3) Publish the certificate template to an enterprise CA
+                <br />
+                4) Enroll the certificate as a privileged user in the target domain
+                <br />
+                5) Authenticate as the privileged user in the target domain using the certificate
```
urangel, line 78: We should try to use semantic markup wherever appropriate. In this case that would be an <ol> tag with <li> children. I can provide the suggested refactor if that helps since I know sometimes the styling within these help texts isn't the most intuitive.

packages/javascript/bh-shared-ui/src/components/HelpTexts/CrossForestTrust/General.tsx
```diff
@@ -0,0 +1,118 @@
...
+                trusting domain into their authentication request. However, this attack can be prevented by SID
+                filtering.
+                <br />
+                <br />
+                SID filtering removes domain SIDs from authentication requests of foreign principals based on the trust
+                configuration. Built-in SIDs (S-1-5-32-*) are always filtered, regardless of the configuration. For
+                cross-forest trusts, blocking spoof SID history attacks requires filtering out SIDs that do not belong
+                to the forest of the trusted domain.
+                <br />
+                <br />
```
urangel, line 50: Recommend using CSS instead of `<br/>` tags to get the spacing the way you would like.

packages/go/analysis/ad/queries.go
```diff
@@ -1628,15 +1628,14 @@ func FetchCertTemplatesPublishedToCA(tx graph.Transaction, ca *graph.Node) (grap
...
+func FetchNodesWithSameForestTrustRelationship(tx graph.Transaction, root *graph.Node) (graph.NodeSet, error) {
 	if pathSet, err := ops.TraversePaths(tx, ops.TraversalPlan{
 		Root:      root,
 		Direction: graph.DirectionInbound,
 		BranchQuery: func() graph.Criteria {
 			return query.And(
 				query.KindIn(query.Start(), ad.Domain),
-				query.KindIn(query.Relationship(), ad.TrustedBy),
-				query.Equals(query.RelationshipProperty(ad.TrustType.String()), "ParentChild"),
+				query.KindIn(query.Relationship(), ad.SameForestTrust),
```
urangel, line 1638: The refactor looks nice in areas like these :D

cmd/api/src/migrations/manifest.go
JonasBK: Fixed in this commit: https://github.com/SpecterOps/BloodHound/pull/1358/commits/2dcfdbd46622e17f40c98d1dacfcce83482b7024

packages/go/analysis/ad/queries.go
JonasBK: Yup 😎

packages/javascript/bh-shared-ui/src/components/HelpTexts/CrossForestTrust/General.tsx
JonasBK: Would that be something like this?
```                
<p style={{ marginTop: '1em', marginBottom: '1em' }}>
                    SID filtering removes domain SIDs from authentication requests of foreign principals based on the trust
                    configuration. Built-in SIDs (S-1-5-32-*) are always filtered, regardless of the configuration. For
                    cross-forest trusts, blocking spoof SID history attacks requires filtering out SIDs that do not belong
                    to the forest of the trusted domain.
                </p>
```
I have no CSS experience. Let me know if there is another way you want me to fix it :)

This is what I'm going for:
image

packages/javascript/bh-shared-ui/src/components/HelpTexts/SameForestTrust/Abuse.tsx
JonasBK: Makes sense!
Fixed in this commit: https://github.com/SpecterOps/BloodHound/pull/1358/commits/0e1e8c9f5a566e858dce2f21df2bf21800d48596

packages/cue/bh/ad/ad.cue
```diff
@@ -1741,7 +1778,7 @@ InboundRelationshipKinds: list.Concat([SharedRelationshipKinds,[Contains]])
 OutboundRelationshipKinds: list.Concat([SharedRelationshipKinds,[Contains, DCFor]])
 
 // Edges that are used in pathfinding
-PathfindingRelationships: list.Concat([SharedRelationshipKinds,[Contains, DCFor, TrustedBy]])
+PathfindingRelationships: list.Concat([SharedRelationshipKinds,[Contains, DCFor, SameForestTrust, SpoofSIDHistory, AbuseTGTDelegation]])
```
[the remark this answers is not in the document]
JonasBK: Good call.
I added them to SharedRelationshipKinds by mistake. Fixed here: https://github.com/SpecterOps/BloodHound/pull/1358/commits/aa7efe0fdcaa5f0cee60cd52a7c1adc10922db0c

cmd/api/src/daemons/datapipe/convertors.go
```diff
@@ -175,20 +176,20 @@ func convertRootCAData(rootca ein.RootCA, converted *ConvertedData) {
...
 
-func convertEnterpriseCAData(enterpriseca ein.EnterpriseCA, converted *ConvertedData) {
-	baseNodeProp := ein.ConvertEnterpriseCAToNode(enterpriseca)
-	converted.NodeProps = append(converted.NodeProps, ein.ConvertEnterpriseCAToNode(enterpriseca))
+func convertEnterpriseCAData(enterpriseca ein.EnterpriseCA, converted *ConvertedData, ingestTime time.Time) {
+	baseNodeProps := ein.ConvertEnterpriseCAToNode(enterpriseca, ingestTime)
+	converted.NodeProps = append(converted.NodeProps, baseNodeProps)
 	converted.NodeProps = append(converted.NodeProps, ein.ParseCARegistryProperties(enterpriseca))
-	converted.RelProps = append(converted.RelProps, ein.ParseACEData(baseNodeProp, enterpriseca.Aces, enterpriseca.ObjectIdentifier, ad.EnterpriseCA)...)
+	converted.RelProps = append(converted.RelProps, ein.ParseACEData(baseNodeProps, enterpriseca.Aces, enterpriseca.ObjectIdentifier, ad.EnterpriseCA)...)
```
urangel, line 183: Ah, I was hoping we could avoid sending ingestTime details down here. My understanding came to be that `collected` is when we resolve a node from ingest, i.e., the collection scope included the node that we are marking as collected and that it isn't coming from a relationship. I suppose I am still fuzzy on the `collected` vs `lastseen` details. Apologies for my misunderstanding.

urangel (approved):
Left some follow up comments on discussions but this should be good to merge when you are ready. It looks like there are some minor import updates on the go side that pop up when running prepare for codereview. Suggest running and adding those imports when before merging.

packages/cue/bh/ad/ad.cue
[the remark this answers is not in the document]
urangel: I believe not including these in either the Inbound or Outbound rel kinds lists means they will be excluded from finding analysis in BHE. Is this the intended behavior?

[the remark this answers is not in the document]
JonasBK: Yes, that is intended. 
It's similar to how we don't have TrustedBy in the Inbound/Outbound lists currently. We analyze each domain individually, so we don't want to traverse across these domain -> domain edges in analysis.

packages/javascript/bh-shared-ui/src/components/HelpTexts/CrossForestTrust/General.tsx
JonasBK: Gotcha! Fixed here: https://github.com/SpecterOps/BloodHound/commit/8dc05ead3937f996c205db024dce984d30ce1ae7

cmd/api/src/daemons/datapipe/convertors.go
JonasBK: No worries at all! It is a bit confusing.

- LastCollected: Last time we collected the node in the sense that we found the underlying object in the directory, and collected all the relevant info about it e.g. properties, inbound permissions, etc.

- LastSeen: Updated when collecting a node, but also when we just see a reference to a node. 

Example: If we collect a domain, we get its trust relationships to other domains. The domain we collect will have its LastCollected and LastSeen timestamps updated, and the other domains where we found a trust relationship to will just have their LastSeen timestamp updated, as we only collected a reference (the trust relationship) to those domains.

Let me know if it makes sense, and if you want me to make any changes to the code :)

urangel (approved):

PR #1362, opened 2025-04-15
Kpom/bh rfc 2

kpom-specter:
## Description

## Motivation and Context

This PR addresses: [GitHub issue or Jira ticket number]

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/rfc/bh-rfc-2.md
+++ b/rfc/bh-rfc-2.md
@@ -1,0 +1,114 @@
+---
+bh-rfc: 2
+title: Early Return Pattern
+authors: |
+    [Holms, Alyx](aholms@specterops.io)
+    [Pomeroy, Kaleb](kpomeroy@specterops.io)
+status: DRAFT
+created: 2025-04-15
+---
+
+# Early Return Pattern
+
+Go encourages keeping blocks small, variables scoped tightly, and failure paths easy to follow. While Inline Initializer Statements help with scoping, they should not be combined with chained nesting patterns, which obscure control flow and increase complexity. Nesting hides the success case in the deepest block, when that’s actually the flow we care about most.
+
+## 1. Overview
+
+This proposal recommends adopting the Early Return Pattern as the default style for handling sequential workflows and error propagation in Go. Early returns flatten control flow, make error handling explicit, and align with Go’s idiomatic practices as described in Effective Go. This change improves readability, reduces indentation depth, and creates more maintainable and composable functions.
+
+### 1.1 Example of Early Return Pattern
+
+```
+func flat() {
+	a, err := stepA()
+	if err != nil {
+		fmt.Println("A had an error")
+		return
+	}
+
+	b, err := stepB()
+	if err != nil {
+		fmt.Println("B had an error")
+		return
+	}
+	c, err := stepC()
+	if err != nil {
+		fmt.Println("C had an error")
+		return
+	}
+
+	fmt.Println(a + b + c)
+}
+```
+
+### 1.2 Example of an Early Return Anti-pattern
+
+```
+func nesting() {
+	if a, err := stepA(); err != nil {
+		fmt.Println("A had an error")
+	} else if b, err := stepB(); err != nil {
+		fmt.Println("B had an error")
+	} else if c, err := stepC(); err != nil {
+		fmt.Println("C had an error")
+	} else {
+		fmt.Println(a + b + c)
+	}
+}
+```
+
+## 2. Motivation & Goals
+
+The goal of this RFC is to establish a consistent and idiomatic approach to control flow and error handling in sequential Go functions. Specifically, it promotes the use of early returns over chained nested if statements when dealing with dependent operations that may return errors.
+
+-   **Readability** Early returns keep the success path at the top level, making it easier to understand the primary logic of a function at a glance.
+-   **Maintainability** Flattened control flow reduces indentation and visual noise, making future changes and code reviews easier.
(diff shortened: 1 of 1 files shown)


rfc/bh-rfc-2.md
```diff
@@ -0,0 +1,114 @@
...
+authors: |
+    [Holms, Alyx](aholms@specterops.io)
+    [Pomeroy, Kaleb](kpomeroy@specterops.io)
+status: DRAFT
+created: 2025-04-15
+---
+
+# Early Return Pattern
+
+Go encourages keeping blocks small, variables scoped tightly, and failure paths easy to follow. While Inline Initializer Statements help with scoping, they should not be combined with chained nesting patterns, which obscure control flow and increase complexity. Nesting hides the success case in the deepest block, when that’s actually the flow we care about most.
```
superlinkx, line 13: This comes across a bit strong and I'm not sure I 100% agree with the premise. Chained nesting when done correctly is pretty consistent with only having error cases until the final else, and it left aligns the setup while right aligning the failure cases. It should be noted that even with chaining, you're still getting early returns, they may just be implicit (which is its own argument to be made).

As I noted in our original conversation, if/else chaining is _semantically_ the same as nesting, but the style doesn't make it look like nesting or allow for all the normal behaviors of nesting. It properly left aligns the actions being taken, right aligns the error path, and puts the happy path at the very end.

I think a better introduction would focus on where if/else chaining is causing direct pain in our upkeep of the code. Things like misuse of the pattern that leads to conditional branching _within_ the chain, having the happy path split into nested chaining that becomes harder to follow, or the fact that chaining when using a DAWGS ReadTransaction with a large closure becomes impossible to quickly check the error handling clause.

rfc/bh-rfc-2.md
```diff
@@ -0,0 +1,114 @@
...
+
+## 1. Overview
+
+This proposal recommends adopting the Early Return Pattern as the default style for handling sequential workflows and error propagation in Go. Early returns flatten control flow, make error handling explicit, and align with Go’s idiomatic practices as described in Effective Go. This change improves readability, reduces indentation depth, and creates more maintainable and composable functions.
```
superlinkx, line 17: This is an alright overview currently, but some of these things are true with the existing pattern as well. Error handling is still explicit in the nesting version, you're still early returning (implicitly, though it can also be made explicit, it just looks weird), and really control flow isn't that much more  flat, because again it's semantically not flat, but visually it's pretty flat.

Maintenance is something that could be improved by moving away from the chained if/else pattern, but we need more concrete information on _why_ that is and what specific footguns we run into today that would be easier to avoid with this approach.

rfc/bh-rfc-2.md
```diff
@@ -0,0 +1,114 @@
...
+
+## 2. Motivation & Goals
+
+The goal of this RFC is to establish a consistent and idiomatic approach to control flow and error handling in sequential Go functions. Specifically, it promotes the use of early returns over chained nested if statements when dealing with dependent operations that may return errors.
+
+-   **Readability** Early returns keep the success path at the top level, making it easier to understand the primary logic of a function at a glance.
+-   **Maintainability** Flattened control flow reduces indentation and visual noise, making future changes and code reviews easier.
+-   **Idiomatic Go** This approach aligns with community conventions and guidance from Effective Go and widely adopted Go style guides.
+-   **Error clarity** By separating failure paths early, each step’s error handling becomes clearer and less entangled with the main logic.
+-   **Reduced coupling**: Avoids tightly chaining variable scopes, enabling greater reuse and flexibility of intermediate values.
```
superlinkx, line 68: I know the kind of refactors this would be helpful for, but messing with intermediates can just as easily become spaghetti as what we have now if not guarded against abuse. How will we allow for this reduced coupling while promoting a style guide that helps curb poor use of access to variables? Specifically, how do we help shape code such that you either build control flow or business logic in a function, but never both at the same time?

rfc/bh-rfc-2.md
```diff
@@ -0,0 +1,114 @@
...
+
+This change supports writing code that is easier to reason about, debug, and refactor over time.
```
superlinkx, line 70: Let's cite why this is

rfc/bh-rfc-2.md
```diff
@@ -0,0 +1,114 @@
...
+Additionally, this does not guard us completely from scoping issues. If, for example stepC returned only a value (and not an error), the following two lines are now valid.
+
+`} else if c := stepC(); err != nil {`
+`c := stepC() {`
+
+In a stack of nested if statements, this nuance would be easy to miss. The err from stepB is still in scope. While the bug would still compile in the idiomatic flat style, it’s more obvious during review because each variable declaration is isolated and visually prominent
+
+## 4. Summary
+
+Chained nesting offers real scoping guarantees at the compiler level, but it comes at a readability cost for humans. In Go, clarity and simplicity should take precedence. Early returns flatten logic, highlight the success path, and align with the language’s philosophy of small, focused functions and explicit control flow. The goal isn’t just correctness at the machine's level. it’s maintainability at the developer level. And for that, flat wins.
```
superlinkx, line 114: I'd really like some additional depth in _why_ the idiomatic option is more readable/easier to maintain than nesting. I think there's some valid reasons that can be brought up, but ultimately this is a style discussion, so bringing good points on why something is easier than another is probably the most important piece of the puzzle

PR #1377, opened 2025-04-18
BED-5696 chore: Add migrations for existing agi to agt systems

mistahj67:
## Description

- Add migration for custom asset tier zero and owned selectors to new Tiering engine

## Motivation and Context
Addresses: BED-5696

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/database/migration/migrations/v7.4.0.sql
+++ b/cmd/api/src/database/migration/migrations/v7.4.0.sql
@@ -38,39 +38,34 @@ CREATE TABLE IF NOT EXISTS asset_group_tag_selector_nodes
 
 -- Migrate existing Tier Zero selectors
 WITH inserted_selector AS (
-INSERT INTO asset_group_tag_selectors (asset_group_tag_id, created_at, created_by, updated_at, updated_by, name, description, is_default, allow_disable, auto_certify)
-SELECT agt.id, current_timestamp, 'SYSTEM', current_timestamp, 'SYSTEM', s.name, s.selector, false, true, false
-FROM asset_group_selectors s
-         JOIN asset_groups ag ON ag.id = s.asset_group_id
-         JOIN asset_group_tags agt ON agt.name = 'Tier Zero'
-         LEFT JOIN asset_group_tag_selectors agts ON agts.name = s.name
-WHERE ag.tag = 'admin_tier_0' and agts.name IS NULL
-    RETURNING id, description
+  INSERT INTO asset_group_tag_selectors (asset_group_tag_id, created_at, created_by, updated_at, updated_by, name, description, is_default, allow_disable, auto_certify)
+  SELECT (SELECT id FROM asset_group_tags WHERE name = 'Tier Zero'), current_timestamp, 'SYSTEM', current_timestamp, 'SYSTEM', s.name, s.selector, false, true, false
+  FROM asset_group_selectors s JOIN asset_groups ag ON ag.id = s.asset_group_id
+  WHERE ag.tag = 'admin_tier_0' and NOT EXISTS(SELECT 1 FROM asset_group_tag_selectors WHERE name = s.name)
+  RETURNING id, description
 )
 INSERT INTO asset_group_tag_selector_seeds (selector_id, type, value) SELECT id, 1, description FROM inserted_selector;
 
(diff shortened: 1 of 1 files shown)


cmd/api/src/database/assetgrouptags_test.go
```diff
@@ -402,7 +403,7 @@ func TestDatabase_GetAssetGroupTagSelectors(t *testing.T) {
 	require.NoError(t, err)
 
 	t.Run("successfully returns an array of selectors, no filters", func(t *testing.T) {
-		results, err := dbInst.GetAssetGroupTagSelectorsByTagId(testCtx, 1, model.SQLFilter{}, model.SQLFilter{})
+		results, err := dbInst.GetAssetGroupTagSelectorsByTagId(testCtx, 1, nameFilter, model.SQLFilter{})
```
computator, line 406: This test description doesn't seem accurate anymore? Also with this change I don't think we have any tests for when there's no filters at all.

cmd/api/src/database/migration/migrations/v7.4.0.sql
computator, line 1: Honestly, a lot of this is very hard to read. Perhaps we can reformat it to be more legible? I can if you want.

cmd/api/src/database/migration/migrations/v7.4.0.sql
```diff
@@ -34,4 +34,417 @@ CREATE TABLE IF NOT EXISTS asset_group_tag_selector_nodes
...
+                    LEFT JOIN asset_group_tag_selectors agts on s.name = agts.name WHERE agts.name IS NULL
+    RETURNING id
+) INSERT INTO asset_group_tag_selector_seeds (selector_id, type, value) SELECT * FROM (SELECT(SELECT id FROM inserted_selector_id) id, 2, 'MATCH (n:AZRole)
+WHERE n.objectid STARTS WITH ''B5A8DCF3-09D5-43A9-A639-8E29EF291470@''
+RETURN n;') s WHERE id IS NOT NULL;
```
computator, line 98: If we're clever I think it might be possible to do a single multi-insert for these. One for all the selectors, and another for all the seeds. Either that or we should see if we could put it into a for-loop somehow.

Both solutions can be done with multi-inserts, CTEs with [value expressions](https://stackoverflow.com/a/6324833), and/or temporary tables. We can use a unique column from the selectors to look up the correct ID to do a multi-insert for the seeds.

cmd/api/src/database/migration/migrations/v7.4.0.sql
AD7ZJ, line 1: I verified that all the stuff in the excel file made it into the migration SQL. I agree the default cypher selector stuff could be formatted to make the lines shorter and more indenting.

cmd/api/src/database/migration/migrations/v7.4.0.sql
```diff
@@ -34,4 +34,43 @@ CREATE TABLE IF NOT EXISTS asset_group_tag_selector_nodes
 	updated_at timestamp with time zone,
 	CONSTRAINT fk_asset_group_tag_selectors_asset_group_tag_selector_nodes FOREIGN KEY (selector_id) REFERENCES asset_group_tag_selectors(id) ON DELETE CASCADE,
 	PRIMARY KEY (selector_id, node_id)
-	);
+);
+
+-- Migrate existing Tier Zero selectors
+WITH inserted_selector AS (
+INSERT INTO asset_group_tag_selectors (asset_group_tag_id, created_at, created_by, updated_at, updated_by, name, description, is_default, allow_disable, auto_certify)
+SELECT agt.id, current_timestamp, 'SYSTEM', current_timestamp, 'SYSTEM', s.name, s.selector, false, true, false
```
computator, line 42: I'm not sure which way is more readable, but `agt.id` could be a subquery so it's not part of the join and makes it more clear that it's static
```suggestion
SELECT (SELECT name FROM asset_group_tags WHERE name = 'Tier Zero'), current_timestamp, 'SYSTEM', current_timestamp, 'SYSTEM', s.name, s.selector, false, true, false
```

computator (approved):

PR #1404, opened 2025-04-25
BED 5594: Generic Ingest Datapipe Updates

brandonshearin:
## Description

- PR scope: reads a generic-formatted file from the file system and converts it into ingestible nodes/rels, and writes them to the graph.
- ingest code path: `jobs.go` -> `ingest.go` -> `decoders.go` -> `convertors.go`
- genericingest.go, contains functions that are specific to supporting the match-by-name use case for relationship-endpoint resolution
- files that contain a node/rel entity that cannot be processed (due to invalid or incomplete information) get logged as a warning, and skipped. ingest will continue processing the file containing the malformed entity, ingesting all entities that it can, but the ingest system will mark the file as failed at the end of processing. This is desirable because we want to show a yellow status (partially complete) on the UI for ingest jobs that contained files that were not processed in their entirety.

## Motivation and Context
Resolves BED-5594

We want to ingest arbitrary data into the graph. These changes update the relevant datapipe parts that manage ingestion. 

## How Has This Been Tested?
- integration tests 
- manual testing through the file upload UI. valid/invalid scenarios for zips and regular jsons. 
- this has been tested on both neo4j and pg drivers.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


.github/config/integration.config.json
```diff
@@ -9,7 +9,7 @@
     "log_path": "bhapi.log",
     "enable_startup_wait_period": false,
     "datapipe_interval": 1,
-    "graph_driver": "neo4j",
+    "graph_driver": "pg",
```
superlinkx, line 12: Revert, we need to be testing BloodHound against Neo4j while we still support it

cmd/api/src/daemons/datapipe/convertors.go
```diff
@@ -59,33 +106,37 @@ func convertComputerData(computer ein.Computer, converted *ConvertedData) {
 
 	converted.NodeProps = append(converted.NodeProps, ein.ParseDCRegistryData(computer))
 	converted.NodeProps = append(converted.NodeProps, baseNodeProp)
+
+	return nil
```
kpowderly, line 110: I'm not a big fan of these functions that were changed to declare that an error can be returned but to only return nil every time. I don't see anywhere that an error is actually returned so was it done just to fit into `decodeBasicData`. Just wondering if there's a cleaner way to handle it, thoughts?

docker-compose.dev.yml
```diff
@@ -132,7 +132,7 @@ services:
       bhe_graph_query_memory_limit: ${bhe_graph_query_memory_limit:-2}
       bhe_enable_text_logger: ${bhe_enable_text_logger:-true}
       bhe_recreate_default_admin: ${bhe_recreate_default_admin:-false}
-      bhe_graph_driver: ${bhe_graph_driver:-neo4j}
+      bhe_graph_driver: ${bhe_graph_driver:-pg}
```
kpowderly, line 135: should we keep this on neo4j?

.github/config/integration.config.json
brandonshearin: fixed

docker-compose.dev.yml
brandonshearin: yes, reverted

cmd/api/src/daemons/datapipe/convertors.go
brandonshearin: all of these functions implement the type `ConversionFunc` defined at the top of decoders.go. i extended that type to return an error so that convertGenericNode and convertGenericEdge could return errors. but you're right-- i had to add a bunch of return nil's to all the other implementing functions.

a made a new interface that only the two convertors that need to return errors implement, and left all the existing convertors as is!

PR #1408, opened 2025-04-28
BED-5816 (chore) adding tests for GetEdgeRelayTargets & changing TestVersion_730_Migration to integration test

bsheth711:
## Description

I added tests for the handler `GetEdgeRelayTargets` and changed the test `TestVersion_730_Migration` to be an integration test.

## Motivation and Context

Resolves [BED-5816](https://specterops.atlassian.net/browse/BED-5816?atlOrigin=eyJpIjoiZTQyZjZmYjQ5M2U3NGJhOTk3NjI1MzJlNDg4ZTAwMTQiLCJwIjoiaiJ9)

Add code coverage for `GetEdgeRelayTargets` and fix the test `TestVersion_730_Migration` being miscategorized as a unit test when it should be an integration test. 

## How Has This Been Tested?

Unit tests and integration tests have been run locally.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-5816]: https://specterops.atlassian.net/browse/BED-5816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/api/src/api/v2/edge_test.go
+++ b/cmd/api/src/api/v2/edge_test.go
@@ -535,7 +535,7 @@ func TestResources_GetEdgeRelayTargets_CannotMatchEdge(t *testing.T) {
 
 			assert.Equal(t, testCase.Expected.Code, actualCode)
 			assert.Equal(t, testCase.Expected.Header, actualHeader)
-			assert.Equal(t, testCase.Expected.Body, actualBody)
+			assert.JSONEq(t, testCase.Expected.Body, actualBody)
 		})
 	}
 }
@@ -599,7 +599,7 @@ func TestResources_GetEdgeRelayTargets_CannotGetNodes(t *testing.T) {
 
 			assert.Equal(t, testCase.Expected.Code, actualCode)
 			assert.Equal(t, testCase.Expected.Header, actualHeader)
-			assert.Equal(t, testCase.Expected.Body, actualBody)
+			assert.JSONEq(t, testCase.Expected.Body, actualBody)
 		})
 	}
 }
@@ -663,7 +663,7 @@ func TestResources_GetEdgeRelayTargets_PositiveTest(t *testing.T) {
 
 			assert.Equal(t, testCase.Expected.Code, actualCode)
 			assert.Equal(t, testCase.Expected.Header, actualHeader)
-			assert.Equal(t, testCase.Expected.Body, actualBody)
+			assert.JSONEq(t, testCase.Expected.Body, actualBody)
 		})
 	}
 }
--- a/cmd/api/src/api/v2/edge_test.go
+++ b/cmd/api/src/api/v2/edge_test.go
@@ -619,7 +619,7 @@ func TestResources_GetEdgeRelayTargets_PositiveTest(t *testing.T) {
 		Expected httpValues
 	}{
 		{
-			Name: "Error Trying to get Nodes",
+			Name: "Successful Request",
 			Request: http.Request{
 				URL: &url.URL{
 					RawQuery: "edge_type=AZBase&source_node=1&target_node=2",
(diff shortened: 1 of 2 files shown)


bsheth711:
recheck

cmd/api/src/api/v2/edge_test.go
```diff
@@ -306,3 +307,363 @@ func TestManagementResource_GetEdgeComposition(t *testing.T) {
...
+			actualCode, actualHeader, actualBody := test.ProcessResponse(t, response)
+
+			assert.Equal(t, testCase.Expected.Code, actualCode)
+			assert.Equal(t, testCase.Expected.Header, actualHeader)
+			assert.Equal(t, testCase.Expected.Body, actualBody)
+		})
+	}
+}
+
+func TestResources_GetEdgeRelayTargets_CannotGetNodes(t *testing.T) {
```
superlinkx, line 543: I haven't had a chance to play with the pattern here, so it may be a bit more nuanced than I expect, but collapsing all of these test cases into a single TestResources_GetEdgeRelayTargets() would make this a lot easier to maintain long term and decreases copy and paste. You'd have to make sure that the graph is included in the `requestInternalState`. The mocks might throw a bit of a wrench since we have t.Parallel(), so the ordering they get called in could change, but that can be solved by adding a 4th field to the cases struct that adds a callback function to contain the specific mock call. You should also only need to set up the resources once, since the mocks can be dynamically added as you go.

It's possible the t.Parallel breaks my assumptions here. If that's the case, these tests are so fast that I'd rather remove t.Parallel than deal with the bloat of splitting each mock test case out into its own test.

cmd/api/src/api/v2/edge_test.go
```diff
@@ -306,3 +307,363 @@ func TestManagementResource_GetEdgeComposition(t *testing.T) {
 		})
 	}
 }
+
+func TestResources_GetEdgeRelayTargets_BadParameters(t *testing.T) {
```
kpowderly, line 311: I know there's not a standard but since you're already doing table tests, I feel like it might be cleaner/more readable for all of GetEdgeRelayTargets tests to be done in one singular function and then each test case would be a subtest. 

Also, I think the values in your structs can be lowercase so that they are private since there's no need to export them.

juggernot325 (approved):
Nice first PR!

cmd/api/src/api/v2/edge_test.go
bsheth711: This is just a personal preference, but I split the tests into multiple functions based on the required internal state for the test. Each subtest then reflects a request made to that internal state. It does mean more copy and paste, but I do not think copy and paste for tests is a bad thing, and it clearly separates internal vs. external state for test cases. If you would still prefer me to refactor, let me know.

superlinkx: I definitely get where you're coming from, and it's similar to our past testing strategies (though we often didn't even do table driven testing). The primary reason we'd prefer to keep these tests bundled is to maximize the usage of the table driven tests and to make maintenance a bit more reliable. Having tons of tests with similar copy and paste strewn about has led to some rough refactor work in the past that we'd like to avoid. We are currently lacking official guidance for this, but the integration testing initiative will be adding more documentation for the styles and best practices we've begun adopting, with this being one of them.

Scoping the internal setup to each test case, rather than requiring separate tests still gives the same level of control and highlights what each test case requires. I'm also not against copy and paste with tests, but we should save ourselves code where it's unnecessary.

bsheth711: Sounds good, I'll get to it!

kpowderly (approved):

PR #1410, opened 2025-04-29
BED-5641: Add Unit Contract Tests for Auth Resources

iustinum:
## Description

Added contract tests for several auth API handlers including SAML, OIDC, and SSO login endpoints. Created comprehensive tests for handlers by following established contract testing patterns.

## Motivation and Context

Addresses BED-5641

This PR replaces integration tests with focused unit contract tests, making our test suite more maintainable and reliable. Contract tests provide better coverage of edge cases while being faster and less brittle than E2E tests.

## How Has This Been Tested?

All tests were run locally and verified to pass. Each test covers multiple scenarios including error cases and happy paths where applicable. The tests use mock objects to isolate the handlers from external dependencies.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/api/v2/auth/login_test.go
```diff
@@ -132,3 +137,85 @@ func TestLoginExpiry(t *testing.T) {
...
+			reqCtx := testCase.userContext(t)
+			req = req.WithContext(reqCtx)
+
+			response := httptest.NewRecorder()
+			resource.Logout(response, req)
+
+			assert.Equal(t, testCase.expected.responseCode, response.Code)
+
+			_, expectedRedirectURL := mockHost(t)
+			assert.Equal(t, expectedRedirectURL, response.Header().Get("Location"))
```
kpowderly, line 218: Normally, contract testing is meant to test the _contract_ between the consumer and the provider. 

You can think of the consumer as the code that initiates the request to perform an action and the provider as the service that responds to that request. 

So with that being said, the response is what we want to make sure that we're asserting. Ideally, we would be able to see in our tests what the request shape looks like, and what the expexted response shape would be (status code, body response, and headers) similar to what you would receive when calling the API yourself via Bruno etc.

cmd/api/src/api/v2/auth/login_test.go
```diff
@@ -132,3 +137,85 @@ func TestLoginExpiry(t *testing.T) {
...
+		responseCode int
+	}
+	type testData struct {
+		name        string
+		setupMocks  func(*testing.T, *api_mocks.MockAuthenticator)
+		expected    expected
+		userContext func(*testing.T) context.Context
+	}
+
+	mockHost := func(t *testing.T) (*url.URL, string) {
```
kpowderly, line 154: I'm slightly confused about what's being done with this. It looks like you're just checking that one of the response headers contains this file path.

Are you okay with just checking the response headers 

```
mux.NewRouter().ServeHTTP(response, req)
status, header, body := test.ProcessResponse(t, response)
```

cmd/api/src/api/v2/auth/oidc_test.go
```diff
@@ -280,3 +289,442 @@ func TestManagementResource_UpdateOIDCProvider(t *testing.T) {
...
+				responseCode:    http.StatusFound,
+				locationHeader:  true,
+				redirectToLogin: false,
+			},
+			setupHost: mockHost,
+		},
+	}
+
+	for _, testCase := range tt {
+		t.Run(testCase.name, func(t *testing.T) {
```
kpowderly, line 380: When looping through each subtest, we want each subtest to follow the same validation logic. Each individual subtest set up should be done separately and then we would assert on the same response fields for all of them. If you have any questions, let me know -- I'd be happy to help :)

cmd/api/src/api/v2/auth/sso_test.go
```diff
@@ -296,3 +302,157 @@ func TestManagementResource_SanitizeAndGetRoles(t *testing.T) {
...
+
+			response := httptest.NewRecorder()
+			resources.SSOLoginHandler(response, req)
+
+			assert.Equal(t, testCase.expected.responseCode, response.Code)
+		})
+	}
+}
+
+func TestManagementResource_SSOCallbackHandler(t *testing.T) {
```
kpowderly, line 383: Probably need to run the code coverage report on all of these because we're missing a lot of coverage

cmd/api/src/api/v2/auth/oidc_test.go
iustinum: Refactored the subtests so they all follow the same validation logic, with setup handled individually per case.

iustinum:
I ran the coverage report, and it's currently at 68.1% for all tests in the auth resource files. I noticed many uncovered handlers aren't part of the contract test list. I'm working on improving coverage for missed branches in the new contract tests, but we might also need to refactor some of the other tests not scoped in the original ticket.

PR #1411, opened 2025-04-29
BED-5845 - Rework DAWGS Value Mapping

zinic:
## Description

Generic value mapping in DAWGS is poorly supported and was hacked together in order to support the first iteration of generic Cypher search for users.

Originally this logic was written for Neo4j with its caveats and quirks. Addition of the PostgreSQL driver introduced significant drift and technical debt in several of these paths.

The PostgreSQL driver for DAWGS currently targets upstream OID matching to utilize the object mapping cache found in the pgx library. This can be examined in the Scan(...) function of the PostgreSQL DAWGS driver type. The performance benefit of this optimization is not great enough to justify both the Neo4j and PostgreSQL driver deviating on mapping to DAWGS types. This deviation also made generic mapping of query results far more difficult.

The Result and Scanner interface have been refactored to remove the current ValueMapper idiom and value mapping between both the Neo4j and PostgreSQL driver is now unified.

## Motivation and Context

Resolves BED-5845

Value mapping in DAWGS contains a complexity trap related to certain query return shapes. This is most prevalent when using the generic Cypher search functionality. This trap leads to CPU saturation.

## How Has This Been Tested?

This refactor is covered by our existing tests.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/queries/graph.go
+++ b/cmd/api/src/queries/graph.go
@@ -147,7 +147,7 @@ type Graph interface {
 	FetchNodesByObjectIDsAndKinds(ctx context.Context, kinds graph.Kinds, objectIDs ...string) (graph.NodeSet, error)
 	ValidateOUs(ctx context.Context, ous []string) ([]string, error)
 	BatchNodeUpdate(ctx context.Context, nodeUpdate graph.NodeUpdate) error
-	RawCypherQuery(ctx context.Context, pQuery PreparedQuery, includeProperties bool) (model.UnifiedGraph, error)
+	RawCypherQuery(ctx context.Context, requestID string, pQuery PreparedQuery, includeProperties bool) (model.UnifiedGraph, error)
 	PrepareCypherQuery(rawCypher string, queryComplexityLimit int64) (PreparedQuery, error)
 	UpdateSelectorTags(ctx context.Context, db agi.AgiData, selectors model.UpdatedAssetGroupSelectors) error
 	FetchNodeByGraphId(ctx context.Context, id graph.ID) (*graph.Node, error)
@@ -427,7 +427,10 @@ func (s *GraphQuery) PrepareCypherQuery(rawCypher string, queryComplexityLimit i
 	return graphQuery, nil
 }
 
-func (s *GraphQuery) RawCypherQuery(ctx context.Context, pQuery PreparedQuery, includeProperties bool) (model.UnifiedGraph, error) {
+// RawCypherQuery executes the given PreparedQuery and returns a model.UnifiedGraph or any error encountered during
+// query execution. This function takes in a generic request ID string to help match intent and execution results in
+// the application log.
+func (s *GraphQuery) RawCypherQuery(ctx context.Context, requestID string, pQuery PreparedQuery, includeProperties bool) (model.UnifiedGraph, error) {
 	var (
 		err error
 
@@ -445,6 +448,13 @@ func (s *GraphQuery) RawCypherQuery(ctx context.Context, pQuery PreparedQuery, i
 		}
 	)
 
+	slog.Info(
+		"Preparing user cypher query",
+		"query", pQuery.StrippedQuery,
+		"fitness", pQuery.complexity.RelativeFitness,
+		"request_id", requestID,
+	)
+
 	if pQuery.HasMutation {
 		// If the mutation is complex it is still worth spinning it into a write transaction in case it fails,
 		// deadlocks or otherwise rolls back
@@ -453,12 +463,12 @@ func (s *GraphQuery) RawCypherQuery(ctx context.Context, pQuery PreparedQuery, i
 		err = s.Graph.ReadTransaction(ctx, txDelegate)
 	}
 
-	runtime := time.Since(start)
-
 	slog.Info(
-		fmt.Sprintf("Executed user cypher query with cost %d in %.2f seconds", pQuery.complexity.RelativeFitness, runtime.Seconds()),
+		"Executed user cypher query",
 		"query", pQuery.StrippedQuery,
-		"query cost", fmt.Sprintf("%d", pQuery.complexity.RelativeFitness),
+		"fitness", pQuery.complexity.RelativeFitness,
+		"elapsed", time.Since(start),
+		"request_id", requestID,
 	)
 
 	if err != nil {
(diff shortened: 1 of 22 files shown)


cmd/api/src/queries/graph.go
```diff
@@ -444,6 +447,13 @@ func (s *GraphQuery) RawCypherQuery(ctx context.Context, pQuery PreparedQuery, i
 		}
 	)
 
+	slog.Info(
```
superlinkx, line 450: We should probably use `slog.InfoContext` since we have a valid context to include for additional information. This will also remove the need to plumb requestID down, since the slog Handler will inject that information if the context is already present.

cmd/api/src/queries/graph.go
```diff
@@ -444,6 +447,13 @@ func (s *GraphQuery) RawCypherQuery(ctx context.Context, pQuery PreparedQuery, i
...
+		"Preparing user cypher query",
+		"query", pQuery.StrippedQuery,
+		"fitness", pQuery.complexity.RelativeFitness,
+		"request_id", requestID,
```
superlinkx, line 454: This is fine, but I've had some improvement with using slog typed parameters:

`slog.String("query", pQuery.StrippedQuery)`

They help with type safety to an extent and help a little with avoiding mismatched arguments

cmd/api/src/queries/graph.go
```diff
@@ -452,12 +462,12 @@ func (s *GraphQuery) RawCypherQuery(ctx context.Context, pQuery PreparedQuery, i
 		err = s.Graph.ReadTransaction(ctx, txDelegate)
 	}
 
-	runtime := time.Since(start)
-
 	slog.Info(
```
superlinkx, line 465: Same thing here, we should use a slog.InfoContext to capture the context and avoid the need for request_id. Can also clean up these parameters in the same way as above too.

superlinkx (changes requested):
Mostly nits, really appreciated having the context specific commits. Made it really easy to see what was changing without feeling overloaded with context, and gave me a good idea of your process too!

packages/go/dawgs/drivers/pg/mapper.go
```diff
@@ -23,62 +23,83 @@ import (
 	"github.com/specterops/bloodhound/dawgs/graph"
 )
 
-func mapValue(ctx context.Context, kindMapper KindMapper) func(rawValue, target any) (bool, error) {
-	return func(rawValue, target any) (bool, error) {
+func mapKinds(ctx context.Context, kindMapper KindMapper, untypedValue any) (graph.Kinds, error) {
+	var kindIDs []int16
+
+	switch typedValue := untypedValue.(type) {
```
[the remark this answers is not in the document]
zinic: Uninitialized slices are zero-length slices. This would have no ill effect downstream to the KindMapper. The critical sections in the KindMapper are also guarded by a RW mutex, making lock contention from this unlikely.

Maybe a length guard before calling the mapper is a good idea here, regardless.

I'm torn here between preserving and relaying call expectations to the caller of this particular function. Failure to map to a single type is not an error. Failure to map to *any available type* however. All current callers of this function will simply throw the error away. I'm considering eliding the error completely here as we do in other parts of the mapping logic.

packages/go/dawgs/drivers/neo4j/mapper.go
```diff
@@ -43,57 +42,48 @@ func AsTime(value any) (time.Time, error) {
...
 }
 
-func mapValue(rawValue, target any) (bool, error) {
+func mapValue(rawValue, target any) bool {
 	switch typedTarget := target.(type) {
 	case *time.Time:
-		if value, err := AsTime(rawValue); err != nil {
-			return false, err
-		} else {
+		if value, err := AsTime(rawValue); err == nil {
```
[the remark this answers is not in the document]
zinic: Cleaning this up with the like named Neo4j version of this function as well.

packages/go/dawgs/drivers/neo4j/node.go
```diff
@@ -185,7 +185,7 @@ func (s *NodeQuery) First() (*graph.Node, error) {
 
 func (s *NodeQuery) Fetch(delegate func(cursor graph.Cursor[*graph.Node]) error, finalCriteria ...graph.Criteria) error {
 	return s.Query(func(result graph.Result) error {
-		cursor := graph.NewResultIterator(s.ctx, result, func(scanner graph.Scanner) (*graph.Node, error) {
+		cursor := graph.NewResultIterator(s.ctx, result, func(scanner graph.Result) (*graph.Node, error) {
```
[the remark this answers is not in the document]
zinic: :facepalm: Yes.

packages/go/dawgs/drivers/pg/mapper.go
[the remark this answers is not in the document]
superlinkx: This seems reasonable, especially if we're eliding to match the other logic. 0-length was my primary concern, as I didn't actually dig into what happens if the valuemapper gets nothing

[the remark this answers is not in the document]
zinic: Did both. Give it a squzz :pray:

cmd/api/src/queries/graph.go
zinic: :+1:

cmd/api/src/queries/graph.go
zinic: :+1:

cmd/api/src/queries/graph.go
zinic: :+1:

packages/go/dawgs/drivers/neo4j/relationship.go
```diff
@@ -211,7 +211,7 @@ func (s *RelationshipQuery) FetchAllShortestPaths(delegate func(cursor graph.Cur
 
 func (s *RelationshipQuery) FetchTriples(delegate func(cursor graph.Cursor[graph.RelationshipTripleResult]) error) error {
 	return s.Query(func(result graph.Result) error {
-		cursor := graph.NewResultIterator(s.ctx, result, func(scanner graph.Scanner) (graph.RelationshipTripleResult, error) {
+		cursor := graph.NewResultIterator(s.ctx, result, func(scanner graph.Result) (graph.RelationshipTripleResult, error) {
```
superlinkx, line 214: Still a few missed renames for scanner

superlinkx (commented):
Changes look good, there's still quite a few places with the renamed graph.Result that name the variable scanner, may be worth the cleanup for consistency, but that's all that I found

packages/go/dawgs/drivers/neo4j/relationship.go
zinic: Snagged em

superlinkx (approved):
Thanks for all the extra diligence, love this changeset

PR #1412, opened 2025-04-29
feat(Generic Ingest): BED-5600 Add UI Custom Node Icons

wes-mil:
## Description

Add custom UI icons to the front-end.

## Motivation and Context

Resolves BED-5600

Adds custom UI icons to the front-end. Primarily to allow users to add icons to generic ingest data.

## How Has This Been Tested?

Tested locally

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/ui/src/views/Content.tsx
```diff
@@ -50,6 +50,7 @@ const Content: React.FC = () => {
         if (isFullyAuthenticated) {
             dispatch(fetchAssetGroups());
             dispatch(ListAssetGroups());
+            dispatch(fetchCustomNodeInformation());
```
benwaples, line 53: running this at this level in the component tree would result on this request occurring on pages that dont require custom nodes. Do we need to run it at this level, or is there a more focused/narrow component that would work?

cmd/ui/src/hooks/useCustomNodeKinds.tsx
```diff
@@ -0,0 +1,9 @@
+import { apiClient } from 'bh-shared-ui';
+import { GetCustomNodeKindsResponse, RequestOptions } from 'js-client-library';
+import { UseQueryResult, useQuery } from 'react-query';
+
+export const getCustomNodeKinds = async (options: RequestOptions): Promise<GetCustomNodeKindsResponse> =>
+    apiClient.getCustomNodeKinds(options).then((res) => res.data);
+
+export const useGetCustomNodeKinds = (): UseQueryResult<GetCustomNodeKindsResponse> =>
+    useQuery<GetCustomNodeKindsResponse>(['getCustomNodeKinds'], ({ signal }) => getCustomNodeKinds({ signal }));
```
benwaples, line 9: Great call pulling this out into its own hook using react-query. We use these all the time and are very helpful. I have 2 suggestion:
- I think you can drop the Get from useGetCustomNodeKinds
- Plumb thru the react-query query options so that this hook is more flexible:

```suggestion
export const useGetCustomNodeKinds = (queryOptions?: GenericQueryOptions<GetCustomNodeKindsResponse>): UseQueryResult<GetCustomNodeKindsResponse> => {
    return useQuery({
        queryKey: ['getCustomNodeKinds'],
        queryFn: ({ signal }) => getCustomNodeKinds({ signal }),
        ...queryOptions
    })
}
```

cmd/ui/src/rootSaga.ts
```diff
@@ -17,10 +17,11 @@
 import { all, call } from 'redux-saga/effects';
 import assetgroupsSaga from 'src/ducks/assetgroups/saga';
 import exploreSaga from 'src/ducks/explore/saga';
+import globalSaga from 'src/ducks/global/saga';
 import searchSaga from 'src/ducks/searchbar/saga';
 import tierZeroSaga from 'src/ducks/tierzero/saga';
 
-const sagas = [call(exploreSaga), call(searchSaga), call(tierZeroSaga), call(assetgroupsSaga)];
+const sagas = [call(exploreSaga), call(searchSaga), call(tierZeroSaga), call(assetgroupsSaga), call(globalSaga)];
```
benwaples, line 24: Nice work adding a saga! Idk if you were familiar with them previously, but there is a bit of a learning curve to them for sure.

However, we have been moving away from the redux+saga implementation when we have the ability to use server state which I think we have that opportunity here.

We should be able to build upon the `useGetCustomNodeKinds` hook and create something like `useCustomNodeIcons` that would do something like:
```tsx

const transformNodes = () => {} // same function you have in your saga

const useCustomNodeIcons = () => {
 return useGetCustomNodeKinds({ select: transformNodes })
}
```

And then in GraphViewV2 and NodeIcon:
```
const customIcons = useCustomNodeIcons() // instead of the redux selector
```
This would also allow us to not need to request custom nodes on every page, only on pages that are displaying nodes (either on the graph or stand alone input adornments)

I do think this pattern would require us to dig a little more into the timing of when we need to make this request and populate the object urls.

cmd/ui/src/components/NodeIcon.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <Box className={classes.root}>
+                <Box className={classes.container}>
+                    <FontAwesomeIcon icon={icon.icon} transform='shrink-2' />
+                </Box>
+            </Box>
+        </Tooltip>
+    );
+};
+
+export const GetIconInfo = (iconName: string, customIcons: IconDictionary): IconInfo => {
```
benwaples, line 70: Is this a react component or a util? Seems like its not returning a `ReactNode`... if its a util, mind moving to perhaps the Explore/utils or somewhere in shared-ui if its going to be used in bhe. React gets fussy with hot reloading when a component file exports more than components 🙄

packages/javascript/bh-shared-ui/src/components/NoDataDialog/NoDataDialog.tsx
```diff
@@ -22,7 +22,7 @@ type NoDataDialogProps = PropsWithChildren<{ open: boolean }>;
 export const NoDataDialog: React.FC<NoDataDialogProps> = ({ open, children }) => {
     return (
         <Dialog
-            open={open}
+            open={false}
```
benwaples, line 25: Is this needed? or was it for testing purposes?

packages/javascript/bh-shared-ui/src/utils/icons.ts
```diff
@@ -47,9 +48,12 @@ import {
     faUser,
     faUsers,
     faWindowRestore,
+    fas,
 } from '@fortawesome/free-solid-svg-icons';
 import { ActiveDirectoryNodeKind, AzureNodeKind } from '../graphSchema';
 
+library.add(fas);
```
benwaples, line 55: Just curious, is this what adds all icon definitions to the bundle?

packages/javascript/bh-shared-ui/src/components/NodeIcon/NodeIcon.tsx
```diff
@@ -48,17 +51,24 @@ const useStyles = makeStyles<Theme, NodeIconProps, string>({
...
 const NodeIcon: React.FC<NodeIconProps> = ({ nodeType }) => {
-    const classes = useStyles({ nodeType });
+    const icon = GetIconInfo(nodeType);
+    const classes = useStyles({ icon });
 
     return (
         <Tooltip title={nodeType || ''} describeChild={true}>
             <Box className={classes.root}>
                 <Box className={classes.container}>
-                    <FontAwesomeIcon icon={NODE_ICON[nodeType]?.icon || faQuestion} transform='shrink-2' />
```
benwaples, line 57: do we lose this fallback from these changes? If so, is there something we are replacing it with?

brandonshearin: question mark fallback logic is extracted into the GetIconInfo() helper

cmd/api/src/api/v2/customnode.go
```diff
@@ -74,8 +72,8 @@ func validateCreateCustomNodeRequest(customNodeKindRequest CreateCustomNodeReque
 func validateConfig(config model.CustomNodeKindConfig) error {
 	if config.Icon.Type != "font-awesome" {
 		return fmt.Errorf("custom node kind config type (%s) is not supported", config.Icon.Type)
```
brandonshearin, line 74: "Invalid icon type. Only Font Awesome icons are supported."

cmd/api/src/api/v2/customnode.go
```diff
@@ -74,8 +72,8 @@ func validateCreateCustomNodeRequest(customNodeKindRequest CreateCustomNodeReque
...
-	} else if !validIconName.MatchString(config.Icon.Name) {
-		return fmt.Errorf("custom node kind config name (%s) is not valid", config.Icon.Name)
+	} else if !validColorString.MatchString(config.Icon.Color) && config.Icon.Color != "" {
+		return fmt.Errorf("custom node kind config color (%s) is invalid", config.Icon.Color)
```
brandonshearin, line 76: make error string more specific about its invalidity--

"Icon color must be a valid hexadecimal color string starting with '#' followed by 3 or 6 hex digits."

also, update the regex to support 3 or 6 character hex strings. 3 char strings are valid

cmd/api/src/api/v2/customnode.go
brandonshearin, line 74: "Invalid icon type. Only Font Awesome icons are supported."

cmd/ui/src/views/Explore/svgIcons.ts
```diff
@@ -16,14 +16,14 @@
...
-import { GLYPHS, GlyphDictionary, IconDictionary, NODE_ICON, UNKNOWN_ICON } from 'bh-shared-ui';
+import { GLYPHS, GetIconInfo, GlyphDictionary, IconDictionary, NODE_ICON, UNKNOWN_ICON } from 'bh-shared-ui';
 
 const NODE_SCALE = '0.6';
 const GLYPH_SCALE = '0.5';
 const DEFAULT_ICON_COLOR = '#000000';
 
 // Adds object URLs to all icon and glyph definitions so that our fontawesome icons can be used by sigmajs node programs
-const appendSvgUrls = (icons: IconDictionary | GlyphDictionary, scale: string): void => {
+export const appendSvgUrls = (icons: IconDictionary | GlyphDictionary, scale: string): void => {
```
brandonshearin, line 26: i remember we exported this temporarily while we figuring out a solution, but remove this export if it is no longer consumed by any code outside of this file.

packages/go/openapi/doc/openapi.json
```diff
@@ -16655,6 +16655,9 @@
               },
               "name": {
                 "type": "string"
+              },
```
brandonshearin, line 16658: one more docs-related improvement--
Screenshot 2025-05-19 at 10 00 03 AM

change the "examples" section for the POST (create) endpoint to include a valid payload that would actually produce a 200 success response. "kindA, kindB, kindC" with three valid icons and maybe 2 valid colors, demonstrating that color is optional

brandonshearin: lets also update the description of this endpoint now that its final behavior has settled

"This endpoint registers new node kinds along with optional display metadata such as icons and colors. The provided configuration controls how nodes of the specified kinds are interpreted and rendered in the UI. 

Only free, solid-style Font Awesome icons are supported. Hex color codes must follow the format #RGB or #RRGGBB. 

If no valid icon or color is provided, the system will fall back to a question mark icon and a white background color.

Batch input is supported; you may submit multiple configuration objects in a single request."

cmd/api/src/api/v2/customnode.go
brandonshearin, line 1: it would be ideal to write some tests for the POST/PUT endpoints in this file, especially because these endpoints contain some validation logic that we want to assert against. look to `cypherquery_test.go` for a good api testing example from recent times

packages/go/openapi/doc/openapi.json
brandonshearin, line 1: update the documentation for the PUT endpoint in a similar manner-- include an example value for that endpoint that would be valid. eg--

{
    "config": {
            "icon": {
                "type": "font-awesome",
                "name": "bell",
                 "color": "#0000FF"
            }
    }
}

packages/javascript/bh-shared-ui/src/components/NodeIcon/NodeIcon.tsx
brandonshearin, line 1: can we add some unit tests for `NodeIcon` that poke at the custom kinds logic, the fallback logic, and the color logic?

packages/go/openapi/doc/openapi.json
brandonshearin: manual tested the PUT request in the swagger docs. the example request is slightly off. it is currently written as
```{
  "config": {
    "type": "font-awesome",
    "name": "house",
    "color": "#333"
  }
}```

but needs to have an `icon` key underneath the `config` top level key. the original comment i left regarding this change explains one such valid example

packages/javascript/bh-shared-ui/src/components/NodeIcon/NodeIcon.tsx
brandonshearin: nice!

cmd/api/src/api/v2/customnode.go
brandonshearin: really solid tests. #font-stupid

brandonshearin (approved):
great job pulling all this together

PR #1418, opened 2025-04-30
Bed 5737

LawsonWillard:
## Description

- Updated filter_internal_test.go to use table driven tests and assert instead of require.
- Move counts parameter to ignored filters as this filter should not be included in the standard query parameter filters.
- Updated ParseQueryParameterFilter to return the equals operator if the provided query parameter value does not match the 'predicate_operator:value' format.
- Removed the now unused "ErrNotFiltered" error

The above changes have been made to ensure that the default eq operator is used in the event that the provided API call does not supply a predicate in the query parameter value.

## Motivation and Context

Resolves BED-5737

If a call to any API endpoint includes a string param without a predicate qualifier prefix (eq:, neq:), the call ignores the filter value rather than assume the default operation of “equals” (eq:).

## How Has This Been Tested?

- Added unit test to cover scenario + regression unit testing to ensure the change did not break any other tests
- Manual testing against API on local dev setup

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/model/filter_internal_test.go
+++ b/cmd/api/src/model/filter_internal_test.go
@@ -97,6 +97,19 @@ func TestQueryParameterFilterParser_ParseQueryParameterFilter(t *testing.T) {
 			wantErr: assert.NoError,
 		},
 		{
+			name: "success - colon is included in value if predicate is omitted",
+			args: args{
+				name:  "parameter",
+				value: ":hello_world",
+			},
+			expected: expected{
+				name:     "parameter",
+				value:    ":hello_world",
+				operator: Equals,
+			},
+			wantErr: assert.NoError,
+		},
+		{
 			name: "fail - unknown operator",
 			args: args{
 				name:  "parameter",
--- a/cmd/api/src/model/filter_internal_test.go
+++ b/cmd/api/src/model/filter_internal_test.go
@@ -19,42 +19,104 @@ package model
 import (
 	"testing"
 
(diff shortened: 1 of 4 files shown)


LawsonWillard:
recheck

cmd/api/src/model/filter_internal_test.go
```diff
@@ -19,42 +19,104 @@ package model
...
+			wantErr: assert.NoError,
+		},
+		{
+			name: "fail - unknown operator",
+			args: args{
+				name:  "parameter",
+				value: "foo:hello world",
+			},
+			wantErr: assert.Error,
+		},
```
definitelynotagoblin, line 106: Nit: I'd be kinda curious to test what happens when value is ":some value"
I believe it will resolve to `equals: ':some value'`
Which is probably correct, but maybe we instead want to treat this as an error 🤷

LawsonWillard: Just tested and thats the result, was also wondering the same. Would it be best to ask that in the ticket? 

Also semi side note: When testing against the /api/v2/available-domains endpoint and using the collected parameter collected=eq:true - correctly returns all results where collected is true, however, collected=neq:true - returns an empty array. I would expect this to return all values which are false, unsure if this is intended or not

LawsonWillard: @definitelynotagoblin added the new test case, will ask in ticket about its expected outcome

PR #1448, opened 2025-05-12
fix(tier-mgmt): selector form updates

urangel:
## Description

- a run button is added to the object selector
- both object id and cypher seeds are now driven by the preview selector endpoint
- pulse animation is added to run button when seeds change to indicate staleness
- delete button on selector form is hidden for default selectors
- the selector status switch now correctly sends null or an iso time string instead of 'on' and 'off'

## Motivation and Context

closes BED-5872, closes BED-5863, closes BED-5864

## How Has This Been Tested?
Unit tests were updated. 
A new unit test was added for whether the delete button renders for a default selector or not.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


elikmiller (approved):

packages/javascript/bh-shared-ui/src/components/AssetGroupManagement/AssetGroupSelectorObjectSelect.tsx
```diff
@@ -20,35 +20,92 @@ import {
...
+
+    useEffect(() => {
+        const result = previewQuery.data ? previewQuery.data : null;
+
+        setSeedPreviewResults(result);
+    }, [previewQuery.data, setSeedPreviewResults]);
+
+    useEffect(() => {
+        previewQuery.refetch();
+    }, [seedsQuery.data]);
```
maffkipp, line 108: Since you are already running side effects in the queryFn for `seedsQuery`, it might be better to move this refetch there instead of handling it in a separate effect

packages/javascript/bh-shared-ui/src/components/AssetGroupManagement/AssetGroupSelectorObjectSelect.tsx
```diff
@@ -20,35 +20,92 @@ import {
...
+        enabled: seeds.length !== 0,
+    });
+
+    const handleRun = useCallback(() => {
+        previewQuery.refetch();
+        setStalePreview(false);
+    }, [previewQuery]);
+
+    useEffect(() => {
+        const result = previewQuery.data ? previewQuery.data : null;
```
maffkipp, line 101: this can be rewritten as `const result = previewQuery.data ?? null` if you want to save some characters! Not a big deal though

packages/javascript/bh-shared-ui/src/views/TierManagement/Cypher/cypher.css
```diff
@@ -0,0 +1,17 @@
+.dark .cm-editor {
```
maffkipp, line 1: I think this solution is reasonable and definitely looks clean, but if we want to avoid having to maintain a separate css file and the new dependency we are adding, I think we could achieve something similar with tailwind.

It is kind of buried in [these docs](https://v3.tailwindcss.com/docs/hover-focus-and-other-states#using-arbitrary-variants), but I believe setting a class like `dark:[&_.cm-editor]:bg-transparent` would achieve the same goal.

packages/javascript/bh-shared-ui/src/views/TierManagement/Save/SelectorForm/index.tsx
```diff
@@ -14,7 +14,7 @@
 //
```
maffkipp, line 14: I know this decision predates this PR, but the pattern elsewhere in the app is to just use `index` files for a module's top-level exports. Could we possibly rename this file to `SelectorForm.tsx` and then create another `index` file as needed?

packages/javascript/bh-shared-ui/src/views/TierManagement/Cypher/cypher.css
urangel: oo nice!

packages/javascript/bh-shared-ui/src/components/AssetGroupManagement/AssetGroupSelectorObjectSelect.tsx
urangel: This ends up in a race condition where the state isn't updated by the time the refetch happens unfortunately :s I'd love to get rid of this useEffect though! Let me know if you think of anything else please

maffkipp (approved):

PR #1470, opened 2025-05-19
feat: Adding validations to user creations form

LucasParraF:
## Description

Administrators were able to create new users without proper validations in terms of length, special characters and blank spaces. I added validations that prevents this to happen while displaying error messages that guide the administrator on which parameters are not being covered.

## Motivation and Context

Resolves: https://specterops.atlassian.net/browse/BED-5769

Administrators were able to create new users without proper validations in terms of length, special characters and blank spaces.

## How Has This Been Tested?

Created new test files.

## Screenshots (optional):

Create-User1
Create-User2
Create-User3
Update-User1
Update-User2
Update-User3

## Types of changes

- [ ] - Chore (a change that does not modify the application functionality)
- [x]  Bug fix (non-breaking change which fixes an issue)
- [ ] - New feature (non-breaking change which adds functionality)
- [ ] - Breaking change (fix or feature that would cause existing functionality to change)
- [ ] - Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Introduced a utility component for testing with pre-populated query data.
- **Bug Fixes**
  - Enhanced input validation in user creation and update forms, including stricter length limits, email format checks, and prevention of leading/trailing spaces.
- **Tests**
  - Added comprehensive tests for user creation and update forms to verify new validation rules.
- **Style**
  - Minor formatting improvements in backend files (no user-facing changes).


packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -136,6 +144,18 @@ const CreateUserForm: React.FC<{
...
+                                            message: 'Principal Name must be less than 1000 characters.',
+                                        },
+                                        minLength: {
+                                            value: 2,
+                                            message: 'Principal name must be 2 letters or more',
+                                        },
+                                        pattern: {
+                                            value: /^[A-Za-z]+$/,
+                                            message: 'No spaces or special characters are allowed',
+                                        },
```
elikmiller, line 158: Some messages have a period and some do not. Can we pick one option and make them all consistent?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -113,6 +113,14 @@ const CreateUserForm: React.FC<{
                                     control={control}
                                     rules={{
                                         required: 'Email Address is required',
+                                        maxLength: {
+                                            value: 310,
+                                            message: 'Email address too long.',
```
elikmiller, line 118: I think we should specify the upper limit of characters in the message.

Suggest "Email Address must be less than 310 characters"

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -157,6 +177,18 @@ const CreateUserForm: React.FC<{
...
+                                            value: 1000,
+                                            message: 'First Name must be less than 1000 characters.',
+                                        },
+                                        minLength: {
+                                            value: 2,
+                                            message: 'First Name must be 2 letters or more',
+                                        },
+                                        pattern: {
+                                            value: /^[A-Za-z]+$/,
+                                            message: 'No spaces or special characters are allowed',
```
elikmiller, line 190: This rule seems too restrictive. Many people have spaces or hyphens in their name (first or last.) Suggest removing this rule or tweaking it to ensure that the field does not contain exclusively white space characters.

elikmiller (changes requested):
If there is not a separate ticket to make the same form changes for the "UpdateUserForm" component then I would additionally suggest we make those changes now to make the UX consistent between adding and updating users.

elikmiller:
@coderabbitai review

packages/javascript/bh-shared-ui/src/components/QueryTestWrapper.tsx
```diff
@@ -0,0 +1,25 @@
...
+            },
+        },
+    });
+
+    stateMap.forEach(({ key, data }) => {
+        queryClient.setQueryData(key, data);
+    });
+
+    return <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>;
+};
```
urangel: This utility is nice for setting the query data and I like the default options that are set as they seem well oriented for testing. It probably isn't well documented but the render function from the `test-utils` file has a QueryClientProvider wrapper itself and accepts a queryClient as part of the params. Would it make sense to repurpose this utility to return the `queryClient` instead of the `QueryClientProvider` that wraps the children?

Here is an example of how to pass the queryClient to the custom render function from test-utils:

```typescript
        const queryClient = setupQueryClient(stateMap) // setupQueryClient doesn't exist currently but the QueryTestWrapper can be turned into it

        render(
                <UpdateUserForm {...DEFAULT_PROPS} />, {queryClient}
        );
```

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -136,6 +144,21 @@ const CreateUserForm: React.FC<{
...
                                     rules={{
                                         required: 'Principal Name is required',
+                                        maxLength: {
+                                            value: 1000,
+                                            message: 'Principal Name must be less than 1000 characters',
+                                        },
+                                        minLength: {
+                                            value: 2,
+                                            message: 'Principal Name must be 2 characters or more',
+                                        },
```
urangel: Since the min and max length requirements are the same for Principal Name, First Name, and Last Name for both the create and update forms it may be useful to create constants for these values so that they can be updated all at once more easily if the need arises in the future. These constants could be defined somewhere like bhce/packages/javascript/bh-shared-ui/src/constants.ts and imported into both the create and update forms.

In constants.ts:
```typescript
export const MAX_NAME_LENGTH = 1000;
export const MIN_NAME_LENGTH = 2;
```

In CreateUserForm.tsx (and also in UpdateUserForm.tsx):
```typescript
import { MAX_NAME_LENGTH, MIN_NAME_LENGTH } from 'the/relative/path/to/constants.ts'

...

                                        maxLength: {
                                            value: MAX_NAME_LENGTH,
                                            message: `Principal Name must be less than ${MAX_NAME_LENGTH} characters`,
                                        },
                                        minLength: {
                                            value: MIN_NAME_LENGTH,
                                            message: `Principal Name must be ${MIN_NAME_LENGTH} characters or more`,
                                        },


```

packages/javascript/bh-shared-ui/src/components/QueryTestWrapper.tsx
LucasParraF: I would check this out. Thanks for the feedback!

elikmiller: I like this idea @urangel !

urangel (approved):
Thank you for all the tests you included! 🚀

elikmiller (approved):

PR #1483, opened 2025-05-21
fix: BHE/BHCE UX parity for handling node select

TheNando:
## Description

* Unselect node on stage click
* Continue to show info panel of last selected item, even if unselected
* Simplify drag logic
* Add timeout to drag start to prevent accidental drags
* Unselect selected node on tab or Cypher query change
* Don't highlight on hover, to match BHE
* Add background color to node label
* Reuse label draw logic for nodes and edges

## Motivation and Context

Resolves BED-4761

## How Has This Been Tested?

Manually tested each bullet point above.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Added support for independent highlighting of graph items, allowing highlighted and selected states to be managed separately in the graph view.
  - Enhanced event handling for graph interactions, including improved drag-and-drop behavior and consistent prevention of unwanted default actions during mouse events.

- **Refactor**
  - Simplified and unified label rendering for nodes and edges, improving label alignment, background rendering, and color handling.
  - Streamlined event handler interfaces and props for graph components, reducing complexity and improving maintainability.
  - Updated utility functions and constants for more accurate label and node positioning.
  - Refactored edge event handling to use external callbacks instead of internal selection hooks.
  - Improved hover and highlight rendering logic for nodes, removing redundant background shapes.
  - Delegated edge label rendering to an external function, removing direct canvas drawing logic.
  - Enhanced label bounding box calculations and padding for better visual consistency.
  - Consolidated event state management for dragging and highlighting, removing legacy timers and redundant state.

- **Bug Fixes**
  - Fixed issues with label rendering and highlight visuals during node dragging and selection.
  - Improved synchronization of highlighted and selected items, ensuring consistent UI updates during navigation and interaction.

### DIFF
--- a/cmd/ui/src/components/GraphEvents.tsx
+++ b/cmd/ui/src/components/GraphEvents.tsx
@@ -16,17 +16,20 @@
 
 import { useRegisterEvents, useSetSettings, useSigma } from '@react-sigma/core';
 import { useExploreSelectedItem } from 'bh-shared-ui';
-import { AbstractGraph, Attributes } from 'graphology-types';
-import { forwardRef, useCallback, useEffect, useImperativeHandle, useRef, useState } from 'react';
-import { SigmaNodeEventPayload } from 'sigma/sigma';
+import type { Attributes } from 'graphology-types';
+import { forwardRef, useCallback, useEffect, useImperativeHandle, useState } from 'react';
+import type { SigmaNodeEventPayload } from 'sigma/sigma';
+import type { Coordinates } from 'sigma/types';
 import {
+    MOUSE_BUTTON_PRIMARY,
     getEdgeDataFromKey,
     getEdgeSourceAndTargetDisplayData,
+    getNodeOffset,
     graphToFramedGraph,
     resetCamera,
 } from 'src/ducks/graph/utils';
 import { bezier } from 'src/rendering/utils/bezier';
-import { getNodeRadius } from 'src/rendering/utils/utils';
+import { getNodeRadius, preventAllDefaults } from 'src/rendering/utils/utils';
 import { useAppSelector } from 'src/store';
 import { sequentialLayout, standardLayout } from 'src/views/Explore/utils';
 
@@ -36,48 +39,50 @@ interface SigmaChartRef {
     runStandardLayout: () => void;
 }
 
-export interface GraphEventProps {
-    onDoubleClickNode?: (id: string) => void;
+interface GraphEventProps {
     onClickNode?: (id: string) => void;
     onClickEdge?: (id: string, relatedFindingType?: string | null) => void;
     onClickStage?: () => void;
-    edgeReducer?: (edge: string, data: Attributes, graph: AbstractGraph) => Attributes;
     onRightClickNode?: (event: SigmaNodeEventPayload) => void;
     showNodeLabels?: boolean;
     showEdgeLabels?: boolean;
 }
 
+/** Meta info about the currently dragged node */
+type DragMetadata = {
+    /** Id of the dragged graph node. */
+    id: string;
+
+    /**
+     * Vector from node's center to cusor position at mouse down. Used to
+     * prevent the node from jumping to the cursor at drag start.
+     */
+    offset: Coordinates;
+};
+
 export const GraphEvents = forwardRef(function GraphEvents(
     {
-        onDoubleClickNode,
         onClickNode,
         onClickEdge,
         onClickStage,
         onRightClickNode,
-        edgeReducer,
         showNodeLabels = true,
         showEdgeLabels = true,
     }: GraphEventProps,
     ref
 ) {
     const exploreLayout = useAppSelector((state) => state.global.view.exploreLayout);
-
-    const { selectedItem } = useExploreSelectedItem();
+    const { cancelHighlight, highlightedItem, selectedItem } = useExploreSelectedItem();
 
     const sigma = useSigma();
+    const graph = sigma.getGraph();
+
     const registerEvents = useRegisterEvents();
     const setSettings = useSetSettings();
 
-    const [hoveredNode, setHoveredNode] = useState<string | null>(null);
-    const [draggedNode, setDraggedNode] = useState<string | null>(null);
-    const [isDragging, setIsDragging] = useState(false);
+    const [draggedMeta, setDraggedMeta] = useState<DragMetadata | null>(null);
+    const draggedNode = draggedMeta && graph.getNodeAttributes(draggedMeta.id);
 
-    const isLongPress = useRef(false);
-    const dragTimerRef = useRef<ReturnType<typeof setTimeout>>();
-    const clickTimerRef = useRef<ReturnType<typeof setTimeout>>();
-    const prevent = useRef(false);
-
-    const graph = sigma.getGraph();
     const sigmaChartRef = ref as React.MutableRefObject<SigmaChartRef | null>;
 
     useImperativeHandle(
@@ -123,13 +128,6 @@ export const GraphEvents = forwardRef(function GraphEvents(
     const sigmaContainer = document.getElementById('sigma-container');
     const { getControlAtMidpoint, getLineLength, calculateCurveHeight } = bezier;
 
-    const clearDraggedNode = () => {
-        setDraggedNode(null);
-        setIsDragging(false);
-        clearTimeout(dragTimerRef.current);
-        isLongPress.current = false;
-    };
-
     const curvedEdgeReducer = useCallback(
         (edge: string, data: Attributes, newData: Attributes) => {
             // We calculate control points for all curved edges here and pass those along as edge attributes in both viewport and framed graph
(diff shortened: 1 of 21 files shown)


cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -26,47 +26,39 @@ import {
     resetCamera,
 } from 'src/ducks/graph/utils';
 import { bezier } from 'src/rendering/utils/bezier';
-import { getNodeRadius } from 'src/rendering/utils/utils';
+import { DRAG_DELAY, MOUSE_BUTTON_PRIMARY, getNodeRadius, preventAllDefaults } from 'src/rendering/utils/utils';
 import { sequentialLayout, standardLayout } from 'src/views/Explore/utils';
 
 export interface GraphEventProps {
-    onDoubleClickNode?: (id: string) => void;
```
TheNando, line 33: Removed unused props.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -26,47 +26,39 @@ import {
...
         onClickStage,
         onRightClickNode,
-        edgeReducer,
         showNodeLabels = true,
         showEdgeLabels = true,
     }: GraphEventProps,
     ref
 ) {
-    const { selectedItem } = useExploreSelectedItem();
+    const { cancelHighlight, highlightedItem, selectedItem } = useExploreSelectedItem();
```
TheNando, line 52: The GraphItemInformationPanel is linked to the selectedItem state. In order to keep this panel open while still allowing for the highlighted item to be unhighlighted, a separate state is used. The state is added to this hook because it needs to react to edge selection which is not triggered at this level.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -26,47 +26,39 @@ import {
...
 
     const sigma = useSigma();
     const registerEvents = useRegisterEvents();
     const setSettings = useSetSettings();
 
-    const [hoveredNode, setHoveredNode] = useState<string | null>(null);
```
TheNando, line 62: BHE only sets a highlighted state when an item is clicked, not on hover. I removed the hover highlights to match the UX.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -26,47 +26,39 @@ import {
...
     const [draggedNode, setDraggedNode] = useState<string | null>(null);
-    const [isDragging, setIsDragging] = useState(false);
+    const isDragging = !!draggedNode;
```
TheNando, line 59: Tracking this as a calculated state reduces the number of re-renders.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -26,47 +26,39 @@ import {
...
 
-    const isLongPress = useRef(false);
```
TheNando, line 66: Refactored the logic on this page to not need all these extra states.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -181,104 +161,80 @@ export const GraphEvents = forwardRef(function GraphEvents(
...
-                    // Prevent sigma to move camera:
-                    event.preventSigmaDefault();
-                    event.original.preventDefault();
-                    event.original.stopPropagation();
+                    // Prevent Sigma from moving camera
+                    preventAllDefaults(event);
                 }
             },
             doubleClickNode: (event) => {
-                if (!onDoubleClickNode) return;
```
TheNando, line 191: There was never a doubleClick handler passed in, so this extra logic could be removed.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -181,104 +161,80 @@ export const GraphEvents = forwardRef(function GraphEvents(
...
+            // Despite the name, this event only triggers on camera update, not any Sigma update
             updated: () => sigma.refresh(),
-            clickStage: () => onClickStage && onClickStage(),
+            clickStage: () => {
+                cancelHighlight();
+                onClickStage?.();
+            },
         });
     }, [
-        registerEvents,
```
TheNando, line 220: Alphabetized.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -181,104 +161,80 @@ export const GraphEvents = forwardRef(function GraphEvents(
...
-        isDragging,
         sigmaContainer,
     ]);
 
     useEffect(() => {
         setSettings({
             nodeReducer: (node, data) => {
                 const camera = sigma.getCamera();
-                const newData: Attributes = {
+                return {
```
TheNando, line 209: Type doesn't matter here as the reducer prop overrides the return value type.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -181,104 +161,80 @@ export const GraphEvents = forwardRef(function GraphEvents(
...
-                if (node === selectedItem) {
-                    newData.highlighted = true;
-                }
-
-                return newData;
             },
             edgeReducer: (edge, data) => {
-                const graph = sigma.getGraph();
                 const camera = sigma.getCamera();
                 const newData: Attributes = {
```
TheNando, line 217: However, type matters here as data is passed to custom reducers.

cmd/ui/src/components/GraphEvents.tsx
urangel: Thanks for cleaning up the cruft. To give you some context, we had attempted to switch over to sigmajs for both community and enterprise originally but the sigmajs implementation needed more time and had various limitations compared to using regraph. The attack paths page in BHE used to have double click interactivity and so from what I recall the reason this is got left here is due to trying to reach feature parity with sigmajs but then later walking that back.

cmd/ui/src/components/GraphEvents.tsx
```diff
@@ -145,33 +135,23 @@ export const GraphEvents = forwardRef(function GraphEvents(
...
-                // https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value
-                if (event.event.original.button === 0) {
-                    setDraggedNode(event.node);
-
-                    dragTimerRef.current = setTimeout(() => {
-                        isLongPress.current = true;
-                    }, 150);
+                if (event.event.original.button === MOUSE_BUTTON_PRIMARY) {
+                    // Drag after a timeout, to prevent accidental dragging
+                    dragTimerRef.current = setTimeout(() => setDraggedNode(event.node), DRAG_DELAY);
```
urangel, line 147: Attempting to drag a node seems to drag the whole graph if done quickly. I wonder if there is some different delay value or a different approach to where this feels snappier and doesn't drag the whole graph inadvertently. I worry this may be a regressive change as it feels like a poorer experience.


https://github.com/user-attachments/assets/29814ff9-348a-45b5-8385-45368b05a0cc

packages/javascript/bh-shared-ui/src/hooks/useExploreSelectedItem.tsx
```diff
@@ -20,27 +20,36 @@ import { useExploreParams } from './useExploreParams';
...
         },
-        [selectedItem, setExploreParams]
+        [setExploreParams]
     );
 
+    const cancelHighlight = useCallback(() => {
+        setExploreParams({
+            highlightedItem: null,
+        });
+    }, [setExploreParams]);
```
benwaples, line 43: Something that is not entirely apparent in BHE is that a graph item will re-highlight on the initial render of a page if that item is selected in the panel. Which is nice if you refresh and lose the graph item you were inspectin! Its also helpful when sharing a link between two users and the receiver of the link will see what item was selected! 

My other concern with tracking this state as a search param is that when a user unselects a graph item, a new entry will push to the browser history. Meaning if I back button thru history, one step will be unselecting a graph item. Which doesnt deviate from our pattern of search params matching user actions, but it does feel slightly like overkill to me, but that is just my opinion!

Rather than tracking what is highlighted, what are your thoughts on a piece of component state that would track if highlighting is enabled or disabled? Its default would be enabled, and selecting new nodes/edges would also enable highlighting as well, but selecting the stage would set highlighting to disabled. From the looks of it, the state could live in the `GraphEvents` component??

Just a suggestion tho!

benwaples (commented):
Thank you so much for all the clean up!! This is a dense part of the codebase so removing any extra unused code is super helpful and ALSO complicated. TY TY!

cmd/ui/src/components/GraphEvents.tsx
TheNando: I see. That makes sense. Thank you for the extra info.

cmd/ui/src/components/GraphEvents.tsx
TheNando: I'll try to refine this a bit more, comparing to how BHE acts.

urangel: I reread my comment today and it sounded somewhat offensive so I wanted to apologize if it came across that way. 🙇 

I haven't looked too closely at the details but however this is currently handled on main seems to handle node vs stage dragging as expected. Maybe the click handler or dragging state that were refactored afforded us the node drag behavior that we want.

cmd/ui/src/components/SigmaChart/GraphEvents.tsx
```diff
@@ -174,143 +170,107 @@ export const GraphEvents = forwardRef(function GraphEvents(
...
+        registerEvents,
         sigma,
-        draggedNode,
-        isDragging,
         sigmaContainer,
     ]);
 
-    useEffect(() => {
+    // Toggle off edge labels when dragging a node to avoid performance hit
+    useLayoutEffect(() => {
```
elikmiller, line 240: Can you tell me about using useLayoutEffect over useEffect in this case?

elikmiller: And throughout this file?

TheNando:
Comments have gotten too noisy. Closing this and reopening here: https://github.com/SpecterOps/BloodHound/pull/1549

PR #1523, opened 2025-05-30
BED-5954 feat(AGT): Add multi-tier analysis support

mistahj67:
## Description

- Update helpers for use in future multi-tier support
- Add `tier_id` query parameter for retrieval of future multi-tier analysis results
- Update tier tagging to prevent nodes from being tagged in additional tiers

## Motivation and Context

Resolves BED-5954

## How Has This Been Tested?
Locally tested

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Enhanced support for tier-based asset group tags, including new handling for tier zero and ordered tier retrieval.
  - Expanded query capabilities with new parameters for filtering by finding type and tier ID.

- **Improvements**
  - Generalized tier node search to support multiple tier types, not just tier zero.
  - Improved asset group node tagging logic to prevent nodes from being assigned multiple tiers.

- **Bug Fixes**
  - Disabled a flaky test related to selector creation in the UI.

- **Chores**
  - Updated and clarified comments and documentation in several files.
  - Removed unnecessary test data from file upload integration tests.

### DIFF
--- a/cmd/api/src/api/v2/file_uploads_integration_test.go
+++ b/cmd/api/src/api/v2/file_uploads_integration_test.go
@@ -186,7 +186,6 @@ func Test_FileUploadWorkFlowVersion6(t *testing.T) { //***
 		"v6/ingest/groups.json",
 		"v6/ingest/ous.json",
 		"v6/ingest/users.json",
-		"v6/ingest/deleted.json",
 		"v6/ingest/sessions.json",
 	})
 
@@ -261,7 +260,6 @@ func Test_CompressedFileUploadWorkFlowVersion6(t *testing.T) { //***
 		"v6/ingest/groups.json",
 		"v6/ingest/ous.json",
 		"v6/ingest/users.json",
-		"v6/ingest/deleted.json",
 		"v6/ingest/sessions.json",
 	})
 
--- a/packages/javascript/bh-shared-ui/src/views/TierManagement/Save/SelectorForm/SelectorForm.test.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/TierManagement/Save/SelectorForm/SelectorForm.test.tsx
@@ -269,7 +269,7 @@ describe('Selector Form', () => {
         });
     });
 
-    it('handles creating a new selector', async () => {
+    it.skip('handles creating a new selector', async () => {
         // Because there is no selector id path parameter in the url, the form is a create form
         // This means that none of the input fields should have any value aside from default values
         await act(async () => {


mistahj67:
@coderabbit review

cmd/api/src/api/attackpaths.go
```diff
@@ -0,0 +1,45 @@
...
+	"fmt"
+	"strconv"
+
+	"github.com/specterops/bloodhound/src/database"
+	"github.com/specterops/bloodhound/src/model"
+)
+
+func ParseTierIdWithTierZeroFallback(ctx context.Context, db database.Database, maybeTierIdParam []string) (int, error) {
+	if len(maybeTierIdParam) != 0 {
+		if tierIdParam, err := strconv.Atoi(maybeTierIdParam[0]); err != nil {
```
superlinkx, line 30: I know this makes the call site a little harder, but ultimately we'll be wanting to clean up how we handle parameters in the API anyway. We should really just send in the string, so parse out what might be a tierID ahead of time. This won't change much about this function, but it will make the signature a little more ergonomic in the future and prevents us from relying on effectively a raw query param.

cmd/api/src/api/attackpaths.go
```diff
@@ -0,0 +1,45 @@
...
+			return 0, err
+		} else if _, err = db.GetAssetGroupTag(ctx, tierIdParam); err != nil {
+			return 0, err
+		} else {
+			return tierIdParam, nil
+		}
+	}
+	if agt, err := db.GetAssetGroupTags(ctx, model.SQLFilter{SQLString: "type = ? AND position = ?", Params: []interface{}{model.AssetGroupTagTypeTier, model.AssetGroupTierZeroPosition}}); err != nil {
```
superlinkx: Style: Please use `any` instead of `interface`

This way of building the model is a little cluttered, but I don't believe we have a nice way to handle that currently and this seems to be the only place we're creating a SQL Filter by hand, so abstraction would likely be premature. No real reason to change other than to match our `any` style preference.

cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -632,27 +628,23 @@ func tagAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph
...
 				defer wg.Done()
-				if err = tagAssetGroupNodesForTag(ctx, db, graphDb, tag, additionalFilters...); err != nil {
+				if err = tagAssetGroupNodesForTag(ctx, db, graphDb, tag, cardinality.NewBitmap64(), additionalFilters...); err != nil {
 					slog.Error("AGT: Error tagging nodes", tag.ToType(), tag, "err", err)
 				}
 			}()
 		}
 
 		// Process the tier tagging synchronously
+		nodesSeen := cardinality.NewBitmap64()
```
superlinkx: It doesn't appear we ever use nodesSeen in this context. Why did we need to pull it up from the function definition if one call site is just passing a NewBitmap64 and the other is effectively doing the same thing?

cmd/api/src/database/assetgrouptags.go
```diff
@@ -210,6 +211,19 @@ func (s *BloodhoundDB) GetAssetGroupTag(ctx context.Context, assetGroupTagId int
 	}
 }
 
+func (s *BloodhoundDB) GetOrderedAssetGroupTagTiers(ctx context.Context) ([]model.AssetGroupTag, error) {
+	var tags model.AssetGroupTags
+	if result := s.db.WithContext(ctx).Raw(
+		fmt.Sprintf(
```
superlinkx, line 218: This should really be the string and use positional argument syntax to inject the tablename

cmd/api/src/api/attackpaths.go
mistahj67: Slated for fast followup, 👍

cmd/api/src/daemons/datapipe/agt.go
mistahj67: Explained offline but this is to prevent lower tiers from tagging nodes in upper tiers

cmd/api/src/database/assetgrouptags.go
mistahj67: Determined this is sadly not possible through the gorm wrapper support so to be consistent with other db callsites, this is left unchanged

superlinkx:
Once we have a successful pipeline and the PR body is filled out properly with testing instructions, I'll give an approval so we can merge early

superlinkx (approved):

packages/go/analysis/tiering/helpers.go
```diff
@@ -23,17 +23,33 @@ import (
...
+	SearchPrimaryTierNodes    graph.Criteria
+	SearchPrimaryTierNodesRel graph.Criteria
+}
+
+func SearchTierNodes(tieringEnabled bool, tierKinds ...graph.Kind) graph.Criteria {
 	if tieringEnabled {
-		return query.Kind(query.Node(), KindTagTierZero)
+		// Default to tier zero in the event no tierKinds are supplied
+		if len(tierKinds) == 0 {
+			tierKinds = append(tierKinds, KindTagTierZero)
```
computator, line 38: You already know this is empty, so you could just do a straight assignment of a new list instead of an append 🙂

cmd/api/src/database/assetgrouptags.go
```diff
@@ -210,6 +211,20 @@ func (s *BloodhoundDB) GetAssetGroupTag(ctx context.Context, assetGroupTagId int
 	}
 }
 
+// TODO Add databases tests for this
+func (s *BloodhoundDB) GetOrderedAssetGroupTagTiers(ctx context.Context) ([]model.AssetGroupTag, error) {
```
computator, line 215: This method (and it's definition in the interface) could both return `model.AssetGroupTags` instead of the explicit slice (which matches with it's implementation anyway).

cmd/api/src/database/assetgrouptags.go
```diff
@@ -210,6 +211,20 @@ func (s *BloodhoundDB) GetAssetGroupTag(ctx context.Context, assetGroupTagId int
...
+	var tags model.AssetGroupTags
+	if result := s.db.WithContext(ctx).Raw(
+		fmt.Sprintf(
+			"SELECT id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify FROM %s WHERE type = 1 AND deleted_at IS NULL ORDER BY position ASC",
```
computator, line 219: `type` shouldn't be hardcoded but we should instead use the `model.AssetGroupTagTypeTier` const, either in the sprintf, or as a query param.

cmd/api/src/api/v2/file_uploads_integration_test.go
```diff
@@ -186,7 +186,6 @@ func Test_FileUploadWorkFlowVersion6(t *testing.T) { //***
 		"v6/ingest/groups.json",
 		"v6/ingest/ous.json",
 		"v6/ingest/users.json",
-		"v6/ingest/deleted.json",
```
computator, line 189: Just verifying this was intentional?

cmd/api/src/api/attackpaths.go
computator: Query params have a `.Get("param_name")` method that returns the first value if it exists. You are only supposed to access the slice behind it if you specifically need multiple values. Would any of that affect this?

cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -632,27 +629,24 @@ func tagAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph
...
 		// Fire off the label tagging
 		wg := sync.WaitGroup{}
 		for _, tag := range labelsOrOwned {
 			// Parallelize the tagging of label nodes
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
-				if err = tagAssetGroupNodesForTag(ctx, db, graphDb, tag, additionalFilters...); err != nil {
+				// Nodes can contain multiple labels therefore there is no need to exclude here
+				if err = tagAssetGroupNodesForTag(ctx, db, graphDb, tag, cardinality.NewBitmap64(), additionalFilters...); err != nil {
```
computator, line 640: Should this be using `nodesSeen`, or is that what your comment about exclusion is saying?

computator: Ah after reading the below code it seems so. Perhaps the comment could mention that's what the empty cardinality bitmap is for.

PR #1532, opened 2025-06-02
Bed 5281 create tier

stephanieslamb:
## Description
Added database functions to support creating new tags

## Motivation and Context

Resolves BED-5281

This is needed for the multi-tier initiative.  

## How Has This Been Tested?
New unit tests have been added for additional functionality and all existing tests pass. 

Should be tested via BHE
## Screenshots (optional):
Screenshot 2025-06-04 at 10 35 59 AM
Screenshot 2025-06-04 at 11 05 49 AM
Screenshot 2025-06-04 at 11 06 33 AM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Added the ability to create asset group tags (such as tiers or labels) via a new API endpoint.
  - Introduced support for specifying tag name, type, description, certification requirement, and position when creating asset group tags.
  - Added new endpoints to retrieve asset group tags, members, and selectors with enhanced access controls.
  - Implemented position management for tier-type tags, including shifting existing tags to maintain order.
  - Added audit logging for updates related to asset group tag positions.

- **Bug Fixes**
  - Improved validation to ensure tag name and type are required when creating asset group tags.
  - Enforced position management rules for tier-type tags to prevent invalid insertions.

- **Documentation**
  - Updated API documentation to include the new endpoint and request schema for creating asset group tags.

- **Tests**
  - Expanded test coverage for creating asset group tags, including scenarios for tier and label types with position management.


cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +445,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
 	}
 	return nodes, CheckError(s.db.WithContext(ctx).Raw(fmt.Sprintf("SELECT selector_id, node_id, certified, certified_by, source, created_at, updated_at FROM %s WHERE selector_id IN ?", model.AssetGroupSelectorNode{}.TableName()), selectorIds).Find(&nodes))
 }
+
+func (s *BloodhoundDB) GetMaxTierPosition(ctx context.Context) (int32, error) {
```
computator: Just wanted to note that this function should also use the current `tx`. Otherwise it could see a different DB state since it's outside the transaction.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +445,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+		Scan(&max); result.Error != nil {
+		return 0, CheckError(result)
+	}
+
+	if !max.Valid {
+		// no rows in table
+		return 0, nil
+	}
+
+	return max.Int32, nil
```
computator: This might be easier
```suggestion
	return max.ValueOrZero(), nil
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +445,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+	case ShiftDown:
+		positionOp = ">"
+	default:
+		return fmt.Errorf("invalid shift direction")
+	}
+
+	// get affected rows
+	var tags []model.AssetGroupTag
+	if err := s.db.WithContext(ctx).
+		Where(fmt.Sprintf("type = ? AND position %s ? AND position > 1", positionOp), 1, position.Int32).
```
computator: This should use the const for the type ID

computator (commented):
Added a few notes while I was reading thru the shifting code 🙂

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,19 +293,49 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
-		var kindId int
-		if result := tx.Raw(fmt.Sprintf("INSERT INTO %s (name) VALUES (?) RETURNING id", kindTable), tag.ToKind()).Scan(&kindId); result.Error != nil {
-			return CheckError(result)
-		} else if result := tx.Raw(fmt.Sprintf(`
-			INSERT INTO %s (type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify)
-			VALUES (?, ?, ?, ?, NOW(), ?, NOW(), ?, ?, ?)
-			RETURNING id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify`,
-			tag.TableName()),
-			tagType, kindId, name, description, userIdStr, userIdStr, position, requireCertify).Scan(&tag); result.Error != nil {
+		if tag.Type == 1 {
```
computator: Oh this should use the const too

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +445,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+
+	if !max.Valid {
+		// no rows in table
+		return 0, nil
+	}
+
+	return max.Int32, nil
+}
+
+func (s *BloodhoundDB) CascadeShiftTierPositions(ctx context.Context, tx *gorm.DB, user model.User, position null.Int32, direction ShiftDirection) error {
```
computator: ok one more 😅. I'm just noticing that even though this is passed `tx` it doesn't use it but just accesses the DB outside the transaction instead.

cmd/api/src/database/assetgrouptags.go
computator: Ok so I realized that since these methods are being called using the `bhdb` object (from `bhdb := NewBloodhoundDB(tx, s.idResolver)`) that it means the `BloodhoundDB` being passed to the function receiver as `s` is referring to `bhdb` and thus already referencing the `tx` instead of the global database reference outside of the transaction. In other words, never mind, this is correct!

cmd/api/src/database/assetgrouptags.go
computator: Per my other comment about using the transaction, in this context `s` actually refers to the transaction so this is correct! The `tx` parameter is unneeded tho!

https://github.com/SpecterOps/BloodHound/pull/1532/files/61670744ea726f8663b28a2c132e9206a5a36a2e#r2131103626

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +295,48 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
-		if result := tx.Raw(fmt.Sprintf("INSERT INTO %s (name) VALUES (?) RETURNING id", kindTable), tag.ToKind()).Scan(&kindId); result.Error != nil {
-			return CheckError(result)
-		} else if result := tx.Raw(fmt.Sprintf(`
-			INSERT INTO %s (type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify)
-			VALUES (?, ?, ?, ?, NOW(), ?, NOW(), ?, ?, ?)
-			RETURNING id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify`,
-			tag.TableName()),
-			tagType, kindId, name, description, userIdStr, userIdStr, position, requireCertify).Scan(&tag); result.Error != nil {
+		if tag.Type == model.AssetGroupTagTypeTier {
+			// positionProvided := tag.Position.Valid
```
AD7ZJ:
```suggestion

```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +295,48 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
+		}
+
+		query := fmt.Sprintf(`
+	WITH inserted_kind AS (
+		INSERT INTO %s (name) VALUES (?) RETURNING id
+	)
+	INSERT INTO %s (type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify)
+	VALUES (?, (SELECT id FROM inserted_kind), ?, ?, NOW(), ?, NOW(), ?, ?, ?)
+	RETURNING id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify
+`, kindTable, tag.TableName())
```
AD7ZJ: Sorry to nit but you might indent this another level or two :)

cmd/api/src/api/registration/v2.go
```diff
@@ -171,18 +171,20 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
...
+		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupTag).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
+		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupMembersByTag).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
+		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/{%s}", api.URIPathVariableAssetGroupTagID, api.URIPathVariableAssetGroupTagMemberID), resources.GetAssetGroupTagMemberInfo).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
+		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/counts", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupTagMemberCountsByKind).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
```
mistahj67: Thanks for order these easier to consume, might even add line breaks for the different resources. 
That said, order here matters with the first taking precedence. If GET `/members/{%s}` is first, GET `/counts` will be handled by the former, we'll want to swap this

```suggestion
		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/counts", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupTagMemberCountsByKind).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/{%s}", api.URIPathVariableAssetGroupTagID, api.URIPathVariableAssetGroupTagMemberID), resources.GetAssetGroupTagMemberInfo).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +451,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+	// update each and create history record
+	for _, tag := range tags {
+		var (
+			auditEntry = model.AuditEntry{
+				Action: model.AuditLogActionUpdateAssetGroupTag,
+				Model:  &tag, // Pointer is required to ensure success log contains updated fields after transaction
+			}
+		)
+
+		if err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
```
AD7ZJ, line 469: I actually am not sure so take this with a grain of salt - is it OK to have nested `AuditableTransaction`s? This would be inside of the one on line 295.

cmd/api/src/database/assetgrouptags.go
mistahj67:
```suggestion
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +295,48 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
+
+			if !positionProvided {
+				if position, err := bhdb.GetMaxTierPosition(ctx, tx); err != nil {
+					return err
+				} else {
+					tag.Position = null.Int32From(position + 1)
+				}
+			}
+
+			if positionProvided && tag.Position == null.Int32From(1) {
```
mistahj67: Not sure its possible but if the above somehow maths to position 1 such as if an error occurs and it returns 0, this check would be skipped
```suggestion
			if tag.Position == null.Int32From(1) {
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +295,48 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
+				return fmt.Errorf("cannot create a new tier 0")
```
mistahj67: nit 
```suggestion
				return fmt.Errorf("cannot create a tier at position 1")
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +295,48 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
+			}
+
+			if positionProvided {
```
mistahj67: Not related but this was the line that got me thinking, we might need to calculate the max position regardless and ensure that the position provided is not more than 1 increment higher than the max position

cmd/api/src/database/assetgrouptags.go
```diff
@@ -19,15 +19,20 @@ package database
...
 	"gorm.io/gorm"
 )
 
+type ShiftDirection int
+
 const (
-	kindTable = "kind"
+	kindTable                = "kind"
+	ShiftUp   ShiftDirection = 1
+	ShiftDown ShiftDirection = -1
```
mistahj67: If these are only limited to db pkg, it'd be a good idea to keep it unexported
```suggestion
	shiftUp   ShiftDirection = 1
	shiftDown ShiftDirection = -1
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -36,6 +41,8 @@ type AssetGroupTagData interface {
 	GetAssetGroupTag(ctx context.Context, assetGroupTagId int) (model.AssetGroupTag, error)
 	GetAssetGroupTags(ctx context.Context, sqlFilter model.SQLFilter) (model.AssetGroupTags, error)
 	GetAssetGroupTagForSelection(ctx context.Context) ([]model.AssetGroupTag, error)
+	GetMaxTierPosition(ctx context.Context, tx *gorm.DB) (int32, error)
+	CascadeShiftTierPositions(ctx context.Context, tx *gorm.DB, user model.User, position null.Int32, direction ShiftDirection) error
```
mistahj67: These two methods are used only inside of the db packages and only called with the `bhdb := NewBloodhoundDB(tx, s.idResolver)` workaround, I think the tx is likely redundant here and we can safely remove them. 

```suggestion
	GetMaxTierPosition(ctx context.Context) (int32, error)
	CascadeShiftTierPositions(ctx context.Context, user model.User, position null.Int32, direction ShiftDirection) error
```

cmd/api/src/database/assetgrouptags.go
stephanieslamb: 🙃

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +451,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+	for _, tag := range tags {
+		var (
+			auditEntry = model.AuditEntry{
+				Action: model.AuditLogActionUpdateAssetGroupTag,
+				Model:  &tag, // Pointer is required to ensure success log contains updated fields after transaction
+			}
+		)
+
+		if err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+			bhdb := NewBloodhoundDB(tx, s.idResolver)
```
mistahj67, line 470: If it's possible, I think we should confirm that nesting tx's cascade's rollbacks up in the event this tx fails. I _think_ it'll be safe but want to verify all the same

cmd/api/src/database/assetgrouptags.go
stephanieslamb: So I wasn't sure on this either 😅  I always thought no.  This was the shared function that @ALCooper12 and I were working on together and I pulled this latest update from her changes.  It does work as expected, but unsure if it's best practice?

cmd/api/src/model/audit.go
```diff
@@ -207,11 +208,12 @@ type Auditable interface {
...
-	Action   AuditLogAction
-	Model    Auditable
-	Status   AuditLogEntryStatus
-	ErrorMsg string
+	CommitID    uuid.UUID
+	Action      AuditLogAction
+	Model       Auditable
+	Status      AuditLogEntryStatus
+	ErrorMsg    string
+	ShouldAudit func() bool
```
mistahj67: 👀

cmd/api/src/api/registration/v2.go
stephanieslamb: updated

cmd/api/src/database/assetgrouptags.go
stephanieslamb: removed

cmd/api/src/database/assetgrouptags.go
stephanieslamb: updated

cmd/api/src/database/assetgrouptags.go
stephanieslamb: updated

cmd/api/src/database/assetgrouptags.go
stephanieslamb: updated to always check the max value.  nice catch!

cmd/api/src/database/assetgrouptags.go
stephanieslamb: updated

cmd/api/src/database/assetgrouptags.go
stephanieslamb: removed

cmd/api/src/model/audit.go
stephanieslamb: this was left over.  removing

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +457,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+			if direction == shiftUp {
+				tag.Position.Int32++
+			} else {
+				tag.Position.Int32--
+			}
+
+			tag.UpdatedAt = time.Now()
+			tag.UpdatedBy = user.ID.String()
+
+			if err := s.db.WithContext(ctx).Save(&tag).Error; err != nil {
```
computator: Just noting that this needs to use `tx` like it does in @ALCooper12's PR otherwise this gets run in the outer transaction's context instead of the inner one

stephanieslamb: my bad. got this one confused with another code review comment.  reverted!

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +457,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+	case shiftDown:
+		positionOp = ">"
+	default:
+		return fmt.Errorf("invalid shift direction")
+	}
+
+	// get affected rows
+	var tags []model.AssetGroupTag
+	if err := s.db.WithContext(ctx).
+		Where(fmt.Sprintf("type = ? AND position %s ? AND position > 1", positionOp), model.AssetGroupTagTypeTier, position.Int32).
```
mistahj67: We likely should put an `AND deleted_at IS NULL` here just to be safe.

cmd/api/src/database/assetgrouptags.go
mistahj67: Confimed, can disregard

cmd/api/src/database/assetgrouptags.go
mistahj67: Can confirm this triggers a rollback, so all good here 👍

cmd/api/src/database/assetgrouptags.go
```diff
@@ -408,3 +457,78 @@ func (s *BloodhoundDB) GetSelectorNodesBySelectorIds(ctx context.Context, select
...
+			}
+
+			tag.UpdatedAt = time.Now()
+			tag.UpdatedBy = user.ID.String()
+
+			if err := tx.WithContext(ctx).Save(&tag).Error; err != nil {
+				return fmt.Errorf("failed to update tag position: %w", err)
+			}
+
+			if err := bhdb.CreateAssetGroupHistoryRecord(ctx, user, tag.Name, model.AssetGroupHistoryActionUpdateTag, tag.ID, null.String{}, null.StringFrom(fmt.Sprintf("original position %d, updated positon %d", originalPosition, tag.Position.Int32))); err != nil {
```
mistahj67: Will verify that notes are focused on end user notes and not update diffs:

```suggestion
			if err := bhdb.CreateAssetGroupHistoryRecord(ctx, user, tag.Name, model.AssetGroupHistoryActionUpdateTag, tag.ID, null.String{}, ""); err != nil {
```

cmd/api/src/api/error.go
```diff
@@ -72,6 +72,10 @@ const (
 	ErrorResponseUserDuplicateEmail                 = "email must be unique"
 	ErrorResponseDetailsUniqueViolation             = "unique constraint was violated"
 	ErrorResponseDetailsNotImplemented              = "All good things to those who wait. Not implemented."
+	ErrorResponseAssetGroupName                     = "asset group tag name is required and no more than 250 characters"
```
mistahj67: nit
```suggestion
	ErrorResponseAssetGroupTagExceededNameLimit                     = "asset group tag name is limited to 250 characters"
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +295,54 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
+				tag.Position = null.Int32From(newMax)
+			} else {
+				// ensure the provided position is the next sequential value
+				if tag.Position.Int32 > newMax {
+					return fmt.Errorf("cannot insert tier at position %d — must be next in sequence (position %d)", tag.Position.Int32, newMax)
+				}
+
+				if tag.Position.Int32 == 1 {
+					return fmt.Errorf("cannot explicitly create a tier at position 1")
+				}
```
mistahj67: I think we need to protect against negative numbers as well
```suggestion
				if tag.Position.Int32 <= 1 {
					return fmt.Errorf("cannot explicitly create a tier at or below position 1")
				}
```

cmd/api/src/database/assetgrouptags.go
mistahj67: Can confirm, notes are for end user, not for update diffs, safe to remove that here

cmd/api/src/database/assetgrouptags.go
stephanieslamb: updated

cmd/api/src/api/error.go
stephanieslamb: updated

cmd/api/src/database/assetgrouptags.go
stephanieslamb: updated

cmd/api/src/database/assetgrouptags.go
stephanieslamb: added

cmd/api/src/database/assetgrouptags.go
```diff
@@ -286,15 +290,52 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
...
+			tagType,
+			name,
+			description,
+			userIdStr,
+			userIdStr,
+			tag.Position,
+			requireCertify,
+		).Scan(&tag); result.Error != nil {
+			if strings.Contains(result.Error.Error(), "duplicate key value violates unique constraint \"kind_name_key\"") {
+				fmt.Printf("dupe error: %v", result.Error)
```
mistahj67:
```suggestion
```

PR #1535, opened 2025-06-02
BED-3538: Explore Table View setup

benwaples:
## Description
This PR refactors graph controls into a shared component called GraphControls and implements the basic setup for ExploreTable so that we can split work between BED-3538 and BED-5891.

## Motivation and Context

Resolves: [BED-3538](https://specterops.atlassian.net/browse/BED-3538)

This PR is needed so we can split up the work for Explore Table and its Manage Columns functionality

## How Has This Been Tested?
TBD

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -0,0 +1,46 @@
...
+import useToggle from '../../hooks/useToggle';
+import { cn } from '../../utils/theme';
+import ManageColumns from './ManageColumns';
+
+interface ExploreTableProps {
+    open: boolean;
+    onClose: () => void;
+}
+
+const ExploreTable: React.FC<ExploreTableProps> = (props) => {
```
benwaples, line 28: Place holder component to be fully built in future tickets

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumns.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Button } from '@bloodhoundenterprise/doodleui';
+import React from 'react';
+import { cn } from '../../utils/theme';
+
+interface ManageColumnsProps {
+    open: boolean;
+    onClose: () => void;
+}
+
+const ManageColumns: React.FC<ManageColumnsProps> = (props) => {
```
benwaples, line 26: Place holder component to be fully built in future tickets

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+import * as exportUtils from '../../utils/exportGraphData';
+import GraphControls from './GraphControls';
+
+const exportToJsonSpy = vi.spyOn(exportUtils, 'exportToJson');
+
+const server = setupServer(
+    rest.get('/api/v2/features', (_req, res, ctx) => {
+        return res(ctx.json({}));
+    }),
+    rest.get('api/v2/customnode', (_, res, ctx) => {
```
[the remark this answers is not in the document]
benwaples: thank you

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const handleToggleAllLabels = () => {
+        if (!showNodeLabels || !showEdgeLabels) {
+            // Show All
+            if (!showNodeLabels) onToggleNodeLabels();
+            if (!showEdgeLabels) onToggleEdgeLabels();
+        } else {
+            // Hide All
+            onToggleNodeLabels();
+            onToggleEdgeLabels();
+        }
```
definitelynotagoblin, line 74: Nit: if you flip these it can read
```javascript
if (showNodeLabels && showEdgeLabels) {
    // Hide All
    onToggleNodeLabels();
    onToggleEdgeLabels();
} else {
    // Show All
    if (!showNodeLabels) onToggleNodeLabels();
    if (!showEdgeLabels) onToggleEdgeLabels();
}
```

definitelynotagoblin: I just find A && B easier to process than !A || !B personally but entirely personal pref

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
```diff
@@ -0,0 +1,41 @@
...
+    useEffect(() => {
+        if (featureFlag?.enabled && autoDisplayTableQueryCandidate) {
+            const emptyEdges = isEmpty(graphData.data.edges);
+            const containsNodes = !isEmpty(graphData.data.nodes);
+
+            if (emptyEdges && containsNodes) {
+                setSelectedLayout('table');
+            }
+        }
+    }, [autoDisplayTableQueryCandidate, featureFlag?.enabled, graphData, setSelectedLayout]);
```
definitelynotagoblin: Not an issue but actually just a question for my own edufication:

My knee-jerk reaction looking at this is that there's a useEffect dependent on a couple "use" hook values, and the effect itself updates a state var.

It looks like this causes a rerender in response to a rerender.  I think technically this is correct, but since `setSelectedLayout` is itself within a condition check, and I don't expect `featureFlag`, `graphData`, or `setSelectedLayout` to change often I also don't think this is a concern worth worrying about.

BUT - in a case where this _were_ a `useEffect` with frequently-updating dependencies and frequently updating state therein - what _would_ be the solution to address rendering on rendering?

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.tsx
benwaples: Great suggestion! Thank you!

updated

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
benwaples: Id say your initial reaction is quite accurate, although with useEffects like this where its setting state and causing a rerender, its not necessarily triggering rerenders in response to rerenders, but instead triggering rerenders in response to data changes. However, that same data changing could cause other rerenders as well which would appear as rerenders responding to rerenders.

Is one concern that `featureFlag`, and `graphData` are in the dependency array but not needed? I would agree that they dont change often, however if either did change then we would in fact want to rerun the effect. As far as `setSelectedLayout`, i agree it wont ever change in this case but thats more of a react best practice and eslint thing :/ 

Also before diving into the theoretical case, note that a useEffect running doesnt always trigger a rerender, and a rerender doesnt trigger a useEffect. The useEffect will only execute if a value in its dep array changes, and the component will rerender if its state or props change, or if the parent rerenders.

There are a couple solutions to your useEffect updating state frequently case!
1. Make sure your effect isnt updating state AND has that piece of state inside its own useEffect dep array, thats an immediate recipe for an infinite loop.
2. Conditionally check everything you can before setting the piece of state within the useEffect. That way we arent updating state unnecessarily and triggering another rerender.
3. You can try simplify any conditionals defined in the useEffect, so instead of requiring multiple values in the dep array, you only need 1. For example, `if (cond1 && cond2 && cond3) {}` would require all 3 values in the dep array. Instead of that, you can define a boolean variable outside the useEffect and run the conditional on that. Now the useEffect only updates when the calculated value changes. This method breaks if your calculation is expensive and you actually only want to run it in the effect!
```
const bool = cond1 && cond2 && cond3
useEffect(() => {
  if (bool) {}
},[bool]) 
```
4. See if any of the state being set inside your effect can be converted to determined state (aka just a normal variable that is calculated on each rerender). Now your effect wont have to set that piece of state causing a rerender.

Those would be my suggestions, but im sure there are more! Hope that helps!

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.tsx
```diff
@@ -0,0 +1,149 @@
...
+                <div className='pointer-events-auto' data-testid='explore_graph-controls'>
+                    <SearchCurrentNodes
+                        sx={{
+                            padding: 1,
+                            marginBottom: 1,
+                        }}
+                        currentNodes={currentNodes}
+                        onSelect={(node) => {
+                            onSearchedNodeClick(node);
+                            toggleCurrentSearch();
```
dcairnsspecterops: For what it's worth, I tend to prefer explicitly setting `false` for something like this, since it seems like there's never a time when we want onSelect to cause the current search to open up. The fear is that we may have situations in the future where the current search happens to already be closed when this onSelect fires, so the user ends up accidentally opening it on search and they end up stuck there.

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
dcairnsspecterops:
```
            if (emptyEdges && containsNodes && selectedLayout !== 'table') {
                setSelectedLayout('table');
            }
```

Would it be worth checking whether the layout is already `table`?

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.tsx
benwaples: ahhh great call, thank you!!

Updated!!

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
benwaples: updated the auto display logic! Part of those updated included enabling the hook only when there isnt a layout already, so checking for `selectedLayout` occurs higher up in the callstack!

cmd/ui/src/ducks/global/reducer.ts
```diff
@@ -40,6 +42,9 @@ const globalViewReducer = (state = initialGlobalState, action: types.GlobalViewA
             draft.darkMode = action.darkMode;
         } else if (action.type === types.GLOBAL_SET_EXPLORE_LAYOUT) {
             draft.exploreLayout = action.exploreLayout;
+            draft.isExploreTableSelected = false;
+        } else if (action.type === types.GLOBAL_SET_IS_EXPLORE_TABLE_SELECTED) {
+            draft.isExploreTableSelected = action.isExploreTableSelected;
         }
```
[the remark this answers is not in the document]
benwaples: The values of exploreLayout does not include 'table', so that race condition shouldnt be possible. However, resettting the isExploreTableSelected at this point is unnecessary. Ill update to remove the `draft.isExploreTableSelected = false;`

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
```diff
@@ -0,0 +1,87 @@
...
+            const shouldAutoDisplay = emptyEdges && containsNodes;
+
+            if (shouldAutoDisplay) {
+                // set triggered to true so that we dont try to reopen the table until a new query is executed.
+                setHasTriggered(true);
+            }
+            // This will automatically open/close the table
+            setAutoDisplayTable(shouldAutoDisplay);
+        }
+    }, [autoDisplayTableQueryCandidate, containsNodes, emptyEdges, enabled, featureFlag?.enabled, setAutoDisplayTable]);
```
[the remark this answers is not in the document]
benwaples: We want to skip all codepaths that could call setAutoDisplay if the query stops being a candidate, otherwise we would be toggling that state for queries that were not candidates for auto display

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+        await waitFor(() => expect(result.current[0]).toBe(false));
+    });
+
+    it('sets autoDisplayTable to false if query returns no results', async () => {
+        const initialRoute = '?searchType=cypher&cypherSearch=YQ%3D%3D';
+        const { result } = setup({ initialRoute });
+
+        await waitFor(() => expect(result.current[0]).toBe(false));
+    });
```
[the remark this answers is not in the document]
benwaples: thanks!

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/queries/cypherSearch.ts
```diff
@@ -37,7 +37,7 @@ export const cypherSearchGraphQuery = (paramOptions: Partial<ExploreQueryParams>
     return {
         ...sharedGraphQueryOptions,
         queryKey: [ExploreGraphQueryKey, searchType, cypherSearch],
-        queryFn: ({ signal }) => apiClient.cypherSearch(decoded, { signal }).then((res) => res.data),
+        queryFn: ({ signal }) => apiClient.cypherSearch(decoded, { signal }, true).then((res) => res.data),
         retry: false,
         enabled: !!(searchType && cypherSearch),
```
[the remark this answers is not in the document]
benwaples: thanks!

jvacca-specterops (approved):

PR #1542, opened 2025-06-03
BED-5660 - Fix pinch zoom issues in BHCE Explore tab

dcairnsspecterops:
## Description
This PR prevents zoom magnifying by overriding the `wheel` browser magnify event (and making the event non-passive), which is the browser's default response to the pinch zoom gesture. Note that this PR only prevents pinch magnification if the pinch occurs within the graph: for the rest of the app, to prevent accessibility concerns, the user can still pinch to magnify.

This PR slows the zoom speed when the user uses pinch zoom or scroll zoom.

This meant overriding the default wheel handler to change the ZOOMING_RATIO constant, which is a private member of in the sigma source code, inaccessible by parameters.

## Motivation and Context
Resolves: https://specterops.atlassian.net/browse/BED-5660

## How Has This Been Tested?
This PR has been manually tests, with a video recording to prove the results.

https://github.com/user-attachments/assets/c5b421c8-d3a0-433b-8956-65d5e44ae0f1

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

## Summary by CodeRabbit

- **New Features**
  - Introduced a new hook to disable browser zoom on mouse wheel events for specific elements.
  - Improved mouse wheel zoom handling in chart components, including smoother zoom animation and reduced zoom speed.
- **Refactor**
  - Centralized and updated wheel event handling logic for chart edge interactions to enhance user experience.


cmd/ui/src/components/GraphEdgeEvents.tsx
```diff
@@ -160,8 +161,17 @@ const GraphEdgeEvents: FC = () => {
         [sigma, mouseCanvas, edgeLabelsCanvas, setExploreSelectedItem, sigmaContainer]
     );
 
+    useEffect(() => {
+        egdeEventsRef?.current?.addEventListener('wheel', (e) => {
+            e.preventDefault();
+            handleEdgeEvents(e);
+        }, { passive: false });
```
dcairnsspecterops: Turns out `{ passive: false }` was the secret sauce. We can only prevent default if the event listener is explicitly non-passive, which cannot be achieved (far as I can tell) without imperatively setting the event listener like so

packages/javascript/bh-shared-ui/src/hooks/useDisablePinchZoomOnElement.ts
```diff
@@ -0,0 +1,27 @@
+import { useCallback } from 'react';
+
+/*
+ *  This function returns a callback to be applied to the ref={} prop of an HTML element
+ *  which you want to disable browzer magnifying zoom (in our case, the explore graph).
+ *  This function accepts a callback to be applied to the wheel handler.
+ */
```
dcairnsspecterops: Waiting for a ref.current HTML element to exist was causing problems. This blog recommends useCallback as a solution:
https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs

(And even recommends pulling the function outside the scope of react, but I wanted to enable a callback argument that can depend on hooks)

cmd/ui/src/components/sigma-functions.ts
```diff
@@ -0,0 +1,99 @@
+import { MouseCaptor } from 'sigma';
+import { Coordinates, MouseCoords, WheelCoords } from 'sigma/types';
+
+/*
+ * This code in this file is copied from the sigma source code, deprecated-v2 branch:
```
dcairnsspecterops: :'(

cmd/ui/src/components/sigma-functions.ts
```diff
@@ -0,0 +1,99 @@
...
+function getWheelDeltaFromSigma(e: WheelEvent): number {
+    // TODO: check those ratios again to ensure a clean Chrome/Firefox compat
+    if (typeof e.deltaY !== 'undefined') return (e.deltaY * -3) / 360;
+
+    if (typeof e.detail !== 'undefined') return e.detail / -9;
+
+    throw new Error('Captor: could not extract delta from event.');
+}
+
+export default function handleWheelFromSigma(this: MouseCaptor, e: WheelEvent): void {
```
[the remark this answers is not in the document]
dcairnsspecterops: image

Seems to me these properties exist in our sigma version

cmd/ui/src/components/GraphEdgeEvents.tsx
```diff
@@ -154,14 +155,21 @@ const GraphEdgeEvents: FC = () => {
...
             });
+
             mouseCanvas.dispatchEvent(customEvent);
             sigma.scheduleRefresh();
         },
         [sigma, mouseCanvas, edgeLabelsCanvas, setExploreSelectedItem, sigmaContainer]
     );
 
+    const edgeEventsRef = useCreateDisableZoomRef<HTMLCanvasElement>((e) => {
+        handleWheelFromSigma.call(sigma.getMouseCaptor(), e);
```
dcairnsspecterops: Blast from the past right here. I needed to make sure that `this` pointed to the correct instance so that function calls deep in the library would have the correct context.

jvacca-specterops (approved):
well done!

StephenHinck:
This works _soooo_ much better than it does today - well done!

PR #1561, opened 2025-06-06
feat(bed-5656): create AZRoleApprover edge for post processing

neumachen:
## Description

## Motivation and Context

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/go/analysis/azure/post.go
```diff
@@ -930,3 +930,33 @@ func UserRoleAssignments(ctx context.Context, db graph.Database) (*analysis.Atom
...
 }
+
+// CreateAZRoleApproverEdge ... TBD
+func CreateAZRoleApproverEdge(ctx context.Context, db graph.Database) (graph.NodeSet, error) {
+	var nodeSet graph.NodeSet
+	if err := db.ReadTransaction(ctx, func(tx graph.Transaction) error {
+		var err error
+		if nodeSet, err = ops.FetchNodeSet(tx.Nodes().Filterf(func() graph.Criteria {
+			return query.And(
+				query.Kind(query.Node(), azure.Tenant),
```
mvlipka: I think you can get rid of the `Tenant` here
The requirements ask for iterating over the role policies for a tenant, but we don't seem to use any information from the tenant for this edge.
Might be worth checking in with Research, but I think we can just simplify this to all `AZRole` nodes

packages/go/analysis/azure/post.go
```diff
@@ -930,3 +930,33 @@ func UserRoleAssignments(ctx context.Context, db graph.Database) (*analysis.Atom
...
+				query.Kind(query.Node(), azure.Role),
+				query.Equals(
+					query.NodeProperty(azure.EndUserAssignmentRequiresApproval.String()),
+					true,
+				),
+				query.IsNotNull(
+					query.NodeProperty(azure.EndUserAssignmentUserApprovers.String()),
```
mvlipka: We'll also need to check the `EndUserAssignmentGroupApprovers` and grab their IDs

Might be easier to do in a second query, though

packages/go/analysis/azure/post.go
```diff
@@ -930,3 +930,33 @@ func UserRoleAssignments(ctx context.Context, db graph.Database) (*analysis.Atom
...
+				),
+			)
+		})); err != nil {
+			return err
+		}
+		// TODO: create edge for AZRoleApprover from the nodeSet
```
mvlipka: You can view the NTLM post to view how to do this:

https://github.com/SpecterOps/BloodHound/blob/c0511982c478940ae9411bb9d5aa5d2472181970/packages/go/analysis/ad/ntlm.go#L190

To create the channel

And then pipe in the relationship:
https://github.com/SpecterOps/BloodHound/blob/c0511982c478940ae9411bb9d5aa5d2472181970/packages/go/analysis/ad/ntlm.go#L792-L795

mvlipka: There's also some examples in the Azure `post.go`
https://github.com/SpecterOps/BloodHound/blob/7705d4a5bb31b4db4ec1989a2cf3185eb4dad975/packages/go/analysis/azure/post.go#L261-L282

packages/go/analysis/azure/post.go
```diff
@@ -930,3 +930,153 @@
...
+									azure.Entity,
+								)
+								if err != nil {
+									return err
+								}
+								fetchedNode = createdNode
+							} else {
+								return err
+							}
+						}
```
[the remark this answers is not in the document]
neumachen: Can you elaborate with an example?

packages/go/analysis/azure/post.go
```diff
@@ -930,3 +930,159 @@ func UserRoleAssignments(ctx context.Context, db graph.Database) (*analysis.Atom
...
+			principalIDs := append(userApproversID, groupApproversID...)
+			if len(principalIDs) == 0 {
+				// Step 3b: primaryApprovers is null/empty
+				// Step 3b.i: Use tenantId from this AZRole (already have tenantID)
+				// Step 3b.ii: Find GlobalAdmin AND PrivilegedRoleAdmin roles in this tenant
+				//             and create an AZRoleApprover edge from each to this role.
+				err = db.ReadTransaction(ctx, func(tx graph.Transaction) error {
+					fetchedNodes, err := ops.FetchNodeSet(tx.Nodes().Filterf(func() graph.Criteria {
+						return query.And(
+							query.Equals(query.NodeProperty(azure.TenantID.String()), tenantID),
```
mvlipka: I don't think this will return the expected results.
This is stating that the node must be multiple types, which shouldn't return any results since a node can't be all 3 of these at once.

I'd double check with research as well, but I believe this should be checking for a Role that is either GlobalAdmin or PrivilegedRoleAdmin, not where the node's kind is equal to both.

mvlipka (changes requested):

packages/go/analysis/azure/azrole/approver.go
```diff
@@ -0,0 +1,150 @@
...
+		return err
+	}
+
+	var fetchedAZRoles graph.NodeSet
+	err = db.ReadTransaction(ctx, func(tx graph.Transaction) error {
+		fetchedNodes, err := ops.FetchNodeSet(tx.Nodes().Filterf(func() graph.Criteria {
+			return query.And(
+				// Step 1: Kind = AZRole and tenantId matches
+				query.Kind(query.Node(), azure.Role),
+				query.Equals(query.NodeProperty(azure.TenantID.String()), tenantID),
```
mvlipka: This should be the `objectid` instead of the tenant id

```
tenantObjectID, err := tenantNode.Properties.Get(common.ObjectID.String()).String()
	if err != nil {
		return err
	}
```

packages/go/analysis/azure/azrole/approver.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func handleDefaultAdminRoles(ctx context.Context, tx graph.Transaction, outC chan<- analysis.CreatePostRelationshipJob, principalID string, fetchedAZRole *graph.Node) error {
```
mvlipka: `principalID` -> `tenantID`

packages/go/analysis/azure/azrole/approver.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, principalID := range principalIDs {
+		fetchedNode, err := tx.Nodes().Filterf(func() graph.Criteria {
+			return query.And(
+				query.Kind(query.Node(), azure.Entity),
+				query.Equals(query.NodeProperty(common.ObjectID.String()), principalID),
+			)
+		}).First()
+
+		var nodeID graph.ID
+		if err != nil {
```
mvlipka: This can be cleaned up:

```
		var nodeID graph.ID
		if err != nil {
			if graph.IsErrNotFound(err) {
				slog.WarnContext(ctx, fmt.Sprintf("Entity node not found for principal ID: %s, skipping edge creation", principalID))
				continue
			}

			return err
		}

		nodeID = fetchedNode.ID
```

packages/go/analysis/azure/role_approver.go
```diff
@@ -0,0 +1,179 @@
...
+				),
+				// Step 2: primaryApprovers (user or group) is not null
+				query.Or(
+					query.IsNotNull(
+						query.NodeProperty(azure.EndUserAssignmentUserApprovers.String()),
+					),
+					query.IsNotNull(
+						query.NodeProperty(azure.EndUserAssignmentGroupApprovers.String()),
+					),
+				),
```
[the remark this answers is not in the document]
mvlipka: @neumachen I believe we talked about this last week and you confirmed the behavior in DAWGs was correct and that the current query should be fine, but want to add reasoning here for later context

PR #1562, opened 2025-06-06
Make datapipe a bit more reusable

kpom-specter:
## Description
Refactor(datapipe): Move the datapipe logic that is the pipeline awayfrom the daemon/state managment. pipeline.go includes all business logic how how the steps execute. datapipe.go owns the order and timing.

## Motivation and Context
Resolves BED-5905

## How Has This Been Tested?

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Improved data pipeline management with a new pipeline interface for streamlined ingestion, analysis, pruning, and deletion operations.
  * Enhanced status tracking with new pipeline status values for more granular monitoring.

* **Bug Fixes**
  * Improved error logging consistency in pipeline-related operations.

* **Refactor**
  * Simplified and modularized the data pipeline daemon, delegating responsibilities to a dedicated pipeline component.
  * Refactored status update methods for clearer separation of concerns and easier maintenance.
  * Updated method signatures and return types for improved clarity and consistency.
  * Removed obsolete feature flag to streamline configuration.

* **Tests**
  * Updated and extended tests to reflect changes in status handling and method signatures.

### DIFF
--- a/cmd/api/src/model/appcfg/flag.go
+++ b/cmd/api/src/model/appcfg/flag.go
@@ -31,7 +31,6 @@ const (
 	FeatureAzureSupport               = "azure_support"
 	FeatureEntityPanelCaching         = "entity_panel_cache"
 	FeatureAdcs                       = "adcs"
-	FeaturePGMigrationDualIngest      = "pg_migration_dual_ingest"
 	FeatureClearGraphData             = "clear_graph_data"
 	FeatureRiskExposureNewCalculation = "risk_exposure_new_calculation"
 	FeatureFedRAMPEULA                = "fedramp_eula"


cmd/api/src/daemons/datapipe/pipeline.go
```diff
@@ -0,0 +1,170 @@
...
+		} else {
+			job.TotalFiles += totalFiles
+			job.FailedFiles += totalFailed
+
+			if err = db.UpdateIngestJob(ctx, job); err != nil {
+				slog.ErrorContext(ctx, fmt.Sprintf("Failed to update number of failed files for ingest job ID %d: %v", job.ID, err))
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
kpom-specter: This is a single worker environment

cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -139,118 +68,53 @@
...
-			job.TotalFiles += totalFiles
-			job.FailedFiles += totalFailed
-
-			if err = db.UpdateIngestJob(ctx, job); err != nil {
-				slog.ErrorContext(ctx, fmt.Sprintf("Failed to update number of failed files for ingest job ID %d: %v", job.ID, err))
-			}
-		}
-	}
+	action(pipelineContext)
 }
```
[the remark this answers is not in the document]
kpom-specter: This is not how the system is designed. Any status changes to the job will be communicated elsewhere. Datapipe is the only thing that manages it's state.

cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -18,115 +18,44 @@ package datapipe
...
-	"github.com/specterops/bloodhound/src/services/job"
-	"github.com/specterops/bloodhound/src/services/upload"
 )
 
 const (
 	pruningInterval = time.Hour * 24
 )
 
+// PipelineInterface defines methods that operate on instance state.
+// These methods are not static and require a fully initialized pipeline.
```
brandonshearin: this comment is a little obtuse-- can you clarify what you mean by "instance state" and "these methods are not static."

consider colocating this interface definition with the implementation in `pipeline.go`

kpom-specter: I'll fix the comment. 

The argument against this is that that pipeline.go shouldn't define the interface. The datapipe says "give me an object with this behavior". The fact that they are both called pipeline is convention, but the contract is with the consumer.  (I had it there initially as well; it's somewhate non-intuitive)

cmd/api/src/services/entrypoint.go
```diff
@@ -104,10 +104,12 @@ func Entrypoint(ctx context.Context, cfg config.Configuration, connections boots
 	} else if ingestSchema, err := upload.LoadIngestSchema(); err != nil {
 		return nil, fmt.Errorf("failed to load OpenGraph schema: %w", err)
 	} else {
+
+		var pipeline = datapipe.NewPipeline(ctx, cfg, connections.RDMS, connections.Graph, graphQueryCache, ingestSchema)
```
urangel: This would be nice to move into the var block below imo

cmd/api/src/services/entrypoint.go
```diff
@@ -104,10 +104,12 @@ func Entrypoint(ctx context.Context, cfg config.Configuration, connections boots
...
 		var (
 			graphQuery     = queries.NewGraphQuery(connections.Graph, graphQueryCache, cfg)
 			authorizer     = auth.NewAuthorizer(connections.RDMS)
-			datapipeDaemon = datapipe.NewDaemon(ctx, cfg, connections, graphQueryCache, time.Duration(cfg.DatapipeInterval)*time.Second, ingestSchema)
+			datapipeDaemon = datapipe.NewDaemon(pipeline, 0, time.Duration(cfg.DatapipeInterval)*time.Second, connections.RDMS)
```
urangel: The random `0` argument is a bit opaque. While it is easy enough to hover on the `NewDaemon` func or go to it's definition to understand what the second parameter relates to I would lean toward declaring a local var named `startDelay` with a value of `0` and pass that instead.

cmd/api/src/daemons/datapipe/pipeline.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2023 Specter Ops, Inc.
```
ddlees: Update copyright year

cmd/api/src/daemons/datapipe/pipeline.go
```diff
@@ -0,0 +1,190 @@
...
+	return s.PruneData(ctx)
+}
+
+// This handles the deletion of data if the customer requests it. I'm not sure why the defers
+// are stacked the way the are, but this should keep all current functionality.
+func (s *BHCEPipeline) DeleteData(ctx context.Context) bool {
+
+	defer func() {
+		_ = s.db.DeleteAnalysisRequest(ctx)
+		_ = s.db.RequestAnalysis(ctx, "datapipe")
```
ddlees, line 77: Consider either incorporating recovery logic or adding a comment to point out where recovery is happening

cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -18,240 +18,110 @@ package datapipe
...
-	"github.com/specterops/dawgs/graph"
 )
 
 const (
 	pruningInterval = time.Hour * 24
 )
 
+// Pipeline defines instance methods that operate on pipeline state.
+// These methods require a fully initialized Pipeline instance including
+// graph and db connections. Whatever is neeeded by the pipe to do the work
```
bsheth711: "neeeded" typo

cmd/api/src/daemons/datapipe/pipeline.go
superlinkx: As in panic recovery? We likely do need to improve that but it wasn't something datapipe daemon was responsible for previously. I agree in general we could do better panic recovery in the app, but most panics are the sorts of things that should propagate up to the daemon root and restart the daemon when encountered imo, which may also be a bit out of scope.

cmd/api/src/daemons/datapipe/pipeline.go
```diff
@@ -0,0 +1,190 @@
...
+			// This is cacheclearing. The analysis is still successful here
+			if _, err := s.db.GetFlagByKey(ctx, appcfg.FeatureEntityPanelCaching); err != nil {
+				slog.ErrorContext(ctx, fmt.Sprintf("Error retrieving entity panel caching flag: %v", err))
+			} else {
+				if err := s.cache.Reset(); err != nil {
+					slog.Error(fmt.Sprintf("Error while resetting the cache: %v", err))
+				} else {
+					slog.Info("Cache successfully reset by datapipe daemon")
+				}
+			}
```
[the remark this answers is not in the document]
superlinkx: CodeRabbit is right, I'm going to add a check that the flag is actually enabled here

cmd/api/src/daemons/datapipe/pipeline.go
ddlees: I'm assuming this was copy-pasta'd from the original logic so definitely a comment and not a required change for this PR.

superlinkx: Makes sense, just wanted to check. I think I might create a ticket for us to eventually do a proper sweep of our code to identify good panic recovery locations

cmd/api/src/daemons/datapipe/pipeline.go
```diff
@@ -0,0 +1,190 @@
...
+			if err = db.UpdateIngestJob(ctx, job); err != nil {
+				slog.ErrorContext(ctx, fmt.Sprintf("Failed to update number of failed files for ingest job ID %d: %v", job.ID, err))
+			}
+		}
+	}
+}
+
+// If the pipeline needs to do anything to the context, this is called before each other pipeline stage
+func (s *BHCEPipeline) IsActive(ctx context.Context, status model.DatapipeStatus) (bool, context.Context) {
+	return true, ctx
```
ddlees, line 163: Was this meant to have more to it? Always returning `true` means the early exit in `Daemon.WithDatapipeStatus` will never happen.

If this is a placeholder for out of scope work please leave a comment.

cmd/api/src/daemons/datapipe/datapipe.go
```diff
@@ -18,240 +18,110 @@ package datapipe
...
-		expectedFiles := make([]string, len(ingestTasks))
-
-		for idx, ingestTask := range ingestTasks {
-			expectedFiles[idx] = ingestTask.FileName
-		}
-
-		go s.orphanedFileSweeper.Clear(s.ctx, expectedFiles)
+// Any function can be wrapped with a datapipe lock, giving it the status. If everything locks
+// the datapipe through this same wrapper, it should always defer the idle status after.
+func (s *Daemon) WithDatapipeStatus(ctx context.Context, status model.DatapipeStatus, action func(context.Context) bool) {
```
ddlees: There's a redundant pattern I see in the actions passed into `WithDataPipeStatus`:

``` go
slog.ErrorContext(ctx, fmt.Sprintf("I'm logging an error: %v", err))
return false
```

Consider changing the action type to `func(context.Context) error`. This way actions can simply return an error:

``` go
return fmt.Errorf("I encountered an error: %v", err)
```

And `WithDatapipeStatus` can handle the logging:

``` go
defer func() {
    successful := actionErr == nil
    if !successful {
        slog.ErrorContext(pipelineContext, "datapipe action failed", "error", actionErr)
    }
    ...
}
```

cmd/api/src/daemons/datapipe/pipeline.go
bsheth711: I believe it is like this because in BHCE there is only a single pipeline, so the singular pipeline will always be active by default. This is in contrast to BHE, where there may be multiple pipelines and the implementation is different.

superlinkx: Correct, this is due to BHCE not having an HA component and is expected to work exactly as written.

cmd/api/src/daemons/datapipe/datapipe.go
superlinkx: Agreed, I'll take a swing at this (I'm taking this over for kpom while he's out)

ddlees (approved):

cmd/api/src/daemons/datapipe/pipeline.go
ddlees: Fair enough. Might be worth changing the method name/interface to `IsPrimary` for clarity. This reads "Is the pipeline active?" which I carries a different meaning IMO.

PR #1564, opened 2025-06-07
BED-3538 - Explore Tab Table View Table

dcairnsspecterops:
## Motivation and Context

Resolves  [BED-3538](https://specterops.atlassian.net/browse/BED-5992)

This PR adds a table controls header, hidden behind a feature flag, and accounts for initial munging of data for the table view.

## How Has This Been Tested?

This has been tested manually.

https://github.com/user-attachments/assets/5dfeccaa-9a49-4693-befe-62db8782b365

Feature flag off:

https://github.com/user-attachments/assets/f0eb0924-0614-4cab-a8e9-ff75595e21dc

## Screenshots (optional):

image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Introduced a feature flag to control access to the Explore Table view in the graph exploration interface.
  - Enhanced the Explore Table with dynamic column generation, integrated table controls (search, download, expand, manage columns, close), and improved UI with icons and action buttons.
  - Added support for including node properties in table data when enabled.

- **Bug Fixes**
  - Table view selection and rendering are now correctly disabled when the feature flag is off.

- **Documentation**
  - Updated component and type definitions for improved clarity and type safety.

- **Refactor**
  - Improved data handling and type support in table and graph query components.

[BED-3538]: https://specterops.atlassian.net/browse/BED-3538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
```diff
@@ -44,6 +44,8 @@ export const useExploreTableAutoDisplay = ({ enabled }: UseExploreTableAutoDispl
     const isCypherSearch = searchType === 'cypher';
     const autoDisplayTableQueryCandidate = !!(
         isCypherSearch && // auto display only on cypher search
+        // TODO: was this deleted? It was part of a merge conflict
+        !hasTriggered && // check that it hasnt already triggered for this query
```
dcairnsspecterops: TODO: This was part of a merge conflict. Check: Was this deleted in a prior PR?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+                    resultsCount={mungedData?.length}
+                    SearchInputProps={{
+                        onChange: (e) => setSearchInput(e.target.value),
+                        value: searchInput,
+                        placeholder: 'Search',
+                    }}
+                />
+                <DataTable
+                    className='h-full'
+                    // TableProps={{
```
dcairnsspecterops, line 109: To be uncommented when https://github.com/SpecterOps/DoodleUI/pull/52 merges

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -90,7 +97,7 @@ const GraphView: FC = () => {
         const graph = new MultiDirectedGraph();
 
         const hideNodes = displayTable;
-        initGraph(graph, items, theme, darkMode, customIcons.data ?? {}, hideNodes);
+        if (!hideNodes) initGraph(graph, items, theme, darkMode, customIcons.data ?? {}, hideNodes);
```
dcairnsspecterops, line 100: Does it seem okay to conditionally call this function? If so, I'll remove the hideNodes parameter in the initGraph function definition.

Previously, this rendered grey lines on my view behind the table, but conditionally calling the function seems to work.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+    open?: boolean;
+    onClose?: () => void;
+    data?: Record<string, TData>;
 }
 
-const ExploreTable: React.FC<ExploreTableProps> = (props) => {
-    const { open, onClose } = props;
+const ExploreTable = <TData extends HasData>({ data, open, onClose }: ExploreTableProps<TData>) => {
+    const [searchInput, setSearchInput] = useState('');
+    const mungedData = useMemo(
```
dcairnsspecterops, line 36: Will move this to a function later--this is the initial PR to get the data rendering in a table.

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -53,15 +54,21 @@ const GraphView: FC = () => {
...
     const exploreLayout = useAppSelector((state) => state.global.view.exploreLayout);
-    const isExploreTableSelected = useAppSelector((state) => state.global.view.isExploreTableSelected);
+    let isExploreTableSelected = useAppSelector((state) => state.global.view.isExploreTableSelected);
 
+    if (!tableViewFeatureFlag?.enabled) {
+        isExploreTableSelected = false;
+    }
+
+    const includeProperties = !!isExploreTableSelected;
+    const graphQuery = useSigmaExploreGraph(includeProperties);
```
benwaples, line 71: should we move up the auotDisplayTable state definition so we can includeProperties if the the table end up autodisplaying?

packages/javascript/bh-shared-ui/src/components/ExploreTable/TableControls.tsx
```diff
@@ -0,0 +1,78 @@
+import { Button, Input, InputProps } from '@bloodhoundenterprise/doodleui';
+import { faClose, faDownload, faExpand, faSearch } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import React from 'react';
+import { cn } from '../../utils';
+
+const ICON_CLASSES = 'cursor-pointer bg-slate-200 p-2 h-4 w-4 rounded-full';
+
+export const TableControls = React.forwardRef<
```
benwaples, line 9: I see this component is set up to receive a ref, however, im not seeing one passed to it in `ExploreTable`. Just checking, Is the forwardRef still needed here?

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -33,26 +34,32 @@ export function exploreGraphQueryFactory(paramOptions: Partial<ExploreQueryParam
...
+        case 'cypher':
+            return cypherSearchQuery;
         default:
             return fallbackQuery;
     }
 }
 
 // Hook for maintaining the top level graph query powering the explore page
-export const useExploreGraph = () => {
+export const useExploreGraph = (includeProperties?: boolean) => {
```
benwaples, line 49: is the reason for plumbing `includeProperties` through to the cypherSearch because we are thinking that the added properties calc should not execute when we dont need it because its expensive? Definitely would be nice to have this mechanism as we add support for the rest of the APIs!

packages/javascript/js-client-library/src/client.ts
```diff
@@ -122,7 +122,7 @@ class BHEAPIClient {
     cypherSearch = (query: string, options?: RequestOptions, includeProperties?: boolean) => {
         return this.baseClient.post<GraphResponse>(
             '/api/v2/graphs/cypher',
-            { query, include_properties: includeProperties },
+            { query, include_properties: includeProperties || false },
```
benwaples, line 125: Are we setting the fallback here to be more explicit? Or is there another reason that its needed that im missing?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
-import { cn } from '../../utils/theme';
-import ManageColumns from './ManageColumns';
-
-interface ExploreTableProps {
-    open: boolean;
-    onClose: () => void;
+import { ColumnDef, DataTable } from '@bloodhoundenterprise/doodleui';
+import { faEllipsis } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { capitalize } from 'lodash';
```
benwaples, line 20: I think we need to add this to the shared-ui rollup to avoid an unresolved dep during build, mind changing the import to `import capitalize from 'lodash/capitalize'` and `'lodash/capitalize'`?

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
dcairnsspecterops: That was my belief! Seems like if it's not more expensive, we would just add the property to all responses and not wait for the parameter

packages/javascript/js-client-library/src/client.ts
dcairnsspecterops: Good question. I can probably remove this, huh? I was probably trying to satisfy some typing thing, but if it wasn't breaking before, there's no reason for it to be breaking now

cmd/ui/src/views/Explore/GraphView.tsx
dcairnsspecterops: Yes! If it's okay with you, I've done this in the follow-up PR:

image

https://github.com/SpecterOps/BloodHound/pull/1602/files#diff-e4c6f9fa00aa3adb90c61e5307227a33ac52491ec5d26ddeedc414f4484912eaL76

cmd/ui/src/views/Explore/GraphView.tsx
maffkipp: This seems fine to me, that function is adding all our incoming API data to the empty `MultiDirectedGraph` that we just initialized before it gets set to the state that our `SigmaChart` component reads from. So in the case where you skip it, sigma ends up just trying to render an empty graph object.

That being said, the fact that we are mutating that `graph` variable does make this initialization step somewhat hard to parse, so it may be worth a minor refactor if that seems viable for this PR.

cmd/ui/src/views/Explore/GraphView.tsx
benwaples: 🤌  thank you thats great!

cmd/ui/src/views/Explore/GraphView.tsx
dcairnsspecterops: Sounds like a good idea! I'm wrapping up a PR dependent on this one. Would it be okay with you if I add this suggestion to the scope of the one I'm working on? @maffkipp https://github.com/SpecterOps/BloodHound/pull/1602

benwaples: Agreed with Wes and this seems like a better solution that physically hiding the nodes! It also matches how we have to set up BHE to hide nodes on the graph.

@maffkipp The refactor is a great suggestion, im going to say its out of scope for this PR because this PR is a bit of a blocker for some other work and we could complete that refactor separately!

benwaples: 👉 👈

packages/javascript/js-client-library/src/client.ts
benwaples: IMO we can remove it because its already an optional param so its fallback is naturally falsey, unless you are getting a type issue in which case keep it!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+    onClose?: () => void;
+    data?: Record<string, TData>;
 }
 
-const ExploreTable: React.FC<ExploreTableProps> = (props) => {
-    const { open, onClose } = props;
+const ExploreTable = <TData extends HasData>({ data, open, onClose }: ExploreTableProps<TData>) => {
+    const [searchInput, setSearchInput] = useState('');
+    const mungedData = useMemo(
+        () => (data && Object.keys(data).map((id) => ({ ...data?.[id]?.data, id }))) || [],
```
benwaples, line 37: This `data` will include edges (if I scroll to the bottom of the table I see rows of undefined), and we want to only display nodes for MVP. Do you mind adding a filter for only nodes here, we have a `isNode` util that could work here!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+                    onExpandClick={() => console.log('expand icon clicked')}
+                    onManageColumnsClick={() => console.log('manage columns button clicked')}
+                    onCloseClick={onClose}
+                    tableName='Results'
+                    resultsCount={mungedData?.length}
+                    SearchInputProps={{
+                        onChange: (e) => setSearchInput(e.target.value),
+                        value: searchInput,
+                        placeholder: 'Search',
+                    }}
```
benwaples, line 105: Something about this is causing the search input to update really slowly. The display is to spec (which is the intent of this PR) so when we go to wire up these other callbacks, we should take a minute to dive into the search input!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+                );
+            },
+        },
+    ];
+
+    const columns: ColumnDef<any, any>[] = useMemo(
+        () =>
+            firstItem
+                ? // If column order exists in redux/localStorage, use that
+                  Object.keys(firstItem).map((key: any) => {
```
benwaples, line 74: We will need to remove `nodetype` from the possible keys here as its already displayed but via the icon!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+        },
+        {
+            accessorKey: 'nonTierZeroPrincipal',
+            header: () => {
+                return <span className='dark:text-neutral-light-1'>Non Tier Zero Principal</span>;
+            },
+            cell: ({ row }) => {
+                return (
+                    <div className='flex justify-center items-center relative'>
+                        <NodeIcon nodeType={row?.original?.nodetype || 'N/A'} />
```
benwaples, line 63: does "N/A" render? IIRC NodeIcon has a fallback (i believe just a ? icon), which we should be able to fallback to here if the nodetype is not recognized!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+            cell: () => (
+                <button className='pl-4'>
+                    <FontAwesomeIcon icon={faEllipsis} className='rotate-90 dark:text-neutral-light-1' />
+                </button>
+            ),
+        },
+        {
+            accessorKey: 'nonTierZeroPrincipal',
+            header: () => {
+                return <span className='dark:text-neutral-light-1'>Non Tier Zero Principal</span>;
```
benwaples, line 58: Should the header text here be `Type`?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
 
-    const [openManageColumns, toggleOpenManageColumns] = useToggle(false);
+    const labelsMap = makeFormattedObjectInfoFieldsMap(firstItem);
+
+    const initialColumns: ColumnDef<any, any>[] = [
+        {
+            accessorKey: '',
+            id: 'action-menu',
+            cell: () => (
+                <button className='pl-4'>
```
benwaples, line 50: I believe the rest of the buttons in the app are Doodle `Button` components (with a few exceptions), however our current variants dont support this mostly unstyled `button` so it would be more effort to use the Doodle Button because we would need to remove all the styles and only use `pl-4`. 

I am tempted to add a plain/empty/no-style variant so we at least have a common component in the off chance we need to make a change to all buttons across the app.

Ill run this by the rest of the UI folks so we have a consensus, but no worries making a change on this PR! If we need to update this, we can do it in future work.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops: Probably!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops, line 50: Good call, this was an oversight.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,31 +14,111 @@
...
+
+    const columns: ColumnDef<any, any>[] = useMemo(
+        () =>
+            firstItem
+                ? // If column order exists in redux/localStorage, use that
+                  Object.keys(firstItem).map((key: any) => {
+                      return {
+                          accessorKey: key,
+                          header: labelsMap?.[key]?.label || capitalize(key),
+                          cell: (info: any) => String(info.getValue()),
```
benwaples, line 78: We should use the `format` util from `packages/javascript/bh-shared-ui/src/utils/entityInfoDisplay.ts` so that the cell values appear the same as they would in the entity panel!

benwaples (approved):
Looks and works great! Ive left comments for the things that we will need to address, however the base functionality of the PR is working as intended so Im okay with merging as is and addressing changes with upcoming PRs because this codepath is safely behind a development feature flag.

PR #1567, opened 2025-06-09
BED-5946 Ingests test data, retrieves nodes and edges, & transforms into generic ingestible JSON file

kpowderly:
## Description

An additional CLI subcommand was added to the StBernard Package to ingest test files, queries nodes and edges from database, and transform this data into a generic ingestible file used for Integration testing.

## Motivation and Context

Resolves [BED-5976](https://specterops.atlassian.net/browse/BED-5976)

Integration test tooling

## How Has This Been Tested?

Will be tested with new integration tests

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-5976]: https://specterops.atlassian.net/browse/BED-5976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **New Features**
  - Added a CLI command to ingest JSON test files into a PostgreSQL database and export a generic graph file.
  - Support for specifying input directory and output file for graph data transformation.
- **Documentation**
  - Updated README with usage instructions, environment variable requirements, and examples for the graph command.
- **Chores**
  - Added a debug configuration for the graph command to aid development and testing.
  - Introduced a new environment variable for PostgreSQL connection configuration.
  - Added a new recipe to automate graph ingestion with default settings.

### DIFF
--- a/packages/go/stbernard/command/graph/graph.go
+++ b/packages/go/stbernard/command/graph/graph.go
@@ -118,7 +118,6 @@ func (s *command) Run() error {
 	} else if err := os.MkdirAll(filepath.Dir(s.outfile), 0755); err != nil {
 		return fmt.Errorf("creating output directory: %w", err)
 	} else {
-		slog.Warn("outfile: %v", s.outfile)
 		return os.WriteFile(s.outfile, jsonBytes, 0644)
 	}
 }


arrows.json
bsheth711, line 1: Should this be added to the .gitignore?

kpowderly:
@coderabbitai summary

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,344 @@
...
+	for _, node := range nodes {
+		name, err := node.Properties.Get(common.Name.String()).String()
+		if err != nil || name == "" {
+			name, err = node.Properties.Get(common.ObjectID.String()).String()
+			if err != nil {
+				return Arrows{}, err
+			}
+		}
+
+		var labels = make([]string, 0, 4)
```
bsheth711: Can we initialize with capacity of len(node.Kinds)?

bsheth711 (approved):

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+	return nil
+}
+
+// Run generate command
+func (s *command) Run() error {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var (
+		outPath = filepath.Join(s.path, "arrows.json")
```
superlinkx: We should allow for the filename to also be part of the path we take in, rather than expecting a directory and appending our own filename to it.

.gitignore
```diff
@@ -267,6 +267,9 @@ go.work.sum
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+# St Bernard
+arrows.json
```
superlinkx: Since we're abandoning arrows.app as a viable strategy for viewing these results, we should instead focus on a format that's generic to our own database. Perhaps `graph.json`.

Additionally, if we set the default path to `workspaceRoot/tmp/...` we won't need to worry about git ignoring the output. Most usages are likely to specify exactly where the output goes, since the current target is for test fixtures.

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+	"github.com/specterops/bloodhound/slicesext"
+	"github.com/specterops/bloodhound/src/migrations"
+	"github.com/specterops/bloodhound/src/model"
+	"github.com/specterops/bloodhound/src/services/graphify"
+	"github.com/specterops/bloodhound/src/services/upload"
+)
+
+const (
+	Name  = "graph"
+	Usage = "Ingest test files and transform graph data into an arrows file"
```
superlinkx: Suggestion: Ingest valid collection files and transform graph data into a generic graph file

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	if err := cmd.Parse(os.Args[cmdIndex+1:]); err != nil {
+		cmd.Usage()
+		return fmt.Errorf("parsing %s command: %w", Name, err)
+	}
+
+	if *path != "" {
+		s.path = *path
+	}
```
superlinkx: We can avoid the pointer pain by using `cmd.StringVar` to get our string. Since we're just trying to put the string data into `s.path`, we can remove this code altogether if we use the following for capturing the flag:

```go
cmd.StringVar(&s.path, "path", workspacePaths.Root, "destination path for generic graph file, default is {root}/tmp/graph.json")
```

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+		return fmt.Errorf("error getting ingest file paths from test directory %v", err)
+	} else if err = ingestData(ctx, ingestFilePaths, database); err != nil {
+		return fmt.Errorf("error ingesting data %v", err)
+	} else if nodes, edges, err := getNodesAndEdges(ctx, database); err != nil {
+		return fmt.Errorf("error retrieving nodes and edges from database %w", err)
+	} else if arrows, err := transformToArrows(nodes, edges); err != nil {
+		return fmt.Errorf("error transforming nodes and edges to arrows format %w", err)
+	} else if jsonBytes, err := json.MarshalIndent(arrows, "", "  "); err != nil {
+		return fmt.Errorf("error occurred while marshalling arrows into bytes %w", err)
+	} else if err := os.MkdirAll(filepath.Dir(outPath), 0o755); err != nil {
```
superlinkx: I don't think the `o` is necessary for these, 0755 is more conventional

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+		}
+	}
+
+	if len(errs) > 0 {
+		var errStrings []string
+		for _, err := range errs {
+			errStrings = append(errStrings, err.Error())
+		}
+		slog.Warn("errors occurred while ingesting files", "errors", errStrings)
+	}
```
superlinkx: We could just `return errors.Join(errs...)` and log the error further up. I'd consider any errors to be `error` level in this context, since missing ingest files is going to make the final result unusable for intended purposes

superlinkx: If we really want them to be a warn for some reason, we could `return fmt.Errorf("%w: %w", ErrIngestWarning, errors.Join(errs...))` to give a sentinel for the caller to distinguish between generic errors and this warning case (requires defining a sentinel called ErrIngestWarning)

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+func transformToArrows(nodes []*graph.Node, edges []*graph.Relationship) (Arrows, error) {
+	var arrowNodes = make([]Node, 0, len(nodes))
+	var arrowEdges = make([]Relationship, 0, len(edges))
+
+	for _, node := range nodes {
+		name, err := node.Properties.Get(common.Name.String()).String()
+		if err != nil || name == "" {
+			name, err = node.Properties.Get(common.ObjectID.String()).String()
+			if err != nil {
+				return Arrows{}, err
```
superlinkx: We could probably wrap this error to make it clear what failed:

```go
fmt.Errorf("could not retrieve name or objectid for node %d: %w", node.ID, err)
```

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+	Label      string            `json:"label"` // kind
+	From       string            `json:"fromId"`
+	To         string            `json:"toId"`
+	Properties map[string]string `json:"properties"`
+}
+
+type Arrows struct {
+	Nodes         []Node         `json:"nodes"`
+	Relationships []Relationship `json:"relationships"`
+}
```
superlinkx: Since Arrows isn't an appropriate tool for visualizing these graphs, and we ultimately need to import them, we should restructure to make compatible with our generic graph tools

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	return Arrows{
+		Nodes:         arrowNodes,
+		Relationships: arrowEdges,
+	}, nil
+}
+
+func convertProperties(input map[string]any) map[string]string {
+	output := make(map[string]string)
```
superlinkx: We know the output map should be the same size as input:

```go
var output = make(map[string]string, len(input))
```

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+		return nodes, edges, fmt.Errorf("error occurred reading the database %w", err)
+	} else {
+		return nodes, edges, nil
+	}
+}
+
+func (s *command) getIngestFilePaths() ([]string, error) {
+	var (
+		testIngestFilePath = filepath.Join("cmd", "api", "src", "test", "fixtures", "fixtures", "v6", "ingest")
+		testIngestFileDir  = filepath.Join(s.root, testIngestFilePath)
```
superlinkx: We're going to want to make this configurable, as there are multiple directories that contain ingest files we care about. Perhaps the flag should allow you to specify a directory `collectionDir` and we traverse it looking for json files and attempting to ingest them?

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+
+	if ingestFiles, err := os.ReadDir(testIngestFileDir); err != nil {
+		return []string{}, fmt.Errorf("error reading  directory %s: %w", testIngestFileDir, err)
+	} else {
+		var paths = make([]string, 0, len(ingestFiles))
+		for _, path := range ingestFiles {
+			paths = append(paths, filepath.Join(testIngestFileDir, path.Name()))
+		}
+		return paths, nil
+	}
```
superlinkx: It would be more resilient if we do a filepath.WalkDir, filtering out directories and non-JSON files, and appending only json files to the list of paths.

packages/go/stbernard/graph/transform.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+func (s *command) initializeDatabase(ctx context.Context) (graph.Database, error) {
+	var (
+		connection = s.env[environment.PostgresConnectionVarName]
+	)
+
+	if pool, err := pg.NewPool(connection); err != nil {
+		return nil, err
```
superlinkx: Adding `fmt.Errorf` wrapping will help us troubleshoot which part of initialization broke in the future

superlinkx (changes requested):

kpowderly:
@coderabbitai full review

packages/go/stbernard/command/graph/graph.go
```diff
@@ -0,0 +1,375 @@
...
+	return Graph{
+		Nodes: graphNodes,
+		Edges: graphEdges,
+	}, nil
+}
+
+func removeNullMapValues[K comparable](m map[K]any) map[K]any {
+	newMap := make(map[K]any)
+	for k, v := range m {
+		if v != nil {
```
superlinkx: Minor nit: we use a not nil check here, but a nil check in slices. We should probably standardize on one or the other (not nil allows us to avoid a `continue` so it's technically a bit more compact.

cmd/api/src/test/fixtures/fixtures/v6/ingest/deleted.json.bad
wes-mil, line 1: Can this be deleted?

packages/go/stbernard/command/graph/graph.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+}
+
+func transformGraph(nodes []*graph.Node, edges []*graph.Relationship) (Graph, error) {
+	var graphNodes = make([]Node, 0, len(nodes))
+	var graphEdges = make([]Edge, 0, len(edges))
+	var isAZBase bool
+	var isBase bool
+
+	var nodeObjectIDs = make(map[graph.ID]string, len(nodes))
```
superlinkx: Our style guide says that when possible, variables should be grouped together at the top of their function with a `var` block:

```suggestion
	var (
		isAZBase bool
		isBase bool

		graphNodes = make([]Node, 0, len(nodes))
		graphEdges = make([]Edge, 0, len(edges))
		nodeObjectIDs = make(map[graph.ID]string, len(nodes))
	)
```

superlinkx (commented):
Only other feedback is that we may want to write a `just` command to handle the default env var and simplifies the parameters needed to run the command. Something like:

`just graphify-ingest cmd/api/src/test/fixtures/fixtures/v6/ingest v6-post-ingest.json`

This will require wiring up another modifiable env var in the justfile (like we do for a couple of environment variables up at the top of the file). If we call with `@just stbernard ...` it will always call the binary from the workspace root, so keep in mind to use `just stbernard` if we want to support users of the command giving relative paths to their current working directory and not necessarily workspace root.

superlinkx (approved):

cmd/api/src/test/fixtures/fixtures/v6/ingest/deleted.json.bad
bsheth711: No, we actually found out that there's an issue ingesting deleted type, so for now we just appended .bad so it is not picked up by the tool we have created. Once ingest is fixed, we'll remove the suffix again.

packages/go/stbernard/command/graph/graph.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	return Graph{
+		Nodes: graphNodes,
+		Edges: graphEdges,
+	}, nil
+}
+
+func removeNullMapValues[K comparable](m map[K]any) map[K]any {
+	newMap := make(map[K]any)
```
bsheth711:
```suggestion
	newMap := make(map[K]any, len(m))
```

justfile
```diff
@@ -9,9 +9,14 @@
 export GOARCH := env_var_or_default("GOARCH", host_arch)
 export INTEGRATION_CONFIG_PATH := env_var_or_default("INTEGRATION_CONFIG_PATH", absolute_path("./local-harnesses/integration.config.json"))
 export SB_LOG_LEVEL := env_var_or_default("SB_LOG_LEVEL", "info")
+export SB_PG_CONNECTION := env_var_or_default("SB_PG_CONNECTION", "user=bloodhound password=bloodhoundcommunityedition dbname=bloodhound host=localhost port=65432")
 
 set positional-arguments
 
+# generate generic ingestible file with default CLI arguments
+graphify-ingest:
+  @just stbernard graph --path=cmd/api/src/test/fixtures/fixtures/v6/ingest
```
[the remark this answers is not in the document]
superlinkx: This was valid, we needed something more like:
```go
graphify-ingest path="" outfile="":
    just stbernard --path {{path}} --outfile {{outfile}}
```

Using `just` instead of `@just` will allow us to use the tool with relative paths without respect to the workspace root, which will likely be helpful.

PR #1552, opened 2025-06-05
BED-5282 feat: tier update

computator:
## Description

Added `UpdateAssetGroupTag`

## Motivation and Context

Resolves BED-5282

Adds the DB method necessary to make a BHE endpoint work

## How Has This Been Tested?

- Go Tests
- Some local testing

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- [x] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Associated an issue: https://github.com/SpecterOps/BloodHound/issues/672
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [x] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments (GoDocs / JSDocs)
- [x] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

PR #1568, opened 2025-06-09
BED-5961 feat: add analysis_enabled flag to asset_group_tags

mistahj67:
## Description

Adds the analysis_enabled flag to asset_group_tags table and sets default to false.
Enabled analysis_enabled flag for tier zero default tier

Note this is blocked until https://github.com/SpecterOps/BloodHound/pull/1552 is merged

## Motivation and Context

Resolves BED-5961

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Introduced a new "analysis_enabled" attribute for asset group tags, enabling viewing, filtering, creation, and updates where applicable.
- **Bug Fixes**
  - Updated error messages to clarify restrictions when modifying certain fields for non-tier tags.
- **Documentation**
  - Enhanced API documentation to include the new "analysis_enabled" property for asset group tags.

### DIFF
--- a/cmd/api/src/database/assetgrouptags.go
+++ b/cmd/api/src/database/assetgrouptags.go
@@ -299,10 +299,10 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
 
 	if tag.ToType() == "unknown" {
 		return model.AssetGroupTag{}, fmt.Errorf("unknown asset group tag")
-	}
-
-	if tagType != model.AssetGroupTagTypeTier && (position.Valid || requireCertify.Valid) {
+	} else if tagType != model.AssetGroupTagTypeTier && (position.Valid || requireCertify.Valid) {
 		return model.AssetGroupTag{}, fmt.Errorf("position and require_certify are limited to tiers only")
+	} else if tagType == model.AssetGroupTagTypeTier {
+		tag.AnalysisEnabled = null.BoolFrom(false)
 	}
 
 	if err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
@@ -335,8 +335,8 @@ func (s *BloodhoundDB) CreateAssetGroupTag(ctx context.Context, tagType model.As
 			WITH inserted_kind AS (
 				INSERT INTO %s (name) VALUES (?) RETURNING id
 			)
-			INSERT INTO %s (type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify)
-			VALUES (?, (SELECT id FROM inserted_kind), ?, ?, NOW(), ?, NOW(), ?, ?, ?)
+			INSERT INTO %s (type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify, analysis_enabled)
+			VALUES (?, (SELECT id FROM inserted_kind), ?, ?, NOW(), ?, NOW(), ?, ?, ?, ?)
 			RETURNING id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify, analysis_enabled
 			`, kindTable, tag.TableName())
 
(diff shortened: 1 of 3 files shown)


cmd/api/src/database/assetgrouptags.go
```diff
@@ -377,8 +377,8 @@ func (s *BloodhoundDB) UpdateAssetGroupTag(ctx context.Context, user model.User,
 		if !tag.Position.Valid {
 			return model.AssetGroupTag{}, fmt.Errorf("position is required for an existing tier")
 		}
-	} else if tag.Position.Valid || tag.RequireCertify.Valid {
-		return model.AssetGroupTag{}, fmt.Errorf("position and require_certify are limited to tiers only")
+	} else if tag.Position.Valid || tag.RequireCertify.Valid || tag.AnalysisEnabled.Bool {
+		return model.AssetGroupTag{}, fmt.Errorf("position, require_certify, analysis_enabled are limited to tiers only")
 	}
```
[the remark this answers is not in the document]
urangel: Thoughts on this? Seems we're using `.Valid` for the others. Require certify is a null.Bool too so it seems like these should follow the same pattern.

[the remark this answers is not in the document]
mistahj67: This particular db column is set to not null, so there wouldn't be a case where we'd want it to be null. I'll go ahead and swap this to just bool to prevent confusion

cmd/api/src/database/assetgrouptags.go
```diff
@@ -377,8 +377,8 @@ func (s *BloodhoundDB) UpdateAssetGroupTag(ctx context.Context, user model.User,
 		if !tag.Position.Valid {
 			return model.AssetGroupTag{}, fmt.Errorf("position is required for an existing tier")
 		}
-	} else if tag.Position.Valid || tag.RequireCertify.Valid {
-		return model.AssetGroupTag{}, fmt.Errorf("position and require_certify are limited to tiers only")
+	} else if tag.Position.Valid || tag.RequireCertify.Valid || tag.AnalysisEnabled.Bool {
+		return model.AssetGroupTag{}, fmt.Errorf("position, require_certify, analysis_enabled are limited to tiers only")
```
computator: - We do want to use `.Valid` as CR and @urangel mentioned because otherwise it is checking the value of `analysis_enabled` instead of whether or not it was specified. If we do want to allow it to be specified but only if it's false, then we should use `.ValueOrZero`
- Grammar for a list
```suggestion
	} else if tag.Position.Valid || tag.RequireCertify.Valid || tag.AnalysisEnabled.Valid {
		return model.AssetGroupTag{}, fmt.Errorf("position, require_certify, and analysis_enabled are limited to tiers only")
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -590,7 +592,7 @@ func (s *BloodhoundDB) GetAssetGroupTagForSelection(ctx context.Context) ([]mode
 		), owned AS (
 			SELECT id FROM asset_group_tags WHERE type = 3 AND deleted_at IS NULL LIMIT 1
 		)
-		SELECT id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, deleted_at, deleted_by FROM %s WHERE id IN ((SELECT id FROM tier), (SELECT id FROM owned))`,
+		SELECT id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify, analysis_enabled FROM %s WHERE id IN ((SELECT id FROM tier), (SELECT id FROM owned))`,
```
computator, line 596: Good catch with the extra fields. In my personal projects I like to use a const string for all the field names, then all the queries can reference it to make sure it's easy to update and they are all the same.

cmd/api/src/database/assetgrouptags.go
mistahj67: I replied to that thread but the field is a not null column, I swapped it to just a bool and then kept the update struct as null.Bool for valid checking

cmd/api/src/database/assetgrouptags.go
mistahj67: I can't claim all the credit here, the tests written by all were extensive enough to highlight these 🎉

cmd/api/src/database/assetgrouptags.go
mistahj67: Discussed offline, opted to swap it to null.Bool for consistency sake

PR #1583, opened 2025-06-12
BED-5278: Tiering list history records

AD7ZJ:
## Description

Creates an endpoint to return history records associated with the asset group tag feature. 

## Motivation and Context
Resolves https://specterops.atlassian.net/browse/BED-5278

## How Has This Been Tested?

Tested locally using bruno. Unit tests have also been added to verify the endpoint and database functions. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Introduced a new API endpoint to retrieve asset group tag history with support for filtering, sorting, and pagination.
  - Added comprehensive OpenAPI documentation detailing query parameters and response schemas for the history endpoint.

- **Bug Fixes**
  - Enhanced validation and error handling for invalid filters and query parameters in history retrieval requests.

- **Tests**
  - Added extensive tests covering filtering, sorting, pagination, and error cases for the asset group tag history API.

- **Documentation**
  - Updated API specifications and documentation to include the new asset group tag history endpoint and its features.

### DIFF
--- a/packages/go/openapi/doc/openapi.json
+++ b/packages/go/openapi/doc/openapi.json
@@ -6096,6 +6096,96 @@
         }
       }
     },
+    "/api/v2/asset-group-tags/search": {
+      "post": {
+        "operationId": "AssetGroupTagSearch",
+        "summary": "Search Asset Group Tags",
+        "description": "Supports searching by name on asset group tags and selectors, and by name or object ID on members.",
+        "tags": [
+          "Asset Isolation",
+          "Enterprise",
+          "Community"
+        ],
+        "requestBody": {
+          "description": "The request body for searching asset group tags.",
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "query": {
+                    "type": "string"
+                  },
+                  "tag_type": {
+                    "type": "integer"
+                  }
+                },
+                "required": [
+                  "query",
+                  "tag_type"
+                ]
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "OK",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "data": {
+                      "type": "object",
+                      "properties": {
+                        "tags": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/model.asset-group-tag"
+                          }
+                        },
+                        "selectors": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/model.asset-group-tags-selector-response"
+                          }
+                        },
+                        "members": {
+                          "type": "array",
+                          "items": {
+                            "$ref": "#/components/schemas/model.asset-group-tags-member"
+                          }
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          },
+          "400": {
+            "$ref": "#/components/responses/bad-request"
+          },
+          "401": {
+            "$ref": "#/components/responses/unauthorized"
+          },
+          "403": {
+            "$ref": "#/components/responses/forbidden"
+          },
+          "404": {
+            "$ref": "#/components/responses/not-found"
+          },
+          "429": {
+            "$ref": "#/components/responses/too-many-requests"
+          },
+          "500": {
+            "$ref": "#/components/responses/internal-server-error"
+          }
+        }
+      }
+    },
     "/api/v2/asset-group-tags/history": {
       "parameters": [
         {


cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +731,44 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+				}
+			}
+		}
+
+		if sqlFilter, err := queryFilters.BuildSQLFilter(); err != nil {
+			api.WriteErrorResponse(rCtx, api.BuildErrorResponse(http.StatusBadRequest, "error building SQL for filter", request), response)
+		} else if historyRecs, err := s.DB.GetAssetGroupHistoryRecords(rCtx, sqlFilter, false, skip, limit); err != nil && !errors.Is(err, database.ErrNotFound) {
+			api.HandleDatabaseError(request, response, err)
+		} else {
+			api.WriteResponseWrapperWithPagination(rCtx, historyRecs, limit, skip, len(historyRecs), http.StatusOK, response)
```
mistahj67: The length provided in paginated responses is intended to be the full length of the filtered responses not just the length of the specific page I believe. We'll want to update the db method to kick out both I think

cmd/api/src/database/assetgrouphistory.go
```diff
@@ -27,15 +27,50 @@ import (
 // AssetGroupHistoryData defines the methods required to interact with the asset_group_history table
 type AssetGroupHistoryData interface {
 	CreateAssetGroupHistoryRecord(ctx context.Context, actor model.User, target string, action model.AssetGroupHistoryAction, assetGroupTagId int, environmentId, note null.String) error
-	GetAssetGroupHistoryRecords(ctx context.Context) ([]model.AssetGroupHistory, error)
+	GetAssetGroupHistoryRecords(ctx context.Context, sqlFilter model.SQLFilter, sortDirectionAscending bool, skip, limit int) ([]model.AssetGroupHistory, error)
```
mistahj67: Might be worth elevating the sort up to an API parameter as well, so the endpoint can support sorting by created_at asc or desc

cmd/api/src/api/v2/assetgrouptags.go
AD7ZJ: Good catch, ty! Fixed!

cmd/api/src/database/assetgrouphistory.go
AD7ZJ: Also a good idea, I implemented the sorting logic that we used in other places. Do you think it's worth allowing sort on anything besides `created_at`? It would be easy to add the other columns although not sure why we would need them.

cmd/api/src/database/assetgrouphistory.go
```diff
@@ -27,15 +28,66 @@ import (
...
+		"SELECT COUNT(*) FROM %s%s",
+		(model.AssetGroupHistory{}).TableName(),
+		sqlFilter.SQLString)
+
+	if result := s.db.WithContext(ctx).Raw(sqlStr, sqlFilter.Params...).Find(&historyRecs); result.Error != nil {
+		return []model.AssetGroupHistory{}, 0, CheckError(result)
+	}
+
+	if limit > 0 || skip > 0 {
+		if result := s.db.WithContext(ctx).Raw(sqlCountStr, sqlFilter.Params...).Scan(&rowCount); result.Error != nil {
```
mistahj67: Nit but we can define this closer to where it's used instead of above
```suggestion
	if limit > 0 || skip > 0 {
	sqlCountStr := fmt.Sprintf(
		"SELECT COUNT(*) FROM %s%s",
		(model.AssetGroupHistory{}).TableName(),
		sqlFilter.SQLString)
		if result := s.db.WithContext(ctx).Raw(sqlCountStr, sqlFilter.Params...).Scan(&rowCount); result.Error != nil {
```

cmd/api/src/database/assetgrouphistory_test.go
```diff
@@ -41,18 +41,62 @@ func TestDatabase_CreateAndGetAssetGroupHistory(t *testing.T) {
...
+	// verify skip
+	require.Equal(t, model.AssetGroupHistoryActionCreateTag, records[0].Action)
+	require.Equal(t, model.AssetGroupHistoryActionDeleteTag, records[1].Action)
+
+	records, _, err = dbInst.GetAssetGroupHistoryRecords(testCtx, model.SQLFilter{SQLString: "action = ?", Params: []any{model.AssetGroupHistoryActionCreateTag}}, model.Sort{{Column: "created_at", Direction: model.AscendingSortDirection}}, 0, 0)
+	require.NoError(t, err)
+
+	// verify SQL filter
+	require.Len(t, records, 1)
+	require.Equal(t, model.AssetGroupHistoryActionCreateTag, records[0].Action)
```
mistahj67: Each of these should probably be a separate `t.Run()` with the comment as the test description string

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +731,50 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+
+	if queryFilters, err := model.NewQueryParameterFilterParser().ParseQueryParameterFilters(request); err != nil {
+		api.WriteErrorResponse(rCtx, api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsBadQueryParameterFilters, request), response)
+		return
+	} else if skip, err := ParseSkipQueryParameter(queryParams, 0); err != nil {
+		api.WriteErrorResponse(rCtx, ErrBadQueryParameter(request, model.PaginationQueryParameterSkip, err), response)
+	} else if limit, err := ParseOptionalLimitQueryParameter(queryParams, 100); err != nil {
+		api.WriteErrorResponse(rCtx, ErrBadQueryParameter(request, model.PaginationQueryParameterLimit, err), response)
+	} else if sort, err := api.ParseSortParameters(model.AssetGroupHistory{}, queryParams); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsColumnNotFilterable, request), response)
```
mistahj67:
```suggestion
		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsNotSortable, request), response)
```

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +731,50 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+		if len(sort) == 0 {
+			sort = model.Sort{{Column: "created_at", Direction: model.DescendingSortDirection}}
+		}
+
+		if sqlFilter, err := queryFilters.BuildSQLFilter(); err != nil {
+			api.WriteErrorResponse(rCtx, api.BuildErrorResponse(http.StatusBadRequest, "error building SQL for filter", request), response)
+		} else if historyRecs, count, err := s.DB.GetAssetGroupHistoryRecords(rCtx, sqlFilter, sort, skip, limit); err != nil && !errors.Is(err, database.ErrNotFound) {
+			api.HandleDatabaseError(request, response, err)
+		} else {
+			api.WriteResponseWrapperWithPagination(rCtx, historyRecs, limit, skip, count, http.StatusOK, response)
```
mistahj67: The response would be more extensible if we tacked on `{records: []}` instead of a pure array

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.history.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - name: environment_id
+      in: query
+      required: false
+      schema:
+        $ref: './../schemas/api.params.predicate.filter.string.yaml'
+    - name: note
+      in: query
+      required: false
+      schema:
+        $ref: './../schemas/api.params.predicate.filter.string.yaml'
```
ALCooper12: (I could be wrong) But according to your ticket: BED-5278, your inputs are supposed to be: `Start Date` and `End Date`, however the response variables are all listed as input parameters instead. I'm comparing your endpoint in the UI of the API Explorer to GET `/api/v2/posture-history/{data_type}` for reference

packages/go/openapi/src/schemas/model.asset-group-tags-history.yaml
```diff
@@ -0,0 +1,49 @@
...
+        action:
+          type: string
+        target:
+          type: string
+        asset_group_tag_id:
+          type: integer
+        environment_id:
+          type: integer
+        note:
+          type: string
```
ALCooper12: Just a few baby corrections!

```suggestion
        id:
          type: integer
          format: int64
        created_at:
          type: string
          format: date-time
        actor:
          type: string
        email:
          $ref: './null.string.yaml'
          format: email
        action:
          type: string
        target:
          type: string
        asset_group_tag_id:
          type: integer
        environment_id:
          $ref: './null.string.yaml'
        note:
          $ref: './null.string.yaml'
```

cmd/api/src/api/v2/assetgrouptags.go
AD7ZJ: Good idea, I'll do that!

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.history.yaml
AD7ZJ: I implemented the `Start Date` and `End Date` requirement by doing a filter on created_at - you would pass in `created_at=gt:<start time>&created_at=lt:<end time>`. This is the best way I know to document the filters, I just referenced how it was done on other endpoints like `/paths/asset-isolation.asset-group-tags.yaml`

packages/go/openapi/src/schemas/model.asset-group-tags-history.yaml
AD7ZJ: Good catch, ty! Fixed!

cmd/api/src/database/assetgrouphistory.go
```diff
@@ -27,15 +28,65 @@ import (
...
+	}
+
+	for _, item := range sortItems {
+		dirString := "ASC"
+		if item.Direction == model.DescendingSortDirection {
+			dirString = "DESC"
+		}
+		sortColumns = append(sortColumns, fmt.Sprintf("%s %s", item.Column, dirString))
+	}
+	sortString := "ORDER BY " + strings.Join(sortColumns, ", ")
```
mistahj67: This variable assumes the sortColumns slice is never empty. This likely needs to be built out similar to the `skipLimitString` where if nothing is supplied it will be an empty string

packages/go/openapi/src/schemas/model.asset-group-tags-history.yaml
```diff
@@ -0,0 +1,50 @@
...
+type: object
+properties:
+  count:
+    type: integer
+  limit:
+    type: integer
+  skip:
+    type: integer
+  data:
+    type: array
```
mistahj67, line 26: When the response key `records` was added it looks like the docs didn't get updated to reflect that as well.

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.history.yaml
mistahj67: Good points all around, on the one end, the ticket has clear requirements of start and end date while the `created_at` filter is more extensible in the long run. I would recommend bringing this question up to the greater engineering team / product to see which is the preferred implementation.

cmd/api/src/database/assetgrouphistory.go
mistahj67: Would also benefit from a db test for the empty sortColumns argument to prevent regression as well

packages/go/openapi/src/schemas/model.asset-group-tags-history.yaml
AD7ZJ: Oops, duh! Got it now :)

cmd/api/src/database/assetgrouphistory.go
AD7ZJ: That's a good catch, thanks!

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.history.yaml
AD7ZJ: Product and the team were OK with it as-is, so let's run with this for now :) Thanks to all for the comments and thorough checking!

PR #1601, opened 2025-06-23
build(bed-6015): optimize container build time

ddlees:
## Description

- Update Dockerfile to optimize container build time
  - Leverage remote content caching for *Hound collector downloads
  - Replace stbernard with base tooling to take better advantage of cache layers and decoupling the UI static asset build from the multi-arch container build
  - Create separate build stage for generating ldflags

## Motivation and Context

Resolves BED-6015

## How Has This Been Tested?

Successfully builds container and executes via `examples/docker-compose/docker-compose.yml`

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


dockerfiles/bloodhound.Dockerfile
```diff
@@ -88,27 +46,83 @@ RUN wget \
...
+RUN git --no-pager -c 'versionsort.suffix=-rc' tag --list v*.*.* --sort=-v:refname | head -n 1 | sed 's/^v//' | awk \
+  -F'[.+-]' \
+  -v pkg="$VERSION_PKG" \
+  '{ major = $1; minor = $2; patch = $3; pre = ""; if ($4) pre = $4; \
+    printf("-X '\''%s.majorVersion=%s'\'' ", pkg, major); \
+    printf("-X '\''%s.minorVersion=%s'\'' ", pkg, minor); \
+    printf("-X '\''%s.patchVersion=%s'\''", pkg, patch); \
+    if (pre != "") \
+      printf(" -X '\''%s.prereleaseVersion=%s'\''", pkg, pre); \
+  }' > LDFLAGS
```
superlinkx, line 92: I don't believe this is the same as what St Bernard was doing to get the correct tag. I'm fairly certain we want to be very explicit with tags, so we should validate that approach and make sure we can replicate here (even if that means having a St Bernard action we can source it from)

dockerfiles/bloodhound.Dockerfile
```diff
@@ -88,27 +46,83 @@ RUN wget \
...
+RUN --mount=type=cache,target=/go/pkg/mod go build -C cmd/api/src -o /bloodhound -ldflags "$(cat LDFLAGS)" github.com/specterops/bloodhound/src/cmd/bhapi
+
 ########
 # Package Bloodhound
 ################
 FROM gcr.io/distroless/static-debian11 AS bloodhound
 ARG SHARPHOUND_VERSION
 ARG AZUREHOUND_VERSION
 
+COPY --from=api-builder /bloodhound /opt/bloodhound /etc/bloodhound /var/log /
```
superlinkx, line 121: Do we know why we copy `/var/log`? It seems a bit odd, though it's been that way for a while from the looks

superlinkx (commented):
Just a couple things stick out, otherwise, I don't see any significant issues

dockerfiles/bloodhound.Dockerfile
superlinkx: I was wrong, this is exactly how stbernard and by extension beagle sorted out current tag

dockerfiles/bloodhound.Dockerfile
ddlees: My guess is it's an artifact from a long time ago when we were building a debian package for enterprise. I'd remove it but I'm not 100% sure we won't break something.

superlinkx: Yeah, that's fair. I'm not really sure what could possibly rely on it, but we'd need a good way to test that it doesn't break anything

superlinkx (approved):

PR #1602, opened 2025-06-23
Bed 6021 combo box

dcairnsspecterops:
## Description

## Motivation and Context

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/ui/src/views/Explore/GraphView.tsx
+++ b/cmd/ui/src/views/Explore/GraphView.tsx
@@ -49,9 +49,9 @@ import ExploreSearch from './ExploreSearch/ExploreSearch';
 import GraphItemInformationPanel from './GraphItemInformationPanel';
 import { transformIconDictionary } from './svgIcons';
 
-export const makeMap = (items) =>
+const makeMap = (items: any[]) =>
     items.reduce((acc, col) => {
-        return { ...acc, [col?.accessorKey || col?.id]: true };
+        return { ...acc, [col?.accessorKey || col?.id || 'accessor_key_unavailable']: true };
     }, {});
 
 const GraphView: FC = () => {
@@ -71,6 +71,7 @@ const GraphView: FC = () => {
 
     const handleManageColumnsChange = useCallback((items) => {
         const newItems = makeMap(items);
+
         dispatch(setVisibleExploreTableColumns(newItems));
     }, []);
 
--- a/cmd/ui/src/views/Explore/GraphView.tsx
+++ b/cmd/ui/src/views/Explore/GraphView.tsx
@@ -36,11 +36,11 @@ import { MultiDirectedGraph } from 'graphology';
 import { Attributes } from 'graphology-types';
 import { GraphNodes } from 'js-client-library';
 import isEmpty from 'lodash/isEmpty';
-import { FC, useEffect, useRef, useState } from 'react';
+import { FC, useCallback, useEffect, useRef, useState } from 'react';
 import { SigmaNodeEventPayload } from 'sigma/sigma';
 import { NoDataDialogWithLinks } from 'src/components/NoDataDialogWithLinks';
 import SigmaChart from 'src/components/SigmaChart';
-import { setExploreLayout, setIsExploreTableSelected } from 'src/ducks/global/actions';
+import { setExploreLayout, setIsExploreTableSelected, setVisibleExploreTableColumns } from 'src/ducks/global/actions';
 import { useSigmaExploreGraph } from 'src/hooks/useSigmaExploreGraph';
 import { useAppDispatch, useAppSelector } from 'src/store';
 import { initGraph } from 'src/views/Explore/utils';
@@ -48,6 +48,11 @@ import ContextMenu from './ContextMenu/ContextMenu';
 import ExploreSearch from './ExploreSearch/ExploreSearch';
 import GraphItemInformationPanel from './GraphItemInformationPanel';
 import { transformIconDictionary } from './svgIcons';
+
+export const makeMap = (items) =>
+    items.reduce((acc, col) => {
+        return { ...acc, [col?.accessorKey || col?.id]: true };
+    }, {});
 
 const GraphView: FC = () => {
     /* Hooks */
@@ -62,6 +67,12 @@ const GraphView: FC = () => {
     const darkMode = useAppSelector((state) => state.global.view.darkMode);
     const exploreLayout = useAppSelector((state) => state.global.view.exploreLayout);
     let isExploreTableSelected = useAppSelector((state) => state.global.view.isExploreTableSelected);
+    const visibleColumns = useAppSelector((state) => state.global.view.visibleExploreTableColumns);
+
+    const handleManageColumnsChange = useCallback((items) => {
+        const newItems = makeMap(items);
+        dispatch(setVisibleExploreTableColumns(newItems));
+    }, []);
 
     if (!tableViewFeatureFlag?.enabled) {
         isExploreTableSelected = false;
@@ -207,6 +218,8 @@ const GraphView: FC = () => {
                 <ExploreTable
                     data={graphQuery.data}
                     open={displayTable}
+                    visibleColumns={visibleColumns}
+                    onManageColumnsChange={handleManageColumnsChange}
                     onClose={() => {
                         setAutoDisplayTable(false);
                         dispatch(setIsExploreTableSelected(false));
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
@@ -18,12 +18,17 @@ import { ColumnDef, DataTable } from '@bloodhoundenterprise/doodleui';
 import { faEllipsis } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { capitalize } from 'lodash';
-import { useMemo, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 import { makeFormattedObjectInfoFieldsMap } from '../../utils';
 import NodeIcon from '../NodeIcon';
 import { TableControls } from './TableControls';
 
 type HasData = { data?: object };
+
+const makeMap = (items) =>
+    items.reduce((acc, col) => {
+        return { ...acc, [col?.accessorKey || col?.id]: true };
+    }, {});
 
 interface ExploreTableProps<TData extends HasData> {
     open?: boolean;
@@ -37,10 +42,35 @@ const ExploreTable = <TData extends HasData>({ data, open, onClose }: ExploreTab
         () => (data && Object.keys(data).map((id) => ({ ...data?.[id]?.data, id }))) || [],
         [data]
     );
-
     const firstItem = mungedData?.[0];
 
     const labelsMap = makeFormattedObjectInfoFieldsMap(firstItem);
+
+    const columns: ColumnDef<any, any>[] = useMemo(
+        () =>
+            firstItem
+                ? // If column order exists in redux/localStorage, use that
+                  Object.keys(firstItem).map((key: any) => {
+                      return {
+                          accessorKey: key,
+                          header: labelsMap?.[key]?.label || capitalize(key),
+                          cell: (info: any) => String(info.getValue()),
+                          id: key,
+                          size: 150,
+                      } as ColumnDef<any, any>;
+                  })
+                : [],
+        [labelsMap, firstItem]
+    );
+
+    const initialVisibleColumns = useMemo(() => makeMap(columns), [columns]);
+
+    const [visibleColumns, setVisibleColumns] = useState(initialVisibleColumns);
+
+    const handleManageColumnsChange = useCallback((items) => {
+        const newItems = makeMap(items);
+        setVisibleColumns(newItems);
+    }, []);
 
     const initialColumns: ColumnDef<any, any>[] = [
         {
@@ -67,23 +97,6 @@ const ExploreTable = <TData extends HasData>({ data, open, onClose }: ExploreTab
         },
     ];
 
-    const columns: ColumnDef<any, any>[] = useMemo(
-        () =>
-            firstItem
-                ? // If column order exists in redux/localStorage, use that
-                  Object.keys(firstItem).map((key: any) => {
-                      return {
-                          accessorKey: key,
-                          header: labelsMap?.[key]?.label || capitalize(key),
-                          cell: (info: any) => String(info.getValue()),
-                          id: key,
-                          size: 150,
-                      } as ColumnDef<any, any>;
-                  })
-                : [],
-        [labelsMap, firstItem]
-    );
-
     if (!open || !data) return null;
 
     const finalColumns = [...initialColumns, ...columns];
@@ -92,9 +105,12 @@ const ExploreTable = <TData extends HasData>({ data, open, onClose }: ExploreTab
             className={`border-2 overflow-hidden absolute z-10 bottom-16 left-4 right-4 max-h-1/2 h-[475px] bg-neutral-light-2`}>
             <div className='explore-table-container w-full h-full'>
                 <TableControls
+                    columns={columns}
+                    visibleColumns={visibleColumns}
                     onDownloadClick={() => console.log('download icon clicked')}
                     onExpandClick={() => console.log('expand icon clicked')}
                     onManageColumnsClick={() => console.log('manage columns button clicked')}
+                    onManageColumnsChange={handleManageColumnsChange}
                     onCloseClick={onClose}
                     tableName='Results'
                     resultsCount={mungedData?.length}
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
@@ -24,12 +24,12 @@ import NodeIcon from '../NodeIcon';
 import { ManageColumnsComboBoxOption } from './ManageColumnsComboBox';
 import { TableControls } from './TableControls';
 
-export const makeMap = (items) =>
-    items.reduce((acc, col) => {
-        return { ...acc, [col?.accessorKey || col?.id]: true };
-    }, {});
-
 type HasData = { data?: object };
+
+const makeMap = (items: any[]) =>
+    items.reduce((acc, col) => {
+        return { ...acc, [col?.accessorKey || col?.id || 'accessor_key_unavailable']: true };
+    }, {});
 
 interface ExploreTableProps<TData extends HasData> {
     open?: boolean;
@@ -55,7 +55,7 @@ const ExploreTable = <TData extends HasData>({
 
     const labelsMap = makeFormattedObjectInfoFieldsMap(firstItem);
 
-    const columns: ColumnDef<any, any>[] = useMemo(
+    const allColumnDefinitions: ColumnDef<any, any>[] = useMemo(
         () =>
             firstItem
                 ? // If column order exists in redux/localStorage, use that
@@ -71,6 +71,27 @@ const ExploreTable = <TData extends HasData>({
                 : [],
         [labelsMap, firstItem]
     );
+
+    const visibleColumnDefinitions = allColumnDefinitions.filter(
+        (columnDef) => visibleColumns?.[columnDef?.accessorKey]
+    );
+
+    const fallbackInitialVisibleColumns = makeMap(allColumnDefinitions);
+
+    const handleManageColumnsChange = (columns: ManageColumnsComboBoxOption[]) => {
+        if (typeof onManageColumnsChange === 'function') {
+            const parsedColumns =
+                // TODO: reconcile ColumnDef and ManageColumnsComboBoxOption types?
+                columns?.length > 0
+                    ? columns
+                    : (allColumnDefinitions.map((def) => ({
+                          ...def,
+                          id: def.accessorKey,
+                      })) as ManageColumnsComboBoxOption[]);
+
+            onManageColumnsChange(parsedColumns);
+        }
+    };
 
     const initialColumns: ColumnDef<any, any>[] = [
         {
@@ -97,22 +118,19 @@ const ExploreTable = <TData extends HasData>({
         },
     ];
 
-    const fallbackInitialColumns = makeMap(columns);
-
     if (!open || !data) return null;
 
-    const finalColumns = [...initialColumns, ...columns];
+    const finalColumns = [...initialColumns, ...visibleColumnDefinitions];
     return (
         <div
             className={`border-2 overflow-hidden absolute z-10 bottom-16 left-4 right-4 max-h-1/2 h-[475px] bg-neutral-light-2`}>
             <div className='explore-table-container w-full h-full'>
                 <TableControls
-                    columns={columns}
-                    visibleColumns={visibleColumns || fallbackInitialColumns}
+                    columns={allColumnDefinitions}
+                    visibleColumns={visibleColumns || fallbackInitialVisibleColumns}
                     onDownloadClick={() => console.log('download icon clicked')}
                     onExpandClick={() => console.log('expand icon clicked')}
-                    onManageColumnsClick={() => console.log('manage columns button clicked')}
-                    onManageColumnsChange={onManageColumnsChange}
+                    onManageColumnsChange={handleManageColumnsChange}
                     onCloseClick={onClose}
                     tableName='Results'
                     resultsCount={mungedData?.length}
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
@@ -18,25 +18,34 @@ import { ColumnDef, DataTable } from '@bloodhoundenterprise/doodleui';
 import { faEllipsis } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { capitalize } from 'lodash';
-import { useCallback, useMemo, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { makeFormattedObjectInfoFieldsMap } from '../../utils';
 import NodeIcon from '../NodeIcon';
+import { ManageColumnsComboBoxOption } from './ManageColumnsComboBox';
 import { TableControls } from './TableControls';
 
-type HasData = { data?: object };
-
-const makeMap = (items) =>
+export const makeMap = (items) =>
     items.reduce((acc, col) => {
         return { ...acc, [col?.accessorKey || col?.id]: true };
     }, {});
+
+type HasData = { data?: object };
 
 interface ExploreTableProps<TData extends HasData> {
     open?: boolean;
     onClose?: () => void;
     data?: Record<string, TData>;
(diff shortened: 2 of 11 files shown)


packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -21,27 +21,78 @@ import { capitalize } from 'lodash';
 import { useMemo, useState } from 'react';
 import { makeFormattedObjectInfoFieldsMap } from '../../utils';
 import NodeIcon from '../NodeIcon';
+import { ManageColumnsComboBoxOption } from './ManageColumnsComboBox';
 import { TableControls } from './TableControls';
 
 type HasData = { data?: object };
 
+const makeMap = (items: any[]) =>
```
dcairnsspecterops: TODO: pull this in a util with better typing

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox.tsx
```diff
@@ -0,0 +1,211 @@
+import { Button, Checkbox, Input } from '@bloodhoundenterprise/doodleui';
+import { faMinus, faPlus, faRefresh, faSearch, faThumbTack } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { UseComboboxPropGetters, useCombobox, useMultipleSelection } from 'downshift';
+import React, { useEffect, useMemo, useRef } from 'react';
+
+export type ManageColumnsComboBoxOption = { id: string; value: string; isPinned?: boolean };
```
dcairnsspecterops: Should this be reconciled with ColumnDef<TData, TKey>? We can add properties to the meta of a column def, and it might be nice to have a single interface

packages/javascript/bh-shared-ui/src/hooks/useCustomNodeKinds.tsx
```diff
@@ -51,5 +51,7 @@ export const useCustomNodeKinds = (
         staleTime: 2 * (60 * 1000),
         cacheTime: 5 * (60 * 1000),
         ...queryOptions,
+        // TODO: remove this! Custom-nodes call kept failing and causing rerenders
```
dcairnsspecterops: Just here because I was having an issue with this endpoint for some reason

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -66,5 +76,6 @@ export const useExploreGraph = (includeProperties?: boolean) => {
                 autoHideDuration: SNACKBAR_DURATION_LONG,
             });
         },
+        enabled,
```
benwaples: it seems that adding this `enabled` field at the end of the properties will override the `enabled` defined within each `queryConfig`, which results in the query trying to execute without a `queryFn` yet.

Perhaps we need to move the enabled field up above the queryConfig, or pass it to `getQueryConfig`?

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -45,8 +50,13 @@ export function exploreGraphQueryFactory(paramOptions: Partial<ExploreQueryParam
...
 }
 
+const DEFAULT_USE_EXPLORE_GRAPH_PARAMS = { includeProperties: false, enabled: true };
+
 // Hook for maintaining the top level graph query powering the explore page
-export const useExploreGraph = (includeProperties?: boolean) => {
+export const useExploreGraph = ({
+    includeProperties,
+    enabled,
+}: UseExploreGraphParams = DEFAULT_USE_EXPLORE_GRAPH_PARAMS) => {
```
benwaples, line 61: If i pass `includeProperties: true`, will enabled then be false because the fallback object isnt triggered? If so, is that intended?

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
```diff
@@ -26,13 +26,18 @@ interface UseExploreTableAutoDisplayParams {
...
+     */
+    includeProperties: boolean;
 }
 
 // This should be able to detect when the Explore Table should display automatically and when NOT to display it.
 // Auto display when current search is cypher and returned data contains nodes but not edges.
 // And dont auto display if the auto display has been closed.
 export const useExploreTableAutoDisplay = ({ enabled }: UseExploreTableAutoDisplayParams) => {
-    const { data: graphData, isFetching } = useExploreGraph();
+    const { data: graphData, isFetching } = useExploreGraph({ includeProperties: true, enabled });
```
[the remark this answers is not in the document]
benwaples: It appears the `UseExploreTableAutoDisplayParams` interface doesnt have the `includeProperties` field so I dont think this comment is accurate

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
```diff
@@ -32,7 +32,7 @@ interface UseExploreTableAutoDisplayParams {
...
```
benwaples, line 35: This appears to be the only place we are setting enabled param for the useExploreGraph hook, and im thinking it may not be needed, but I totally could be missing a use case!

If auto display table is enabled, then yes we want the useExploreGraph to be enabled. If auto display table is disabled we dont necessarily want/care to execute a graph query from this hook, however because react-query will cache the results of the graph query already, this extra call wont make an extra network request. But IMO it add a little more complexity to the useExploreGraph which Im hoping to avoid.

again its totally possible that im missing an edge case here and we need to plumb this enabled field thru!

dcairnsspecterops: From slack, for those looking in:

```
Whenever a cypher query of any kind is run, we ask "is this an auto table situation?"
When we "close" the table, we make this call again.
 Because includeProperties wasn't included, the queryKey updates
After all, our table explore query does include properties.
Which forces reevaluation of the useExploreGraph query, since the query key is new
That evaluation then triggers the table to autoDisplay
was my experience, I think
```

packages/javascript/bh-shared-ui/src/hooks/useExploreTableAutoDisplay/useExploreTableAutoDisplay.tsx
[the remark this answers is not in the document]
dcairnsspecterops: @benwaples You are right, Code Rabbit seems to be looking at an old commit, where I was passing `includeProperties` from above.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,201 @@
...
     onClose?: () => void;
-    data?: Record<string, TData>;
+    data?: Record<string, WrappedExploreTableItem>;
+    selectedColumns?: Record<string, boolean>;
+    allColumnKeys?: string[];
+    onManageColumnsChange?: (columns: ManageColumnsComboBoxOption[]) => void;
 }
 
-const ExploreTable = <TData extends HasData>({ data, open, onClose }: ExploreTableProps<TData>) => {
+const MemoDataTable = memo(DataTable);
```
benwaples: I see a handful of memoized things in here, just curious if you ran into performance issues or if they are in anticipation of performance issues?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,201 @@
...
+    allColumnKeys,
+    selectedColumns,
+}: ExploreTableProps) => {
     const [searchInput, setSearchInput] = useState('');
+    const [isExpanded, toggleIsExpanded] = useToggle(false);
+
     const mungedData = useMemo(
-        () => (data && Object.keys(data).map((id) => ({ ...data?.[id]?.data, id }))) || [],
+        () =>
+            // TODO: remove id and just use objectid for onRowClick/getRowId?
```
benwaples: this should work because the `selectedItem` param works with a db id or objectid!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops: I did! I can't remember the details, but this was an important fix for, uh, something.

dcairnsspecterops: Something along the lines of "GraphView was updating, which was forcing unnecessary re-renders of DataTable"

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,201 @@
...
+                    return (
+                        <div className='flex justify-center items-center relative'>
+                            <NodeIcon nodeType={(info.getValue() as string) || ''} />
+                        </div>
+                    );
+                },
             },
-        },
-    ];
+            ...['objectid', 'displayname'].map(makeColumnDef),
```
benwaples: looks like some nodes use the `name` property, and some use `displayname` 🙄  so even for nodes that we know the name of, im seeing a `-` under the name field. 

Looks like the data prop has a label field that is used by the graph and is present for both graph libraries (perhaps we can use some shared type for those transforming functions so any future changes know of this unexpected dependency on the `label.text` field of a node!). Could we set `displayname` to `label.text` in the mungedData objects? something like:

```
Object.entries(data).map(([key, value]) => ({
                    ...value.data,
                    id: key,
                    displayname: value.label.text,
                }))
```

dcairnsspecterops: Good call!

cmd/ui/src/hooks/useSigmaExploreGraph/useSigmaExploreGraph.tsx
```diff
@@ -15,7 +15,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { isGraphResponse, transformToFlatGraphResponse, useExploreGraph } from 'bh-shared-ui';
-import { FlatGraphResponse, GraphResponse } from 'js-client-library';
+import { FlatGraphResponse, GraphData, GraphResponse } from 'js-client-library';
```
TheNando: This is a minor optimization, but it adds up in the long run; when importing a type, you can use the "type" import syntax. This will tell the TypeScript compiler to only import the type data and not any value data. This helps with tree shaking to reduce bundle size.

```ts
import type { FlatGraphResponse, GraphData, GraphResponse } from 'js-client-library';
```

In this case, the type can be included at the front since all the module imports are types. But if there was a mix of values as well (functions, constants) you could preface just the types. In the example below, `AssetGroupTagTypesMap` is a object constant, i.e. a value, but GraphData is a type.

```ts
import { AssetGroupTagTypesMap, type GraphData } from 'js-client-library';
```

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -61,48 +63,41 @@ const GraphView: FC = () => {
...
-    const customIcons = useCustomNodeKinds({ select: transformIconDictionary });
-
-    const [autoDisplayTable, setAutoDisplayTable] = useExploreTableAutoDisplay({
-        enabled: !exploreLayout,
-    });
-    const displayTable = autoDisplayTable || !!isExploreTableSelected;
-
     useEffect(() => {
-        let items: any = graphQuery.data;
+        let items: any = graphQuery.data?.nodes;
```
TheNando, line 99: This whole block is a little goofy, not your fault since you didn't touch much here. Not a fan of reusing `items` for 3 different contexts (nodes value, empty default value, or graph items value). If this is working maybe it's best to leave it alone, but I would assume at the very least, `isEmpty(items.nodes)` can be removed since there should never be `graphQuery.data.nodes.nodes`.

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -236,8 +237,11 @@ const GraphView: FC = () => {
             <NoDataDialogWithLinks open={!graphHasData} />
             {tableViewFeatureFlag?.enabled && (
                 <ExploreTable
-                    data={graphQuery.data}
+                    data={graphQuery.data?.nodes as Record<string, StyledGraphNode>}
```
TheNando: When you use the `value as type` syntax, this is called a type assertion. You are telling TypeScript language server to forget what type it has calculated and go forward acting as if it is the provided type. Typically, this should only be reserved for when you are coming from a generic to a more specific type, like when using JSON.parse or getting URLSearchParams. In this case, it's probably more likely that the ExploreTable data prop needs its type updated to match the expected incoming data type, in this case `FlatGraphResponse`.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,201 @@
...
-import { TableControls } from './TableControls';
+import { ManageColumnsComboBoxOption } from './ManageColumnsComboBox/ManageColumnsComboBox';
+import TableControls from './TableControls';
 
-type HasData = { data?: object };
+const REQUIRED_EXPLORE_TABLE_COLUMN_KEYS = ['nodetype', 'objectid', 'displayname'];
 
-interface ExploreTableProps<TData extends HasData> {
+const requiredColumns = REQUIRED_EXPLORE_TABLE_COLUMN_KEYS.reduce(
+    (acc, curr) => ({ ...acc, [curr]: true }),
```
TheNando: CodeRabbit is on point here. If you just have a dozen entries here, no big deal, but this operation, while pure and functional, will have an exponential performance impact. The other benefit is that with CR's suggestion, you don't need the type assertion since you are no longer widening the type through the `reduce` method. So CR's sounds solid, just double check that it results in the same result.

```ts
const requiredColumns = Object.fromEntries(
    REQUIRED_EXPLORE_TABLE_COLUMN_KEYS.map(key => [key, true])
);

```

TheNando (commented):
I am still digging through this, but I'll leave some notes of what I've seen so far.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops: `fromEntries` is new to me! Good call

cmd/ui/src/views/Explore/GraphView.tsx
dcairnsspecterops: Looks like this is a vestige of a moment in development where I hadn't munged correctly. Removing the casting altogether seems to do the trick. :)

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -61,54 +62,46 @@ const GraphView: FC = () => {
...
     useEffect(() => {
-        let items: any = graphQuery.data;
+        let items: any = graphQuery.data?.nodes;
 
         if (!items && !graphQuery.isError) return;
         if (!items) items = {};
 
-        // `items` may be empty, or it may contain an empty `nodes` object
-        if (isEmpty(items) || isEmpty(items.nodes)) items = transformFlatGraphResponse(items);
+        items = transformFlatGraphResponse(items);
```
dcairnsspecterops, line 104: Removing `isEmpty(items.nodes)` (which always evaluates to `true`) breaks this. Seems to me that we always need to run items through this function? I'll keep thinking about this.

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -45,8 +50,13 @@ export function exploreGraphQueryFactory(paramOptions: Partial<ExploreQueryParam
...
 }
 
+const DEFAULT_USE_EXPLORE_GRAPH_PARAMS = { includeProperties: false, enabled: true };
+
 // Hook for maintaining the top level graph query powering the explore page
-export const useExploreGraph = (includeProperties?: boolean) => {
+export const useExploreGraph = ({
+    includeProperties = DEFAULT_USE_EXPLORE_GRAPH_PARAMS.includeProperties,
+    enabled = DEFAULT_USE_EXPLORE_GRAPH_PARAMS.enabled,
+}: UseExploreGraphParams = DEFAULT_USE_EXPLORE_GRAPH_PARAMS) => {
```
TheNando, line 61: Constants are typically a good idea, but in this case, I'd argue for simplicity.

```ts
export const useExploreGraph = ({ includeProperties = false, enabled = true }: UseExploreGraphParams) => {
```

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -28,6 +28,11 @@ import {
     relationshipSearchQuery,
 } from './queries';
 
+type UseExploreGraphParams = {
+    includeProperties?: boolean;
+    enabled?: boolean;
+};
+
 export function exploreGraphQueryFactory(paramOptions: Partial<ExploreQueryParams>): ExploreGraphQuery {
```
TheNando: It looks like `cypherSearchGraphQuery` unfortunately does not completely align with the others queries. It has an extra param of `includeProperties` which means its type is not actually `ExploreGraphQuery`. It looks like there is no easy way around that. But we can make a small improvement to the typing to reflect what's happening.

```ts
export function exploreGraphQueryFactory(paramOptions: Partial<ExploreQueryParams>): ExploreGraphQuery | CypherExploreGraphQuery {
```

With that, we could make a second small improvement to not need the type assertion on line 68

```ts
    const queryConfig =
        params?.searchType === 'cypher'
            ? query.getQueryConfig(params, includeProperties)
            : query.getQueryConfig(params);

File: packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx [hunk 18]
```diff
@@ -14,109 +14,199 @@
...
+                return value ? (
+                    <div className='h-full w-full flex justify-center items-center text-center'>
+                        <FontAwesomeIcon icon={faCheck} color='green' className='scale-125' />{' '}
+                    </div>
+                ) : (
+                    <div className='h-full w-full flex justify-center items-center text-center'>
+                        <FontAwesomeIcon icon={faCancel} color='lightgray' className='scale-125' />{' '}
+                    </div>
+                );
+            }
```
TheNando: You can use ternaries to keep this a little DRYer.
```suggestion
            if (typeof value === 'boolean') {
                return (
                    <div className='h-full w-full flex justify-center items-center text-center'>
                        <FontAwesomeIcon
                            icon={value ? faCheck : faCancel}
                            color={value ? 'green' : 'lightgray'}
                            className='scale-125'
                        />{' '}
                    </div>
                );
            }
```

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,199 @@
...
-    const firstItem = mungedData?.[0];
+    const filteredData = useMemo(
+        () =>
+            mungedData?.filter((potentialRow) => {
+                const filterKeys: (keyof GraphNode)[] = ['displayname', 'objectid'];
+                const filterTargets = filterKeys.map((filterKey) => {
+                    const stringyValue = String(potentialRow?.[filterKey]);
+
+                    return stringyValue?.toLowerCase();
+                });
```
TheNando: Two small suggestions here. The definition on line 94 can be moved outside of the component as a module constant. This way, it's not recreated for every row. If you use the ordering on line 141, you could use the const there as well, since the order would matter in that case: `...['objectid', 'displayname'].map(makeColumnDef),`.

And lines 95-99 can be compressed into a 1 liner. `String()` will always return a string value no matter what you throw at it, so no need for the optional chaining operator (?.). The optional chains are another place where the small performance overhead adds up, so ideally, they should only be used when really needed.

```ts
const filterTargets = filterKeys.map((filterKey) => String(item[filterKey]).toLowerCase());
```

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,199 @@
...
+        () => allColumnKeys?.filter((key) => !requiredColumns[key]).map(makeColumnDef) || [],
+        [allColumnKeys]
+    );
+
+    const selectedColumnDefinitions = useMemo(
+        () => nonRequiredColumnDefinitions.filter((columnDef) => selectedColumns?.[columnDef?.id || '']),
+        [nonRequiredColumnDefinitions, selectedColumns]
+    );
+
+    const requiredColumnDefinitions = useMemo(
```
TheNando: Since all the items in this value are static, instead of the memo, you can just move this to a constant outside of the component within the module scope. Then you could remove the value from the deps below to reduce some renders.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,199 @@
...
+    );
+
+    const searchInputProps = useMemo(
+        () => ({
+            onChange: handleSearchInputChange,
+            value: searchInput,
+            placeholder: 'Search',
+        }),
+        [handleSearchInputChange, searchInput]
+    );
```
TheNando: If the handleSearchInputChange method is only used in the searchInputProps, you should be able to just inline it there and it should stay static within the memo.

```tsx
    const searchInputProps = useMemo(
        () => ({
            onChange: (e: ChangeEvent<HTMLInputElement>) => setSearchInput(e.target.value),
            value: searchInput,
            placeholder: 'Search',
        }),
        [ searchInput]
    );
```

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,109 +14,199 @@
...
+        }),
+        []
+    );
+
+    const tableHeadProps: DataTableProps['TableHeadProps'] = useMemo(
+        () => ({
+            className: 'pr-4',
+        }),
+        []
     );
```
TheNando: I'm assuming these were memoized because they are object values that were causing the data table to re-render on every ExploreTable render. If so, this is another scenario you could move the values to module-scope constants above the component. `useMemo` isn't free, so in cases where you aren't using component-scope members, constants are a great go-to over memos.

packages/javascript/bh-shared-ui/src/components/ExploreTable/index.ts
```diff
@@ -15,3 +15,5 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export { default as ExploreTable } from './ExploreTable';
+export type { ManageColumnsComboBoxOption } from './ManageColumnsComboBox/ManageColumnsComboBox';
```
TheNando, line 18: This is correct, but be aware that if the exports were to include non-types, you'd want to only preface the types themselves, otherwise you would not be able to use the non-type members for anything other than static type analysis.

```ts
// myFunc will not be emitted in js, can only be used for type info
export type { MyType, myFunc } from 'utils';

// Can use myFunc for either, as it will be emitted in js
export { type MyType, myFunc } from 'utils';
```

packages/javascript/js-client-library/src/types.ts
```diff
@@ -267,12 +267,18 @@ export type GraphNode = {
     label: string;
     kind: string;
     objectId: string;
+    objectid?: string;
```
TheNando: Is it possible to manually transform these keys? I feel like this is really compromising the type's integrity, specifically this key which is just a lower-case version of another key.

packages/javascript/js-client-library/src/types.ts
```diff
@@ -267,12 +267,18 @@ export type GraphNode = {
...
     isOwnedObject: boolean;
+    nodetype?: string;
+    displayname?: string;
+    enabled?: boolean;
+    pwdlastset?: number;
+    lastlogontimestamp?: number;
     descendent_count?: number | null;
     properties?: Record<string, any>;
-};
+} & Record<string, any>;
```
TheNando: This addendum effectively says that a GraphNode object can have any combination of key-value pairs outside the definition here, really weakening the type. Could you elaborate on the issue you encountered that lead you to this? If this is to solve a specific case, we may be better served converting this to a interface or creating an extended type as to not so drastically hinder the original type.

packages/javascript/js-client-library/src/types.ts
```diff
@@ -285,15 +291,15 @@ export type GraphEdge = {
     impactPercent?: number;
     exploreGraphId?: string;
     data?: Record<string, any>;
-};
+} & Record<string, any>;
```
TheNando: Same thing here. We should avoid this, if possible. Maybe there is another solution.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops: Wow, yes, totally slipped my mind. `useMemo` is ridiculous here for simple consts, thanks for the catch

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops: Yep, an obviously good idea which, I feel like if I were the reviewer I would have caught, but somehow slipped my awareness in my own code. Thanks much.

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
dcairnsspecterops: The issue here is that the parameter needs a default, since this existing hook is already called throughout the app with no parameters. The alternative is to find every use of this hook and pass it an empty object(?), or check for the existence of the object inside the function, which ended up feeling messier than this (since now the object needs a name and we need to pull the vars inside the call)

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
dcairnsspecterops: That's much nicer, thanks.

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
TheNando: That's a fair point. 👍

packages/javascript/js-client-library/src/types.ts
dcairnsspecterops: Chatted about this on slack; this key comes from the properties object. We now type the properties object separately, and export a SpreadNodeWithProperties interface to find this compromise.

TheNando (approved):

PR #1607, opened 2025-06-24
fix: readd necessary identity kinds to fix indexed lookups for ingest

superlinkx:
## Description

Fixes non-indexed lookups for ingest queries regression introduced in #1404 

## Motivation and Context

Resolves BED-6059

A performance regression that only affected Neo4j was introduced with the mentioned PR that was expanding ingest functionality. Postgres was not affected since the way indexing works for our Postgres graphs does not rely on a Kind label to be explicitly added.

This fix patches the hole, while adding some guards to keep the newer ingest work functional.

## How Has This Been Tested?

Ran through both our example azure dataset (this was a reliable way to trigger the performance issues) and a generic ingest payload through both BloodHound and BHE with this set of fixes without errors and with expected performance and results.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/go/dawgs/drivers/neo4j/batch.go
+++ b/packages/go/dawgs/drivers/neo4j/batch.go
@@ -18,9 +18,12 @@ package neo4j
 
 import (
 	"context"
+	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/neo4j/neo4j-go-driver/v5/neo4j"
+	"github.com/specterops/bloodhound/bhlog/measure"
 	"github.com/specterops/bloodhound/dawgs/graph"
 	"github.com/specterops/bloodhound/dawgs/util/size"
 )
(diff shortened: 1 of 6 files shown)


superlinkx:
I left in a pruned set of debug logging that I think might be helpful in the future. If there are any concerns, let me know, but otherwise I think they're a good addition.

brandonshearin (approved):

packages/go/dawgs/drivers/neo4j/batch.go
```diff
@@ -72,6 +75,7 @@ func (s *batchTransaction) UpdateNodeBy(update graph.NodeUpdate) error {
 }
 
 func (s *batchTransaction) UpdateRelationshipBy(update graph.RelationshipUpdate) error {
+	defer measure.ContextMeasure(s.innerTx.ctx, slog.LevelDebug, "update relationship by", slog.String("pattern", fmt.Sprintf("(%d)-[%s]->(%d)", update.Start.ID, update.Relationship.Kind, update.End.ID)))()
```
zinic: This is done in batch which may make this output confusing as it would appear a single query might be blamed for the latency tail.

superlinkx: That's fair. It's also in dawgs so won't survive the merge anyway. I'm just gonna pull it since I need to push a new commit anyway

PR #1628, opened 2025-06-30
BED-5725 Payload Ingest Integration Test

kpowderly:
## Description

Replacement of previous integration tests to test ingest.

## Motivation and Context

Resolves [BED-5725](https://specterops.atlassian.net/browse/BED-5725)

Two integration tests were created to test ingestion.
1. Asserts the json data files with graph fixture
2. Asserts the zip file (containing the json data files) with graph fixture

A bug was identified when testing ingest so slight changes were made to ad.go. The bug caused a mismatched `name` property to be different based on the order the files were ingested. It has since been resolved.

**Flow**
a. Reuse payload files from IngestJob Test as input
b. Set up database with a consistent fake job that appropriately points to the input files
c. Validate graph matches some graph fixture

## How Has This Been Tested?

Passing integration tests

## Screenshots (optional):

Screenshot 2025-06-30 at 11 47 37 AM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-5725]: https://specterops.atlassian.net/browse/BED-5725?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive test fixture files for version 6 ingestion, covering computers, containers, domains, GPOs, groups, organizational units (OUs), sessions, and users.
  * Introduced integration tests validating ingestion of individual JSON files and ZIP archives, verifying database state against expected graphs.
  * Added a full integration test suite setup for the Graphify service with PostgreSQL and graph database initialization.

* **Bug Fixes**
  * Improved path handling and error messages during file ingestion in the command-line interface.

* **Refactor**
  * Enhanced test utilities for graph loading and assertions with improved performance and parallelism.
  * Adjusted local group property handling in GPO parsing.
  * Renamed and exported ingestion processing function for wider accessibility.

* **Chores**
  * Reordered import statements across multiple files for consistency.
  * Enhanced build script ingestion recipe with flexible parameters.
  * Updated module dependencies and added new testing dependencies.


packages/go/ein/ad.go
```diff
@@ -717,12 +717,9 @@ func ParseGPOChanges(changes GPOChanges) ParsedLocalGroupData {
 			groupID := computer.ObjectIdentifier + "-" + group.RID
 			if len(group.Members) > 0 {
 				parsedData.Nodes = append(parsedData.Nodes, IngestibleNode{
-					ObjectID: groupID,
-					// TODO: look up computer name?
-					PropertyMap: map[string]any{
-						"name": groupID,
-					},
-					Labels: []graph.Kind{ad.LocalGroup},
```
wes-mil, line 725: What was this doing before and why was it removed? I'm having trouble gathering it from context clues

packages/go/lab/generic/graph.go
```diff
@@ -90,76 +99,115 @@ func WriteGraphToDatabase(db graph.Database, g *Graph) error {
...
 			}
-		}
-	}
 
+			// assert properties
+			for expectedProperty, expectedValue := range expectedNode.Properties {
+				value, ok := node.Properties.Map[expectedProperty]
+				assert.Truef(t, ok, "could not find expected property `%s` on node `%s`", expectedProperty, node.ID)
+
+				if ok && !slices.Contains(excludedProperties, expectedProperty) {
```
wes-mil: Could consider making excluded properties a `map[string]{}`. I generally prefer hashmap lookups, but with two elements in the list I think this is just as good

packages/go/ein/ad.go
bsheth711: Setting the name property to the groupID here was causing a bug where the order of files during ingestion mattered, which should not be the case. This is because if the ous.json file was ingested first, which has the GPO changes, the resulting node's name would be the groupID, not the actual name for the group, which was the case if the computers.json file was ingested first instead. The ticket has some comments with more specific details on the bug.

packages/go/lab/generic/graph.go
bsheth711: This is something I too considered doing. However, for very small values of N, checking list membership is actually faster than the overhead of hashing

bsheth711: I tested both ways, and it is indeed faster to use a hashmap, so we will go with that

superlinkx (approved):
I'm happy with how this turned out. I'd want us to be careful to guard against these setup/teardowns getting abstracted to an outside package (the temptation will always be there, and there's a graveyard of our "integration frameworks" to show why that's a bad idea), but I think this kind of abstraction tailored per package-under-test makes a lot of sense and I like the attention to detail in how to make it reusable without over abstracting.

We may want to consider moving the setup/teardown helpers to a `graphify_integration_test.go` file, solely to help with discovery of package-wide testing tools in a common file, rather than stacked on specific tests, but I'll leave that for future discussions when the time comes.

PR #1648, opened 2025-07-03
BED-5784 - Search Asset Group Tags Endpoint

stephanieslamb:
## Description

Adding an endpoint for the ability to search tags or labels, selectors and members

## Motivation and Context

Resolves BED-5784

This change will allow searching on the privileged zones page whether on the tiers or labels page.   

## How Has This Been Tested?
New unit tests have been added for new functionality and existing unit tests pass.  (will post screenshots)

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added a new API endpoint to search asset group tags, selectors, and members by name or object ID with results limited to 20 items each.
* **Documentation**
  * Updated API documentation to include the new search endpoint and introduced a standardized schema for asset group tag members.
* **Bug Fixes**
  * Enhanced error messages for invalid tag types and queries shorter than three characters.
* **Tests**
  * Added extensive tests covering the new search endpoint and database selector retrieval.
* **Chores**
  * Added Apache 2.0 license headers to multiple UI component files.
* **Style**
  * Improved code formatting in test files for consistency.

### DIFF
--- a/packages/go/openapi/src/paths/asset-isolation.asset-group-tags.search.yaml
+++ b/packages/go/openapi/src/paths/asset-isolation.asset-group-tags.search.yaml
@@ -1,3 +1,18 @@
+# Copyright 2025 Specter Ops, Inc.
+#
+# Licensed under the Apache License, Version 2.0
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
 post:
   operationId: AssetGroupTagSearch
   summary: Search Asset Group Tags
--- a/cmd/api/src/database/mocks/db.go
+++ b/cmd/api/src/database/mocks/db.go
@@ -1162,18 +1162,18 @@ func (mr *MockDatabaseMockRecorder) GetAssetGroupTagSelectorCounts(ctx, tagIds a
 }
 
 // GetAssetGroupTagSelectors mocks base method.
-func (m *MockDatabase) GetAssetGroupTagSelectors(ctx context.Context, sqlFilter model.SQLFilter) (model.AssetGroupTagSelectors, error) {
+func (m *MockDatabase) GetAssetGroupTagSelectors(ctx context.Context, sqlFilter model.SQLFilter, limit int) (model.AssetGroupTagSelectors, error) {
 	m.ctrl.T.Helper()
-	ret := m.ctrl.Call(m, "GetAssetGroupTagSelectors", ctx, sqlFilter)
+	ret := m.ctrl.Call(m, "GetAssetGroupTagSelectors", ctx, sqlFilter, limit)
 	ret0, _ := ret[0].(model.AssetGroupTagSelectors)
 	ret1, _ := ret[1].(error)
 	return ret0, ret1
 }
 
 // GetAssetGroupTagSelectors indicates an expected call of GetAssetGroupTagSelectors.
-func (mr *MockDatabaseMockRecorder) GetAssetGroupTagSelectors(ctx, sqlFilter any) *gomock.Call {
+func (mr *MockDatabaseMockRecorder) GetAssetGroupTagSelectors(ctx, sqlFilter, limit any) *gomock.Call {
 	mr.mock.ctrl.T.Helper()
-	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetAssetGroupTagSelectors", reflect.TypeOf((*MockDatabase)(nil).GetAssetGroupTagSelectors), ctx, sqlFilter)
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetAssetGroupTagSelectors", reflect.TypeOf((*MockDatabase)(nil).GetAssetGroupTagSelectors), ctx, sqlFilter, limit)
 }
 
 // GetAssetGroupTagSelectorsByTagId mocks base method.
--- a/cmd/api/src/services/graphify/analysis_integration_test.go
+++ b/cmd/api/src/services/graphify/analysis_integration_test.go
@@ -35,7 +35,7 @@ func TestVersion5Analysis(t *testing.T) {
 		ctx = context.Background()
 
 		analysisFilePath = path.Join("fixtures", "Version5JSON", "analysis")
-		ingestFilePath = path.Join("fixtures", "Version5JSON", "ingest")
+		ingestFilePath   = path.Join("fixtures", "Version5JSON", "ingest")
 
 		testSuite = setupIntegrationTestSuite(t, ingestFilePath)
 	)
@@ -63,7 +63,7 @@ func TestVersion6ADCSAnalysis(t *testing.T) {
 		ctx = context.Background()
 
 		analysisFilePath = path.Join("fixtures", "Version6ADCSJSON", "analysis")
-		ingestFilePath = path.Join("fixtures", "Version6ADCSJSON", "ingest")
+		ingestFilePath   = path.Join("fixtures", "Version6ADCSJSON", "ingest")
 
 		testSuite = setupIntegrationTestSuite(t, ingestFilePath)
 	)
@@ -91,7 +91,7 @@ func TestVersion6AllAnalysis(t *testing.T) {
 		ctx = context.Background()
 
 		analysisFilePath = path.Join("fixtures", "Version6AllJSON", "analysis")
-		ingestFilePath = path.Join("fixtures", "Version6AllJSON", "ingest")
+		ingestFilePath   = path.Join("fixtures", "Version6AllJSON", "ingest")
 
 		testSuite = setupIntegrationTestSuite(t, ingestFilePath)
 	)
@@ -119,7 +119,7 @@ func TestVersion6Analysis(t *testing.T) {
 		ctx = context.Background()
 
 		analysisFilePath = path.Join("fixtures", "Version6JSON", "analysis")
-		ingestFilePath = path.Join("fixtures", "Version6JSON", "ingest")
+		ingestFilePath   = path.Join("fixtures", "Version6JSON", "ingest")
 
 		testSuite = setupIntegrationTestSuite(t, ingestFilePath)
 	)
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableHeaderCell.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableHeaderCell.tsx
@@ -1,3 +1,18 @@
+// Copyright 2025 Specter Ops, Inc.
+//
+// Licensed under the Apache License, Version 2.0
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
 import { faCaretDown, faCaretUp } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { cn, formatPotentiallyUnknownLabel } from '../../utils';
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
@@ -1,3 +1,18 @@
+// Copyright 2025 Specter Ops, Inc.
+//
+// Licensed under the Apache License, Version 2.0
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
 import { Button, Input } from '@bloodhoundenterprise/doodleui';
 import { faMinus, faPlus, faRefresh, faSearch } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsListItem.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsListItem.tsx
@@ -1,3 +1,18 @@
+// Copyright 2025 Specter Ops, Inc.
+//
+// Licensed under the Apache License, Version 2.0
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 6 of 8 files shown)


cmd/api/src/model/assetgrouptags.go
```diff
@@ -148,6 +148,10 @@ func (s AssetGroupTag) ToType() string {
 	}
 }
 
+func IsValidTagType(tagType AssetGroupTagType) bool {
+	return tagType == AssetGroupTagTypeTier || tagType == AssetGroupTagTypeLabel || tagType == AssetGroupTagTypeOwned
+}
```
computator: Maybe this is a question for @mistahj67 but now that we have `0` for hygiene should that be included here too?

cmd/api/src/database/assetgrouptags.go
```diff
@@ -238,16 +239,18 @@ func (s *BloodhoundDB) GetAssetGroupTags(ctx context.Context, sqlFilter model.SQ
 	}
 
 	var tags model.AssetGroupTags
+
 	if result := s.db.WithContext(ctx).Raw(
 		fmt.Sprintf(
-			"SELECT id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify, analysis_enabled FROM %s WHERE deleted_at IS NULL%s",
+			"SELECT id, type, kind_id, name, description, created_at, created_by, updated_at, updated_by, position, require_certify, analysis_enabled FROM %s WHERE deleted_at IS NULL%s ORDER BY name ASC, id ASC",
```
computator: I believe sorting by `id` is redundant since `name` is unique and therefore sorting by just name will already result in a fully sorted list.

cmd/api/src/model/assetgrouptags.go
computator: I guess for the particular usage here this is what we want, but just based on the method name I'm still wondering if we should handle `0` somehow.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +733,98 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+		matchedTags      = model.AssetGroupTags{}
+		matchedSelectors = model.AssetGroupTagSelectors{}
+	)
+
+	if err := json.NewDecoder(request.Body).Decode(&reqBody); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if !model.IsValidTagType(reqBody.TagType) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseAssetGroupTagInvalid, request), response)
+	} else if reqBody.Query == "" || len(reqBody.Query) < 3 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf(api.FmtErrorResponseDetailsMissingRequiredQueryParameter, "query"), request), response)
```
computator: `len(reqBody.Query) < 3` also covers the case of query being `""` so the extra check is redundant.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +733,98 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+	} else if tags, err := s.DB.GetAssetGroupTags(request.Context(), model.SQLFilter{}); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else if selectors, err := s.DB.GetAssetGroupTagSelectors(request.Context(), model.SQLFilter{}); err != nil {
+		api.HandleDatabaseError(request, response, err)
```
computator: #`HandleDatabaseError` returns a 404 if there's no results, but in this case that would be incorrect because the URL does exist plus no results is an unexpected condition. If there is a `database.ErrNotFound` we would probably want to either return a 500 or just continue on even with an empty result list. `HandleDatabaseError` would still be correct for other errors though.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +733,98 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+	} else if sort, err := api.ParseGraphSortParameters(AssetGroupMember{}, queryParams); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsColumnNotFilterable, request), response)
+	} else if skip, err := ParseSkipQueryParameter(queryParams, 0); err != nil {
+		api.WriteErrorResponse(request.Context(), ErrBadQueryParameter(request, model.PaginationQueryParameterSkip, err), response)
+	} else {
+		var (
+			kinds []graph.Kind
```
computator:
```suggestion
			kinds graph.Kinds
```

This will also allow you to use `kinds.Add(...)` instead of append if you want.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +733,98 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+		filter := query.And(
+			query.Or(
+				query.CaseInsensitiveStringContains(query.NodeProperty(common.Name.String()), reqBody.Query),
+				query.CaseInsensitiveStringContains(query.NodeProperty(common.ObjectID.String()), reqBody.Query),
+			),
+			query.KindIn(query.Node(), kinds...),
+		)
+
+		if nodes, err := s.GraphQuery.GetFilteredAndSortedNodesPaginated(sort, filter, skip, assetGroupTagsSearchLimit); err != nil {
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("Error getting members: %v", err), request), response)
```
computator, line 811: CodeRabbit is correct that this falls through to `WriteBasicResponse` so it does need a return statement.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -731,3 +733,98 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+			}
+		}
+
+		filter := query.And(
+			query.Or(
+				query.CaseInsensitiveStringContains(query.NodeProperty(common.Name.String()), reqBody.Query),
+				query.CaseInsensitiveStringContains(query.NodeProperty(common.ObjectID.String()), reqBody.Query),
+			),
+			query.KindIn(query.Node(), kinds...),
+		)
```
computator: I am not sure how we would want to handle this case, or what would be "correct", but if we hit the `assetGroupTagsSearchLimit` before this point, then the graph query would not have the full list of kinds for `KindIn`, and therefore would be returning incomplete results. This would be a different result set from the graph than if it was a query with all the kinds but limited to a certain number of nodes.

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -2251,3 +2252,344 @@ func TestResources_PreviewSelectors(t *testing.T) {
...
+	var (
+		mockCtrl    = gomock.NewController(t)
+		mockDB      = mocks_db.NewMockDatabase(mockCtrl)
+		mockGraphDb = mocks_graph.NewMockGraph(mockCtrl)
+		resources   = v2.Resources{
+			DB:         mockDB,
+			GraphQuery: mockGraphDb,
+		}
+		handler  = http.HandlerFunc(resources.SearchAssetGroupTags)
+		endpoint = "/api/v2/asset-group-tags/search"
```
computator, line 2264: I would either remove this var or use it for all the endpoint urls

cmd/api/src/api/v2/assetgrouptags.go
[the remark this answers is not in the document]
mistahj67: Intentionally needed for the member filter below

cmd/api/src/model/assetgrouptags.go
mistahj67: I'd likely nix this entirely in favor of leveraging the `ToType()` method and kicking out if it's unknown. Findings arent a factor for this endpoint, it's just the search for tags / selectors / members

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.search.yaml
```diff
@@ -0,0 +1,88 @@
...
+                          type: string
+                        primary_kind:
+                          type: string
+                        name:
+                            type: string
+                        properties:
+                            type: object
+                            additionalProperties: true
+                        source:
+                          type: integer
```
mistahj67: Can we abstract this into a reference for use by other endpoints like the `/members` endpoints?

cmd/api/src/api/v2/assetgrouptags.go
mistahj67: Nice catch! so gotta remove the early break but just stop checking contains / adding to the list

cmd/api/src/api/v2/assetgrouptags.go
mistahj67: Should have an empty result set for each, the response shouldn't 404 basically (barring the FF being disabled)

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.search.yaml
```diff
@@ -0,0 +1,88 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+post:
+  operationId: AssetGroupTagSearch
+  summary: Search Asset Group Tags
+  description: Supports searching by name on asset group tags and asset group tag selectors, and by name or object ID on members.
```
ALCooper12: Maybe `Supports searching by name for asset group tags and selectors, and by name or object ID for members`? However it's a very small nit, and def could be ignored if you don't like this suggestion!

cmd/api/src/api/v2/assetgrouptags_test.go
ALCooper12: I agree on using your `endpoint` variable in place of all the:
```
req := httptest.NewRequest(http.MethodPost, endpoint, strings.NewReader(reqBody))
```

cmd/api/src/api/v2/assetgrouptags.go
stephanieslamb: I am not sure I am following?  Currently, kinds are being appended and used the in the filter regardless of the `assetGroupTagsSearchLimit`.  

```
if reqBody.TagType == model.AssetGroupTagTypeTier && t.Type == model.AssetGroupTagTypeTier {
	kinds = append(kinds, t.ToKind())
	if strings.Contains(strings.ToLower(t.Name), strings.ToLower(reqBody.Query)) {
		matchedTags = append(matchedTags, t)

		if len(matchedTags) >= assetGroupTagsSearchLimit {
		        break
		}
	}
}

```
The `assetGroupTagsSearchLimit` only limit to the number of items we are returning for each type(tags, selectors and members).

cmd/api/src/api/v2/assetgrouptags.go
stephanieslamb: Are you suggesting something like this: 
```
else if selectors, err := s.DB.GetAssetGroupTagSelectors(request.Context(), model.SQLFilter{}); errors.Is(err, database.NotFound)  {
  selectors = assetGroupTagSelectors{}
else if err != nil {
}
```

cmd/api/src/api/v2/assetgrouptags_test.go
stephanieslamb: updated to use the variable! forgot it was there 😄

cmd/api/src/api/v2/assetgrouptags.go
stephanieslamb: Ignore me 😅  I understand what you are saying now! Nice catch.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -736,3 +738,88 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+	} else {
+		var (
+			kinds graph.Kinds
+		)
+
+		for _, s := range selectors {
+			if strings.Contains(strings.ToLower(s.Name), strings.ToLower(reqBody.Query)) && len(matchedSelectors) < assetGroupTagsSearchLimit {
+				matchedSelectors = append(matchedSelectors, s)
+			}
+		}
```
mistahj67: Since we aren't needing every selector, it might be worth using the sqlFilter above to have the db do the filtering for us and adding in limit support to the db method. I realize this is a bigger ask, lmk if it gets too complex. At a minimum we should break out of this for loop once we hit the limit

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -736,3 +738,88 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+		reqBody          = AssetGroupTagSearchRequest{}
+		queryParams      = request.URL.Query()
+		members          = []AssetGroupMember{}
+		matchedTags      = model.AssetGroupTags{}
+		matchedSelectors = model.AssetGroupTagSelectors{}
+	)
+
+	if err := json.NewDecoder(request.Body).Decode(&reqBody); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if (model.AssetGroupTag{Type: reqBody.TagType}).ToType() == "unknown" {
```
mistahj67: I didn't realize at the time that this wasn't loading in any tags when I asked to do the type check, let's pair and I think I have an idea on how we can simplify this a bit more!

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -736,3 +738,88 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+
+		for _, s := range selectors {
+			if strings.Contains(strings.ToLower(s.Name), strings.ToLower(reqBody.Query)) && len(matchedSelectors) < assetGroupTagsSearchLimit {
+				matchedSelectors = append(matchedSelectors, s)
+			}
+		}
+
+		for _, t := range tags {
+			isLabelType := t.Type == model.AssetGroupTagTypeLabel
+			isOwnedType := t.Type == model.AssetGroupTagTypeOwned
```
mistahj67: nit, I think this can be simplified with the OR higher up than instantiating 2 variables and checking below
```suggestion
			isLabelType := t.Type == model.AssetGroupTagTypeLabel || t.Type == model.AssetGroupTagTypeOwned
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -728,3 +731,24 @@ func (s *BloodhoundDB) UpdateTierPositions(ctx context.Context, user model.User,
...
+func (s *BloodhoundDB) GetAssetGroupTagSelectors(ctx context.Context, sqlFilter model.SQLFilter) (model.AssetGroupTagSelectors, error) {
+	var selectors model.AssetGroupTagSelectors
+
+	if sqlFilter.SQLString != "" {
+		sqlFilter.SQLString = " WHERE " + sqlFilter.SQLString
+	}
+
+	if result := s.db.WithContext(ctx).Raw(
+		fmt.Sprintf(
+			"SELECT id, asset_group_tag_id, created_at, created_by, updated_at, updated_by, disabled_at, disabled_by, name, description, is_default, allow_disable, auto_certify FROM %s%s ORDER BY name ASC, asset_group_tag_id ASC, id ASC",
```
mistahj67: We may need some indexes to performantly order these

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -736,3 +738,84 @@ func (s *Resources) PreviewSelectors(response http.ResponseWriter, request *http
...
+		}
+		var (
+			nodeFilter = query.And(
+				query.Or(
+					query.CaseInsensitiveStringContains(query.NodeProperty(common.Name.String()), reqBody.Query),
+					query.CaseInsensitiveStringContains(query.NodeProperty(common.ObjectID.String()), reqBody.Query),
+				),
+				query.KindIn(query.Node(), kinds...),
+			)
+			selectorFilter = model.SQLFilter{SQLString: "name ILIKE ? AND asset_group_tag_id IN ?", Params: []any{fmt.Sprintf("%%%s%%", reqBody.Query), tagIds}}
```
computator: Using a plain string concatenation will be a lot easier to understand I think instead of counting how many percent symbols are in a format string 😆. Besides, there's one less function call too!
```suggestion
			selectorFilter = model.SQLFilter{SQLString: "name ILIKE ? AND asset_group_tag_id IN ?", Params: []any{"%" + reqBody.Query + "%", tagIds}}
```

PR #1651, opened 2025-07-03
DRAFT - BED-6138 - StBernard Error Handling

zinic:
## Description

* chore (stbernard): BED-6138 - refactor exec pattern to make error reporting more consistent; minor logic fixups
* chore (pfc): BED-6138 - add license headers

## Motivation and Context

Resolves BED-6138

This should make error reporting and investigation into StBernard a little easier. Added a bunch of minor nit fixes along the way.

## How Has This Been Tested?

It builds.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/go/stbernard/git/git.go
```diff
@@ -84,14 +80,18 @@ func CheckClean(cwd string, env environment.Environment) (bool, error) {
...
-	} else if err := cmd.Run(); err != nil {
-		if exiterr, ok := err.(*exec.ExitError); !ok || exiterr.ExitCode() != 1 {
-			return false, fmt.Errorf("git diff-index --quiet HEAD --: %w", err)
-		} else {
+	}
+
+	if _, err := cmdrunner.Run("git", []string{"diff-index", "--quiet", "HEAD", "--"}, cwd, env); err != nil {
+		var errResult *cmdrunner.ExecutionResult
+
+		if errors.As(err, &errResult) && errResult.ReturnCode == 1 {
```
superlinkx, line 90: You know, if we're using %w to wrap the underlying errors, we wouldn't _technically_ need our custom errResult type as we could just hinge on the underlying exec.ExitError (basically convert the existing check to an As like this, but with the exec.ExitError type.

packages/go/stbernard/command/license/internal/cmd.go
```diff
@@ -129,7 +129,12 @@ func Run(env environment.Environment) error {
 		if ext == "" {
 			ext = filepath.Base(path)
 		}
-		if !info.IsDir() && !slices.Contains(disallowedExtensions, ext) {
+
+		// Ensure that we're not attempting to scan symbolic links
+		if lstatInfo, err := os.Lstat(path); err != nil {
```
superlinkx: We're in a filewalk, so you can get the info from the `info` variable. This first part could then be removed and you'd run `if info.Mode()&os.ModeSymlink != os.ModeSymlink && ...`

packages/go/stbernard/cmdrunner/cmdrunner.go
```diff
@@ -17,64 +17,122 @@
...
+	slog.Info("exec", slog.String("command", commandStr))
+
+	return func() {
+		var (
+			formattedArgs = strings.Join(result.Arguments, " ")
+			elapsed       = time.Since(started)
+		)
+
+		if result.ReturnCode != 0 {
+			if _, err := io.Copy(os.Stderr, result.CombinedOutput); err != nil {
```
superlinkx: I would encourage us to use `result.ErrorOutput()` instead here, since combined output will be more noisy in cases like `golangci-lint` outputting valid results (it will give a non-zero exit and put the json on stdout with additional information on stderr). Most standard commands are really good about proper stdout/stderr usage, especially anything that expects to run in CI environments

superlinkx (approved):
I'm approving, but added a few nits that may be worth exploring, especially the ErrorOutput vs CombinedOutput. Ran this myself in a few scenarios that I knew would give valid and invalid error feedback and I'm loving the improved ergonomics overall.

packages/go/stbernard/cmdrunner/cmdrunner.go
```diff
@@ -17,64 +17,122 @@
...
+			result.ReturnCode = exitErr.ExitCode()
+			return nil, result
+		}
+
+		// Likely a system fault that prevented the command from ever running
+		return nil, err
+	}
 
-	return exitErr
+	return result, nil
```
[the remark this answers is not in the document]
superlinkx: This suggestion from coderabbit might be more reason to consider not wrapping errors in a struct like this, and instead using traditional error wrapping, and allowing the caller to infer the correct exec error itself (can run errors.As on wrapped errors)

packages/go/stbernard/cmdrunner/cmdrunner.go
```diff
@@ -17,64 +17,122 @@
...
+// The combined output buffer preserves the chronological order of both streams at time of collection from their
+// respective file descriptors.
+type ExecutionResult struct {
+	Command        string
+	Arguments      []string
+	Path           string
+	StandardOutput *bytes.Buffer
+	ErrorOutput    *bytes.Buffer
+	CombinedOutput *bytes.Buffer
+	ReturnCode     int
```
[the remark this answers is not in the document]
zinic: Fixing up

packages/go/stbernard/cmdrunner/cmdrunner.go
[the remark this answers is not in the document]
zinic: This was how I had this originally, however, I would argue it's just as confusing to encourage users to exercise the left-hand tuple value when the right-hand tuple value, the error, is not `nil`. Typically the expectation is:

```
if value, err := func(); err != nil {
    // Accessing value in this context could result in a null dereference - should we then task all users of this interface with having to perform a nil check _just to make sure_? Error negotiation with a struct elides this by setting the expectation that the error contains all of the information necessary
}
```

packages/go/stbernard/git/git.go
zinic: I'm not a fan of this, personally. The entire point was to communicate both failure but allow for introspection without violating the existing golang idiom: nil or default left-hand tuple value when right-hand error value is not nil.

I'm not convinced that the way Golang stdlib does this is appropriate. `ExitErr` is useless return code introspection and a more complete command result struct would have been better. It is also a departure from the aforementioned expectation. I would argue that the `exec` call itself should never return an error unless something happened with `execev` or some other system-level failure that's non-recoverable.

If there really is a strong desire for this change then I would suggest reworking the `Run` function to honor the above expectation. Failed commands are not errors, they are results with data requiring introspection.

packages/go/stbernard/command/license/internal/cmd.go
zinic: Will fixup - this was me trying to debug a really weird symlink issue locally and I forgot to pull it

packages/go/stbernard/cmdrunner/cmdrunner.go
zinic: Fixing up

packages/go/stbernard/cmdrunner/cmdrunner.go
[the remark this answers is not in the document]
zinic: @coderabbitai weigh in on my latest comment

packages/go/stbernard/git/git.go
zinic: I'm going to keep this as-is. If there remains a compelling reason to refactor we can cover it in a follow-up PR.

PR #1666, opened 2025-07-09
BED-6022 - Table kebab menu, item selection, JSON download

dcairnsspecterops:
## Description

This PR adds to the explore table view:
1) Row selection on clicking a table row
2) Table sorting
3) Moves row/column munging to its own hook outside ExploreTable component
4) Download JSON

## Motivation and Context

Resolves BED-6022

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added ability to export graph node data as JSON from the Explore Table.
  * Enhanced table row interaction with new options for row selection, context menu, and download actions.
  * Introduced a new table header component with improved sort indicators.
  * Added support for dark mode styling in table controls and dropdowns.

* **Refactor**
  * Simplified Explore Table logic by extracting data processing and column management into a custom hook.
  * Improved event handling and memoization for table interactions.
  * Removed the Manage Columns component for a streamlined UI.

* **Style**
  * Updated styling for dark mode compatibility in table components and dropdowns.
  * Adjusted dropdown and list item styles for better dark mode support.

* **Chores**
  * Updated dependencies for improved UI components.


packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -14,76 +14,25 @@
...
-const requiredColumns = Object.fromEntries(REQUIRED_EXPLORE_TABLE_COLUMN_KEYS.map((key) => [key, true]));
+const REQUIRED_EXPLORE_TABLE_COLUMN_KEYS = ['nodetype', 'objectid', 'displayname'];
 
-type MungedTableRowWithId = GraphNodeSpreadWithProperties & { id: string };
+export const requiredColumns = Object.fromEntries(REQUIRED_EXPLORE_TABLE_COLUMN_KEYS.map((key) => [key, true]));
 
-const columnHelper = createColumnHelper();
+export type MungedTableRowWithId = GraphNodeSpreadWithProperties & { id: string };
 
-const makeColumnDef = (key: string) =>
```
benwaples, line 38: Very clean refactor!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableHeaderCell.tsx
```diff
@@ -0,0 +1,31 @@
...
+    sortBy?: keyof MungedTableRowWithId;
+    sortOrder?: string;
+    onClick: () => void;
+}) => {
+    return (
+        <div
+            className='flex items-center p-1 m-0 cursor-pointer h-full hover:bg-neutral-100 dark:hover:bg-neutral-dark-4'
+            onClick={onClick}>
+            <div>{formatPotentiallyUnknownLabel(String(headerKey))}</div>
+            <div className={cn('pl-2', sortBy !== headerKey ? 'opacity-0' : '')}>
```
benwaples: `cn` can also take an object where the key will be added to the className if the value is truethy, which is handy for multiple conditionals! It wouldnt change much in this case, but possibly useful in the future.

for example:

```suggestion
            <div className={cn('pl-2', {['opacity-0']: sortBy !== headerKey})}>
```

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -0,0 +1,195 @@
+import { createColumnHelper, DataTable } from '@bloodhoundenterprise/doodleui';
+import { faCancel, faCheck, faEllipsis } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { Button } from '@mui/material';
+import { useCallback, useMemo, useState } from 'react';
+import { EntityField, format } from '../../utils';
+import NodeIcon from '../NodeIcon';
+import { ExploreTableProps, MungedTableRowWithId, requiredColumns } from './ExploreTable';
```
benwaples: Looks like this will cause a circular dependency because ExploreTable is also importing from `useExploreTableRowsAndColumns.tsx`, perhaps `requiredColumns` moves to `explore-table-utils.ts` and we make a types file in /ExploreTable for shared types??

cmd/ui/src/views/Explore/GraphView.tsx
```diff
@@ -123,6 +126,47 @@ const GraphView: FC = () => {
...
+        (nodeInfo: NodeClickInfo) => {
+            if (!nodeInfo.id) return;
+            handleContextMenu({ event: { x: nodeInfo.x, y: nodeInfo.y }, node: nodeInfo.id } as SigmaNodeEventPayload);
+        },
+        [handleContextMenu]
+    );
+
+    const handleDownloadClick = useCallback(() => {
+        if (graphQuery.data) {
+            exportToJson({ nodes: graphQuery.data.rawNodes });
```
benwaples, line 168: There is the `exportJsonData` that is used for the download json functionality in `GraphControls`, would that work in this case? That way we dont have to add to the return type of `useSigmaExploreGraph`.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -175,27 +96,29 @@ const ExploreTable = ({
 
     return (
         <div
-            className={`border-2 overflow-hidden absolute z-10 bottom-16 left-4 right-4 bg-neutral-light-2 ${isExpanded ? `h-[calc(100%-72px)]` : 'h-1/2'}`}>
+            className={`dark:bg-neutral-dark-5 border-2 overflow-hidden absolute z-10 bottom-16 left-4 right-4 bg-neutral-light-2 ${selectedNode ? 'w-[calc(100%-450px)]' : ''} ${isExpanded ? `h-[calc(100%-72px)]` : 'h-1/2'}`}>
```
benwaples: potentially a good use case for cns computed classnames!

```suggestion
            className={'dark:bg-neutral-dark-5 border-2 overflow-hidden absolute z-10 bottom-16 left-4 right-4 bg-neutral-light-2', {'w-[calc(100%-450px)]':selectedNode, 'h-[calc(100%-72px)]':isExpanded, 'h-1/2': !isExpanded}}>
```

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -0,0 +1,195 @@
...
+                                <FontAwesomeIcon
+                                    icon={value ? faCheck : faCancel}
+                                    color={value ? 'green' : 'lightgray'}
+                                    className='scale-125'
+                                />
+                            </div>
+                        );
+                    }
+
+                    return format({ keyprop: String(key), value, label: String(key) }) || '--';
```
benwaples: I typically see `.toString()` used so I took a look at the difference between that and `String()` and it looks like one big difference is that `String()` will gracefully handle undefined and null values (as 'null' and 'undefined'), is that by chance a case youre expecting?

benwaples: Perhaps pulling '--' into a const would make it easier to test against? And then if we need to change that fallback into a different `ReactNode` we could change it in one place!

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -0,0 +1,195 @@
...
+                const filterTargets = filterKeys.map((filterKey) => {
+                    const stringyValue = String(item?.[filterKey]);
+
+                    return stringyValue?.toLowerCase();
+                });
+
+                return filterTargets.some((filterTarget) => filterTarget?.includes(searchInput?.toLowerCase()));
+            }),
+        [searchInput, rows]
+    );
```
benwaples: Im thinking we can reduce the number of iterations we have in this calculation because if the `toLowerCase()` method iterates that would a mean complexity of O(`rows` * `length_of_<displayname | objectId>` * `lenght_of_searchInput`)?  Which probably isnt insane, but a smol win would be to assign `searchInput?.toLowerCase()` to a const above the .filter and move the `filterKey` definition above the .filter as well!

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
dcairnsspecterops: I avoid `toString()` just because I dislike what it does to objects: `[Object object]`.  The only reason I'm parsing it to a string is that the `format` function wants it like that. I'm happy to make it `toString` if you feel like that's more consistent with the rest of the app/a better decision.

How feel you feel about replacing the `--` with the 🚫 font awesome icon?

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
dcairnsspecterops: oops, yep, great catch.

cmd/ui/src/views/Explore/GraphView.tsx
dcairnsspecterops: It looks to me like `exportToJson` is the same function used in `GraphControls`, and `exportJsonData` is a piece of state sent down as `jsonData` to `GraphControls`. I think the `rawNodes `here are better for user to download than the styled nodes, since users probably don't care about color, fontSize, etc? I could be misunderstanding

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
dcairnsspecterops: Good idea! You noticed this with just your eyes? Remarkable

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
benwaples: ohhh interesting!! That does make me realize we may want to consider the values that we are trying to render in these fields almost like `unknown`. Looks like the `format` function handles values that are arrays as well but idk if we have a way to render those in the table yet! 

Lets leave the `String()` use here for this PR because it handles more types gracefully, and in our tweaks effort we spend a little time investigating what it would look like if anything other than booleans and strings got rendered!

IMO we MVP with `--` but have 🚫  in our back pocket if we people need more indication that a value doesnt exist for a given field.  Unless youre feeling string about the no-entry sign, in which case we just gotta swing it by product before adding it!

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
benwaples: hahah nar nar im not built like that! i had to rebuild because my previous branch was off stage!

cmd/ui/src/views/Explore/GraphView.tsx
benwaples: Im not seeing the styling included in `exportJsonData`, however, I also dont see the extra properties bag either :/ makes sense to include the properties bag in the json tho so good call not using `exportJsonData`!

If we are downloading the raw json from the API, could we skip the intermediary and use the `useExploreGraph` directly within the `ExploreTable`?

dcairnsspecterops: I'm open to moving this callback into the ExploreTable, [Edit: Maybe never mind this bit, since I'm now seeing our API calls live in the shard-ui library too: _though it does feel a little weird to have ExploreTable making api calls, since it lives in the shared-ui library, and it feels to me like components from another library should be more agnostic and take more props rather than rely on context from outside_]. 

Could we push this to the next ticket, since I'm pretty deep into that one?

benwaples: yup im down to push it from scope of this PR!

benwaples (approved):
LGTM, thanks for the changes!

maffkipp (approved):
Pulled this down and everything works as expected!

PR #1685, opened 2025-07-15
BED-3538 - Add testing for ExploreTable

dcairnsspecterops:
## Description

## Motivation and Context

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -44,6 +52,78 @@ const useExploreTableRowsAndColumns = ({
     const [sortBy, setSortBy] = useState<keyof MungedTableRowWithId>();
     const [sortOrder, setSortOrder] = useState<'asc' | 'desc'>();
 
+    const rows = useMemo(
+        () =>
+            (data &&
+                [
+                    ...Object.entries(data),
```
dcairnsspecterops: Dramatically exemplifying a large data set for testing, development. To be removed.

cmd/ui/src/hooks/useSigmaExploreGraph/useSigmaExploreGraph.tsx
```diff
@@ -30,17 +30,25 @@ export const normalizeGraphDataForSigma = (
...
+export const useSigmaExploreGraph = () => {
+    const graphQuery = useExploreGraph();
     const normalizedGraphData = useMemo(() => normalizeGraphDataForSigma(graphQuery.data), [graphQuery.data]);
-
     const castGraphData = graphQuery?.data?.data as GraphData;
+
+    const nodeKeys = useMemo(
+        () =>
+            Array.isArray(castGraphData?.node_keys)
+                ? ['nodetype', 'isTierZero'].concat(castGraphData?.node_keys)
```
dcairnsspecterops, line 41: We get a node_keys array to tell use which column options should be available in the manage columns combobox. Kind (i.e., 'user' etc) and isTierZero are not included in the node_keys array, because they are not "real" properties, but properties computed/added on the front end in the convertGraphResponseToRegraphItems and transformToFlatGraphResponse functions. The AC does designate them as "default columns", however.

The "easiest" solution from my perspective is to simply prepend those keys to the node_keys array is it exists.

Why not use 'kind', which is on the top level of our response data, instead of nodetype, which is created on the front end? By the time the data gets to the GraphView, it has lost its kind property, renamed as nodetype. I propose we just use nodetype rather than change the interface to include the kind property. This interface is shared between the graph visualization and the table visualization (perhaps a mistake?), and the keys were changed intentionally for the graph library input, I assume some good reason which I do not want to regress against.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -26,18 +26,26 @@ const MemoDataTable = memo(DataTable<MungedTableRowWithId, any>);
 
 type DataTableProps = React.ComponentProps<typeof MemoDataTable>;
 
+const tableProps: DataTableProps['TableProps'] = {
+    className: 'w-[default] w-full',
```
dcairnsspecterops: Note: we will want to uncomment these style changes when the DoodleUI code is merged; I commented this out to confirm CI passes.

cmd/ui/src/views/Explore/GraphView.test.tsx
```diff
@@ -56,4 +72,58 @@ describe('GraphView', () => {
 
         expect(errorAlert).toBeInTheDocument();
     });
+
+    it('renders a table if the query has NO node edges', async () => {
+        const urlSearchParams = new URLSearchParams();
+        urlSearchParams.append('searchType', 'cypher');
+        urlSearchParams.append('cypherSearch', 'encodedquery');
+        vi.mocked(useSearchParams).mockReturnValue([urlSearchParams, vi.fn()]);
```
benwaples: We have the ability to plumb an initial route thru the render function, would that help simplify the url mocking in this test file?

For example:
```js
render(<GraphView />, { route: `/graphview?searchType=cypher&cypherSearch=encodedquery` })
```

cmd/ui/src/views/Explore/graph-view-test-data.ts
```diff
@@ -0,0 +1,638 @@
+import { GraphData } from 'js-client-library';
+
+export const featureFlagsResponse = {
```
benwaples, line 18: we have a util function called `createFeatureFlags` that you can pass in an array like `[{ key: explore_table_view, enabled: true }]` and itll generate the rest of the needed info!

cmd/ui/src/views/Explore/graph-view-test-data.ts
```diff
@@ -0,0 +1,638 @@
...
+            'serviceprincipalnames',
+            'sidhistory',
+            'system_tags',
+            'title',
+            'trustedtoauth',
+            'unconstraineddelegation',
+            'whencreated',
+        ],
+        edges: [],
+        nodes: {
```
benwaples, line 305: Do we need all 10 of these nodes? Or would a sample size of 2 or 3 suffice?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.test.tsx
```diff
@@ -0,0 +1,224 @@
...
+
+const setup = async () => {
+    render(<WrappedExploreTable />);
+
+    const user = userEvent.setup();
+
+    return { user };
+};
+
+describe('ExploreTable', async () => {
```
benwaples, line 76: Nice coverage!!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+const ExploreTableDataCell = ({ value, columnKey }: { value: EntityField['value']; columnKey: string }) => {
+    if (columnKey === 'nodetype') {
+        return (
+            <div className='flex justify-center'>
+                <NodeIcon nodeType={value?.toString() || ''} />
+            </div>
+        );
+    }
+    if (isIconField(value)) {
```
benwaples: It looks like for the entity panel we handle `undefined` separately from `false` and im thinking we should match that behavior here. That probably explains why we had a mix of `-` and `X` in boolean columns :/ 

Lets keep the isIconField check because we can still use the outer wrapper div that centers the icon but we could add a check like: 
```tsx
{typeof value === 'boolean' ?  <FontAwesomeIcon />: FALLBACK_STRING }

```

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableHeaderCell.tsx
```diff
@@ -15,31 +15,45 @@
...
-                {!sortOrder && <FontAwesomeIcon icon={faCaretDown} />}
-                {sortOrder === 'asc' && <FontAwesomeIcon icon={faCaretUp} />}
-                {sortOrder === 'desc' && <FontAwesomeIcon icon={faCaretDown} />}
+        <Tooltip title={<p>{label}</p>}>
+            <div
+                className={cn(
+                    'flex items-center m-0 cursor-pointer h-full w-full hover:bg-neutral-100 dark:hover:bg-neutral-dark-4',
+                    {
+                        'justify-center':
+                            dataType === 'boolean' || KEYS_TO_RENDER_AS_ICON.includes(headerKey.toString()),
```
benwaples, line 45: Awesome! this will be a nice control in the future as we probably find more columns that people will want icons rendered for.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
```diff
@@ -63,6 +63,17 @@ export const ManageColumnsComboBox = ({
...
+    const handleResetDefault = useCallback(() => {
+        setSelectedColumns([...pinnedColumns]);
+        onChange([...pinnedColumns]);
+    }, [onChange, pinnedColumns]);
+
+    useEffect(() => {
+        if (selectedColumns.length === 0) {
+            handleResetDefault();
+        }
+    }, [selectedColumns, handleResetDefault]);
```
benwaples: Could this be handled deterministically in the onStateChange handler of useMultipleSelection? Or is there a need to have it as a side-effect running?

```ts
            if (type !== useMultipleSelection.stateChangeTypes.DropdownKeyDownBackspace) {
                if (newSelectedColumns && newSelectedColumns.length !== 0) {
                    setSelectedColumns(newSelectedColumns || []);
                    onChange(newSelectedColumns || []);
                } else {
                    handleResetDefault();
                }
            }
```

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -170,10 +186,14 @@ const useExploreTableRowsAndColumns = ({
         if (sortBy) {
             if (sortOrder === 'asc') {
                 dataToSort.sort((a, b) => {
+                    if (a[sortBy] === true) return 1;
```
benwaples: Do we need to handle booleans as a whole so that false values dont attempt to get stringified? When i sort on a column that contains booleans and undefined, seems like undefined is getting handled in unexpected ways!

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.tsx
```diff
@@ -132,20 +133,27 @@ function GraphControls<T extends readonly string[]>(props: GraphControlsProps<T>
...
-                                selected={featureFlag?.enabled ? selectedLayout === layout : undefined}
-                                onClick={() => onLayoutChange(layout)}>
-                                {capitalize(layout)}
-                            </MenuItem>
-                        ))}
+                        .map((buttonLabel) => {
+                            const tableViewIsSelected = tableViewFeatureFlag?.enabled && isExploreTableSelected;
+                            const isSelected = tableViewIsSelected
+                                ? buttonLabel === 'table'
+                                : selectedLayout === buttonLabel;
```
benwaples: A wee readiblity nit: using buttonLabel as the first value in each comparison
```suggestion
                                ? buttonLabel === 'table'
                                : buttonLabel === selectedLayout;
```

benwaples (changes requested):
Nice clean up!! This is looking really tidy and the test coverage is so good! I left a few comments and am happy to help out on any of them.

cmd/ui/src/views/Explore/graph-view-test-data.ts
dcairnsspecterops: We don't necessarily need them all, though the current tests are written against this assumption, and find ing 10 rows is a more 'false positive' proof test than finding 3 rows.  It's a small enough number that we could conceivably end up with 3 aria row items accidentally in some way we're not thinking of, but we're unlikely to hit the magic number of 10 accidentally.

cmd/ui/src/views/Explore/GraphView.test.tsx
dcairnsspecterops: Oh, nice! That would probably be helpful

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
dcairnsspecterops: Good idea!

cmd/ui/src/views/Explore/graph-view-test-data.ts
benwaples: ahh nice! I do despise false positives lol Im down to keep it. Perhaps when we implement virtualization we can refactor this to generate these rows with deterministic data so we can test against search/filter/sort functionality with large data sets! But that can be a future improvement!

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
dcairnsspecterops: I just created a more robust sort function. Let me know if it seems to behave as you'd expect.

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -169,43 +185,59 @@ const useExploreTableRowsAndColumns = ({
...
+                        : -1;
+                }
+                return -1;
+            }
+
+            return 1;
+        });
+
+        if (columnDefs[columnDefs.length - 1]) {
+            columnDefs[columnDefs.length - 1].size = Infinity;
```
benwaples: mind adding a comment here about why `Infinity` is used?

packages/javascript/bh-shared-ui/src/components/ExploreTable/explore-table-utils.ts
```diff
@@ -29,10 +29,42 @@ export const makeStoreMapFromColumnOptions = (columnOptions: ManageColumnsComboB
...
 export type MungedTableRowWithId = GraphNodeSpreadWithProperties & { id: string };
 
-const REQUIRED_EXPLORE_TABLE_COLUMN_KEYS = ['nodetype', 'objectid', 'displayname'];
+export const REQUIRED_EXPLORE_TABLE_COLUMN_KEYS = ['nodetype', 'isTierZero', 'name', 'objectid'];
 
 export const requiredColumns = Object.fromEntries(REQUIRED_EXPLORE_TABLE_COLUMN_KEYS.map((key) => [key, true]));
 
+export const compareForExploreTableSort = (a: any, b: any) => {
+    if (typeof a === 'number' || typeof b === 'number') {
+        if (typeof a === 'number' && typeof b === 'number') return a > b;
```
benwaples: I think this branch will return true/false, but should return positive or negative. Should the evaluation be be `a-b`?

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -170,10 +186,14 @@ const useExploreTableRowsAndColumns = ({
...
+                    if (a[sortBy] === true) return 1;
+
                     return a[sortBy]?.toString().localeCompare(b[sortBy]?.toString());
                 });
             } else {
                 dataToSort.sort((a, b) => {
+                    if (b[sortBy] === true) return 1;
+
                     return b[sortBy]?.toString().localeCompare(a[sortBy]?.toString());
                 });
```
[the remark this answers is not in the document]
benwaples: this has been updated with a more extensive sorting function called compareForExploreTableSort

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
benwaples: woah nice!! I think the more robust will be much more maintainable and easier to logic thru all the branches. Thank you! I left one comment about a possible return type, otherwise it looks solid and behaves how Id expect

packages/javascript/bh-shared-ui/src/components/ExploreTable/explore-table-utils.ts
dcairnsspecterops: Good call

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -52,25 +47,22 @@ export function exploreGraphQueryFactory(
...
-    enabled = DEFAULT_USE_EXPLORE_GRAPH_PARAMS.enabled,
-}: UseExploreGraphParams = DEFAULT_USE_EXPLORE_GRAPH_PARAMS) => {
+export const useExploreGraph = (enabled = true) => {
     const params = useExploreParams();
 
     const { addNotification } = useNotifications();
 
     const query = exploreGraphQueryFactory(params);
 
+    const includeProperties = true;
```
benwaples: We should remove all the plumbing for includeProperties all the way to the api request in `bhce/packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/queries/cypherSearch.ts`

cmd/ui/src/hooks/useSigmaExploreGraph/useSigmaExploreGraph.tsx
benwaples: resolved for MVP and created this refactor ticket: https://specterops.atlassian.net/browse/BED-6232

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
dcairnsspecterops: Changed this to 'auto`. Had to hack around the types to make it work, but I think it's worth it

cmd/ui/src/views/Explore/graph-view-test-data.ts
benwaples: we can hit this in a follow up effort as well

PR #1694, opened 2025-07-18
feat(Database Management): add deletion by source kind

TheNando:
## Description

* Update Database Management page, adding options to delete by source kind
* Add graph source kind endpoint to JS client
* Update JS client clear-database endpoint to support source kind

## Motivation and Context

Resolves BED-6191

Now that customers can import data via OpenGraph, we need more granular options to remove data from the database other than deleting it all.

## How Has This Been Tested?

Manually tested. Added unit tests.

## Screenshots (optional):

image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting and deleting specific source kinds of graph data in the database management UI.
  * Introduced a new checkbox component for multi-source kind selection.
  * Enhanced API client with a method to fetch available source kinds.

* **Bug Fixes**
  * Ensured that certain data types are always excluded from deletion to prevent accidental data loss.

* **Tests**
  * Added and improved tests for new checkbox functionality, UI permission handling, and alternative deletion scenarios.

* **Documentation**
  * Updated type definitions and interfaces to reflect new deletion options.


cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -31,13 +31,15 @@ import { FC, useReducer } from 'react';
 import { useMutation } from 'react-query';
 import { useSelector } from 'react-redux';
 import { selectAllAssetGroupIds, selectTierZeroAssetGroupId } from 'src/ducks/assetgroups/reducer';
+import { SourceKindsCheckboxes } from './SourceKindsCheckboxes';
 
 const initialState: State = {
+    deleteAllAssetGroupSelectors: false,
```
TheNando, line 37: Just sorting the props here and in a few other places. And adding `deleteSourceKinds`.

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.tsx
```diff
@@ -110,6 +115,14 @@ const reducer = (state: State, action: Action): State => {
                 noSelectionError: false,
             };
         }
+        case 'source_kinds': {
```
TheNando, line 118: Source kinds is an int array while the rest of these are single booleans. So a separate action is needed to manage state.

brandonshearin (approved):
end-end testing is a success!

cweidenkeller (approved):
LGTM

cmd/ui/src/views/DatabaseManagement/DatabaseManagement.test.tsx
```diff
@@ -81,10 +120,11 @@ describe('DatabaseManagement', () => {
         const deleteButton = screen.getByRole('button', { name: 'Delete' });
 
         expect(deleteButton).toBeDisabled();
+        permissionResponse = createAuthStateWithPermissions([Permission.WIPE_DB]).user;
```
wes-mil, line 123: Is this hanging or does this actually do something?

TheNando: Yeah. There is probably a better way to achieve this without the mutated state, by created a test factory that lets you pass in the permission outcome. For most of these tests, we want the permission response to include the WIPE_DB, but for this one test, we are testing when the user does not have perms. So at the start, we change the permissionResponse to not have the perm. The mocked API call will reference the value when that test runs and the test will pass with the buttons disabled. But if we don't set the variable back, it will still have the changed value for the next tests, which expect the permission to be there again.

So effectively, the default state is handled by the test scope variable, but this one tests we update the value, but set it back so to not affect the other tests.

PR #1700, opened 2025-07-21
feat (api): Add Get SavedQuery and Get SavedQueryPermissions Endpoints - BED-6220 and BED-6219

LawsonWillard:
## Description

Added two endpoints to support the Saved Queries initiative, one to retrieve a Saved Query by id and a second to retrieve a Saved Query's permissions by id. The bulk of the added code is in testing and documentation. 

## Motivation and Context
Resolves: BED-6219
Resolves: BED-6220

To accomplish the saved queries initiative we need to be able to get a single saved query along with its permissions set. This way, the front end can correctly show which users a query is shared with. 

## How Has This Been Tested?

- Added unit and integration tests for the added functions
- Tested locally against the endpoints. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/registration/v2.go
+++ b/cmd/api/src/api/registration/v2.go
@@ -208,6 +208,7 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		routerInst.POST("/api/v2/graphs/cypher", resources.CypherQuery).RequirePermissions(permissions.GraphDBRead),
 		routerInst.GET("/api/v2/saved-queries", resources.ListSavedQueries).RequirePermissions(permissions.SavedQueriesRead),
 		routerInst.POST("/api/v2/saved-queries", resources.CreateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
+		routerInst.GET(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.GetSavedQuery).RequirePermissions(permissions.SavedQueriesRead),
 		routerInst.PUT(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.UpdateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.DELETE(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.DeleteSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.GET(fmt.Sprintf("/api/v2/saved-queries/{%s}/permissions", api.URIPathVariableSavedQueryID), resources.GetSavedQueryPermissions).RequirePermissions(permissions.SavedQueriesRead),
--- a/cmd/api/src/api/registration/v2.go
+++ b/cmd/api/src/api/registration/v2.go
@@ -210,6 +210,7 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		routerInst.POST("/api/v2/saved-queries", resources.CreateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.PUT(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.UpdateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.DELETE(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.DeleteSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
+		routerInst.GET(fmt.Sprintf("/api/v2/saved-queries/{%s}/permissions", api.URIPathVariableSavedQueryID), resources.GetSavedQueryPermissions).RequirePermissions(permissions.SavedQueriesRead),
 		routerInst.DELETE(fmt.Sprintf("/api/v2/saved-queries/{%s}/permissions", api.URIPathVariableSavedQueryID), resources.DeleteSavedQueryPermissions).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.PUT(fmt.Sprintf("/api/v2/saved-queries/{%s}/permissions", api.URIPathVariableSavedQueryID), resources.ShareSavedQueries).RequirePermissions(permissions.SavedQueriesWrite),
 		routerInst.GET(fmt.Sprintf("/api/v2/saved-queries/{%s}/export", api.URIPathVariableSavedQueryID), resources.ExportSavedQuery).RequirePermissions(permissions.SavedQueriesRead),
--- a/cmd/api/src/api/v2/saved_queries.go
+++ b/cmd/api/src/api/v2/saved_queries.go
@@ -200,7 +200,7 @@ func (s Resources) ExportSavedQuery(response http.ResponseWriter, request *http.
 	} else if savedQuery, err = s.DB.GetSavedQuery(request.Context(), savedQueryID); err != nil {
 		auditLogEntry.Status = model.AuditLogStatusFailure
 		api.HandleDatabaseError(request, response, err)
-	} else if isAccessibleToUser, err = s.canUserAccessQuery(request.Context(), savedQuery, user.ID); err != nil {
+	} else if isAccessibleToUser, err = s.canUserAccessQuery(request.Context(), savedQuery, user); err != nil {
 		auditLogEntry.Status = model.AuditLogStatusFailure
 		api.HandleDatabaseError(request, response, err)
 	} else if !isAccessibleToUser {
@@ -216,11 +216,12 @@ func (s Resources) ExportSavedQuery(response http.ResponseWriter, request *http.
 	}
 }
 
-func (s Resources) canUserAccessQuery(ctx context.Context, query model.SavedQuery, userId uuid.UUID) (bool, error) {
-	if userId.String() == query.UserID {
+// canUserAccessQuery - Users can access a query if they own it, are an admin, the query is shared to them or is a public query.
+func (s Resources) canUserAccessQuery(ctx context.Context, query model.SavedQuery, user model.User) (bool, error) {
+	if user.ID.String() == query.UserID || user.Roles.Has(model.Role{Name: auth.RoleAdministrator}) {
 		return true, nil
 	}
-	return s.DB.IsSavedQuerySharedToUserOrPublic(ctx, query.ID, userId)
+	return s.DB.IsSavedQuerySharedToUserOrPublic(ctx, query.ID, user.ID)
 }
 
 // ExportSavedQueries - Exports one or more saved queries in a ZIP file. The scope query parameter determines which queries are exported.
--- a/cmd/api/src/api/v2/saved_queries.go
+++ b/cmd/api/src/api/v2/saved_queries.go
@@ -46,6 +46,30 @@ import (
 	"github.com/specterops/bloodhound/packages/go/mediatypes"
 )
 
+func (s Resources) GetSavedQuery(response http.ResponseWriter, request *http.Request) {
+	var (
+		rawSavedQueryID = mux.Vars(request)[api.URIPathVariableSavedQueryID]
+	)
+
+	if user, isUser := auth.GetUserFromAuthCtx(ctx2.FromRequest(request).AuthCtx); !isUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "No associated user found", request), response)
+	} else if savedQueryID, err := strconv.ParseInt(rawSavedQueryID, 10, 64); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
+	} else if savedQuery, err := s.DB.GetSavedQuery(request.Context(), savedQueryID); err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "query not found", request), response)
+		} else {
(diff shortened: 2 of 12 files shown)


packages/go/openapi/doc/openapi.json
```diff
@@ -7001,6 +7001,51 @@
...
+          "404": {
+            "$ref": "#/components/responses/not-found"
+          },
+          "429": {
+            "$ref": "#/components/responses/too-many-requests"
+          },
+          "500": {
+            "$ref": "#/components/responses/internal-server-error"
+          }
+        }
```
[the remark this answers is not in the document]
LawsonWillard: Should be fixed in most recent commit

cmd/api/src/api/registration/v2.go
```diff
@@ -208,8 +208,10 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		routerInst.POST("/api/v2/graphs/cypher", resources.CypherQuery).RequirePermissions(permissions.GraphDBRead),
 		routerInst.GET("/api/v2/saved-queries", resources.ListSavedQueries).RequirePermissions(permissions.SavedQueriesRead),
 		routerInst.POST("/api/v2/saved-queries", resources.CreateSavedQuery).RequirePermissions(permissions.SavedQueriesWrite),
+		routerInst.GET(fmt.Sprintf("/api/v2/saved-queries/{%s}", api.URIPathVariableSavedQueryID), resources.GetSavedQuery).RequirePermissions(permissions.SavedQueriesRead),
```
kpowderly, line 211: Looks like both of these endpoints are using `.RequirePermissions(permissions.SavedQueriesRead),`

Is this endpoint supposed to have the permissions required if we have a separate endpoint for `GetSavedQueryPermissions`?

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -46,6 +46,30 @@ import (
 	"github.com/specterops/bloodhound/packages/go/mediatypes"
 )
 
+func (s Resources) GetSavedQuery(response http.ResponseWriter, request *http.Request) {
+	var (
+		rawSavedQueryID = mux.Vars(request)[api.URIPathVariableSavedQueryID]
+	)
+
+	if user, isUser := auth.GetUserFromAuthCtx(ctx2.FromRequest(request).AuthCtx); !isUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "No associated user found", request), response)
```
kpowderly: It seems like we don't follow this standard here go usually recommends that [error messages be all lowercase with no punctuation](https://go.dev/wiki/CodeReviewComments#error-strings). Feel free to ignore since it's not really an implemented standard here but figured I'd call that out!

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -46,6 +46,30 @@ import (
 	"github.com/specterops/bloodhound/packages/go/mediatypes"
 )
```
kpowderly, line 48: Do you mind adding a godoc comment on this function as well? I think with such similar endpoints, it's probably best to distinguish between the two

cmd/api/src/api/v2/saved_queries_permissions.go
```diff
@@ -78,6 +79,65 @@ func CanUpdateSavedQueriesPermission(user model.User, savedQueryBelongsToUser bo
 	return ErrForbidden
 }
 
+type SavedQueryPermissionResponse struct {
+	QueryID         int64       `json:"query_id"`
+	Public          bool        `json:"public"`
+	SharedToUserIDs []uuid.UUID `json:"shared_to_user_ids"`
+}
+
+func (s *SavedQueryPermissionResponse) AppendUserId(userId uuid.NullUUID) {
```
kpowderly: Can we make this private? I don't see anywhere it needs to be exported

cmd/api/src/api/v2/saved_queries_test.go
```diff
@@ -2951,3 +2951,253 @@ func TestResources_ExportSavedQueries(t *testing.T) {
...
+		},
+		{
+			name: "fail - database error retrieving saved query",
+			fields: fields{
+				setupMocks: func(t *testing.T, mock *mocks.MockDatabase) {
+					mockDB.EXPECT().GetSavedQuery(gomock.Any(), int64(1)).Return(model.SavedQuery{}, fmt.Errorf("test error"))
+				},
+			},
+			args: args{
+				buildRequest: func() (*http.Request, error) {
```
kpowderly: On these buildRequest functions, we don't need to return the error since you're using `require.NoError(t, err)`.

cmd/api/src/api/v2/saved_queries_test.go
```diff
@@ -2951,3 +2951,253 @@ func TestResources_ExportSavedQueries(t *testing.T) {
...
+			},
+			expect: expected{
+				responseCode:  http.StatusOK,
+				responseQuery: testSavedQuery2,
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tt.fields.setupMocks(t, mockDB)
```
kpowderly: I just wrote up a standard that we're working on for contract (unit) tests if you want to give it a shot since we're already moving in that direction.

https://specterops.atlassian.net/wiki/spaces/~71202003066efc8a3e41a0a828eb6d4705a673/pages/1305772102/Golang+Test+Standards#Example-Tests-to-follow

The goal is to parallelize all top-level tests and their subtests, while ensuring that every subtest uses a single, consistent assertion function.

Something like this ->

``` go
                        t.Parallel()

			tt.fields.setupMocks(t, mockDB)

			s := v2.Resources{
				DB: mockDB,
			}

			request := tt.args.buildRequest()
			tt.fields.setupMocks(t, mockDB)

			response := httptest.NewRecorder()

			router := mux.NewRouter()
			router.HandleFunc("/api/v2/saved-queries/{%s}/permissions", s.GetSavedQueryPermissions).Methods(request.Method)
			router.ServeHTTP(response, request)

			status, header, body := test.ProcessResponse(t, response)

			assert.Equal(t, tt.expect.responseCode, status)
			assert.Equal(t, tt.expect.responseHeader, header)
			assert.Equal(t, tt.expect.responseError, body)
```

cmd/api/src/api/v2/saved_queries_permissions.go
```diff
@@ -78,6 +79,65 @@ func CanUpdateSavedQueriesPermission(user model.User, savedQueryBelongsToUser bo
...
+
+	if user, isUser := auth.GetUserFromAuthCtx(ctx2.FromRequest(request).AuthCtx); !isUser {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "No associated user found", request), response)
+	} else if savedQueryID, err := strconv.ParseInt(rawSavedQueryID, 10, 64); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsIDMalformed, request), response)
+	} else if savedQueryPermissions, err := s.DB.GetSavedQueryPermissions(request.Context(), savedQueryID); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else if len(savedQueryPermissions) == 0 {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "no query permissions exist for saved query", request), response)
+	} else if isAccessibleToUser, err := s.canUserAccessSavedQueryPermissions(request.Context(), savedQueryPermissions[0], user); err != nil {
```
kpowderly: Sorry, this is probably a dumb question but should we be checking all query permissions that come back instead of just grabbing the first entry?

cmd/api/src/api/registration/v2.go
LawsonWillard: Yeah, its a bit confusing since they both refer to permissions but they perform different functions. 

The `GetSavedQueryPermissions` endpoint returns if the query is public/private and if private who the query is shared with, while `RequirePermissions` determines if the requesting client has a role that can access the given endpoint.

cmd/api/src/api/v2/saved_queries.go
LawsonWillard: Can do! Also adding another comment on GetSavedQueryPermissions to help clarify what permissions mean for these endpoints.

cmd/api/src/api/v2/saved_queries_permissions.go
LawsonWillard: Not a dumb question! Query permissions can only be viewed if the query is public, the requesting user owns the query or is an admin. Its not apparent in this PR but query permissions will only have 1 record if its public and if private then all other records will have that public value set to false. Because of this, we only need to check the first item to see if a user can access the query permissions.

cmd/api/src/api/v2/saved_queries_test.go
LawsonWillard: Will update both `TestResources_GetSavedQuery` and `TestResources_GetPermissionsForSavedQuery`

cmd/api/src/api/v2/saved_queries_test.go
```diff
@@ -2951,3 +2952,253 @@ func TestResources_ExportSavedQueries(t *testing.T) {
...
+			expect: expected{
+				responseCode:   http.StatusOK,
+				responseBody:   fmt.Sprintf(`{"data":{"user_id":"%s","name":"Test2Query","query":"match (n:Base) return n","description":"test description 2","id":2,"created_at":"0001-01-01T00:00:00Z","updated_at":"0001-01-01T00:00:00Z","deleted_at":{"Time":"0001-01-01T00:00:00Z","Valid":false}}}`, user2Id),
+				responseHeader: http.Header{"Content-Type": []string{"application/json"}},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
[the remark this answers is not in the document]
LawsonWillard: These tests are self contained and will not be flaky in this context so running them parallel works

ALCooper12:
@LawsonWillard I could be wrong since you may have already gotten approval or something, but we usually do 1 PR per jira ticket. If you haven't gotten a :+1:, I'd double check just in case!

cmd/api/src/api/v2/saved_queries_permissions_test.go
```diff
@@ -1535,3 +1537,300 @@ func TestResources_DeleteSavedQueryPermissions(t *testing.T) {
...
+			tt.fields.setupMocks(t, databaseMock)
+			s := v2.Resources{
+				DB: databaseMock.mockDatabase,
+			}
+
+			response := httptest.NewRecorder()
+			request := tt.args.buildRequest()
+			s.GetSavedQueryPermissions(response, request)
+			statusCode, header, body := test.ProcessResponse(t, response)
+			fmt.Println(body)
```
kpowderly: fyi, you forgot a println here

cmd/api/src/database/saved_queries_permissions_test.go
```diff
@@ -343,3 +344,35 @@ func TestSavedQueriesPermissions_IsSavedQuerySharedToUser(t *testing.T) {
 	require.NoError(t, err)
 	assert.True(t, isShared)
 }
+
+func TestSavedQueriesPermissions_GetSavedQueryPermissions(t *testing.T) {
```
kpowderly: I just realized this is an integration test file -- can you update the file name to saved_queries_permissions_integration_test.go ?

Thanks! :) Sorry I missed this on first pass

kpowderly:
> @LawsonWillard I could be wrong since you may have already gotten approval or something, but we usually do 1 PR per jira ticket. If you haven't gotten a 👍, I'd double check just in case!

If you wanna split these after the last couple changes, just let me know and I'll approve both.

kpowderly (approved):
Approving after last few comments are resolved! Great job!

cmd/api/src/database/saved_queries_permissions_test.go
LawsonWillard: Will do!

LawsonWillard:
@ALCooper12 sounds good! I'll get it split out

LawsonWillard:
Got the other PR here: https://github.com/SpecterOps/BloodHound/pull/1707

cmd/api/src/api/v2/saved_queries.go
```diff
@@ -216,11 +241,12 @@ func (s Resources) ExportSavedQuery(response http.ResponseWriter, request *http.
...
 
-func (s Resources) canUserAccessQuery(ctx context.Context, query model.SavedQuery, userId uuid.UUID) (bool, error) {
-	if userId.String() == query.UserID {
+// canUserAccessQuery - Users can access a query if they own it, are an admin, the query is shared to them or is a public query.
+func (s Resources) canUserAccessQuery(ctx context.Context, query model.SavedQuery, user model.User) (bool, error) {
+	if user.ID.String() == query.UserID || user.Roles.Has(model.Role{Name: auth.RoleAdministrator}) {
 		return true, nil
 	}
-	return s.DB.IsSavedQuerySharedToUserOrPublic(ctx, query.ID, userId)
+	return s.DB.IsSavedQuerySharedToUserOrPublic(ctx, query.ID, user.ID)
```
ALCooper12, line 249: _Adding as a single comment so that my question doesn't block you_ 

Just wondering, but why did you add this extra db method that is very similar to: `s.DB.IsSavedQuerySharedToUser`? 

I see that the new method allows you to also check if a saved query is public as well. Was it so that you could do 2 things at once inside your helper function: `canUserAccessQuery`?

ALCooper12 (approved):
Left a question comment, otherwise everything else LGTM! 

And thanks for adding the needed `saved_query_id` parameters in the other endpoints within the openAPI docs! :smile:

cmd/api/src/api/v2/saved_queries.go
LawsonWillard: Yes! Trying to minimize DB calls when possible

ALCooper12: Gotchaa. Thanks for letting me know!

PR #1720, opened 2025-07-24
feat: copy cell text to clipboard

benwaples:
## Description
Adds functionality to copy a cells value to the users clipboard if that cell is a string or array of strings. 

Also includes a fix for BED-5663 that will not add searchType=cypher to the url until there is a valid search to be made

## Motivation and Context

Resolves BED-6261 and BED-5663

BED-6261: give the user a quick way to copy a cells text to the clipboard
BED-5663: cleans up some work from deep linking that would only allow a search to be made if all parameters for that search existed. It became really noticeable with table view because table view is only available for cypherSearches, so if someone previously used explore table -> navigated away -> navigated back to explore -> opened cypher search, the table would automatically open without a query ever executing.

## How Has This Been Tested?
Added unit tests and performed manual tests.

## Screenshots (optional):

https://github.com/user-attachments/assets/a1751c8e-eb3c-4b67-afed-818e87653711

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.tsx
```diff
@@ -0,0 +1,80 @@
...
+                <FontAwesomeIcon
+                    icon={value ? faCheck : faCancel}
+                    color={value ? 'green' : 'lightgray'}
+                    className='scale-125'
+                />
+            </div>
+        );
+    }
+
+    const stringyKey = columnKey?.toString();
```
benwaples, line 47: lines 47-77 are the actual changes in this file. It looks like a completely new file because I moved the file into its own folder to be able to organize it with its tests!

cmd/ui/src/views/Explore/ExploreSearch/ExploreSearch.tsx
```diff
@@ -140,7 +140,9 @@ const ExploreSearch: React.FC = () => {
             params.exploreSearchTab = 'pathfinding';
         }
         if (tab === 'cypher') {
-            params.searchType = 'cypher';
+            if (cypherSearchState.cypherQuery) {
+                params.searchType = 'cypher';
+            }
```
benwaples, line 145: see BED-5663: When changing the exploreSearchTab, we should only update the searchType when values for the searchType exist. Right now, when selecting the cypher tab, we always set the searchTab to cypher, even if a cypherSearch is missing.

benwaples:
@coderabbitai review

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const handleCopyToClipBoard: React.MouseEventHandler<HTMLButtonElement> = (e) => {
+        e.stopPropagation(); // prevents the click event bubbling up the DOM and triggering the row click handler
+        if (typeof formattedValue === 'string') {
+            copyToClipboard(formattedValue);
+        } else {
+            copyToClipboard(formattedValue.join(', '));
+        }
+        setDisplayCopyCheckmark(true);
+        setTimeout(() => setDisplayCopyCheckmark(false), 500);
```
TheNando, line 58: This solution works. You'll probably be ok without needing to clean up the timeout since it's so short lived and only happens on the click. If we decide to use this copy button in multiple places, it may be worth wile to extract this to its own copy component and use CSS animations instead of setting timeouts. Just something to keep in mind if this use case becomes more common.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.tsx
```diff
@@ -0,0 +1,80 @@
...
+                )}>
+                <FontAwesomeIcon icon={displayCopyCheckmark ? faCheck : faCopy} />
+            </button>
+            <span className={cn('group-hover:pl-5 transition-[padding-left] ease-in', transitionDelay)}>
+                {formattedValue}
+            </span>
+        </span>
+    ) : (
+        FALLBACK_STRING
+    );
```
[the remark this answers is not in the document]
TheNando: The functionality works as expected without this change. The group is set in a parent table cell. (This is a note to CodeRabbit)

TheNando (approved):
Looks great!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.tsx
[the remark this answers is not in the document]
benwaples: thank you @TheNando!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.tsx
benwaples: ohhh ive never done that but thats a fantastic suggestion! My understanding is that the product team does want to make this a more common use case so Id love to take a stab at what youre talking about.

If youre cool with this, I would like to tackle this outside of the staging branch. I have a list of improvement that have come out of table view, may I add this suggestion to that?

benwaples: https://specterops.atlassian.net/browse/BED-6263

PR #1724, opened 2025-07-25
bed-6237: update documentation for posture endpoint

irshadaj:
## Description
We originally planned on removing the `from` and `to` filters (and related logic) which were API V1 contracts. However, we still have the filters documented in V2 which prevents us from outright deleting them. I have updated the documentation for the posture endpoint to reflect that these two filters are deprecated and will not be carried over to V3. Additionally, we were missing some other filters from the documentation, which have been added.

## Motivation and Context
Resolves https://specterops.atlassian.net/browse/BED-6237

## How Has This Been Tested?
Verified that the documentation has been updated on the local environment

Screenshot 2025-07-25 at 13 31 21

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added new filters to the posture stats endpoint: asset group tag ID, created_at, updated_at, and deleted_at, allowing more precise filtering by group and date ranges.
  * Changed the attack paths details endpoint method from PUT to GET for improved data retrieval.

* **Documentation**
  * Updated API documentation to mark "from" and "to" parameters as deprecated and recommend using the new created_at filter instead. Descriptions and examples for new filters are included.

### DIFF
--- a/packages/go/openapi/doc/openapi.json
+++ b/packages/go/openapi/doc/openapi.json
@@ -15903,9 +15903,6 @@
             "schema": {
               "$ref": "#/components/schemas/api.params.predicate.filter.time"
             }
-          },
-          {
-            "$ref": "#/components/parameters/query.asset-group-tag-id"
           }
         ],
         "responses": {
--- a/packages/go/openapi/src/paths/risk-posture.posture-stats.yaml
+++ b/packages/go/openapi/src/paths/risk-posture.posture-stats.yaml
@@ -87,7 +87,6 @@ get:
       required: false
       schema:
         $ref: './../schemas/api.params.predicate.filter.time.yaml'
-    - $ref: './../parameters/query.asset-group-tag-id.yaml'
   responses:
     200:
       description: OK


packages/go/openapi/doc/openapi.json
```diff
@@ -15820,7 +15820,7 @@
             "name": "from",
             "deprecated": true,
             "x-go-name": "fromDeprecated",
-            "description": "Lower bound for created_at; to return posture stats starting at a specific date/time",
+            "description": "This filter has been deprecated. Please use the created_at filter instead.",
```
mvlipka: Should we leave the original description and add the deprecation notice to the end of it?
The property is marked deprecated by the `deprecated` property, which will indicate that the user should switch to the new parameter, but I think we'd want to leave in the original documentation for the users still on v1 in order to help them make the switch

AD7ZJ (commented):
I thought openapi.json is auto generated, shouldn't these changes be made in the corresponding yaml files and then run `just gen-spec`?

AD7ZJ (approved):
Look good to me with the exception of code rabbit's note, I think that's a valid concern.  It looks to me like the ref should remain, asset_group_tag_id is not a regular filter?

PR #1734, opened 2025-07-29
BED-5084 Test Helpers for Posture API Tests

kpowderly:
## Description

Test helpers for https://github.com/SpecterOps/bloodhound-enterprise/pull/517

Test helper so we now have the ability to do an assertion on unordered lists. [Assert.json](https://pkg.go.dev/github.com/stretchr/testify/assert#JSONEq) respects list order so we would see intermittent failures if we didn't sort ourselves.

## Motivation and Context

Resolves [BED-5084](https://specterops.atlassian.net/browse/BED-5084)

## How Has This Been Tested?

Test Helper

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-5084]: https://specterops.atlassian.net/browse/BED-5084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Improved test utilities to support order-insensitive JSON comparisons, making tests more robust when comparing JSON data.
* **Tests**
  * Added comprehensive tests to verify correct handling and normalization of JSON structures, including sorting of arrays and nested elements.

### DIFF
--- a/cmd/api/src/utils/test/response.go
+++ b/cmd/api/src/utils/test/response.go
@@ -24,7 +24,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"sort"
+	"slices"
 	"strings"
 	"testing"
 
@@ -129,12 +129,13 @@ func OverwriteQueryParamIfHeaderAndParamExist(headers http.Header, headerKey, pa
 	return headers
 }
 
-// NormalizeJSON parses JSON and sorts arrays inside so order doesn't matter.
-// Returns raw string if input is not valid JSON.
-func NormalizeJSON(t *testing.T, raw string) any {
+// SortJSONArrayElements parses JSON and recursively sorts all arrays so their order doesn't affect equality.
+// Returns the raw string if the input is not valid JSON.
+//
+// Use this to normalize JSON bodies in tests where slice order is nondeterministic.
+func SortJSONArrayElements(t *testing.T, raw string) any {
 	var data any
-	err := json.Unmarshal([]byte(raw), &data)
-	if err != nil {
+	if err := json.Unmarshal([]byte(raw), &data); err != nil {
 		return raw
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/api/src/utils/test/response.go
```diff
@@ -126,3 +128,38 @@ func OverwriteQueryParamIfHeaderAndParamExist(headers http.Header, headerKey, pa
...
+				// Fallback to some consistent ordering when marshal fails
+				return fmt.Sprintf("%v", val[i]) < fmt.Sprintf("%v", val[j])
+			}
+ 			return string(bi) < string(bj)
+ 		})
+	case map[string]any:
+		for _, vv := range val {
+			sortJSONArrays(vv)
+		}
+	}
```
computator: This doesn't seem as if it will sort the map keys? I know Go maps are unordered, but to make "normalized" JSON the object keys would also normally be sorted. I'm not sure if that's relevant here or not.

cmd/api/src/utils/test/response.go
```diff
@@ -126,3 +128,38 @@ func OverwriteQueryParamIfHeaderAndParamExist(headers http.Header, headerKey, pa
...
+	return data
+}
+
+func sortJSONArrays(v any) {
+	switch val := v.(type) {
+	case []any:
+		for _, elem := range val {
+			sortJSONArrays(elem)
+		}
+		sort.SliceStable(val, func(i, j int) bool {
```
computator: I was just reading the [docs](https://pkg.go.dev/sort#SliceStable) for `SliceStable()` and it mentions that perhaps it would be better to use `slices.SortStableFunc()`?

cmd/api/src/utils/test/response.go
```diff
@@ -126,3 +128,38 @@ func OverwriteQueryParamIfHeaderAndParamExist(headers http.Header, headerKey, pa
...
+			bi, err1 := json.Marshal(val[i])
+			bj, err2 := json.Marshal(val[j])
+			if err1 != nil || err2 != nil {
+				// Fallback to some consistent ordering when marshal fails
+				return fmt.Sprintf("%v", val[i]) < fmt.Sprintf("%v", val[j])
```
computator: Perhaps `%+v` or `%#v` would be better so it gets a full representation of the item?

zinic (approved):
lgtm

cmd/api/src/utils/test/response.go
kpowderly: my brain is breaking... let me dig into this more

cmd/api/src/utils/test/response.go
kpowderly: You're right. I guess I should update the naming to be something like SortJSONArrayElements. The purpose was intended only to create a consistent deterministic structure for testing when comparing raw json objects containing unordered lists.

Normalize is a bit flippant here.

cmd/api/src/utils/test/response.go
kpowderly: Updated to use it but it feels overly complex for such a small use-case idk 🤷‍♀️ whatever you prefer!!

bsheth711 (approved):

computator (approved):

PR #1741, opened 2025-07-30
BED-6232: Explore Table data fetch refactor

benwaples:
## Description
We removed some of the mechanisms to control when we asked the cypher endpoint to return all the node properties which allows us to simplify how we fetch data for the explore table. Previously we passed data down from added some transformations to useRegraphExploreGraph and passed that data down to ExploreTable. Now we can leave useRegraphExploreGraph to handle data just for the graph, and execute those transformations directly in ExploreTable.

Additionally, this consolidates labels we use for node keys coming from a UnifiedNode so that the UI uses common labels for those properties. This was left out of the generate graphSchema file because those keys do not come from the graph/AD directly and are specifically added in the API or UI layer.

Side effects of this refactor:
- test log cleanup
- move cypher test response to shared-ui mocks.

## Motivation and Context

Resolves BED-6232

## How Has This Been Tested?
Updated tests, added test coverage, performed manual tests

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Improved property labeling and display in the Explore Table and related UI components for enhanced clarity and consistency.

* **Refactor**
  * Simplified and streamlined data flow in the Explore Table and related components, moving from external to internal state management and updating how data and selection are handled.
  * Updated column keys and labels (e.g., from 'nodetype' to 'kind') for better alignment with underlying data structures.
  * Converted components to fully controlled patterns, removing internal state where appropriate.
  * Replaced multiple conditional UI renderings with dynamic mapping for cleaner code.

* **Bug Fixes**
  * Enhanced reliability and accuracy of table sorting, column management, and row selection.

* **Tests**
  * Expanded and updated test coverage for Explore Table components, integrating dynamic mocked API data and improving test reliability.

* **Chores**
  * Removed obsolete test data files and cleaned up unused code and redundant imports.
  * Updated mock data exports and test utilities for consistency.


packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
```diff
@@ -43,16 +43,12 @@ export const ManageColumnsComboBox = ({
...
+    const selectedColumns = useMemo(
         () => allColumns.filter((item) => selectedColumnsProp[item.id]),
         [allColumns, selectedColumnsProp]
     );
-    useEffect(() => {
-        setSelectedColumns(initialColumns);
-    }, [initialColumns]);
 
     const pinnedColumns = useMemo(() => allColumns.filter((item) => item.isPinned), [allColumns]);
-    const [selectedColumns, setSelectedColumns] = useState<ManageColumnsComboBoxOption[]>(initialColumns);
```
benwaples, line 55: Looks like `selectedColumns` can be determined state from the selectedColumnsProp and we can skip the duplicate state updates and treats this component as always controlled. This investigation spurred from a "uncontrolled component becoming controlled" warning in test logs!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsListItem.tsx
```diff
@@ -37,7 +37,7 @@ const ManageColumnsListItem = ({ isSelected, item, onClick, itemProps }: ManageC
             e.stopPropagation();
             onClick(item);
         }}>
-        <button className="w-full text-left flex justify-between items-center'">
+        <div className="w-full text-left flex justify-between items-center'">
```
benwaples, line 40: This shouldnt change functionality but will address a console/test log warning. The checkbox component is a button and click events will still bubble up to the ListItem onClick, I believe this will be easier for screen readers compared to the nested button, but please correct me if im wrong on that!!

packages/javascript/bh-shared-ui/src/views/Explore/BasicObjectInfoFields.tsx
```diff
@@ -62,14 +62,23 @@ const RelatedKindField = (
...
+
 export const BasicObjectInfoFields: React.FC<BasicObjectInfoFieldsProps> = (props): JSX.Element => {
     return (
         <>
-            {props.nodeType && <Field label='Node Type' value={props.nodeType} />}
-            {props.isTierZero && <Field label='Tier Zero:' value={true} />}
-            {props.isOwned && <Field label='Owned Object:' value={true} />}
-            {props.displayname && <Field label='Display Name:' value={props.displayname} />}
-            <Field label='Object ID:' value={props.objectid} />
+            {basicObjectFields.map((field) => {
```
benwaples, line 76: standardizing how these labels are defined so we dont have to update multiple places in the future.

cmd/ui/src/hooks/useSigmaExploreGraph/useSigmaExploreGraph.tsx
```diff
@@ -33,23 +33,9 @@ export const normalizeGraphDataForSigma = (
...
-
     // return the full query so we can know loading/error state, and use react-query tools. But override the data field with the normalized value
     return {
         ...graphQuery,
-        data: {
-            nodes: normalizedGraphData,
-            node_keys: nodeKeys,
-            rawNodes: castGraphData?.nodes,
-        },
+        data: normalizedGraphData,
```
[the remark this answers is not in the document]
benwaples: transformFlatGraphResponse is still needed to transform the graph data into a format the graphing library can consume. Its not the most clear name because its doing more than just transforming a FlatGraphResponse, but also adding and computing properties.

jvacca-specterops (approved):

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -72,12 +70,12 @@ const ExploreTable = ({
         [setSearchInput]
     );
 
+    const exploreTableData = useMemo(() => getExploreTableData(graphData), [graphData]);
```
dcairnsspecterops, line 79: This is a good move. Makes me a bit nervous to shuffle data around, and I still wonder if we might want separate hooks for the table data, but I think yours is probably the better solution.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
dcairnsspecterops: If you've thoroughly tested it, I believe you! I was nervous about trusting the internal state to sync with our redux state, but I struggle to remember the issue it was causing. I assume something along the lines of initializing columns from redux state not causing `onChange` to fire?.

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -37,44 +37,34 @@ type DataTableProps = React.ComponentProps<typeof DataTable>;
...
-                    nextRow.id = key;
-                    nextRow.displayname = value?.label?.text;
-
-                    acc.push(nextRow as MungedTableRowWithId);
-
-                    return acc;
-                }, [])) ||
-            [],
-        [data]
+            ((exploreTableData?.nodes &&
```
dcairnsspecterops: The reduce upstairs here was to filter out edges at the same time as we create the new data. Does that happen somewhere else now?

I'm a little nervous about all these spread operations, since it looks like I was trying to avoid that in my reduce function while trying to improve performance. But maybe it's not important with virtualization helping?

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -206,14 +199,11 @@ const useExploreTableRowsAndColumns = ({
         const columnDefs = selectedColumnDefinitions.sort((a, b) => {
             const idA = a?.id || '';
             const idB = b?.id || '';
-            const aIsRequired = requiredColumns[idA];
-            const bIsRequired = requiredColumns[idB];
+            const aIsRequired = isRequiredColumn(idA);
+            const bIsRequired = isRequiredColumn(idB);
```
dcairnsspecterops, line 209: I'm curious what this function does for us, since it looks like a passthrough for property access. Were there some typing issues solved by it?

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
benwaples: Good questions!! `exploreTableData?.nodes` here is the equivalent of `GraphResponse['data'][''nodes]`, coming directly from the API. So we no longer need to filter out the edges! Previously, the response from `useSigmaExploreGraph` would mix edges+nodes to pass along to the graph, however GraphResponse will come from the api with edges and nodes separated into their own fields!

I see what youre saying about the extra spread operations! I bet youre right and virtualization is helping reduce any noticeable performance issues, and I agree the extra spread operations aren't idea. We could assign all known properties manually and eliminate the extra rest+spread operation. I think the rest+spread operation of `rest` (in this case, the known properties on a node) would be the extra enumeration in the changes because we previously used `Object.assign` to flatten the property bag with other known node keys. And I believe `Object.assign` enumerates under the hood, but please correct me if im wrong!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
benwaples: Thank you!! Yeah agreed its not ideal to shuffle how we pull in this data, especially because its working great as-is. Im hoping organizing it like this will make it a little easier to maintain and make changes in in the future! 

ohhh I like that suggestion a lot! Are you thinking something similar to how `useSigmaExploreGraph` pulls in `useExploreGraph` + any transformations needed? Essentially combining `useExploreGraph` + `getExploreTableData` for now? Or are you thinking something that would replace `useExploreGraph` entirely with a new `react-query`?

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
benwaples: Ultimately yes, its a type guard for when we use idA/idB below.

Previously `REQUIRED_EXPLORE_TABLE_COLUMN_KEYS` could be indexed with `string` because it was inferred as `string[]`, but I added `satisfies PropertyLabelOverides[]` to give us some autocomplete help, which meant that we could only index with one of the strings found in that array. So `isRequiredColumn` does the same thing it did previously under the hood, but adds a type guard for TS that says idA/idB is `(typeof REQUIRED_EXPLORE_TABLE_COLUMN_KEYS)[number]` if is considered required.

In `isRequiredColumn`, I am realizing I incorrectly casted `value as PropertyLabelOverides` in that function, so Im pushing a change that will remove that cast and change the name of the required keys array back to `REQUIRED_EXPLORE_TABLE_COLUMN_KEYS`

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
benwaples: Looks like we were overriding displayname with the node label (artifact from using the data transformed for the graph), so i also pushed a change to not set displayname to the label and also updates the search functionality to match against label instead of display name!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
benwaples: I did some investigation + psuedo code and Id like to push out this change until we figure out how we will support more APIs because depending on how we do that, it could mean more data transformations/ complexities that we would want to account for with that separate hook. Plus Id rather not break up the explore-table-utils file between a new hook and the component for now!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ManageColumnsComboBox/ManageColumnsComboBox.tsx
benwaples: Agreed! Im thinking the internal state couldve been added when we had the `useEffect` listening to internal state to reset columns to default if non were selected, but im not totally sure!

Ive tested default behavior, reset to default button, select all/none button, and manually unselecting all to reset to default and all are working as expected 👍

PR #1742, opened 2025-07-30
feat(OpenGraph): BED-5383 - Add endpoint for file level ingest error details

wes-mil:
## Description

Modify `/api/v2/file-upload/{id}` to accept an HTTP header to specify file upload name.

Add `/api/v2/file-upload/{id}/completed-tasks` to return details about individual files.

## Motivation and Context

Resolves: BED-5383

Provide more detailed information about errors during ingest.

## How Has This Been Tested?

Start file upload job. 
Upload file using the `POST /api/v2/file-upload/{id}` endpoint specifying the `X-File-Upload-Name` header. 
End file upload job. 
Call `GET /api/v2/file-upload/{id}/completed-tasks`. Ensure that `file_name` is populated on each file. If a zip was uploaded ensure that `parent_file` is also populated on each.

## Screenshots (optional):

Screenshot from 2025-08-19 12-06-00
Screenshot from 2025-08-19 12-06-11

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Optional X-File-Upload-Name header to provide the original filename.
  * New GET endpoint to list completed tasks for a file-upload job, showing per-file name, parent-file, and per-file errors.
  * Job summary now includes total_files and failed_files counts.

* **Improvements**
  * Uploads accepted only when the job is running (otherwise 400).
  * Temporary uploaded files are removed on task-creation failure to avoid leftover files.

* **Documentation**
  * OpenAPI updated with header, endpoint, schemas, and responses.


cmd/api/src/services/graphify/ingest.go
```diff
@@ -104,7 +103,7 @@ func ReadFileForIngest(batch *TimestampedBatch, reader io.ReadSeeker, options Re
 }
 
 func IngestBasicData(batch *TimestampedBatch, converted ConvertedData) error {
-	errs := util.NewErrorCollector()
+	errs := newGraphifyErrorBuilder()
```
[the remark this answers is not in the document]
wes-mil: Hello Mr Rabbit man, I'm pretty sure that if you look A LITTLE BIT harder you'll see the function

cmd/api/src/database/migration/migrations/v8.1.0.sql
```diff
@@ -50,3 +50,7 @@ CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_primary_kind ON asset_group_ta
 CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_environment_id ON asset_group_tag_selector_nodes USING btree (node_environment_id);
 CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_object_id ON asset_group_tag_selector_nodes USING btree (node_object_id);
 CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_name ON asset_group_tag_selector_nodes USING btree (node_name);
+
+-- File Ingest Details
+ALTER TABLE ingest_jobs ADD COLUMN IF NOT EXISTS task_info jsonb NOT NULL DEFAULT '{"completed_tasks": []}'::jsonb;
```
superlinkx: Why are we storing task info as jsonb? This should probably either be additional columns or an additional table

bsheth711: I would prefer we do not use jsonb, unless we have a good reason to do so. Instead we can use traditional relational DB design to capture this.

cmd/api/src/services/graphify/tasks.go
```diff
@@ -71,22 +82,31 @@ func (s *GraphifyService) extractIngestFiles(path string, fileType model.FileTyp
 			}
 		}()
 
-		for _, f := range archive.File {
+		for i, f := range archive.File {
 			//skip directories
 			if f.FileInfo().IsDir() {
 				continue
 			}
```
bsheth711, line 91: fileData is initialized with a predetermined size, but directories are skipped while iterating. This means that if there is a directory, there would be an empty default initialized IngestFileData in the slice at the index the directory was found. Instead, fileData should probably be created with size 0:

`fileData = make([]IngestFileData, 0, len(archive.File))`

Each new IngestFileData should then be appended.

cmd/api/src/services/graphify/tasks_integration_test.go
```diff
@@ -53,10 +53,18 @@ func TestVersion5IngestJSON(t *testing.T) {
...
+		fileData, err := testSuite.GraphifyService.ProcessIngestFile(ctx, model.IngestTask{FileName: file, FileType: model.FileTypeJson}, time.Now())
 		require.NoError(t, err)
+
+		failed := 0
+		for _, data := range fileData {
+			if len(data.Errors) > 0 {
+				failed++
+			}
+		}
+
```
bsheth711, line 65: Perhaps make a helper function for this

cmd/api/src/services/graphify/decoders.go
```diff
@@ -41,7 +40,7 @@ func decodeBasicData[T any](batch *TimestampedBatch, decoder *json.Decoder, conv
 	var (
 		count         = 0
 		convertedData ConvertedData
-		errs          = util.NewErrorCollector()
+		errs          = newGraphifyErrorBuilder()
```
superlinkx: This feels like a good opportunity to just use `[]error`, append errors to `errs`, and then `errors.Join()` at the end to get a singular error value (containing the others). This not only is simpler and doesn't require the new file, but also preserves errors.Is and errors.As semantics if we ever need them in higher code. (`errors.Join` on an empty list or even a list of nils is always `nil` so you don't need to do any extra checks either)

superlinkx (changes requested):

cmd/api/src/services/graphify/decoders.go
superlinkx: Rethinking this now that I've been given more context. It's not ideal, but given the way these functions are wired up and where we need to process the errors, this may be best effort right now.

cmd/api/src/services/graphify/tasks.go
wes-mil: Great call out :smile:

cmd/api/src/api/v2/fileingest_test.go
brandonshearin, line 1: add some cases that cover parsing out the filename header. include a case that captures your fallback-scenario of unknown.json and unknown.zip.

cmd/api/src/services/graphify/error.go
brandonshearin, line 1: why is this abstraction necessary? it seems like the only trick it provides over the util.ErrorCollector() is to flatten errors of type graphifyError.

is it possible to augment the existing util.ErrorCollector with that logic? something like--
```
func flattenErr(dst *[]error, err error) {
	if err == nil {
		return
	}
	switch e := err.(type) {
	case graphify.GraphifyError:
		for _, sub := range e.Errors {
			flattenErr(dst, sub)
		}
	default:
		*dst = append(*dst, e)
	}
}

func (s *errorCollector) Add(err error) {
	s.lock.Lock()
	defer s.lock.Unlock()
	flattenErr(&s.errors, err)
}
```

cmd/api/src/model/ingest.go
```diff
@@ -21,10 +21,11 @@ import (
 )
 
 type IngestTask struct {
-	FileName    string     `json:"file_name"`
-	RequestGUID string     `json:"request_guid"`
-	JobId       null.Int64 `json:"task_id" gorm:"column:task_id"`
-	FileType    FileType   `json:"file_type"`
+	FileName         string     `json:"file_name"`
+	ProvidedFileName string     `json:"provided_file_name"`
```
brandonshearin: it may be hard to quickly glean the difference between`file_name` and `provided_file_name` when we return to this code in the future. can you leave a comment that describes the difference between these two properties?

it would be helpful if the comment mentioned that one is intended as a user-provided name and the other is randomized on the FS to prevent filename collisions. maybe there is more to say but those are two important ones.

brandonshearin: in addition, lets consider renaming these fields on the go type.

FileName --> StoredFileName
ProvidedFileName --> OriginalFileName

to me, this emphasizes the distinction of how we use these fields in the system.

cmd/api/src/services/graphify/error.go
brandonshearin: if this route is feasible, sanity check this update to util.ErrorCollector with unit tests

wes-mil: There could certainly be a better way to do this, but due to both how util.ErrorCollector is written and how ingest is written we do kinda need this abstraction. 

The issue with `errors.Join()` is that it bundles the errors into (potentially) a tree structure. This _can_ be mitigated, but it's somewhat annoying. Not to mention, unwrapping those wrapped errors is pretty ugly. 

By having a custom error builder and consequently custom error type, we're able to bundle all of that relevant information together as it should be. Then upstream all we have to do is `errors.As(GraphifyError)` to capture the information. This means that even if a `GraphifyError` is wrapped into the `util.ErrorCollector`, we can extract it and only it. Rather than having to sift through other potential errors that got thrown onto the collector.

cmd/api/src/api/v2/fileingest_test.go
wes-mil: ~~Testing zip files requires writing a massive hex string in the test, which I think is pretty bad from a security perspective. I'm going to forgo that in the unit tests.~~

Just kidding, I fixed it

superlinkx: Why not load a real zip file from FS rather than encode it?

bsheth711: That could be done but it may be best to avoid interacting with the file system given this is a unit test

brandonshearin (approved):

cweidenkeller (approved):

PR #1745, opened 2025-07-31
BED-6168: Add support for custom icons on AGTs

AD7ZJ:
## Description

Add support for custom icons on asset group tags (privilege zones)

## Motivation and Context

Resolves https://specterops.atlassian.net/browse/BED-6168

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added optional "glyph" field to asset group tags (nullable string); API, database migration, OpenAPI docs and tests updated. Added two default glyph constants.

* **Bug Fixes**
  * Tests extended to verify glyph handling across create/read/update/delete and ordering flows, including null vs non-null behavior.

* **Improvements**
  * Enforced uniqueness for non-null glyphs and added clear user-facing error messages for invalid or duplicate glyph values.

### DIFF
--- a/cmd/api/src/model/assetgrouptags.go
+++ b/cmd/api/src/model/assetgrouptags.go
@@ -78,8 +78,8 @@ const (
 )
 
 const (
-	Tier0Glyph = "gem"
-	OwnedGlyph = "skull"
+	TierZeroGlyph = "gem"
+	OwnedGlyph    = "skull"
 )
 
 type AssetGroupTag struct {


cmd/api/src/database/migration/migrations/v8.1.0.sql
```diff
@@ -28,4 +28,7 @@ ON CONFLICT DO NOTHING;
 -- Targeted Access Control users_roles column additions and defaults
 ALTER TABLE users_roles 
         ADD COLUMN IF NOT EXISTS access_control_list text[] default array ['all_environments'],
-        ADD COLUMN IF NOT EXISTS explore_enabled bool DEFAULT true;
\ No newline at end of file
+        ADD COLUMN IF NOT EXISTS explore_enabled bool DEFAULT true;
+
+ALTER TABLE asset_group_tags
+        ADD COLUMN IF NOT EXISTS glyph TEXT UNIQUE;
```
superlinkx, line 55: Make sure this is not unique for null

superlinkx (changes requested):
Lgtm, let's check the DB constraint to ensure multiple nulls will be writable before pulling this in.

cmd/api/src/database/migration/migrations/v8.1.0.sql
AD7ZJ: added a test to prove that, ty!

cmd/api/src/model/assetgrouptags.go
```diff
@@ -77,6 +77,11 @@ const (
 	AssetGroupExpansionMethodParents  AssetGroupExpansionMethod = 3
 )
 
+const (
+	Tier0Glyph = "gem"
```
urangel: tiny nit: Prefer `TierZeroGlyph` here as it is a more consistent way of writing that out.

cmd/api/src/database/assetgrouptags_test.go
```diff
@@ -308,6 +313,27 @@ func TestDatabase_CreateAssetGroupTag(t *testing.T) {
...
+		dbInst := integration.SetupDB(t)
+
+		_, err := dbInst.CreateAssetGroupTag(testCtx, model.AssetGroupTagTypeTier, testActor, "t1 name", "", null.Int32{}, null.Bool{}, null.String{})
+		require.NoError(t, err)
+
+		_, err = dbInst.CreateAssetGroupTag(testCtx, model.AssetGroupTagTypeTier, testActor, "t2 name", "", null.Int32{}, null.Bool{}, null.String{})
+		require.NoError(t, err)
+	})
+
+	t.Run("Duplicate glyph errors out", func(t *testing.T) {
```
urangel, line 326: Do we need to gate updating the glyph for the default, system created tags (the tag for Tier Zero and the tag for Owned)? This seems like something that might be important to communicate consistent language through design.

cmd/api/src/database/migration/migrations/v8.1.0.sql
```diff
@@ -50,3 +50,6 @@ CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_primary_kind ON asset_group_ta
 CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_environment_id ON asset_group_tag_selector_nodes USING btree (node_environment_id);
 CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_object_id ON asset_group_tag_selector_nodes USING btree (node_object_id);
 CREATE INDEX IF NOT EXISTS idx_agt_selector_nodes_name ON asset_group_tag_selector_nodes USING btree (node_name);
+
+ALTER TABLE asset_group_tags
+        ADD COLUMN IF NOT EXISTS glyph TEXT UNIQUE;
```
urangel, line 55: Any unique text gives me some pause but I don't have a better suggestion at the moment. The UI will give users guardrails at least. Maybe the BH docs need to specify this should be associated with fontawesome icon string names so that API users have guidance.

urangel:
I imagine a large part of why we want custom glyphs for nodes associated with a certain tag is for displaying that on the graph. Is this work supposed to enable the UI for doing that? Do we need to update endpoints associated with graph responses to make that happen?

cmd/api/src/database/assetgrouptags_test.go
mistahj67: Those are and should be gated correct. I tested locally but if it didn't block for you, please lmk

PR #1749, opened 2025-07-31
BED-6287 - Table Tweaks

dcairnsspecterops:
## Description

This PR bundles a few tasks:
1) Remove reference to `explore_table_view` feature flag
2) Move copy to clipboard button to `bh-shared-ui` component, using CSS animation callbacks rather than setTimeout to reset the clipboard icon
3) Download data as CSV, not json and move download logic to ExploreTable, removing the `onDownloadClick` callback from `ExploreTable`
4) Deselect a row on click. This includes explicitly NOT deselecting if user clicks on the kebab menu for a selected row. 
5) Update [DoodleUI with updates to DataTable](https://github.com/SpecterOps/DoodleUI/pull/64
), syncing internal selected row state with selectedRow prop.
## Motivation and Context

Resolves BED-6287, BED-6264, BED-6263, BED-6231.

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Reusable copy-to-clipboard button added.
  * New clear-selection action to reset the current explore selection.

* **Enhancements**
  * Explore table export now produces CSV for download.
  * Clicking an already-selected row clears selection.
  * Kebab menu clicks no longer toggle row selection.
  * Defensive error handling when rendering fields.
  * Clipboard icon reset animation added.

* **Dependency Updates**
  * Added json-2-csv; updated doodleui versions.

* **Tests**
  * Updated tests for CSV export and clipboard; simplified feature-flag mocks.


packages/javascript/bh-shared-ui/src/components/CopyToClipboardButton/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    if (element?.tagName === 'BUTTON') {
+                        setDisplayCopyCheckmark(false);
+                        onAnimationEnd(animationEvent);
+                    }
+                }}
+                className={cn(
+                    'absolute top-1/2 left-2 -translate-x-1/2 -translate-y-1/2 opacity-0 pr-1 group-hover:opacity-100 transition-opacity ease-in',
+                    transitionDelay,
+                    {
+                        'animate-[null-animation_1s]': displayCopyCheckmark,
```
dcairnsspecterops, line 72: If we point the `onAnimationStart` and `onAnimationEnd` callbacks to an empty keyframes animation, we can treat it like a timeout to set the state. Is that our intention? Is there some other way we'd want to manage the "📋 -> click -> ✅ -> wait a second -> 📋 until next click" logic? Seems to me we need to use state to manage this somehow?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.test.tsx
```diff
@@ -38,9 +38,14 @@ describe('ExploreTableDataCell', () => {
 
         await user.click(screen.getByText('copy'));
 
+        screen.debug();
```
urangel: This debug snuck in 🧹

packages/javascript/bh-shared-ui/package.json
```diff
@@ -19,7 +19,7 @@
     "author": "The BloodHound Enterprise Team (https://bloodhoundenterprise.io/)",
     "license": "MIT",
     "dependencies": {
-        "@bloodhoundenterprise/doodleui": "^1.0.0-alpha.22",
+        "@bloodhoundenterprise/doodleui": "/Users/dpcairns/coding/something_30040/package.tgz",
```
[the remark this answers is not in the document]
urangel: ➕

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -102,7 +112,7 @@ const ExploreTable = ({
                     columns={columnOptionsForDropdown}
```
urangel: I noticed some discrepancies in the displayed table values vs the json response and the entity panel display.

image

The display name seems to be showing the same value as name in the table but there is a different value in the response and the entity panel. Last logon replicated seems to be missing some transformation as a value of -1 will display as NEVER in the entity panel. Are these differences intentional? 

We may need to use the same data parsing and transformation that is used in the entity panel if the values should align. This might also be the case for the data output to csv. I'm not sure if it is expected that the values be the raw values from the JSON response or the "pretty" transformed values from the entity panel.

packages/javascript/bh-shared-ui/src/components/CopyToClipboardButton/CopyToClipboardButton.tsx
```diff
@@ -0,0 +1,62 @@
...
+            copyToClipboard(value);
+        } else {
+            copyToClipboard(value.join(', '));
+        }
+        setDisplayCopyCheckmark(true);
+    };
+
+    return (
+        <>
+            <button
```
urangel: It would be ideal to avoid nesting buttons for upholding HTML and accessibility standards. Have we considered a tooltip approach for copying the value?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.test.tsx
dcairnsspecterops: good catch!

packages/javascript/bh-shared-ui/src/components/CopyToClipboardButton/CopyToClipboardButton.tsx
dcairnsspecterops: How would you feel about a div with `role='button'`?

packages/javascript/bh-shared-ui/src/views/Explore/fragments.tsx
```diff
@@ -104,13 +104,17 @@ export const FieldsContainer: React.FC<PropsWithChildren> = ({ children }) => {
...
+        if (
+            value === undefined ||
+            value === '' ||
+            (Array.isArray(value) && value.length === 0) ||
+            (typeof value === 'object' && Object.keys(value).length === 0)
+        )
+            return null;
+    } catch (e) {
         return null;
+    }
```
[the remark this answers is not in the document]
dcairnsspecterops: I don't know why `null` was initially excluded from the check, and don't want to introduce regressions in case it's important to exclude `null` from this check

packages/javascript/bh-shared-ui/src/components/CopyToClipboardButton/index.ts
```diff
@@ -0,0 +1,3 @@
+import { CopyToClipboardButton } from './CopyToClipboardButton';
```
benwaples, line 16: Mind running just prepare-for-codereview to generate the licenses for the new files in this PR?

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -95,21 +97,34 @@ const ExploreTable = ({
...
-            exportToJson({ nodes: exploreTableData?.nodes });
-        }
-    }, [exploreTableData?.nodes]);
-
     const handleRowClick = useCallback(
         (row: MungedTableRowWithId) => {
             if (row.id !== selectedItem) {
                 setSelectedItem(row.id);
+            } else {
+                setSelectedItem('');
```
benwaples: This is so nice!! Id suggest that rather than defining what it means to remove a `selectedItem` here, lets add a  `removeSelectedItem` (or any name lol) function to the `useExploreSelectedItem`  that would set `selectedItem: null` and `expandedSections: null`. That way if/when we need to run this logic in other places, we already have a definition of what it means to remove a `selectedItem`!

Also, during refinement for BED-6264 that I didnt update the ticket with (or convey to you, sorry) was that the minimize button on the entity panel should become an X and would `removeSelectedItem`.  We didnt cover a second click on a given row would  `removeSelectedItem`, however I do really like it and find it ergonomic. Ill reach out to @StephenHinck to confirm if we want to:
1. include the second click of a row to `removeSelectedItem`
2. and check if the edge info panel minimize button should also be an X to match the discussed functionality for the entity panel. 

It may make sense to make the entity panel changes in a separate PR as this one is almost done, but totally up to you!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -95,21 +97,34 @@ const ExploreTable = ({
...
+        if (nodes) {
+            const nodeValues = Object.values(nodes)?.map((node) => {
+                const flattenedNode = Object.assign(node, node.properties);
+
+                delete flattenedNode.properties;
+
+                return flattenedNode;
+            });
+
+            const csv = json2csv(nodeValues, { keys: exploreTableData.node_keys });
```
benwaples: This library looks nice and handles all the weird edge cases of using different data types in csv like arrays and objects (objects arent a concern rn for us)! 

I think one potential improvement would be to replace undefined values with an empty string or something that would treat that value as an empty cell? Im not sure exactly how excel/sheets interpret undefined in a given cell!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
dcairnsspecterops: Good catch! This is fixed in main with @benwaples's changes

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTableDataCell/ExploreTableDataCell.test.tsx
benwaples: Thats my bad!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -95,21 +97,34 @@ const ExploreTable = ({
...
+
+                delete flattenedNode.properties;
+
+                return flattenedNode;
+            });
+
+            const csv = json2csv(nodeValues, { keys: exploreTableData.node_keys, emptyFieldValue: '' });
+
+            fileDownload(csv, 'nodes.csv');
+        }
```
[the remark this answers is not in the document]
dcairnsspecterops: VSCode seems to disagree. It seems like this works without awaiting, which seems unlikely if returning a promise.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -95,21 +97,34 @@ const ExploreTable = ({
...
+        const nodes = exploreTableData?.nodes;
+        if (nodes) {
+            const nodeValues = Object.values(nodes)?.map((node) => {
+                const flattenedNode = Object.assign(node, node.properties);
+
+                delete flattenedNode.properties;
+
+                return flattenedNode;
+            });
+
```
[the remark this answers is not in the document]
dcairnsspecterops: We are mutating a copy of the object, not the original object.

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
benwaples: This is a good catch! Looks like I missed "Last logon replicated" (and 3 others) in my recent changes to sync up table values/headers with the entity panel :/ 

The values we display in the table do go through the same formatting as values for the entity panel, but we arent passing in a given values `kind` (for example: ad, az, cm, ov) which in this case actually tells the formatting that `ADSpecificTimeProperties` values of `-1` that have a kind of `ad` should display as `NEVER`.

The fix would be to add a check for the labels kind:
`ExploreTable/ExploreTableDataCell/ExploreTableDataCell.tsx`
```ts
    const { kind } = validateProperty(stringyKey); // ADDED
    const formattedValue = format({ keyprop: stringyKey, value, label: stringyKey, kind });
```   

@dcairnsspecterops
1. Ill work on a separate PR to get this fixed! I want to spend a moment to investigate any performance implications because `validateProperty` will enumerate a number of lists and I think there could be some optimization for `validateProperty`. 
2. I do agree with Uli's suggestion that the csv should use the "pretty" values, would it make sense to include that in this PR or a separate one? FWIW I think we can re-run the same transformation functions inside the csv handler so we dont have to transform all the data for every single query. That way we get a little more benefit from the virtualization only calling it when needed!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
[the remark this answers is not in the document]
dcairnsspecterops: https://github.com/mrodrig/json-2-csv/blob/main/upgrade_guides/UPGRADE_4_to_5.md

Asynchronous methods are now synchronous
Thanks to [@Nokel81](https://github.com/Nokel81) for the pull request to convert the module to convert the asynchronous internal flow to be entirely synchronous. This helps enable certain use cases where a conversion is needed in a synchronous manner, or where asynchronous code might not be feasible. While these changes shift the module towards a more synchronous focused use case, it's still entirely possible to perform JSON to CSV or CSV to JSON conversions for an asynchronous use case too.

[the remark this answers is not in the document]
benwaples: @coderabbitai I think youre referring to v4 docs

packages/javascript/bh-shared-ui/src/hooks/useExploreSelectedItem.tsx
```diff
@@ -37,6 +37,13 @@ export const useExploreSelectedItem = () => {
         [selectedItem, setExploreParams]
     );
 
+    const clearSelectedItem = useCallback(() => {
+        setExploreParams({
+            expandedPanelSections: null,
+            selectedItem: null,
+        });
+    }, [setExploreParams]);
+
```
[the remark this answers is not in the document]
benwaples: `useGraphItem` will only execute the `queryFn` if itemId is truethy because the enabled field i set to `!!itemId` already. Our version of react-query doesnt have a typeguard for itemId when its used in the enabled field so this suggestion wouldnt change anything besides the fallback of null becoming undefined which we dont need to do in this changeset.

packages/javascript/bh-shared-ui/src/components/CopyToClipboardButton/index.ts
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+import { CopyToClipboardButton } from './CopyToClipboardButton';
+
+export default CopyToClipboardButton;
```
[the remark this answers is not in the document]
benwaples: We dont need to export both default and named from the index, so im going to respectfully decline this suggestions  thank you

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -15,16 +15,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { DataTable } from '@bloodhoundenterprise/doodleui';
+import fileDownload from 'js-file-download';
+import { json2csv } from 'json-2-csv';
 import { ChangeEvent, memo, useCallback, useMemo, useState } from 'react';
```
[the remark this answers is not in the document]
benwaples: rollup has this defined as an external dep so it can be expected to be existing in the environment that imports bh-shared-ui

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
benwaples: Resolving this and addressing the 2 points above in follow up PRs!

packages/javascript/bh-shared-ui/src/components/ExploreTable/ExploreTable.tsx
```diff
@@ -89,6 +90,15 @@ const ExploreTable = ({
...
     );
 
+    const handleDownloadClick = useCallback(() => {
+        if (data) {
+            const nodeValues = Object.values(data)?.map((node) => node.data);
+            const csv = json2csv(nodeValues, { keys: allColumnKeys });
+
+            fileDownload(csv, 'nodes.csv');
+        }
+    }, [data, allColumnKeys]);
```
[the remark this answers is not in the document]
benwaples: is this something to be concerned about? Im not super familiar with these libraries so I lean on the side of caution of adding try/catch with a toast notification if it fails, but perhaps these functions throwing is unlikely/impossible

benwaples (approved):

PR #1767, opened 2025-08-07
BED-6332 feat(AGT): add denormalized property fields to selector nodes table

mistahj67:
## Description

Adds denormalized property fields to selector nodes table with backfill:
- node name
- node object id
- node primary kind
- node environment id

Also added indexes for the new property columns

## Motivation and Context

Resolves BED-6332 

## How Has This Been Tested?
Locally ran with new migrated columns and verified all were populated accordingly

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Selector nodes now store and surface richer metadata (primary kind, environment ID, object ID, display name) and use a centralized helper to extract these properties.

* **Database**
  * Added denormalized columns and indexes to persist and query the new selector-node fields; added users.all_environments (default true).

* **Bug Fixes**
  * More granular detection and updates for selector nodes based on expanded metadata and certify state.

* **Tests**
  * Updated tests and mocks to accommodate extended selector-node data and reordered arguments.

* **Chores**
  * Centralized missing-name/object-id defaults via shared constants.

### DIFF
--- a/cmd/api/src/daemons/datapipe/agt.go
+++ b/cmd/api/src/daemons/datapipe/agt.go
@@ -417,8 +417,7 @@ func SelectNodes(ctx context.Context, db database.Database, graphDb graph.Databa
 		certified   = model.AssetGroupCertificationNone
 		certifiedBy null.String
 
-		nodeIdsToDelete []graph.ID
-		nodesToUpdate   []model.AssetGroupSelectorNode
+		nodesToUpdate []model.AssetGroupSelectorNode
 	)
 
 	if selector.AutoCertify.ValueOrZero() {
@@ -483,7 +482,7 @@ func SelectNodes(ctx context.Context, db database.Database, graphDb graph.Databa
 			}
 		}
 
-		slog.Info("AGT: Completed selecting", "selector", selector.Name, "countTotal", len(nodesWithSrcSet), "countInserted", countInserted, "countUpdated", len(nodesToUpdate), "countDeleted", len(nodeIdsToDelete))
+		slog.Info("AGT: Completed selecting", "selector", selector.Name, "countTotal", len(nodesWithSrcSet), "countInserted", countInserted, "countUpdated", len(nodesToUpdate), "countDeleted", len(oldSelectedNodesByNodeId))
 	}
 	return nil
 }


cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -434,15 +434,15 @@ func SelectNodes(ctx context.Context, db database.Database, graphDb graph.Databa
...
-				if err = db.InsertSelectorNode(ctx, selector.AssetGroupTagId, selector.ID, id, certified, certifiedBy, node.Source, displayName); err != nil {
+			if oldNode, ok := oldSelectedNodesByNodeId[id]; !ok {
+				if err = db.InsertSelectorNode(ctx, selector.AssetGroupTagId, selector.ID, id, certified, certifiedBy, node.Source, primaryKind, envId, objectId, displayName); err != nil {
 					return err
 				}
 				countInserted++
-				// Auto certify is enabled but this node hasn't been certified, certify it
-			} else if selector.AutoCertify.Bool && oldCert == model.AssetGroupCertificationNone {
+				// Auto certify is enabled but this node hasn't been certified, certify it. Further - update any out of sync node properties
+			} else if (selector.AutoCertify.Bool && oldNode.Certified == model.AssetGroupCertificationNone) || oldNode.NodeName != displayName || oldNode.NodePrimaryKind != primaryKind || oldNode.NodeEnvironmentId != envId || oldNode.NodeObjectId != objectId {
```
computator: I'd reformat this to multiple lines so it's easier to follow but the logic seems good

cmd/api/src/daemons/datapipe/agt.go
```diff
@@ -453,8 +453,8 @@ func SelectNodes(ctx context.Context, db database.Database, graphDb graph.Databa
 		// Update the selected nodes that need updating
 		if len(nodeIdsToUpdate) > 0 {
 			for _, nodeId := range nodeIdsToUpdate {
-				displayName, _ := nodesWithSrcSet[nodeId].Properties.GetWithFallback(common.Name.String(), "NO NAME", common.DisplayName.String(), common.ObjectID.String()).String()
-				if err = db.UpdateSelectorNodesByNodeId(ctx, selector.AssetGroupTagId, selector.ID, certified, certifiedBy, nodeId, displayName); err != nil {
+				primaryKind, displayName, objectId, envId := model.GetAssetGroupMemberProperties(nodesWithSrcSet[nodeId].Node)
```
computator: It seems like we just retrieved this same data above. Would it be feasible to store it so we only need to retrieve it once for both steps? Especially if there'll be a lot of nodes updated. If it's just a few it doesn't matter so much.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -691,14 +691,18 @@ func (s *BloodhoundDB) InsertSelectorNode(ctx context.Context, assetGroupTagId,
...
+func (s *BloodhoundDB) UpdateSelectorNodesByNodeId(ctx context.Context, assetGroupTagId, selectorId int, nodeId graph.ID, certified model.AssetGroupCertification, certifiedBy null.String, primaryKind, environmentId, objectId, displayName string) error {
 	return s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		if result := tx.WithContext(ctx).Exec(fmt.Sprintf("UPDATE %s SET certified = ?, certified_by = ?, updated_at = current_timestamp WHERE selector_id = ? AND node_id = ?", model.AssetGroupSelectorNode{}.TableName()), certified, certifiedBy, selectorId, nodeId); result.Error != nil {
+		var oldSelectorNode model.AssetGroupSelectorNode
+		if err := tx.WithContext(ctx).Raw(fmt.Sprintf("SELECT selector_id, node_id, certified, certified_by, node_primary_kind, node_environment_id, node_object_id, node_name, updated_at, created_at FROM %s WHERE selector_id = ? AND node_id = ?", model.AssetGroupSelectorNode{}.TableName()), selectorId, nodeId).First(&oldSelectorNode); err.Error != nil && !errors.Is(err.Error, gorm.ErrRecordNotFound) {
```
computator, line 697: This ignores a record that isn't found. For an update method I would expect it to error when the thing you are trying to update doesn't exist (and I think that's what it used to do as well).

computator: Also I'm not seeing what the data is being selected for? We check the certified field but that's it. The rest appears unused?

cmd/api/src/database/migration/migrations/v8.1.0.sql
```diff
@@ -26,6 +26,22 @@ VALUES (current_timestamp,
...
+
+-- Add denormalized property columns to asset_group_tag_selector_nodes table
+ALTER TABLE ONLY asset_group_tag_selector_nodes
+  ADD COLUMN IF NOT EXISTS node_primary_kind TEXT;
+ALTER TABLE ONLY asset_group_tag_selector_nodes
+  ADD COLUMN IF NOT EXISTS node_environment_id TEXT;
+ALTER TABLE ONLY asset_group_tag_selector_nodes
+  ADD COLUMN IF NOT EXISTS node_object_id TEXT;
+ALTER TABLE ONLY asset_group_tag_selector_nodes
+  ADD COLUMN IF NOT EXISTS node_name TEXT;
```
computator: I don't think we are using table inheritance so `ONLY` should be unnecessary, and these statements can all be combined into a single statement to alter the table all at once instead of multiple times.

cmd/api/src/daemons/datapipe/agt.go
mistahj67: Ideally it's a lot the first time, but after that it should just be any that are different / changed which should be a lot less

cmd/api/src/database/assetgrouptags.go
mistahj67: I can jump back to before but if you issue an update with a where clause and it finds nothing, it shouldn't throw an error right, just like delete's, it just no-ops?

I just threw it all in to be safe, but correct, only certified is being checked. I could just store that instead

cmd/api/src/model/assetgrouptags.go
```diff
@@ -242,15 +246,31 @@ func (s AssetGroupTagSelector) ValidFilters() map[string][]FilterOperator {
...
 func (s AssetGroupSelectorNode) TableName() string {
 	return "asset_group_tag_selector_nodes"
 }
+
+/*
+These are the relevant properties for asset group tags. This method serves to keep consistency across the feature
+*/
+func GetAssetGroupMemberProperties(node *graph.Node) (primaryKind, displayName, objectId, envId string) {
+	primaryKind = analysis.GetNodeKindDisplayLabel(node)
+	displayName, _ = node.Properties.GetWithFallback(common.Name.String(), "NO NAME", common.DisplayName.String(), common.ObjectID.String()).String()
```
sirisjo: **nitpick** (non-blocking): I would move `NO NAME` and `NO OBJECT ID` into variables for maintainability/just in case we end up wanting to use these in other places down the road.

PR #1746, opened 2025-07-31
BED-6211: Migration for users_roles table to support TAC ACL

bsheth711:
## Description

PR to update users_roles table to support TAC ACL

## Motivation and Context

Resolves [BED-6211](https://specterops.atlassian.net/browse/BED-6211)

## How Has This Been Tested?

This has been manually tested

## Types of changes

- Database Migrations

## Checklist:

- [X] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [X] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments (GoDocs / JSDocs)
- [X] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

[BED-6211]: https://specterops.atlassian.net/browse/BED-6211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Introduced support for targeted access control, allowing more granular management of user permissions.
  * Added new settings for user roles to enable or disable exploration features and specify environment access.

* **Chores**
  * Updated database to support new access control features.

PR #1784, opened 2025-08-15
BED-6377 modified UpdateUser & CreateUser to take etac list properties

mvlipka:
## Description

This PR updates the CreateUser and UpdateUser auth endpoints to support ETAC

## Motivation and Context

Resolves: BED-6377

## How Has This Been Tested?
HoundStrike integration tests were covered and the screenshots are listed in the associated BHE PR

I ensured the openapi changes were correct by viewing the documentation in the API Explorer

I also tested creating and updating a user in the UI to ensure nothing was broken there

Then, with the etac feature flag turned on, I was able to test each endpoint

### CreateUser
I created payloads and sent them to `POST:/api/v2/bloodhound-users`

Testing all_environments being set to false by default
```
{
  "first_name": "USR qCaE",
  "last_name": "USR PfgCft",
  "email_address": "TESTME@so.io",
  "principal": "TESTME",
  "roles": [
    2
  ],
  "secret": "testPassword1234!",
  "needs_password_reset": false
}
```
And ensured that `all_environments` was set to `false` in the database for the newly created user

Testing all_environments being set in the payload
```
{
  "first_name": "USR qCaE",
  "last_name": "USR PfgCft",
  "email_address": "TESTME@so.io",
  "principal": "TESTME",
  "roles": [
    2
  ],
  "secret": "testPassword1234!",
  "needs_password_reset": false,
  "environment_access_control": {
    "all_environments": true
  }
}
```
Similarly, I checked the user's all_environments value in the database and ensured it was true

Testing creating a user with an etac list
```
{
  "first_name": "USR qCaE",
  "last_name": "USR PfgCft",
  "email_address": "TESTME@so.io",
  "principal": "TESTME",
  "roles": [
    2
  ],
  "secret": "testPassword1234!",
  "needs_password_reset": false,
  "environment_access_control": {
    "environments": [
      {
        "environment_id": "S-1-5-21-3702535222-3822678775-2090119576"
      },
      {
        "environment_id": "S-1-5-21-2697957641-2271029196-387917394"
      }
    ]
}
```
And I ensured that 2 new rows were created in the `environment_access_control` table were made, with the new user's user_id and the sent environments set

Testing 400 Bad Request when setting etac on a user who is an admin or power user
```
{
  "first_name": "USR qCaE",
  "last_name": "USR PfgCft",
  "email_address": "TESTME@so.io",
  "principal": "TESTME",
  "roles": [
    1
  ],
  "secret": "testPassword1234!",
  "needs_password_reset": false,
  "environment_access_control": {
    "environments": [
      {
        "environment_id": "S-1-5-21-3702535222-3822678775-2090119576"
      },
      {
        "environment_id": "S-1-5-21-2697957641-2271029196-387917394"
      }
    ]
}
```
```
{
  "first_name": "USR qCaE",
  "last_name": "USR PfgCft",
  "email_address": "TESTME@so.io",
  "principal": "TESTME",
  "roles": [
    1
  ],
  "secret": "testPassword1234!",
  "needs_password_reset": false,
  "environment_access_control": {
     "all_environments": true
  }
}
```
And ensured that I received a 400 bad request response for each request

### UpdateUser
I created payloads and sent them to `PATCH:/api/v2/bloodhound-users/<user_id>`

Testing setting all_environments to true on a non-admin
```
{
  "environment_access_control": {
    "all_environments": true
  }
}
```

Testing setting an environment list on a non-admin
```
{
  "environment_access_control": {
    "environments": [
      {
        "environment_id": "S-1-5-21-3702535222-3822678775-2090119576"
      },
      {
        "environment_id": "S-1-5-21-2697957641-2271029196-387917394"
      }
    ]
}
```

Both resulted in a 200 response and successfully applied the correct db changes for the associated user

Testing setting etac on a user who is an admin
```
{
  "environment_access_control": {
  "environment_access_control": {
    "environments": [
      {
        "environment_id": "S-1-5-21-3702535222-3822678775-2090119576"
      },
    ] 
}
}
```
```
{
  "environment_access_control": {
    "all_environments": true
  }
}
```
Both requests result in a 400 bad request

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Per-user Environment Targeted Access Control (ETAC): users can be all_environments or have an environment_access_control list; ETAC enforced on create/update and disallowed for Administrator/Power User roles.

* **Documentation**
  * OpenAPI and schemas updated to expose all_environments, environment_access_control, update payloads, and mutual-exclusivity rules.

* **Chores**
  * Migration defaults existing users to all_environments=true; tests and mocks updated to support ETAC and environment validation.

### DIFF
--- a/cmd/api/src/database/migration/migrations/v8.3.0.sql
+++ b/cmd/api/src/database/migration/migrations/v8.3.0.sql
@@ -22,7 +22,7 @@ DO
 $$
     BEGIN
         ALTER TABLE environment_access_control
-            RENAME COLUMN environment_id TO environment_id;
+            RENAME COLUMN environment TO environment_id;
     EXCEPTION
         WHEN undefined_column THEN
     END;


cmd/api/src/api/v2/auth/auth.go
```diff
@@ -484,6 +528,45 @@ func (s ManagementResource) UpdateUser(response http.ResponseWriter, request *ht
...
+						newEnvironment := model.EnvironmentAccess{
+							UserID:      user.ID.String(),
+							Environment: environment,
+						}
+						environments = append(environments, newEnvironment)
+					}
+					user.EnvironmentAccessControl = environments
+				}
+			}
+		}
```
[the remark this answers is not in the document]
bsheth711: Good catch

cmd/api/src/database/access_control_list.go
```diff
@@ -18,65 +18,60 @@ package database
...
-
-// EnvironmentAccess defines the model for a row in the environment_access_control table
-type EnvironmentAccess struct {
-	UserID      string `json:"user_id"`
-	Environment string `json:"environment"`
-	model.BigSerial
-}
-
-func (EnvironmentAccess) TableName() string {
-	return EnvironmentAccessControlTable
```
bsheth711, line 43: Could elect to keep `TableName()` function here since it is related to the database

mvlipka: It's still there, the `EnvironmentAccess` struct was just moved to the `model` package

packages/go/openapi/doc/openapi.json
```diff
@@ -17982,6 +18004,29 @@
...
+              }
+            },
+            "oneOf": [
+              {
+                "required": "all_environments"
+              },
+              {
+                "required": "environments"
+              }
+            ]
```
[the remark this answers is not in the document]
mvlipka: The openapi.json is generated via our .yaml files and a generate command, direct modifications to this file are not supported and the modification should occur in the .yaml file associated with this object

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -484,6 +528,51 @@ func (s ManagementResource) UpdateUser(response http.ResponseWriter, request *ht
...
+					for _, environment := range updateUserRequest.EnvironmentControlList.Environments {
+						newEnvironment := model.EnvironmentAccess{
+							UserID:      user.ID.String(),
+							Environment: environment,
+						}
+						environments = append(environments, newEnvironment)
+					}
+					user.EnvironmentAccessControl = environments
+				}
+			}
```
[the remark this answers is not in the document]
mvlipka: This is intentional as it lets GORM handle the creation of the etac list

cmd/api/src/database/access_control_list.go
```diff
@@ -18,65 +18,61 @@ package database
...
+	"fmt"
 
 	"github.com/specterops/bloodhound/cmd/api/src/model"
 	"gorm.io/gorm"
 )
 
 const (
-	EnvironmentAccessControlTable = "environment_access_control"
+	EnvironmentAccessControlTable       = "environment_access_control"
+	EnvironmentAccessControlFeatureFlag = "targeted_access_control"
```
stephanieslamb: I think we currently have our feature flags in `appconfig.go`.  Is there a reason we have this one in the database?

mvlipka: Nope! Just an oversight, fixing now

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -371,6 +397,28 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
...
+			if etacFeatureFlag.Enabled {
+				// In order to properly create the environment access rows for the new user, we need to wait for the user to be created
+				// to obtain the user's id to link the user to the new etac list
+				if createUserRequest.UpdateUserRequest.EnvironmentControlList != nil {
+					// If the user isn't granting access to all environments, give them access to each environment requested
+					if createUserRequest.UpdateUserRequest.EnvironmentControlList.AllEnvironments {
+						if err := s.db.DeleteEnvironmentListForUser(request.Context(), newUser); err != nil {
+							api.HandleDatabaseError(request, response, err)
+							return
+						}
```
bsheth711: Is this necessary given we are creating a new user? There should be no records for them in the environment_access_control table

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -371,6 +397,28 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
 				api.HandleDatabaseError(request, response, err)
 			}
 		} else {
+			// eTAC
+			if etacFeatureFlag.Enabled {
+				// In order to properly create the environment access rows for the new user, we need to wait for the user to be created
+				// to obtain the user's id to link the user to the new etac list
+				if createUserRequest.UpdateUserRequest.EnvironmentControlList != nil {
```
bsheth711: These two conditions can be combined:
`
if etacFeatureFlag.Enabled && createUserRequest.EnvironmentControlList != nil {
`

bsheth711:
The User Creation and Updates are happening in separate DB transactions than updates to the `environment_access_control` table. The API will return an error, but the state will be persisted in the database. In the case of User Creation, it is possible that a user will be created and persisted to the database, but their environment list will not. In the case of User Update it is possible that a user's environment list will be created and persisted to the database but the update to the user will fail. 

These operations should happen in a DB transaction together such that if either the update to the `environment_access_control table` or `users` table fails, both operations will be rolled back.

cmd/api/src/api/v2/auth/auth.go
mvlipka: Nope! Good catch, will get it removed

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -484,6 +519,46 @@ func (s ManagementResource) UpdateUser(response http.ResponseWriter, request *ht
 			user.Roles = roles
 		}
 
+		// eTAC
+		if etacFeatureFlag, err := s.db.GetFlagByKey(request.Context(), appcfg.FeatureEnvironmentAccessControl); err != nil {
```
mistahj67, line 508: As both of these are the same, would it make sense to create a singular pathway / abstracted fn here for maintainability?

cmd/api/src/database/access_control_list.go
```diff
@@ -28,76 +29,32 @@ const (
...
-		user.AllEnvironments = false
-
-		err = s.UpdateUser(ctx, user)
-
-		if err != nil {
-			return err
-		}
-
-		return nil
+		result := tx.WithContext(ctx).Delete(&model.EnvironmentAccess{}, "user_id = ?", user.ID.String())
```
mistahj67: Always wary of gorm's delete personally. It might be better to use an exec here with raw sql here

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -362,6 +362,41 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
...
+			// The migration sets the default for all_environments to true, which will enable all users to have access to all environments until ETAC is explicitly enabled
+			userTemplate.AllEnvironments = false
+
+			if createUserRequest.EnvironmentControlList != nil {
+				if roles.Has(model.Role{Name: auth.RoleAdministrator}) || roles.Has(model.Role{Name: auth.RolePowerUser}) {
+					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseETACInvalidRoles, request), response)
+					return
+				}
+
+				if len(createUserRequest.EnvironmentControlList.Environments) != 0 && createUserRequest.EnvironmentControlList.AllEnvironments {
```
mistahj67: I haven't tested locally so please disregard if this is handled but if I fire an empty list at this but have all_env set to false, does this put the user into a bugged state?

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -362,6 +362,41 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
...
+					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseETACBadRequest, request), response)
+					return
+				}
+
+				userTemplate.AllEnvironments = createUserRequest.EnvironmentControlList.AllEnvironments
+
+				environments := make([]model.EnvironmentAccess, 0)
+				for _, environment := range createUserRequest.EnvironmentControlList.Environments {
```
mistahj67: Can we please verify the environment strings prior to inserting these into the array to ensure data integrity?

cmd/api/src/model/access_control_list.go
```diff
@@ -0,0 +1,27 @@
...
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+package model
+
+// EnvironmentAccess defines the model for a row in the environment_access_control table
+type EnvironmentAccess struct {
+	UserID      string `json:"user_id"`
+	Environment string `json:"environment"`
+	BigSerial
```
mistahj67, line 22: Can you please verify that updates to these records also update the created_at / updated_at columns?

cmd/api/src/api/v2/auth/auth.go
mvlipka: They're mostly the same, but with minor differences between the two. I believe I have a way to abstract it though. I'll give that a shot

cmd/api/src/api/v2/auth/auth.go
mvlipka: Good observation! It's an intended state that was discussed with product
Users may have both an empty ETAC list and `all_environments` set to false, effectively denying them access to all environments

cmd/api/src/api/v2/auth/auth.go
mvlipka: This would be a good improvement, I'll look at what we can do here.
There is a phase 2 for ETAC which will re-visit some of the environments logic. Eventually, this `environment` column will be a foreign key to an entry in the domains table once the OpenGraph initiative is finished on partitioning our database by domains

cmd/api/src/model/access_control_list.go
mvlipka: `created_at` it filled in at time of creation, thank you GORM magic

`updated_at` is also filled in at the time of creation, but due to how we setup ETAC, a user's ETAC list is cleared when updating a user and the new ETAC list is applied to them, `updated_at` will reflect the `created_at` timestamp

mistahj67: Sounds good, thanks for verifying

cmd/api/src/api/v2/auth/auth.go
mistahj67: Fair enough, thanks for elucidating

cmd/api/src/api/v2/auth/etac.go
```diff
@@ -0,0 +1,94 @@
...
+			ad.Domain, azure.Tenant,
+		}, etacRequest.Environments...)
+		if err != nil {
+			return fmt.Errorf("error fetching environments: %w", err)
+		}
+
+		objectIDs, err := nodeSetToObjectIDSlice(nodes)
+		if err != nil {
+			return err
+		}
```
mistahj67: As this is going to be used solely to verify inclusion, I think either a map data structure would be more fitting here, thoughts?

cmd/api/src/api/v2/auth/etac.go
```diff
@@ -0,0 +1,94 @@
...
+
+		environments := make([]model.EnvironmentAccess, 0, len(etacRequest.Environments))
+		for _, environment := range etacRequest.Environments {
```
mistahj67: Alternative question to above, as the graph db hit above effectively will filter down the environments in the request to just valid environments, what if we looped that `nodes` response and filled the environments slice that way? Eliminates the map / contains check entirely

Can do a quick length check at the end if we still wanted to kick out an error if they supply a bad id OR can just ignore it completely?

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -362,6 +364,25 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
...
+
+			if createUserRequest.EnvironmentControlList != nil {
+				err := handleETACRequest(request.Context(), *createUserRequest.EnvironmentControlList, roles, &userTemplate, s.GraphQuery)
+				if err != nil {
+					api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+					return
+				}
+			}
+		}
+
```
bsheth711: (Optional) Could consider further moving all ETAC logic entirely to separate function--create and update both have the same checks

cmd/api/src/api/v2/auth/etac.go
mvlipka: I had considered doing the length check, but discussed in our team channel about the response here. It makes sense to me to return the environments that did not exist to the user as it gives immediate feedback to the user that their request had bad data

While this is not an issue in the UI, it could potentially be a sanity check for those using the API as it ensures the data they entered is correct

cmd/api/src/api/v2/auth/etac.go
mvlipka: That's fine
I don't think we have datasets that warrant performance gains that we get from using a map rather than a slice, changing this is quick

cmd/api/src/api/v2/auth/auth.go
mvlipka: Yeah, this is doable

Personally, I like keeping the database check in the respective UpdateUser and CreateUser methods as this allows the user to grok the intent of what's happening here at a glance, rather than having to navigate to the function and see that it is conditionally ran based on an external requirement

This way, we keep all the etac logic isolated to the new function, and the condition to run the function or not resides outside of it

cmd/api/src/api/v2/auth/etac.go
```diff
@@ -0,0 +1,90 @@
...
+		return errors.New(api.ErrorResponseETACBadRequest)
+	} else if etacRequest.AllEnvironments {
+		user.EnvironmentAccessControl = make([]model.EnvironmentAccess, 0)
+	} else {
+		if nodes, err := graphDB.FetchNodesByObjectIDsAndKinds(ctx, graph.Kinds{
+			ad.Domain, azure.Tenant,
+		}, etacRequest.Environments...); err != nil {
+			return fmt.Errorf("error fetching environments: %w", err)
+		} else {
+			if objectIDs, err := nodeSetToObjectIDSlice(nodes); err != nil {
```
mistahj67: nit on name now that it's a map
```suggestion
			if nodesByObjectId, err := nodeSetToObjectIDMap(nodes); err != nil {
```

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -362,6 +364,24 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
...
 		}
 
+		// eTAC
+		etacFeatureFlag, err := s.db.GetFlagByKey(request.Context(), appcfg.FeatureEnvironmentAccessControl)
+		if err != nil {
+			api.HandleDatabaseError(request, response, err)
+			return
+		} else if etacFeatureFlag.Enabled {
+			// Access to all environments will be denied by default
+			// The migration sets the default for all_environments to true, which will enable all users to have access to all environments until ETAC is explicitly enabled
```
mistahj67, line 376: I see this mentions a migration but is / was that in a different PR?

cmd/api/src/api/v2/auth/auth.go
```diff
@@ -362,6 +364,24 @@ func (s ManagementResource) CreateUser(response http.ResponseWriter, request *ht
 			}
 		}
 
+		// eTAC
+		etacFeatureFlag, err := s.db.GetFlagByKey(request.Context(), appcfg.FeatureEnvironmentAccessControl)
+		if err != nil {
```
mistahj67: In keeping with consistency for variable scoping, I think we'll want to rely on the if block here as well please
```suggestion
		if etacFeatureFlag, err := s.db.GetFlagByKey(request.Context(), appcfg.FeatureEnvironmentAccessControl); err != nil {
```

packages/go/openapi/src/schemas/api.requests.user.update.yaml
```diff
@@ -40,3 +40,22 @@ properties:
     description: ID of the SSO provider for this user
   is_disabled:
     type: boolean
+  environment_access_control:
+    type: object
+    additionalProperties: false
+    description: Sets an environment access control list to the user. When all_environments is set to true, a user's existing environment list is cleared and they are granted access to all environments. Administrators and Power Users may not have an ETAC list applied to them.
```
mistahj67:
```suggestion
    description: Sets an environment access control to the user. When all_environments is set to true, a user's existing environment list is cleared and they are granted access to all environments. Administrators and Power Users may not have an ETAC list applied to them.

```

cmd/api/src/api/v2/auth/auth.go
mvlipka: BED-6211 took care of that:
https://github.com/SpecterOps/BloodHound/pull/1746

mistahj67: Nvm, was added in v8.1.0, resolving!

PR #1816, opened 2025-08-22
feat(api): BED-5279 Certify/Revoke Certification for Objects 

sirisjo:
## Description

This PR adds an API endpoint to certify or decertify objects by member ID/object ID. A few notes on functionality: 

- The operation is done in a transaction: if any of the database requests fail, the entire request will fail and nothing will be updated. 
- if there are duplicate entries for a given object ID, only entries associated with the highest priority tag are updated. All other entries are set to pending certification
- if the object is already set to manually certified, it will not be updated again 

## Motivation and Context

Resolves BED-5279

This change will enable manual certification or decertification of objects based on object ID as belonging to the zone they are aligned to. 

## How Has This Been Tested?

Unit tests added, integration tests added, and also tested the following cases manually in the test database: 
- tested happy path (certify/decertify) for a single object ID 
- tested happy path (certify/decertify) for multiple object IDs 
- tested the transaction functionality by manually throwing an error after 2 object IDs are processed -- nothing is committed to the DB if the transaction fails
- confirmed that when there are duplicate object IDs, only those associated with the highest priority tag are updated, and the other entries are set to pending
- confirmed nothing is updated if the object IDs in the request already match the action in the request 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - POST endpoint to batch certify or revoke asset-group members (member_ids, action, optional note).
  - Priority-aware per-node certification so the highest-priority selector controls an object’s certification.
- **Behavioral Changes**
  - New "Pending" certification state used for newly inserted selector nodes and update flows.
  - New user-facing error messages for invalid action, missing member IDs, and unknown user.
- **Documentation**
  - OpenAPI updated with the new certifications endpoint and schema.
- **Tests**
  - Expanded endpoint and database tests covering errors, priority handling, history, and success flows.


cmd/api/src/api/registration/v2.go
```diff
@@ -179,6 +179,7 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupMembersByTag).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
 		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/counts", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupTagMemberCountsByKind).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
 		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/{%s}", api.URIPathVariableAssetGroupTagID, api.URIPathVariableAssetGroupTagMemberID), resources.GetAssetGroupTagMemberInfo).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
+		routerInst.POST("/api/v2/asset-group-tags/members/certification", resources.CertifyMembers).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
```
mistahj67:
```suggestion
		routerInst.POST("/api/v2/asset-group-tags/certifications", resources.CertifyMembers).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBWrite),
```

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +967,83 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+
+// CreateInputsForUpdateCertificationBySelectorNode Visible for Testing
+func CreateInputsForUpdateCertificationBySelectorNode(nodes []model.AssetGroupSelectorNodeExpanded, requestAction model.AssetGroupCertification, userEmail null.String, note null.String) []database.UpdateCertificationBySelectorNodeInput {
+	var (
+		certificationStatus model.AssetGroupCertification
+		certifiedBy         null.String
+	)
+	inputs := []database.UpdateCertificationBySelectorNodeInput{}
+	// Only update the nodes with the highest priority for a given nodeID
+	lastProcessedNodeId, highestPriorityForGivenNode := graph.ID(0), -1
```
mistahj67: nit: can these initializations be moved into the above var block and also separate out these 2 into individual lines

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +967,83 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+	if node.NodeId != lastProcessedNodeId {
+		return true
+	} else if node.Position == highestPriority {
+		return true
+	}
+	return false
+}
+
+// CreateInputsForUpdateCertificationBySelectorNode Visible for Testing
+func CreateInputsForUpdateCertificationBySelectorNode(nodes []model.AssetGroupSelectorNodeExpanded, requestAction model.AssetGroupCertification, userEmail null.String, note null.String) []database.UpdateCertificationBySelectorNodeInput {
```
mistahj67: nit: let's brainstorm possible other names perhaps? this took me a second to parse what it is

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +967,83 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+	switch certType {
+	case model.AssetGroupCertificationManual, model.AssetGroupCertificationRevoked:
+		return true
+	default:
+		return false
+	}
+}
+
+func nodeShouldBeUpdated(node model.AssetGroupSelectorNodeExpanded, lastProcessedNodeId graph.ID, highestPriority int) bool {
+	if node.NodeId != lastProcessedNodeId {
```
mistahj67: I think it's worthwhile tossing a auto certified check and kickout here if we have access to this info easily enough, just to be safe

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +967,83 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+		api.WriteErrorResponse(requestContext, api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseAssetGroupCertTypeInvalid, request), response)
+	} else if memberIds := reqBody.MemberIDs; len(memberIds) == 0 {
+		api.WriteErrorResponse(requestContext, api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseAssetGroupMemberIDsRequired, request), response)
+	} else if nodes, err := s.DB.GetAssetGroupSelectorNodeExpandedIgnoreAutoCertify(requestContext, reqBody.MemberIDs...); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else {
+		dbInputs := CreateInputsForUpdateCertificationBySelectorNode(nodes, reqBody.Action, user.EmailAddress, null.StringFrom(reqBody.Note))
+		if err := s.DB.UpdateCertificationBySelectorNodeTransaction(requestContext, dbInputs); err != nil {
+			api.HandleDatabaseError(request, response, err)
+			return
```
mistahj67:
```suggestion
```

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -3157,3 +3157,496 @@ func TestResources_SearchAssetGroupTagHistory(t *testing.T) {
...
+			NodeId:            nodeId1,
+			Certified:         model.AssetGroupCertificationPending,
+			CertifiedBy:       null.String{},
+			Source:            model.AssetGroupSelectorNodeSourceSeed,
+			CreatedAt:         time.Now(),
+			UpdatedAt:         time.Now(),
+			NodePrimaryKind:   "",
+			NodeEnvironmentId: "",
+			NodeObjectId:      "",
+			NodeName:          "",
```
mistahj67: go tip: the default for strings is "" so initialization of struct fields don't need to be explicitly defined. this can help cut down on some of the boilerplate

cmd/api/src/database/assetgrouptags.go
```diff
@@ -716,6 +720,44 @@ func (s *BloodhoundDB) UpdateSelectorNodesByNodeId(ctx context.Context, assetGro
...
 }
 
+func (s *BloodhoundDB) UpdateCertificationBySelectorNode(ctx context.Context, selectorId int, certified model.AssetGroupCertification, certifiedBy null.String, nodeId graph.ID) error {
+	return s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		if result := tx.WithContext(ctx).Exec(fmt.Sprintf("UPDATE %s SET certified = ?, certified_by = ?, updated_at = current_timestamp WHERE selector_id = ? AND node_id = ?", model.AssetGroupSelectorNode{}.TableName()), certified, certifiedBy, selectorId, nodeId); result.Error != nil {
+			return CheckError(result)
+		}
+		return nil
+	})
+}
```
mistahj67: I haven't checked directly, but is it possible to move this into the below function instead to keep this from requiring 2 db methods?

cmd/api/src/database/db.go
```diff
@@ -263,3 +265,7 @@ func (s *BloodhoundDB) Migrate(ctx context.Context) error {
 
 	return nil
 }
+
+func (s *BloodhoundDB) BeginTransaction(ctx context.Context, fn func(tx *gorm.DB) error) error {
+	return s.db.WithContext(ctx).Transaction(fn)
+}
```
mistahj67: Is this required or can the inside call be used in the above db method instead of adding to this interface 🤔

cmd/api/src/queries/graph.go
```diff
@@ -683,6 +684,16 @@ func (s *GraphQuery) FetchNodeByGraphId(ctx context.Context, id graph.ID) (*grap
 	}
 }
 
+func (s *GraphQuery) FetchNodesByGraphIds(ctx context.Context, ids []graph.ID) ([]*graph.Node, error) {
```
mistahj67: I think this might have been from a previous iteration and is no longer needed? I don't see it used 🤔

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.certify.yaml
```diff
@@ -0,0 +1,61 @@
...
+        schema:
+          type: object
+          properties:
+            member_ids:
+              type: array 
+              items: integer
+            action:
+              type: integer
+              enum:
+                - 1 # decertify
```
mistahj67:
```suggestion
                - 1 # revoke
```

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.certify.yaml
```diff
@@ -0,0 +1,61 @@
...
+post:
+  operationId: CertifyOrDecertifyObjects
+  summary: Certify or Decertify Objects
+  description: Manually certify/decertify objects as belonging to the Zone they're aligned to
+  tags:
+    - Asset Isolation
+    - Community
+    - Enterprise
+  requestBody:
+    description: The request body for certifying or decertifying objects.
```
mistahj67:
```suggestion
    description: The request body for certifying or revoking certification of objects. 
```

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.certify.yaml
```diff
@@ -0,0 +1,61 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+post:
+  operationId: CertifyOrDecertifyObjects
+  summary: Certify or Decertify Objects
+  description: Manually certify/decertify objects as belonging to the Zone they're aligned to
```
mistahj67:
```suggestion
  description: Manually certify/revoke certification of objects as belonging to the Zone they're selected by
```

cmd/api/src/database/assetgrouptags.go
sirisjo: It is possible; however, I think it is cleaner to have it separated out, with one method to manage the logic/transaction and the other method to actually call the DB. This is also more extensible, in case there is a use case in the future where a single node needs to be updated outside of the transaction for some reason. Open to discussion though!

cmd/api/src/queries/graph.go
sirisjo: Oops! Good catch 😅

cmd/api/src/database/assetgrouptags.go
sirisjo: discussed offline, will be refactoring this 👍

cmd/api/src/api/registration/v2.go
```diff
@@ -179,6 +179,7 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupMembersByTag).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
 		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/counts", api.URIPathVariableAssetGroupTagID), resources.GetAssetGroupTagMemberCountsByKind).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
 		routerInst.GET(fmt.Sprintf("/api/v2/asset-group-tags/{%s}/members/{%s}", api.URIPathVariableAssetGroupTagID, api.URIPathVariableAssetGroupTagMemberID), resources.GetAssetGroupTagMemberInfo).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBRead),
+		routerInst.POST("/api/v2/asset-group-tags/certification", resources.CertifyMembers).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBWrite),
```
mistahj67: typically we want to pluralize resources in routes

```suggestion
		routerInst.POST("/api/v2/asset-group-tags/certifications", resources.CertifyMembers).CheckFeatureFlag(resources.DB, appcfg.FeatureTierManagement).RequirePermissions(permissions.GraphDBWrite),
```

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +968,92 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+		certifiedBy         null.String
+
+		lastProcessedNodeId         graph.ID = 0
+		highestPriorityForGivenNode          = -1
+
+		inputs = []database.UpdateCertificationBySelectorNodeInput{}
+	)
+
+	// Only update the nodes with the highest priority for a given nodeID
+	for _, node := range nodes {
```
mistahj67: nit: just realized these are selectorNodeRecords and not nodes themselves, I wonder if changing the name might help readability

packages/go/openapi/src/openapi.yaml
```diff
@@ -356,6 +356,8 @@ paths:
     $ref: './paths/asset-isolation.asset-group-tags.search.yaml'
   /api/v2/asset-group-tags-history:
     $ref: './paths/asset-isolation.asset-group-tags-history.yaml'
+  /api/v2/asset-group-tags/members/certification:
+    $ref: './paths/asset-isolation.asset-group-tags.certify.yaml'
```
mistahj67:
```suggestion
  /api/v2/asset-group-tags/certifications:
    $ref: './paths/asset-isolation.asset-group-tags.certify.yaml'
```

cmd/api/src/database/assetgrouptags.go
```diff
@@ -716,6 +718,35 @@ func (s *BloodhoundDB) UpdateSelectorNodesByNodeId(ctx context.Context, assetGro
...
+func (s *BloodhoundDB) UpdateCertificationBySelectorNode(ctx context.Context, inputs []UpdateCertificationBySelectorNodeInput) error {
+	historyRecordForNode := make(map[graph.ID]bool)
+	return s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		transaction := NewBloodhoundDB(tx, s.idResolver)
+		var err error
+		for _, input := range inputs {
+			if result := tx.WithContext(ctx).Exec(fmt.Sprintf("UPDATE %s SET certified = ?, certified_by = ?, updated_at = current_timestamp WHERE selector_id = ? AND node_id = ?", model.AssetGroupSelectorNode{}.TableName()), input.CertificationStatus, input.CertifiedBy, input.SelectorId, input.NodeId); result.Error != nil {
+				return CheckError(result)
+			} else if !historyRecordForNode[input.NodeId] && input.CertificationStatus != model.AssetGroupCertificationPending {
+				if err = transaction.CreateAssetGroupHistoryRecord(ctx, model.AssetGroupActorSystem, input.CertifiedBy.String, input.NodeName, model.ToAssetGroupHistoryActionFromAssetGroupCertification(input.CertificationStatus), input.AssetGroupTagId, null.String{}, input.Note); err != nil {
```
mistahj67: issue (blocking): This hardcodes the actor to `SYSTEM`, it should be user id

If possible, can the below db tests test this as well

cmd/api/src/api/registration/v2.go
AD7ZJ: I had to put the GET version of this endpoint up above `/api/v2/asset-group-tags/{%s}` to get it to work, might be good to put this there too? Just a thought :)

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -3157,3 +3157,456 @@ func TestResources_SearchAssetGroupTagHistory(t *testing.T) {
...
+		assetGroupTag3                      = 3
+		highestPosition                     = 1
+		mediumPosition                      = 2
+		lowestPosition                      = 3
+		userEmailAddress                    = null.StringFrom("test@testy.com")
+		stringNote                          = "test"
+		note                                = null.StringFrom(stringNote)
+		assetGroupSelectorNode1HighPriority = model.AssetGroupSelectorNodeExpanded{AssetGroupSelectorNode: model.AssetGroupSelectorNode{
+			SelectorId: selectorId1,
+			NodeId:     nodeId1,
```
AD7ZJ, line 3183: I'm not sure this will affect the outcome of your tests, but there can't be multiple nodeIDs with the same selector ID in the table this function pulls from. The nodeID - selectorID combo should be unique. It might be worth re-working this a little so the tests match reality and also to avoid confusion :) I would also add a test in here for a node with `Certified` set to auto to make sure that case is handled correctly.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +968,92 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+			certificationStatus = requestAction
+			highestPriorityForGivenNode = node.Position
+			certifiedBy = userEmail
+			lastProcessedNodeId = node.NodeId
+			// don't actually update nodes that already match the request action
+			if node.Certified == requestAction {
+				continue
+			}
+		}
+		inputs = append(inputs, database.UpdateCertificationBySelectorNodeInput{AssetGroupTagId: node.AssetGroupTagId, SelectorId: node.SelectorId, CertifiedBy: certifiedBy, CertificationStatus: certificationStatus, NodeId: node.NodeId, NodeName: node.NodeName, Note: note})
```
AD7ZJ:
```suggestion
			inputs = append(inputs, database.UpdateCertificationBySelectorNodeInput{AssetGroupTagId: node.AssetGroupTagId, SelectorId: node.SelectorId, CertifiedBy: certifiedBy, CertificationStatus: certificationStatus, NodeId: node.NodeId, NodeName: node.NodeName, Note: note})
		}
```
Should the append be inside the if, so no update happens if the node doesn't need updated? Unless it's intentional to hard-code those dupes to `AssetGroupCertificationPending`

sirisjo: Good question! It is intended to set the lower-priority dupes to pending, only the highest-priority selector nodes for a given node ID should be updated, and the other nodes must be set to pending (just in case they were manually certified at a different point in time).

AD7ZJ: Sounds good, might not be a bad idea to throw that in a comment to aid the next poor soul that comes along :D nice work!

sirisjo: Good idea, I updated the godoc to better explain how the dupes are handled 👍

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +968,93 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+func isValidCertType(certType model.AssetGroupCertification) bool {
+	switch certType {
+	case model.AssetGroupCertificationManual, model.AssetGroupCertificationRevoked:
+		return true
+	default:
+		return false
+	}
+}
+
+func recordShouldBeUpdated(record model.AssetGroupSelectorNodeExpanded, lastProcessedNodeId graph.ID, highestPriority int) bool {
```
ALCooper12: Ignore if you'd rather keep! But maybe change the function name to: `isRecordUpdated` instead?

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.certify.yaml
```diff
@@ -0,0 +1,65 @@
...
+          additionalProperties: false
+          required:
+            - member_ids
+            - action
+
+
+  responses:
+    200:
+      description: OK
+
```
ALCooper12: You forgot to delete this extra line above!

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -967,3 +968,93 @@ func (s *Resources) GetAssetGroupTagHistory(response http.ResponseWriter, reques
...
+		return true
+	} else if record.Position == highestPriority {
+		return true
+	}
+	return false
+}
+
+type setSelectorNodeCertifier interface {
+	UpdateCertificationBySelectorNode(ctx context.Context, input []database.UpdateCertificationBySelectorNodeInput) error
+}
```
ALCooper12, line 998: I might be wrong, but we usually implement interfaces outside of this file. @mistahj67 is this true? Or maybe it's fine here?

packages/go/openapi/src/paths/asset-isolation.asset-group-tags.certify.yaml
```diff
@@ -0,0 +1,65 @@
...
+              minItems: 1 
+              uniqueItems: true
+            action:
+              type: integer
+              enum:
+                - 1 # revoke
+                - 2 # certify 
+            note: 
+              type: string
+          additionalProperties: false
```
ALCooper12: Nice addition here!!

cmd/api/src/api/v2/assetgrouptags.go
sirisjo: To me, `isRecordUpdated` conveys that we are checking if a record has been updated already. This function checks if a record _should be_ updated -- open to naming it something else if the current name is not descriptive enough!

sirisjo: Perhaps something like `recordEligibleForUpdate` ?

cmd/api/src/api/v2/assetgrouptags.go
mistahj67: it's fine here, it's specific to this function so leaving it closer to the consumer I think is preferred.

ALCooper12: Cool cool! Thank you!

cmd/api/src/api/v2/assetgrouptags.go
ALCooper12: Yesss I like that name better! I was a bit confused on the other one lolz

PR #1839, opened 2025-08-29
feat (api): Introduce EnableEmbeddedCollectorsDownload to Config BED-6074

LawsonWillard:
## Description

Adds new config option to enable BHE containers to only download collectors that are bundled with the container. 

## Motivation and Context
Resolves: BED-6074

As a user, I want the option to disable the ability to download collectors which are not bundled inside my provided container image, so that I can prevent the API service from making requests to external services which I do not own. 

## How Has This Been Tested?

- No Op in BHCE

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


elikmiller (approved):

LawsonWillard:
> I previously approved this but then I had a question.
> 
> I think you've explained this to me - but are any changes required here in BHCE to support this flag? I think BHCE always serves pre-bundled collectors. Based on the naming of the config flag I think it could be implied that setting it to `false` (which it is by default) would disable the ability to download collectors from BHCE.
> 
> Should we rename the config? Or does it make sense to plumb through the ability to disable embedded collector downloads in BHCE too?

Yeah I think renaming it would be best, its a no op in BHCE so nothing changes and having a more descriptive flag name would better help users understand this. Hows changing enable_embedded_collectors_download to force_download_embedded_collectors?

elikmiller:
> > I previously approved this but then I had a question.
> > I think you've explained this to me - but are any changes required here in BHCE to support this flag? I think BHCE always serves pre-bundled collectors. Based on the naming of the config flag I think it could be implied that setting it to `false` (which it is by default) would disable the ability to download collectors from BHCE.
> > Should we rename the config? Or does it make sense to plumb through the ability to disable embedded collector downloads in BHCE too?
> 
> Yeah I think renaming it would be best, its a no op in BHCE so nothing changes and having a more descriptive flag name would better help users understand this. Hows changing enable_embedded_collectors_download to force_download_embedded_collectors?

Yeah that sounds better to me!

PR #1858, opened 2025-09-03
BED-6024-two-cards-dialog-env-selector

catsiller:
## Description

1. replaced all MUI components in CreateUserForm with DoodleUI components
2. replaced all MUI components in UpdateUserForm with DoodleUI components
3. replaced all MUI components in CreateUserDialog with DoodleUI components
4. replaced all MUI components in UpdateUserDialog with DoodleUI components
5. existing code was using camel case variables in the API client - replaced with snake case to match database values (updated in the API client as well as the forms/dialogs/types where needed)
6. added new feature in CreateUserForm and UpdateUserForm (BHE only) to include an environment selector panel in create/update user dialogs for read-only and user roles when admin user is created or updating a user
7. added prop to show environment panel mentioned in number 6 ^ to display only in BHE 

## Motivation and Context

Resolves [BED-6024](https://specterops.atlassian.net/browse/BED-6024)

BHE Administrators can create/update users and grant environment access through a panel within the create/update dialogs, admins can grant access to all, some, or no environments for read-only and user roles

## How Has This Been Tested?

## Types of changes

(Breaking change for form updates from MUI components to DoodleUI components)

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Environment Targeted Access Control panel, EnvironmentSelectPanel, role utilities, SSO and single-user data hooks, and users view variant that enables environment-access controls.

* **UI Changes**
  * Redesigned Create/Edit user dialogs and forms with internal dialog handling, permission-aware triggers, auth-method branching, environment access toggle, safer combobox rendering, and updated user actions menu items/test IDs.

* **API Updates**
  * Public request/response shapes moved to snake_case, added environment fields and secret; client now sends user objects directly.

* **Tests**
  * Updated tests, mocks, and helpers to cover ETAC flows and new dialog/form behavior.

* **Chores**
  * Removed legacy authentication type exports.

[BED-6024]: https://specterops.atlassian.net/browse/BED-6024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/justfile
+++ b/justfile
@@ -46,10 +46,7 @@ analyze *FLAGS:
 # Run tests
 test *FLAGS:
   @just stbernard test {{FLAGS}}
-
-# generate Go Coverage
-cover *FLAGS:
-  @just stbernard cover {{FLAGS}}
+  @just stbernard cover
 
 # Build application
 build *FLAGS:
--- a/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/CypherSearch.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/CypherSearch.tsx
@@ -18,7 +18,7 @@ import { useTheme } from '@mui/material';
 import '@neo4j-cypher/codemirror/css/cypher-codemirror.css';
 import { CypherEditor } from '@neo4j-cypher/react-codemirror';
 import { UpdateUserQueryRequest } from 'js-client-library';
-import { useEffect, useRef, useState } from 'react';
+import { useRef, useState } from 'react';
 import { useQuery } from 'react-query';
 import { AppIcon } from '../../../components';
 import { graphSchema } from '../../../constants';
@@ -79,18 +79,7 @@ const CypherSearchInner = ({
     const { checkPermission } = usePermissions();
 
     const cypherEditorRef = useRef<CypherEditor | null>(null);
-    const getCypherValueOnLoadRef = useRef(false);
     const { data: permissions } = useQueryPermissions(selectedQuery?.id);
-
-    useEffect(() => {
-        //Setting the selected query once on load
-        //The cypherQuery dependency is required
-        //check for flag
-        if (!getCypherValueOnLoadRef.current && cypherQuery) {
-            getCypherValueOnLoadRef.current = true;
-            setSelected({ query: cypherQuery, id: undefined });
-        }
-    }, [cypherQuery]);
 
     const handleCypherSearch = () => {
         if (cypherQuery) {
--- a/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/CommonSearches.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/CommonSearches.tsx
@@ -18,9 +18,9 @@ import { Button } from '@bloodhoundenterprise/doodleui';
 import { faChevronDown, faChevronUp } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import fileDownload from 'js-file-download';
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import PrebuiltSearchList from '../../../../components/PrebuiltSearchList';
-import { getExportQuery, useDeleteSavedQuery, usePrebuiltQueries, useSavedQueries } from '../../../../hooks';
+import { getExportQuery, useDeleteSavedQuery, usePrebuiltQueries } from '../../../../hooks';
 import { useSelf } from '../../../../hooks/useSelf';
 import { useNotifications } from '../../../../providers';
 import { QueryLineItem, QueryListSection } from '../../../../types';
@@ -44,7 +44,6 @@ const CommonSearches = ({
 }: CommonSearchesProps) => {
     const { selected, selectedQuery, setSelected } = useSavedQueriesContext();
 
-    const userQueries = useSavedQueries();
     const deleteQueryMutation = useDeleteSavedQuery();
     const { addNotification } = useNotifications();
 
@@ -55,6 +54,7 @@ const CommonSearches = ({
     const [queryId, setQueryId] = useState<number>();
 
     const [categoryFilter, setCategoryFilter] = useState<string[]>([]);
+    const [isFiltered, setIsFiltered] = useState(false);
 
     //master list of pre-made queries
     const queryList: QueryListSection[] = usePrebuiltQueries();
@@ -66,11 +66,6 @@ const CommonSearches = ({
 
     const { getSelfId } = useSelf();
     const { data: selfId } = getSelfId;
-
-    useEffect(() => {
-        setFilteredList(queryList);
-        handleFilter(searchTerm, platform, categoryFilter, source);
-    }, [userQueries.data]);
 
     const handleClick = (query: string, id: number | undefined) => {
         if (selected.query === query && selected.id === id) {
@@ -106,6 +101,7 @@ const CommonSearches = ({
     };
 
     const handleFilter = (searchTerm: string, platform: string, categories: string[], source: string) => {
+        setIsFiltered(true);
         setSearchTerm(searchTerm);
         setPlatform(platform);
         setCategoryFilter(categories);
@@ -205,7 +201,7 @@ const CommonSearches = ({
 
             <div className={cn('grow-1 min-h-0 overflow-auto', { hidden: !showCommonQueries })}>
                 <PrebuiltSearchList
-                    listSections={filteredList}
+                    listSections={isFiltered ? filteredList : queryList}
                     clickHandler={handleClick}
                     deleteHandler={handleDeleteQuery}
                     clearFiltersHandler={handleClearFilters}
--- a/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/SaveQueryDialog.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/SaveQueryDialog.tsx
@@ -88,15 +88,9 @@ const SaveQueryDialog: React.FC<{
             setId(undefined);
             setIsNew(true);
         }
-    }, [selectedQuery]);
-
-    useEffect(() => {
         setName(selectedQuery && selectedQuery.name ? selectedQuery.name : '');
-    }, [selectedQuery, selectedQuery?.name]);
-
-    useEffect(() => {
         setDescription(selectedQuery && selectedQuery.description ? selectedQuery.description : '');
-    }, [selectedQuery?.description]);
+    }, [selectedQuery]);
 
     useEffect(() => {
         setLocalCypherQuery(cypherQuery);
--- a/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/SavedQueryPermissions.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/SavedQueryPermissions.tsx
@@ -16,7 +16,7 @@
 import { Button, Checkbox, ColumnDef, DataTable, Input } from '@bloodhoundenterprise/doodleui';
 import { CheckedState } from '@radix-ui/react-checkbox';
 import { UserMinimal } from 'js-client-library';
-import { useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useQuery } from 'react-query';
 import { AppIcon } from '../../../../components';
 import { useQueryPermissions } from '../../../../hooks';
@@ -54,7 +54,7 @@ const SavedQueryPermissions: React.FC<SavedQueryPermissionsProps> = (props: Save
 
     const { data, isLoading } = useQueryPermissions(queryId as number);
 
-    function idMap() {
+    const idMap = useCallback(() => {
         return listUsersQuery.data
             ?.filter((user: UserMinimal) => user.id !== selfId)
             .map((user: UserMinimal) => {
@@ -64,9 +64,9 @@ const SavedQueryPermissions: React.FC<SavedQueryPermissionsProps> = (props: Save
                     email: user.email_address,
                 };
             });
-    }
+    }, [listUsersQuery.data, selfId]);
 
-    const usersList = useMemo(() => idMap(), [listUsersQuery.data, selfId]);
+    const usersList = useMemo(() => idMap(), [idMap]);
     const allUserIds = useMemo(() => usersList?.map((x) => x.id) ?? [], [usersList]);
 
     useEffect(() => {
@@ -74,7 +74,7 @@ const SavedQueryPermissions: React.FC<SavedQueryPermissionsProps> = (props: Save
         const initialShared = data.public ? allUserIds : data.shared_to_user_ids ?? [];
         setSharedIds(initialShared);
         setIsPublic(Boolean(data.public));
-    }, [data, allUserIds]);
+    }, [data, allUserIds, setSharedIds, setIsPublic]);
 
     const handleCheckAllChange = (checkedState: CheckedState) => {
         const isTrue = checkedState === true;
--- a/packages/javascript/bh-shared-ui/src/views/Explore/providers/SavedQueriesProvider/SavedQueriesProvider.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/providers/SavedQueriesProvider/SavedQueriesProvider.tsx
@@ -13,19 +13,28 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useCypherSearch, useGetSelectedQuery } from '../../../../hooks';
 import { QueryLineItem, SaveQueryAction, SelectedQuery } from '../../../../types';
 import { SavedQueriesContext } from './SavedQueriesContext';
 
 export function SavedQueriesProvider({ children }: { children: any }) {
-    const [selected, setSelected] = useState<SelectedQuery>({ query: '', id: undefined });
+    const { setCypherQuery, performSearch, cypherQuery } = useCypherSearch();
+    const [selected, setSelected] = useState<SelectedQuery>({ query: cypherQuery, id: undefined });
+
     const [showSaveQueryDialog, setShowSaveQueryDialog] = useState(false);
     const [saveAction, setSaveAction] = useState<SaveQueryAction | undefined>(undefined);
 
     const selectedQuery: QueryLineItem | undefined = useGetSelectedQuery(selected.query, selected.id);
 
-    const { setCypherQuery, performSearch } = useCypherSearch();
+    const getCypherValueOnLoadRef = useRef(false);
+
+    useEffect(() => {
+        if (!getCypherValueOnLoadRef.current && cypherQuery) {
+            setSelected({ query: cypherQuery, id: undefined });
+            getCypherValueOnLoadRef.current = true;
+        }
+    }, [cypherQuery]);
 
     const runQuery = (query: string, id?: number) => {
         setSelected({ query, id });
--- a/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
@@ -244,12 +244,15 @@ export const TagForm: FC = () => {
     const handleDeleteCancel = useCallback(() => setDeleteDialogOpen(false), []);
 
     const handleGlyphCancel = useCallback(() => setGlyphDialogOpen(false), []);
-    const handleGlyphSelect = useCallback((iconName?: IconName) => {
-        if (iconName) setValue('glyph', iconName, { shouldDirty: true });
-        else setValue('glyph', '', { shouldDirty: true });
+    const handleGlyphSelect = useCallback(
+        (iconName?: IconName) => {
+            if (iconName) setValue('glyph', iconName, { shouldDirty: true });
+            else setValue('glyph', '', { shouldDirty: true });
 
-        setGlyphDialogOpen(false);
-    }, []);
+            setGlyphDialogOpen(false);
+        },
+        [setValue]
+    );
 
     useEffect(() => {
         if (tagQuery.data) {


packages/javascript/bh-shared-ui/src/views/Users/Users.tsx
```diff
@@ -182,49 +188,76 @@ const Users: FC = () => {
...
+                    {/*
+                    <FeatureFlag
+                    flagKey='PUT_ETAC_FEATURE_FLAG_HERE'
+                        enabled={
+                        }
+                        disabled={
+                        }
+                    />
+                    */}
+                    <Dialog
```
maffkipp: I think it might make the most sense to move all of the doodleui `Dialog` subcomponents (`Dialog`, `DialogPortal`, etc.) down into the `CreateUserDialog` component instead of assembling them in the top-level `Users` component, to prevent future users of this component from needing to set up all the same logic. This would also simplify our test setup in `CreateUserDialog.test.tsx`

packages/javascript/bh-shared-ui/src/views/Users/Users.tsx
```diff
@@ -182,49 +188,76 @@ const Users: FC = () => {
...
+                        open={createUserDialogOpen}
+                        onOpenChange={toggleCreateUserDialog}
+                        data-testid='create-user-dialog'>
+                        <DialogTrigger asChild>
```
maffkipp: I _think_ that when using the `DialogTrigger` component, the open/closed state gets managed internally and you should not need to track that state with `createUserDialogOpen`. For examples see: https://ui.shadcn.com/docs/components/dialog

packages/javascript/bh-shared-ui/src/views/Users/Users.tsx
```diff
@@ -182,49 +188,76 @@ const Users: FC = () => {
...
-                userId={selectedUserId!}
-                hasSelectedSelf={hasSelectedSelf}
-                onSave={updateUserMutation.mutateAsync}
-                isLoading={updateUserMutation.isLoading}
-                error={updateUserMutation.error}
-            />
+            <Dialog open={updateUserDialogOpen} onOpenChange={toggleUpdateUserDialog}>
+                <DialogPortal>
+                    <UpdateUserDialog
+                        error={updateUserMutation.error}
```
maffkipp: It looks like we are pulling a lot of values out of `updateUserMutation` and passing them as individual props, and that mutation is not used anywhere else in the `Users` component. It might be better to pull `updateUserMutation` and `createUserMutation` down into their respective Dialog components to simplify these component interfaces

cmd/api/src/api/v2/version.go
```diff
@@ -42,6 +43,7 @@ func GetVersion(response http.ResponseWriter, request *http.Request) {
 			CurrentVersion:    "v2",
 			DeprecatedVersion: "none",
 		},
-		Server: version.GetVersion().String(),
+		Server:         version.GetVersion().String(),
+		ProductEdition: "community",
```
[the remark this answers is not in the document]
catsiller: *should not be change in pr*

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+        password: '',
+        needs_password_reset: false,
+        roles: [3],
+        sso_provider_id: '',
+        all_environments: false,
+        environment_targeted_access_control: {
+            environments: null,
         },
-    });
+    };
```
benwaples: Looks like this object needs to match the `CreateUserRequestForm` type, mind adding a type constraint to the object so that if the `CreateUserRequestForm` changes, or more likely the underlying `CreateUserRequest` changes, we are forced to account for it here? Something like thus:
```ts
const defaultValues = {
  keys: values
} satisfies CreateUserRequestForm;
```

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+    const handleCheckedChange = (checked: boolean | 'indeterminate') => {
+        setNeedsPasswordReset(checked === true);
+        if (checked === true) {
+            form.setValue('needs_password_reset', true);
+        } else {
+            form.setValue('needs_password_reset', false);
+        }
+    };
+
+    const getRolesQuery = useQuery(['getRoles'], ({ signal }) =>
```
benwaples: It appears we already have a reuseable hook for this: `useListDisplayRoles`, can we use that hook in place of this one?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+    );
+
+    const roleInputValue = form.watch('roles');
+    const selectedETACEnabledRole = roleInputValue.toString() === '2' || roleInputValue.toString() === '3';
+
+    const { data: availableEnvironments } = useAvailableEnvironments();
+
+    const filteredEnvironments = availableEnvironments?.filter((environment: Environment) =>
+        environment.name.toLowerCase().includes(searchInput.toLowerCase())
+    );
```
benwaples: I could see this potentially causing some slow down in a tenant with lots of environments because its a nested loop that runs on every render, regardless if the `searchInput` or `availableEnvironments` are changing. 

But this may not be a problem at all tho, and potentially a premature optimization, but we could mock out the `availableEnvironments` response with 150 envs (which is an actual case right now) and see if there are any slowdowns while using the form. If there is, something like this could be useful:

```ts
const filteredEnvironments = useMemo(() => {
  const searchInputLowered = searchInput.toLowerCase()
  return availableEnvironments?.filter((environment: Environment) =>
        environment.name.toLowerCase().includes(searchInputLowered)
}, [availableEnvironments, searchInput])
```

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+
+    const allEnvironmentsSelected: any =
```
benwaples: could this be typed as a boolean or is there a need for the `: any`? Also this is nit picky but I do love it when boolean names start with is/are (anything that would read like the variable is storing a boolean) like `areAllEnvironmentsSelected`

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+    const getRolesQuery = useQuery(['getRoles'], ({ signal }) =>
+        apiClient.getRoles({ signal }).then((res) => res.data?.data?.roles)
+    );
+
+    const listSSOProvidersQuery = useQuery(['listSSOProviders'], ({ signal }) =>
+        apiClient.listSSOProviders({ signal }).then((res) => res.data?.data)
+    );
+
+    const roleInputValue = form.watch('roles');
+    const selectedETACEnabledRole = roleInputValue.toString() === '2' || roleInputValue.toString() === '3';
```
benwaples: Is this the only place we need to know if a user role is an ETAC enabled role? If not, id suggest we pull out this into a function like `isUserRoleETACEnabled(roleInputValue.toString())` so we can reuse this logic in the future and if a user role becomes ETAC enabled then we can easily add it to the check.

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+                                                                    <FormItem className='flex flex-row items-center'>
+                                                                        <Checkbox
+                                                                            {...field}
+                                                                            checked={selectedEnvironments.includes(
+                                                                                item.id
+                                                                            )}
+                                                                            className='m-3 data-[state=checked]:bg-primary data-[state=checked]:border-[#2C2677]'
+                                                                            id='environments'
+                                                                            onCheckedChange={(checked) =>
+                                                                                handleEnvironmentSelectChange(
```
benwaples: This handler also fires when clicking the label, but it will only select the first item in the list. Perhaps the handler needs to be moved to the FormItem or FormField?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
+                                                                control={form.control}
+                                                                render={({ field }) => (
+                                                                    <FormItem className='flex flex-row items-center'>
+                                                                        <Checkbox
+                                                                            {...field}
+                                                                            checked={selectedEnvironments.includes(
+                                                                                item.id
+                                                                            )}
+                                                                            className='m-3 data-[state=checked]:bg-primary data-[state=checked]:border-[#2C2677]'
+                                                                            id='environments'
```
benwaples: It appears that this id wont be unique because when we map over the `filteredEnvirontments` each environment checkbox will have this id, which can cause a lot of browsers to warn about non-unique ids used. Perhaps we can use something like `environment-${item.id}` so that its unique but still predictable for the `FormLabel`?

packages/javascript/bh-shared-ui/src/components/ExploreSearchCombobox/ExploreSearchCombobox.tsx
```diff
@@ -113,7 +113,7 @@ const ExploreSearchCombobox: React.FC<{
                                 <ListItemText primary={disabledText} />
                             </ListItem>
                         ) : (
-                            data!.map((item: SearchResult, index: any) => {
+                            data?.map((item: SearchResult, index: any) => {
```
benwaples, line 116: TY!

packages/javascript/bh-shared-ui/src/components/UpdateUserDialog/UpdateUserDialog.tsx
```diff
@@ -46,29 +55,30 @@ const UpdateUserDialog: React.FC<{
...
+                        <VisuallyHidden asChild>
+                            <DialogDescription>Update User Dialog</DialogDescription>
+                        </VisuallyHidden>
+                        <UpdateUserForm
+                            error={error}
+                            isLoading={isLoading}
+                            onSubmit={handleOnSave}
+                            hasSelectedSelf={hasSelectedSelf}
+                            showEnvironmentAccessControls={showEnvironmentAccessControls}
+                            userId={userId!}
```
benwaples: Im curious if this non-null assertion is still needed here? It looks like userId comes from props and is typed as `string` so Im thinking it should already be typed correctly to pass to `UpdateUserForm`?

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
```diff
@@ -56,427 +69,681 @@ const UpdateUserForm: React.FC<{
...
-                                    type='email'
-                                    fullWidth
-                                    error={!!errors.emailAddress}
-                                    helperText={errors.emailAddress?.message}
-                                    data-testid='update-user-dialog_input-email-address'
-                                />
+        <Form {...form}>
+            <form autoComplete='off' onSubmit={form.handleSubmit(handleOnSave, onError)}>
+                <div className='flex gap-x-4 justify-center'>
+                    <Card className='p-6 rounded shadow max-w-[600px] w-full'>
```
benwaples: Feel free to shoot down this suggestion, I very well could be missing some context!

It appears the dialog background is transparent and the content has the Card with the form in that, ya? I bring this up because it appears the dialog border still has a box shadow (very faintly) which doesnt seem intended. 

So my suggestion would effectively be to remove the card and use the dialogs background + maxWidth the dialog to `sm`. I think would include a handful of styling changes because I see the card uses flex.

Orrr perhaps use the good old div (without padding, rounding, box-shadow, or max-width) instead of card  🤷

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,696 @@
...
-                                />
-                            </Grid>
+        <Form {...form}>
+            <form
+                autoComplete='off'
+                data-testid='create-user-dialog_form'
+                onSubmit={form.handleSubmit(handleOnSave, onError)}>
+                {!(getRolesQuery.isLoading || listSSOProvidersQuery.isLoading) && (
+                    <div className='flex gap-x-4 justify-center'>
+                        <Card className='p-6 rounded shadow max-w-[600px] w-full'>
```
benwaples: Same comment regarding Card styling inside the Dialog: https://github.com/SpecterOps/BloodHound/pull/1858/files#r2474218509

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
benwaples: Im seeing the chonky CR comment about this and had similar feelings but figured it wasnt much of a risk right not considering those IDs are in the main migration schema. 

If we did want to address this, I think the easiest/safest way is instead of a `isUserRoleETACEnabled` util, we add a hook that pulls in the `useListDisplayRoles` and returns two functions, one that checks if a role id is ETAC enabled, or that a role name is ETAC enabled! And then at this point it would make sense to encode what role names are considered ETAC enabled.

Im happy to help figure out where and what these types+functions should be!

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
```diff
@@ -56,427 +69,681 @@ const UpdateUserForm: React.FC<{
...
-            authenticationMethod: initialData.SSOProviderId ? 'sso' : 'password',
+            authenticationMethod: initialData.sso_provider_id ? 'sso' : 'password',
         },
     });
 
-    const authenticationMethod = watch('authenticationMethod');
+    const [selectedRoleValue, setSelectedRoleValue] = useState<number[]>(initialData.roles);
+    const [searchInput, setSearchInput] = useState<string>('');
+    const roleInputValue = form.watch('roles');
+    const selectedRole = roleInputValue.toString() === '2' || roleInputValue.toString() === '3';
```
benwaples: Looks like this same evaluation is done below for the selectedETACEnabledRole variable, do we need both?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,699 @@
...
+    showEnvironmentAccessControls?: boolean;
+}> = ({ error, isLoading, onSubmit, showEnvironmentAccessControls }) => {
+    const defaultValues = {
+        email_address: '',
+        principal: '',
+        first_name: '',
+        last_name: '',
+        password: '',
+        needs_password_reset: false,
+        roles: [3],
```
benwaples: This seems like another use case for a hook/util/function that would map UserRoles to ids so we could be certain that we are initiating the array with the Read-Only user role. 

Unfortunately that would mean we cant completely define the `defaultValues` until the `useListDisplayRoles` request completed. I think in that case we would have to initiate roles as an empty array (looks like the form field is in a loading state until the `getRolesQuery` completes anyway, yay!), and pass an `onSuccess` handler to the `useQuery` options that would do something like `form.setValue('',[userRoleMap.ReadOnly?.id])`

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,699 @@
...
+            const returnMappedEnvironments: string[] | undefined = availableEnvironments?.map(
+                (environment) => environment.id
+            );
+            returnMappedEnvironments && setSelectedEnvironments(returnMappedEnvironments);
+        } else {
+            setSelectedEnvironments([]);
         }
+    };
 
+    const formatReturnedEnvironments: EnvironmentRequest[] =
```
benwaples: looks like this variable is only used in the handleOnSave function? If so, could we move the definition into that function so we dont redefine the variable on every rerender?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,699 @@
...
+        all_environments: false,
+        environment_targeted_access_control: {
+            environments: null,
         },
-    });
+    };
+
+    const form = useForm<CreateUserRequestForm>({ defaultValues });
 
     const [authenticationMethod, setAuthenticationMethod] = React.useState<string>('password');
```
benwaples: I know this is outside your change set, but do you mind updating this to be useState instead of `React.useState` here for consistency?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,699 @@
...
+    const [needsPasswordReset, setNeedsPasswordReset] = useState(false);
+    const [searchInput, setSearchInput] = useState<string>('');
+    const [selectedEnvironments, setSelectedEnvironments] = useState<string[]>([]);
+    const [selectedRoleValue, setSelectedRoleValue] = useState([3]);
```
benwaples: same comment as this one: https://github.com/SpecterOps/BloodHound/pull/1858/files#r2475736429

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,699 @@
...
-                                        />
-                                    )}
-                                />
-                            </Grid>
+                            <div className='flex flex-col mt-4 w-full' data-testid='create-user-dialog_content'>
+                                <div className='mb-4'>
+                                    <FormField
+                                        name='roles.0'
+                                        control={form.control}
+                                        defaultValue={1}
```
benwaples: just curious, what is defaultValue doing in this case? I would think that it would be something like 3 or '3' to indicate the default value is Read-Only, but Im not totally sure what 1 does.

packages/javascript/bh-shared-ui/src/views/Users/Users.tsx
```diff
@@ -114,14 +113,18 @@ const Users: FC = () => {
...
             const updatedUser: UpdateUserRequest = {
-                emailAddress: user.email_address || '',
+                email_address: user.email_address || '',
                 principal: user.principal_name || '',
-                firstName: user.first_name || '',
-                lastName: user.last_name || '',
-                ...(user.sso_provider_id && { SSOProviderId: user.sso_provider_id }),
+                first_name: user.first_name || '',
+                last_name: user.last_name || '',
+                ...(user.sso_provider_id && { sso_provider_id: user.sso_provider_id }),
```
benwaples: would `sso_provider_id: user.sso_provider_id || undefined` work?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
catsiller: ah good catch - since the default state of roles is set above in a hook, defaultValue={1} is doing nothing/not necessary so removing

specter-flq:
Started the review by running the UI and looked good except there is a shadow being applied to the scrollable dialog contained. See attached Image

Screenshot 2025-11-10 at 4 11 51 PM

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
catsiller: refactored a bit, now selectedRole watches user input on the select

`const selectedRoleValue = form.watch('roles.0');`

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
catsiller: moved into new `EnvironmentSelectPanel` component / CodeRabbit suggested to move it into useEffect so did!

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,208 @@
...
+                                        data-testid='create-user-dialog_select-all-environments-checkbox'
+                                    />
+                                    <FormLabel
+                                        htmlFor='allEnvironments'
+                                        className='ml-3 w-full cursor-pointer font-normal'>
+                                        Select All Environments
+                                    </FormLabel>
+                                </FormItem>
+                            )}
+                        />
```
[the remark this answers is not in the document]
catsiller: if removed, indeterminate not displaying in `areAllEnvironmentsSelected <Checkbox />`

specter-flq (commented):
Sorry if its a bit nitty, but when checking out the ETAC modal when creating user saw that the dark line/border under search does not go the whole width of the container. It jumped at me so wanted to confirm the design and in the design it goes full width, see below for the comparisons. 
Screenshot 2025-11-12 at 3 43 46 PM
vs
(design)
Screenshot 2025-11-12 at 3 38 05 PM

specter-flq:
Not sure if this is intended but will add just in case. Seems like everytime a user opens the update user all the etac choices are checked. In the video below I open the modal, all are check but then I check one and save. Then open again and all are still checked.


https://github.com/user-attachments/assets/f3753cd2-473b-4aa8-8afc-1a6bc95da228

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
+        } else {
+            form.setValue('needs_password_reset', false);
         }
+    };
+
+    const getRolesQuery = useQuery(['getRoles'], ({ signal }) =>
+        apiClient.getRoles({ signal }).then((res) => res.data?.data?.roles)
+    );
+
+    const listSSOProvidersQuery = useQuery(['listSSOProviders'], ({ signal }) =>
```
specter-flq: Seems like this is being repeated in other files, wonder if we should maybe create a hook for this and at least add it to the files in the scope of this ticket and maybe create a tech debt ticket for the others after this is merged.

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
+        apiClient.listSSOProviders({ signal }).then((res) => res.data?.data)
+    );
 
+    const selectedRoleValue = form.watch('roles.0');
+
+    const matchingRole = getRolesQuery.data?.find((item) => selectedRoleValue === item.id)?.name;
```
specter-flq: this maybe could go to a utils file ? Seems like it is repeated and could come in handy in other situations

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
+
+    const selectedETACEnabledRole = matchingRole && ['Read-Only', 'User'].includes(matchingRole);
+    const selectedAdminOrPowerUserRole = matchingRole && ['Administrator', 'Power User'].includes(matchingRole);
```
specter-flq: potentially a utils candidate as well, at least in the scope of this feature.

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
                     message: 'An unexpected error occurred. Please try again.',
                 });
             }
         }
-    }, [authenticationMethod, setValue, error, setError]);
+    };
 
-    const getListDisplayRolesQuery = useListDisplayRoles();
+    const handleOnSave = () => {
+        const values = form.getValues();
```
specter-flq: nit: although values is good and we have already formData below, maybe add another word to the variable name something like formValues, so down the file its still clear what it is.

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
+
+    const getRolesQuery = useQuery(['getRoles'], ({ signal }) =>
+        apiClient.getRoles({ signal }).then((res) => res.data?.data?.roles)
+    );
+
+    const listSSOProvidersQuery = useQuery(['listSSOProviders'], ({ signal }) =>
+        apiClient.listSSOProviders({ signal }).then((res) => res.data?.data)
+    );
 
+    const selectedRoleValue = form.watch('roles.0');
```
specter-flq: wonder if maybe 'roles.0' should be a constant with a more descriptive name.

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
-                                            label='Last Name'
-                                            fullWidth
-                                            error={!!errors.lastName}
-                                            helperText={errors.lastName?.message}
-                                            data-testid='create-user-dialog_input-last-name'
-                                        />
-                                    )}
-                                />
-                            </Grid>
+        <Form {...form}>
```
specter-flq, line 166: Wonder if do to the size of the form we should consider breaking it down into smaller components by field maybe and have those in files in a folder rather than in here.

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.test.tsx
```diff
@@ -89,7 +91,11 @@ const MOCK_ROLES = [
 ];
 
 describe('CreateUserForm', () => {
-    it('should not have less characters than the minimum requirement', async () => {
+    type SetupOptions = {
+        renderShowEnvironmentAccessControls?: boolean;
+    };
+
+    const setup = (options?: SetupOptions) => {
```
specter-flq: nit: maybe we should consider giving this a two letter variable name so down the file in the other tests we would have a bit more descriptive sense of what it is setting up. Maybe formInitSetup or something like that.

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+    isNewUser: boolean;
+    initialData?: UpdateUserRequestForm;
+    form: UseFormReturn;
+}> = ({ isNewUser, initialData, form }) => {
+    const { data: availableEnvironments } = useAvailableEnvironments();
+
+    const initialEnvironmentsSelected = initialData?.environment_targeted_access_control?.environments?.map(
+        (item) => item.environment_id
+    );
+    const returnMappedEnvironments: any = availableEnvironments?.map((environment) => environment.id);
```
specter-flq: maybe we should consider typing it to not have the any in there.

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+            : matchingEnvironmentValues || (isNewUser && []);
+
+    const [searchInput, setSearchInput] = React.useState<string>('');
+    const [selectedEnvironments, setSelectedEnvironments] = React.useState<string[]>(checkedEnvironments);
+
+    const filteredEnvironments = availableEnvironments?.filter((environment: Environment) =>
+        environment.name.toLowerCase().includes(searchInput.toLowerCase())
+    );
+
+    const allEnvironmentsSelected: any =
```
specter-flq: same as above with typing

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+import { UseFormReturn } from 'react-hook-form';
+import { useAvailableEnvironments } from '../../hooks/useAvailableEnvironments/useAvailableEnvironments';
+import { UpdateUserRequestForm } from '../UpdateUserForm';
+
+const EnvironmentSelectPanel: React.FC<{
+    isNewUser: boolean;
+    initialData?: UpdateUserRequestForm;
+    form: UseFormReturn;
+}> = ({ isNewUser, initialData, form }) => {
+    const { data: availableEnvironments } = useAvailableEnvironments();
```
specter-flq: The availableEnviroments data is type Environment[] but i see below alot of string and any typing when working with environments. Maybe it would be possible to keep working with the Environment type?

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+            form.setValue('environment_targeted_access_control.environments', null);
+        }
+    }, [selectedEnvironments]);
+
+    return (
+        <Card className='flex-1 p-4 rounded shadow max-w-[400px]'>
+            <DialogTitle>Environmental Targeted Access Control </DialogTitle>
+            <div className='flex flex-col h-full pb-6' data-testid='create-user-dialog_environments-checkboxes-dialog'>
+                <div className='border border-color-[#CACFD3] mt-3 box-border h-full overflow-y-auto'>
+                    <div className='border border-solid border-color-[#CACFD3] flex'>
```
specter-flq: We should consider replacing the hex value #CACFD3 with the variable, we have them in doodle.  Seems like we have it twice but it should be either neutral-5 or neutral-light-5

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+                                        className={
+                                            allEnvironmentsSelected && '!bg-primary border-[#2C2677] dark:!bg-[#f4f4f4]'
+                                        }
+                                        icon={
+                                            allEnvironmentsIndeterminate && (
+                                                <Minus
+                                                    className='h-full w-full bg-[#f4f4f4] text-neutral-dark-1 dark:bg-[#222222] dark:text-[#f4f4f4]'
+                                                    absoluteStrokeWidth={true}
+                                                    strokeWidth={3}
+                                                />
```
specter-flq: Same as above, 
#f4f4f4 is in doodle under neutral-2 or neutral-light-2. 
#2C2677 is not in there but maybe we can find a similar one? or add it. 
#222222 is neutral-2 or neutral-dark-2

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+                                            name='environment_targeted_access_control.environments'
+                                            control={form.control}
+                                            render={() => (
+                                                <FormItem className='flex flex-row items-center'>
+                                                    <Checkbox
+                                                        checked={
+                                                            selectedEnvironments &&
+                                                            selectedEnvironments.includes(item.id)
+                                                        }
+                                                        className='m-3 data-[state=checked]:bg-primary data-[state=checked]:border-[#2C2677]'
```
specter-flq: same as above, #2C2677, regarding hardcoded hex values

packages/javascript/bh-shared-ui/src/components/UpdateUserDialog/UpdateUserDialog.tsx
```diff
@@ -14,30 +14,38 @@
...
-        if (user.SSOProviderId) {
-            parsedSSOProviderId = parseInt(user.SSOProviderId);
+        if (user.sso_provider_id) {
+            parsedSSOProviderId = parseInt(user.sso_provider_id);
         }
 
         onSave({
             ...user,
-            SSOProviderId: parsedSSOProviderId,
+            sso_provider_id: parsedSSOProviderId,
```
specter-flq: nitish:  Maybe we could reduce to:

```
const handleOnSave = (user: UpdateUserRequestForm) => {
        onSave({
            ...user,
            sso_provider_id: user.sso_provider_id ? parseInt(user.sso_provider_id) : undefined,
        })
            .then(() => {
                onClose();
            })
            .catch((err) => console.error(err));
    };
```

packages/javascript/bh-shared-ui/src/components/CreateUserDialog/CreateUserDialog.tsx
```diff
@@ -14,51 +14,73 @@
...
             ...user,
-            SSOProviderId: parsedSSOProviderId,
+            sso_provider_id: parsedSSOProviderId,
         })
-            .then(() => {
-                onClose();
-            })
+            .then(() => setIsOpen(false))
             .catch((err) => console.error(err));
     };
```
specter-flq: nitish: Maybe we could reduce to:

```
const handleOnSave = (user: UpdateUserRequestForm) => {
        onSave({
            ...user,
            sso_provider_id: user.sso_provider_id ? parseInt(user.sso_provider_id) : undefined,
        })
            .then(() => {
                onClose();
            })
            .catch((err) => console.error(err));
    };
```

packages/javascript/bh-shared-ui/src/components/CreateUserDialog/CreateUserDialog.test.tsx
```diff
@@ -106,46 +122,51 @@ describe('CreateUserDialog', () => {
...
-            />
-        );
-
-        return {
-            user,
-            testUser,
-            testOnClose,
-            testOnSave,
-        };
+    const setup = async (options?: SetupOptions) => {
```
specter-flq: nit: maybe we should consider giving this a two letter variable name so down the file in the other tests we would have a bit more descriptive sense of what it is setting up. Maybe dialogInitSetup or something like that.

packages/javascript/bh-shared-ui/src/components/UpdateUserDialog/UpdateUserDialog.test.tsx
```diff
@@ -126,6 +134,7 @@ describe('UpdateUserDialog', () => {
     type SetupOptions = {
         renderErrors?: boolean;
         renderLoading?: boolean;
+        renderShowEnvironmentAccessControls?: boolean;
     };
 
     const setup = (options?: SetupOptions) => {
```
specter-flq: nit: maybe we should consider giving this a two letter variable name so down the file in the other tests we would have a bit more descriptive sense of what it is setting up. Maybe dialogInitSetup or something like that.

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.test.tsx
```diff
@@ -361,7 +361,11 @@ const MOCK_USER = [
 ];
 
 describe('UpdateUserForm', () => {
-    it('should not allow the input to exceed the allowed length', async () => {
+    type SetupOptions = {
+        renderShowEnvironmentAccessControls?: boolean;
+    };
+
+    const setup = (options?: SetupOptions) => {
```
specter-flq: nit: maybe we should consider giving this a two letter variable name so down the file in the other tests we would have a bit more descriptive sense of what it is setting up. Maybe dialogInitSetup or something like that.

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
```diff
@@ -14,40 +14,49 @@
...
-    onCancel: () => void;
     onSubmit: (user: UpdateUserRequestForm) => void;
     userId: string;
     hasSelectedSelf: boolean;
     isLoading: boolean;
     error: any;
-}> = ({ onCancel, onSubmit, userId, hasSelectedSelf, isLoading, error }) => {
+    showEnvironmentAccessControls?: boolean;
+}> = ({ onSubmit, userId, hasSelectedSelf, isLoading, error, showEnvironmentAccessControls }) => {
     const getUserQuery = useQuery(
```
specter-flq: I know this is here from before, but feels like this should live in a hook

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
```diff
@@ -56,427 +65,522 @@ const UpdateUserForm: React.FC<{
...
-                    type: 'custom',
-                    message: 'An unexpected error occurred. Please try again.',
-                });
             }
         }
-    }, [authenticationMethod, setValue, error, setError]);
+    };
+
+    const handleOnSave = () => {
+        const values = form.getValues();
```
specter-flq: nit: same as in create form, maybe a two letter variable name to be a tad more descriptive

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
```diff
@@ -56,427 +65,522 @@ const UpdateUserForm: React.FC<{
...
-                                    type='email'
-                                    fullWidth
-                                    error={!!errors.emailAddress}
-                                    helperText={errors.emailAddress?.message}
-                                    data-testid='update-user-dialog_input-email-address'
-                                />
-                            )}
-                        />
-                    </Grid>
+        <Form {...form}>
```
specter-flq, line 191: Similar to the Create form, maybe we could break this down due to size?

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
catsiller: resolved in https://github.com/SpecterOps/BloodHound/pull/2050

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
catsiller: resolved in https://github.com/SpecterOps/BloodHound/pull/2050

packages/javascript/bh-shared-ui/src/views/Users/Users.tsx
catsiller: resolved in https://github.com/SpecterOps/BloodHound/pull/2050

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,208 @@
...
+
+    const allEnvironmentsSelected: any =
+        selectedEnvironments &&
+        selectedEnvironments.length === availableEnvironments?.length &&
+        availableEnvironments!.length > 0;
+
+    const allEnvironmentsIndeterminate =
+        selectedEnvironments &&
+        selectedEnvironments.length > 0 &&
+        selectedEnvironments.length < availableEnvironments!.length;
```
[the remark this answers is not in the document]
catsiller: resolved in https://github.com/SpecterOps/BloodHound/pull/2050

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
catsiller: resolved in https://github.com/SpecterOps/BloodHound/pull/2050

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,537 @@
...
+                                    disabled={isLoading}
+                                    role='button'
+                                    type='submit'>
+                                    Save
+                                </Button>
+                            </DialogActions>
+                        </Card>
+                        {showEnvironmentAccessControls && selectedETACEnabledRole && (
+                            <EnvironmentSelectPanel form={form as unknown as UseFormReturn} />
+                        )}
```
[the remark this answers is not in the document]
catsiller: fixed in https://github.com/SpecterOps/BloodHound/pull/2050

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
catsiller: ohhh the colors are in Doodle, thanks for that info!

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+    const handleSelectAllEnvironmentsChange = (allEnvironmentsChecked: string | boolean) => {
+        if (allEnvironmentsChecked || allEnvironmentsIndeterminate) {
+            const returnMappedEnvironments: string[] | undefined = availableEnvironments?.map(
+                (environment) => environment.id
+            );
+            returnMappedEnvironments && setSelectedEnvironments(returnMappedEnvironments);
+        } else {
+            setSelectedEnvironments([]);
+        }
```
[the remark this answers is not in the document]
maffkipp: The intended behavior here is to select all after the indeterminate checkbox is clicked, so this implementation appears to work as expected

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
maffkipp: We are planning on holding off on this refactor for this PR, but we can create a followup ticket to break the rest of the form out into some reusable fields!

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
maffkipp: This is occurring because the items in the list do not have unique IDs, which we are addressing elsewhere!

packages/javascript/bh-shared-ui/src/components/UpdateUserDialog/UpdateUserDialog.tsx
```diff
@@ -46,29 +54,30 @@ const UpdateUserDialog: React.FC<{
...
+                            isLoading={isLoading}
+                            onSubmit={handleOnSave}
+                            hasSelectedSelf={hasSelectedSelf}
+                            showEnvironmentAccessControls={showEnvironmentAccessControls}
+                            userId={userId!}
+                        />
+                    </DialogContent>
+                </DialogOverlay>
+            </DialogPortal>
         </Dialog>
```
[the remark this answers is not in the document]
catsiller: resolved in https://github.com/SpecterOps/BloodHound/pull/2058

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
```diff
@@ -14,426 +14,541 @@
...
-                                                        error={!!errors.password}
-                                                        helperText={errors.password?.message}
-                                                        data-testid='create-user-dialog_input-password'
+                                                    <Tooltip
+                                                        defaultOpen={false}
+                                                        tooltip='Only Read-Only and Users roles contain the environment target access control.'
+                                                        contentProps={{
+                                                            className:
+                                                                'max-w-80 dark:bg-neutral-dark-5 dark:text-white border-0 !z-[2000]',
+                                                        }}
```
benwaples: Love the tooltip! 

I believe the trigger element renders a button for accessibility and because its in a form the default behavior onclick will attempt submitting the form which would be funny lol there is prop that passes props along to the tooltip trigger tho! I think could add this prop to the Tooltip: 
```
triggerProps={{ type: 'submit' }}
```

packages/javascript/bh-shared-ui/src/components/CreateUserForm/CreateUserForm.tsx
benwaples: I tested this out with 150 domains and its not a concern!

packages/javascript/bh-shared-ui/src/components/EnvironmentSelectPanel/EnvironmentSelectPanel.tsx
```diff
@@ -0,0 +1,222 @@
...
+                                    <FormLabel
+                                        htmlFor='allEnvironments'
+                                        className='ml-3 w-full cursor-pointer font-normal'>
+                                        Select All Environments
+                                    </FormLabel>
+                                </FormItem>
+                            )}
+                        />
+                    </div>
+                    <div className='flex flex-col' data-testid='create-user-dialog_environments-checkboxes-div'>
```
benwaples: I tested this with 50 domains and the list of environments expands past the form, i attached a wee video example. Perhaps we lock the height of of the `EnvironmentSelectPanel` to the height of the form?

https://github.com/user-attachments/assets/18a537fe-176e-41f2-baca-fc919347002a

packages/javascript/bh-shared-ui/src/components/UpdateUserForm/UpdateUserForm.tsx
benwaples: The dialog border box shadow is gone, nice work!! 

There are a couple smol unexpected nits Im noticing that I think are related: 
- a scrollbar outside the content but within the dialog area
- The dialog technically extends past the card area so I cant close the dialog by clicking outside the card because im still clicking inside the dialog (words are hard and this is hard for me to explain so I added a video). But if I click a little farther out, it closes.

https://github.com/user-attachments/assets/5dfd59ee-c071-4562-90d5-adebbd529698

packages/javascript/bh-shared-ui/src/views/Users/UsersWithEnvironmentAccessControls.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { FC } from 'react';
+import FeatureFlag from '../../components/FeatureFlag/FeatureFlag';
+import Users from './Users';
+
+const UsersWithEnvironmentAccessControls: FC = () => {
+    return (
+        <FeatureFlag
+            flagKey='environment_targeted_access_control'
+            enabled={<Users showEnvironmentAccessControls={true} />}
+            disabled={<Users showEnvironmentAccessControls={false} />}
```
benwaples: Not requesting any changes here, just wanted to share a fun React thing!

Boolean props that default false can be passed like: `<Users showEnvironmentAccessControls />` and react will interpret that as `<Users showEnvironmentAccessControls={true} /> `!

PR #1862, opened 2025-09-04
BED-6063: Changelog

brandonshearin:
## Summary
This PR introduces a new changelog package to the pg driver, providing a mechanism for deduplicating and reconciling graph changes during ingest.

## Motivation
ingest can enqueue large volumes of redundant node and edge updates. Without deduplication, these updates create unnecessary db writes, increasing ingest time. The changelog provides an in-memory cache and buffered flush pipeline to ensure that only meaningful changes are written.

### Cache
- Maintains a hash map of node/edge identity + content uint64 hashes.
- Determines whether a proposed change is new, modified, or unchanged.
- Tracks hit/miss stats for observability.
- Initialized with a size equal to graph cardinality. Research from @superlinkx found that this improves memory pressure considerably due to the decreased number of allocations needed to grow the map as it is written to.

### Loop
- Buffers deduplicated changes in memory.
- Flushes either on reaching a batch size threshold or after idle timeout.

### Changelog (public API)
- Composes the cache, db, and loop into a single orchestrator.
- Provides ResolveChange (deduplication), Submit (enqueue), and Start (event loop) methods.
- Exposes cache stats for metrics/logging.

## Benefits
Performance: reduces redundant DB writes during ingest.
Correctness: preserves reconciliation behavior via consistent lastseen updates.
Testability: cache, db, and loop have good seams for unit tests


go.mod
brandonshearin, line 1: open question:
dawgs branching strategy. 
there is a sister pr with a couple of changes that this pr depends on

cmd/api/src/daemons/changelog/steelthread/main.go
brandonshearin, line 1: Open Question:
should i leave this in the codebase?
it can serve as a end-end harness- a way to get feedback on updates to the changelog without going through the entire application.

docker-compose.dev.yml
brandonshearin, line 1: remove this before merging.

cmd/api/src/daemons/changelog/loop.go
```diff
@@ -0,0 +1,147 @@
...
+				return err
+			}
+		}
+		return nil
+	})
+}
+
+// changeBuffer is a small helper that accumulates changes and flushes them
+// when size thresholds are hit.
+type changeBuffer struct {
```
brandonshearin: add unit test coverage for changeBuffer, esp. tryFlush()

cmd/api/src/daemons/changelog/changelog.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (s *Changelog) ResolveChange(change Change) (bool, error) {
+	c := s.flagManager.getCache()
+	if c == nil { // treat as pass-through when disabled.
+		return true, nil
+	}
+
+	return c.shouldSubmit(change)
+}
```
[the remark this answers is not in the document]
wes-mil: This nil check if probably a good idea

cmd/api/src/services/graphify/ingest.go
```diff
@@ -53,16 +53,83 @@ type ReadOptions struct {
...
+// IngestContext is a container for dependencies needed by ingest
+type IngestContext struct {
+	Ctx context.Context
+	// Batch is the buffering/flushing mechanism that writes entities to the graph database
+	Batch BatchUpdater
+	// IngestTime is a single timestamp assigned to the lastseen property of every entity ignested per ingest run
 	IngestTime time.Time
+	// Manager is the caching layer that deduplicates ingest payloads across ingest runs
+	Manager ChangeManager
 }
```
wes-mil, line 65: Go pretty heavily discourages embedding a context inside of a struct. [link](https://go.dev/blog/context-and-structs)
> Contexts should not be stored inside a struct type, but instead passed to each function that needs it.

I'm not exactly sure what this could look like instead, but I think it's worth considering a different pattern for this

wes-mil: Perhaps passing context as a separate parameter?

cmd/api/src/services/graphify/ingestnodes.go
```diff
@@ -0,0 +1,166 @@
...
+		} else {
+			slog.Error(fmt.Sprintf("Bad type found for node operating system property during ingest. Expected string, got %T", rawOS))
+		}
+	}
+
+	if rawDN, hasDN := properties[ad.DistinguishedName.String()]; hasDN && rawDN != nil {
+		if dn, typeMatches := rawDN.(string); typeMatches {
+			properties[ad.DistinguishedName.String()] = strings.ToUpper(dn)
+		} else {
+			slog.Error(fmt.Sprintf("Bad type found for node distinguished name property during ingest. Expected string, got %T", rawDN))
```
wes-mil: I'd use slog instead of fmt instead of slog

cmd/api/src/services/graphify/ingest.go
wes-mil: I would still prefer to pass contexts through function parameters, however it does appear that there is more nuance to this rule. As stated by an engineer director at Golang [here](https://github.com/golang/go/issues/22602#issuecomment-344980503), passing contexts through parameter structs is appropriate. In this specific case, I think this is okay but only because of the ergonomics of datapipe/graphify ingest function.

superlinkx: That thread is pretty specific that you still shouldn't do it for public methods/types. I'd rather us move the context out of this struct and use a different name for the configuration struct. Long term, I want us to have a more consistent approach to context, but for now just avoiding footguns with context fields in our package APIs is good enough.

cmd/api/src/services/graphify/ingestnodes.go
superlinkx: Yes please, all slog's should use proper slog patterns with typed parameters, i.e.

```go
slog.Error(
    "Bad type found for node operating system property during ingest",
    slog.Group(
        "type_mismatch",
        slog.String("expected", "string"),
        slog.String("got", fmt.Sprintf("%T", rawOS),
    ),
)
```

Note that we need a fmt.Sprintf here unless we want to use the reflect package directly, which I don't think is warranted. But any values we care about should be structured log fields rather than 

Additionally, because these "errors" do not actually cause any change in behavior downstream, we should probably use Warn to start as it's possible we run into these frequently and clog up our error logs with noise.

brandonshearin: the functions in this file were just moved from a previous home in ingest.go, that file was getting big and i moved node specific handlers here, and relationship specific things to ingestrelationships.go

happy to fix this broken window while we are here

superlinkx: That makes sense. It's a pretty fast fix, I'd appreciate us getting this broken window sorted while we're seeing it. The `fmt.Sprintf()` wrapping was always just to make it so we could ship slog in a reasonable time frame without causing serious disruptions. Ideally we should be fixing them anytime we find them to match our typed, structured logging guidelines.

cmd/api/src/services/graphify/ingest.go
superlinkx: I'm fine making this a suggestion and let Brandon have final say on whether we restructure this now or later. The slog broken window is the only one I was willing to block on.

cmd/api/src/services/graphify/ingestnodes.go
brandonshearin: done!

superlinkx (approved):

PR #1873, opened 2025-09-09
feat(FinishedJobsTable): apply filter to finished jobs hook

TheNando:
## Description

* Allows for filtering finished jobs API client and query
* Moves some types from API client response module to types module

## Motivation and Context

Resolves BED-6407

The last PR for the Finished Jobs Log rework. This applies the state that is built by the FinishedJobsFilterDialog to the API call.

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


packages/javascript/js-client-library/src/client.ts
```diff
@@ -92,6 +92,13 @@ import {
 } from './responses';
 import * as types from './types';
 
+/** Return the value as a string with the given prefix */
+const prefixValue = (prefix: string, value: any) => (value !== undefined ? `${prefix}:${value.toString()}` : undefined);
```
benwaples, line 96: should we require value to be `number | string`? Or is it expected that we could use this with other data types?

packages/javascript/js-client-library/src/client.ts
```diff
@@ -743,22 +750,59 @@ class BHEAPIClient {
...
+                        status: prefixValue('eq', status),
+                        start_time: prefixValue('gte', start_time),
+                        end_time: prefixValue('lte', end_time),
+                        client_id: prefixValue('eq', client_id),
+                        ad_structure_collection: prefixValue('eq', ad_structure_collection),
+                        ca_registry_collection: prefixValue('eq', ca_registry_collection),
+                        cert_services_collection: prefixValue('eq', cert_services_collection),
+                        dc_registry_collection: prefixValue('eq', dc_registry_collection),
+                        local_group_collection: prefixValue('eq', local_group_collection),
+                        session_collection: prefixValue('eq', session_collection),
```
benwaples, line 804: very nice!

benwaples (approved):
Nice job setting all this up so that wiring up the filters was this straight forward! I left one potential no-op comment regarding a type so approving anyway!

packages/javascript/js-client-library/src/client.ts
TheNando: That would definitely improve the type. We'd also have to include undefined and boolean. In this case, it could be useful to keep the `any` as it would support any object with a `toString`. There could be a case where we need to support an array-like params key where it could be easier to make a custom toString that supports this method rather than handle that as a bespoke alternative case.

PR #1878, opened 2025-09-10
feat (api): Add Get Users Minimal Endpoint - BED-6448

LawsonWillard:
## Description

Added new endpoint to get all users without any sensitive data. The endpoint is at `/api/v2/bloodhound-users-minimal`. Added new tests to cover the endpoint as well as updated open api docs to include new endpoint. 

Added new Read Users permission to enable non admins to access the above endpoint. Updated the schema with this new permission and added it to the 8.2.0 migration as well. 

## Motivation and Context
Resolves: BED-6448

As a user without Administrator permissions, I want to be able to share queries with other active users. Right now, only administrator user roles can use the GET /bloodhound-users which is how a given user can get information about other users to share a query with.

## How Has This Been Tested?

- Added new unit tests to cover endpoint
- Local Testing:
- Create two users, 1 can be the initial admin and second should be an upload only user. 
- First user should be able to access the endpoint while the upload only user should see a 403. 
- First user should see both users returned with only email, first name, last name and user id in the payload. 

Migration Testing:
- Switch to v8.1.1 branch
- Using just run-bhce-container, create a v8.1.1 container (v8.1.2 doesnt exist in BHCE), once started, login. 
- Down the v8.1.1 container 
- Switch to feature branch (BED-6448)
- Using just run-bhce-container, create a v8.2.0 container, once started, login. 
- Verify that the new endpoints exist and are usable.
- Verify in the postgres DB that the new permission exists and that they are associated to the appriopriate roles.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- New Features
  - Added endpoint to list active users with minimal, non-sensitive fields: GET /api/v2/bloodhound-users-minimal. Supports sorting (id, first_name, last_name, email_address) and filtering. Excludes disabled users.
- Permissions
  - Introduced a new “Read Users” permission and assigned it to Administrator, User, Read-Only, and Power User roles.
- Documentation
  - OpenAPI updated with the new endpoint and a minimal user schema.
- Tests
  - Added tests covering sorting, filtering, validation errors, and database error handling for the new endpoint.


cmd/api/src/auth/permission.go
```diff
@@ -91,6 +93,7 @@ func Permissions() PermissionSet {
 		AuthManageProviders:                 model.NewPermission("auth", "ManageProviders"),
 		AuthManageSelf:                      model.NewPermission("auth", "ManageSelf"),
 		AuthManageUsers:                     model.NewPermission("auth", "ManageUsers"),
+		ReadUsers:                           model.NewPermission("users", "Read"),
```
mistahj67: I think it might be better to keep this under the same auth flag to match the `ManageUsers` precedence
```suggestion
		AuthReadUsers:                           model.NewPermission("auth", "ReadUsers"),
```

cmd/api/src/api/v2/get_users.go
```diff
@@ -0,0 +1,180 @@
...
+
+type UserMinimal struct {
+	ID            uuid.UUID `json:"id"`
+	PrincipalName string    `json:"principal_name"`
+	FirstName     string    `json:"first_name"`
+	LastName      string    `json:"last_name"`
+}
+
+// ListUsersMinimal - Returns a list of Users without any sensitive data. At the time, this is used in the saved queries
+// to workflow to return a list of users with whom a query can be shared with.
```
AD7ZJ:
```suggestion
// workflow to return a list of users with whom a query can be shared with.
```

cmd/api/src/api/v2/get_users.go
```diff
@@ -0,0 +1,180 @@
...
+			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("%s: %s", api.ErrorResponseDetailsNotSortable, column), request), response)
+			return
+		}
+
+		if descending {
+			order = append(order, column+" desc")
+		} else {
+			order = append(order, column)
+		}
+	}
```
AD7ZJ: There's a function called `api.ParseSortParameters()` which does this, I think you could use that here?

cmd/api/src/api/v2/get_users.go
```diff
@@ -0,0 +1,180 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+// ValidFilters - returns a map of columns and their valid filters
+func (s UserMinimal) ValidFilters() map[string][]model.FilterOperator {
+	return map[string][]model.FilterOperator{
+		"first_name":     {model.Equals, model.NotEquals},
```
AD7ZJ: Consider adding approximate equals for the strings?

cmd/api/src/api/v2/get_users.go
```diff
@@ -0,0 +1,180 @@
...
+		"last_name":      {model.Equals, model.NotEquals},
+		"principal_name": {model.Equals, model.NotEquals},
+		"id":             {model.Equals, model.NotEquals},
+	}
+}
+
+// IsString - determines if the passed column is a string or not
+func (s UserMinimal) IsString(column string) bool {
```
AD7ZJ: nit: Consider renaming to `IsStringColumn()` for consistency with other endpoints.

cmd/api/src/database/migration/migrations/schema.sql
AD7ZJ, line 1: Should we be modifying the schema.sql file here? I'm new to migrations so take this with a grain of salt but I think all schema changes should go in the vX.X.X.sql file and occasionally those will get squashed into this - but only after the fact, when the vX.X.X.sql files had a chance to run and make the actual changes.

cmd/api/src/database/migration/migrations/v8.2.0.sql
```diff
@@ -25,3 +25,33 @@ VALUES (
...
+        ('users', 'Read')
+    ))
+    OR
+    (r.name = 'Read-Only' AND (p.authority, p.name) IN (
+        ('users', 'Read')
+    ))
+    OR
+    (r.name = 'Power User' AND (p.authority, p.name) IN (
+        ('users', 'Read')
+))) ON CONFLICT DO NOTHING;
```
AD7ZJ: I think you can simplify it a bit: 
```suggestion
INSERT INTO roles_permissions (role_id, permission_id)
SELECT r.id, p.id
FROM roles r
JOIN permissions p
  ON (p.authority, p.name) = ('users', 'Read')
WHERE r.name IN ('Administrator', 'User', 'Read-Only', 'Power User')
ON CONFLICT DO NOTHING;
```

cmd/api/src/api/registration/v2.go
```diff
@@ -91,6 +91,7 @@ func registerV2Auth(resources v2.Resources, routerInst *router.Router, permissio
 		// User management for all BloodHound users
 		routerInst.GET("/api/v2/bloodhound-users", managementResource.ListUsers).RequirePermissions(permissions.AuthManageUsers),
 		routerInst.POST("/api/v2/bloodhound-users", managementResource.CreateUser).RequirePermissions(permissions.AuthManageUsers),
+		routerInst.GET("/api/v2/bloodhound-users/minimal", resources.ListUsersMinimal).RequirePermissions(permissions.ReadUsers), // returns user data without any sensitive information.
```
AD7ZJ: Should this endpoint be something else like `/api/v2/bloodhound-users-minimal` so it doesn't conflict with the following endpoint `/api/v2/bloodhound-users/<user-id>`? I got bit by that on PZ initiative...

AD7ZJ (commented):
Here are a few suggestions

cmd/api/src/database/migration/migrations/schema.sql
mistahj67: nice catch, no this should not be modified at this time

cmd/api/src/api/registration/v2.go
LawsonWillard: Yeah that would work, wasnt sure how best to name this endpoint/permissions set

cmd/api/src/database/migration/migrations/schema.sql
LawsonWillard: ~~I'm not sure to be honest, may be worth reaching out to eng as a whole.~~

Thank you guys, will update!

cmd/api/src/api/registration/v2.go
mistahj67: Because this route is above the user id route below, this will technically work as intended, but it is a valid point to discuss all the same. Might be worth asking for wider audience input there

cmd/api/src/auth/permission.go
LawsonWillard: Wasnt sure if we wanted it under auth but I'll swap it over, do you think it would be worth swapping the endpoint to use the managementResource struct over the regular resource struct (in /registration/v2.go)?

cmd/api/src/api/v2/auth/list_users_minimal.go
```diff
@@ -0,0 +1,138 @@
...
+// workflow to return a list of users with whom a query can be shared with.
+func (s ManagementResource) ListActiveUsersMinimal(response http.ResponseWriter, request *http.Request) {
+	var (
+		usersFilter UserMinimal
+		queryParams = request.URL.Query()
+	)
+
+	if orderBy, err := api.ParseSortParameters(UserMinimal{}, queryParams); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+		return
```
mistahj67:
```suggestion
```

cmd/api/src/api/sort.go
```diff
@@ -98,3 +99,35 @@ func ParseGraphSortParameters(s Sortable, params url.Values) (query.SortItems, e
...
+			return nil, fmt.Errorf("%w: column index: %d", ErrResponseDetailsColumnEmpty, index)
+		}
+	}
+
+	if identifierColumn.Column != "" {
+		sqlSort = append(sqlSort, buildSortString(identifierColumn))
+	}
+
+	return sqlSort, nil
+}
```
mistahj67: I'm a bit concerned around the fallback default behavior provided? Adding a no-op to the sql statements feels unnecessarily risky. It may be a no-op now, but who is to say it will be later. What do you think of just taking a generic default / fallback approach in the event no sort is provided?

```suggestion
// BuildSQLSort takes our sort models and converts them into proper SQL form before sending them to the database.
// The defaultSort should be used in the event no sort is provided
func BuildSQLSort(sort model.Sort, defaultSort model.SortItem) ([]string, error) {
	var sqlSort = make([]string, 0, len(sort))
	for index, sortItem := range sort {
		if sortItem.Column != "" {
			sqlSort = append(sqlSort, buildSortString(sortItem))
		} else {
			return nil, fmt.Errorf("%w: column index: %d", ErrResponseDetailsColumnEmpty, index)
		}
	}

	// Add default if it is not present
	if len(sqlSort) == 0 {
		sqlSort = append(sqlSort, buildSortString(defaultSort))
	}

	return sqlSort, nil
}
```

cmd/api/src/api/v2/auth/list_users_minimal.go
```diff
@@ -0,0 +1,138 @@
...
+	} else if order, err := api.BuildSQLSort(orderBy, model.SortItem{Column: "id"}); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else {
+		queryParameterFilterParser := model.NewQueryParameterFilterParser()
+		if queryFilters, err := queryParameterFilterParser.ParseQueryParameterFilters(request); err != nil {
```
mistahj67: if you define the queryParamaterFilterParser in the top var block, then the following if statement can be added to the same else if scoped block and be that much cleaner 🧼

cmd/api/src/api/v2/auth/list_users_minimal.go
```diff
@@ -0,0 +1,138 @@
...
+							return
+						}
+
+						queryFilters[name][i].IsStringData = usersFilter.IsStringColumn(filter.Name)
+					}
+				}
+			}
+			if sqlFilter, err := queryFilters.BuildSQLFilter(); err != nil {
+				api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponseDetailsBadQueryParameterFilters, request), response)
+				return
```
mistahj67:
```suggestion
```

cmd/api/src/database/auth.go
```diff
@@ -344,6 +345,26 @@ func (s *BloodhoundDB) GetAllUsers(ctx context.Context, order string, filter mod
 	return users, CheckError(result)
 }
 
+func (s *BloodhoundDB) GetAllActiveUsers(ctx context.Context, order string, filter model.SQLFilter) (model.Users, error) {
```
mistahj67: I don't think there's a need to create this db method actually. The GetAllUsers is just as effective if you can skip it in the user loop in the handler.

packages/go/openapi/src/paths/bh-users.bloodhound-users-minimal.yaml
```diff
@@ -0,0 +1,72 @@
...
+      schema:
+        $ref: './../schemas/api.params.predicate.filter.string.yaml'
+    - name: last_name
+      in: query
+      schema:
+        $ref: './../schemas/api.params.predicate.filter.string.yaml'
+    - name: principal_name
+      in: query
+      schema:
+        $ref: './../schemas/api.params.predicate.filter.string.yaml'
```
mistahj67: Missed the swap to email I think here. Friendly reminder to generate docs after this change so the json above is also updated!

cmd/api/src/api/v2/auth/list_users_minimal.go
mistahj67, line 1: suggest: let's rename to just `users.go` and hopefully in time more user related auth functions will find there way to this file.

cmd/api/src/api/v2/auth/users.go
```diff
@@ -0,0 +1,134 @@
...
+func (s ManagementResource) ListActiveUsersMinimal(response http.ResponseWriter, request *http.Request) {
+	var (
+		usersFilter                UserMinimal
+		queryParams                = request.URL.Query()
+		queryParameterFilterParser = model.NewQueryParameterFilterParser()
+	)
+
+	if orderBy, err := api.ParseSortParameters(UserMinimal{}, queryParams); err != nil {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+	} else if order, err := api.BuildSQLSort(orderBy, model.SortItem{Column: "email"}); err != nil {
```
mistahj67:
```suggestion
	} else if order, err := api.BuildSQLSort(orderBy, model.SortItem{Column: "email", Direction: model.AscendingSortDirection}); err != nil {
```

cmd/api/src/database/auth.go
mistahj67: This can be deleted now I believe

cmd/api/src/database/mocks/db.go
```diff
@@ -890,6 +890,21 @@ func (mr *MockDatabaseMockRecorder) GetAggregatedSelectorNodesCertification(ctx,
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetAggregatedSelectorNodesCertification", reflect.TypeOf((*MockDatabase)(nil).GetAggregatedSelectorNodesCertification), ctx, sqlFilter, skip, limit)
 }
 
+// GetAllActiveUsers mocks base method.
+func (m *MockDatabase) GetAllActiveUsers(ctx context.Context, order string, filter model.SQLFilter) (model.Users, error) {
```
mistahj67: Same with this as well. PFC should catch it

cmd/api/src/migrations/manifest.go
```diff
@@ -24,15 +24,16 @@ import (
 	"strings"
 	"time"
 
+	"github.com/specterops/dawgs/graph"
+	"github.com/specterops/dawgs/ops"
+	"github.com/specterops/dawgs/query"
+
```
mistahj67: heads up, not blocking but this is not needed and technically deviates from the golang convention here. I suspect your ide is fighting you. I've see these in a few places on these PRs

cmd/api/src/database/db.go
```diff
@@ -26,15 +26,16 @@ import (
 	"time"
 
 	"github.com/gofrs/uuid"
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+
```
mistahj67: same unnecessary change

cmd/api/src/database/auth_test.go
```diff
@@ -24,14 +24,15 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
```
mistahj67: same unnecessary change

cmd/api/src/api/sort_test.go
```diff
@@ -17,14 +17,17 @@
...
 	"net/url"
 	"testing"
 
-	"github.com/specterops/bloodhound/cmd/api/src/api"
-	"github.com/specterops/bloodhound/cmd/api/src/model"
 	"github.com/specterops/dawgs/cypher/models/cypher"
 	"github.com/specterops/dawgs/query"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
```
mistahj67, line 26:
```suggestion
```

cmd/api/src/api/sort.go
```diff
@@ -21,8 +21,9 @@ import (
 	"fmt"
 	"net/url"
 
-	"github.com/specterops/bloodhound/cmd/api/src/model"
 	"github.com/specterops/dawgs/query"
+
```
mistahj67:
```suggestion
```

PR #1901, opened 2025-09-17
Bed 6063 ha dummy

brandonshearin:
make changelog HA aware. CE implementation always reports as primary

non-primary instances will skip these protected codepaths:
- cache enable/disable
- feature flag polling
- clearCache()

### DIFF
--- a/cmd/api/src/daemons/changelog/flag.go
+++ b/cmd/api/src/daemons/changelog/flag.go
@@ -22,32 +22,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/specterops/bloodhound/cmd/api/src/daemons/ha"
 	"github.com/specterops/bloodhound/cmd/api/src/model/appcfg"
 	"github.com/specterops/dawgs/graph"
 )
-
-type LockResult struct {
-	Context   context.Context
-	IsPrimary bool
-}
-
-type HAMutex interface {
-	TryLock() (LockResult, error)
-}
-
-// dummyHA is a no-op implementation for BHCE that always reports as primary
-type dummyHA struct{}
-
-func (d *dummyHA) TryLock() (LockResult, error) {
-	return LockResult{
-		Context:   context.Background(),
-		IsPrimary: true,
-	}, nil
-}
-
-func newDummyHA() HAMutex {
-	return &dummyHA{}
-}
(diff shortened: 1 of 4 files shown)


cweidenkeller (approved):

cmd/api/src/daemons/changelog/flag.go
```diff
@@ -26,19 +26,66 @@ import (
 	"github.com/specterops/dawgs/graph"
 )
 
+type LockResult struct {
```
zinic: These types do not belong in the `changelog` package in my opinion. Please create an adjunct package `ha` to `changelog` in the `daemons` package similar to organization in BHE. Import stutter may occur in BHE which will require import aliasing but that's a lower cost compared to the cognitive dissonance having HA related types colocated in `changelog`.

cmd/api/src/daemons/changelog/flag.go
```diff
@@ -26,19 +26,66 @@ import (
...
 	mu    sync.RWMutex
 	cache *cache
+
+	haMutex HAMutex
 }
 
-func newFeatureFlagManager(flagGetter func(context.Context) (bool, int, error), pollInterval time.Duration) *featureFlagManager {
+// isPrimary checks if this instance is the primary API instance using the HA mutex.
+// Returns true and the primary context if this instance is primary, false otherwise.
+func (s *featureFlagManager) isPrimary(ctx context.Context) (bool, context.Context) {
```
zinic, line 43: This feels a little strange. Why have this wire-up here?

zinic (changes requested):
I have some minor design concerns around where these types live and some functionality threaded through the `featureFlagManager`.

cmd/api/src/daemons/changelog/flag.go
zinic: Nope, never mind. I'm not reading good at code. Disregard this comment.

cmd/api/src/daemons/changelog/flag.go
superlinkx: Agreed, it's likely other daemons might need this wire up in the future. Nothing else stood out to me as must fix right now, we can look at additional refactor work after release

PR #1981, opened 2025-10-17
BED-6631: RODC Default Selector Broken Due To Incorrect Cypher

ALCooper12:
## Description

- Altered v8.3.0 migration file to use the correct attribute name `n.isreadonlydc = true` for the cypher query of the RO-DC default selector

## Motivation and Context

This PR addresses:
- BED-6631

- At the moment, the default selector for Read-Only DCs does not return objects because the cypher query references the attribute incorrectly as `isReadOnlyDC`. The attribute in BloodHound is case-sensitive and is actually stored as `isreadonlydc`. This fix will allow results to be returned when running the cypher query

## How Has This Been Tested?

- Within my local env and the test env _(still working on this)*_

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added a default "Read‑Only DCs" selector so read‑only data centers are automatically identified for grouping and filtering.

* **Chores**
  * Seeded the default selector with the query used to populate it during setup (no changes to existing selectors).
  * Minor formatting adjustment in migration scripts (no behavioral change).

### DIFF
--- a/cmd/api/src/database/migration/migrations/v8.3.0.sql
+++ b/cmd/api/src/database/migration/migrations/v8.3.0.sql
@@ -124,7 +124,6 @@ WHERE agts.id = duplicate_selectors.id AND duplicate_selectors.rowNumber > 1;
 ALTER TABLE IF EXISTS asset_group_tag_selectors DROP CONSTRAINT IF EXISTS asset_group_tag_selectors_unique_name_asset_group_tag;
 ALTER TABLE IF EXISTS asset_group_tag_selectors ADD CONSTRAINT asset_group_tag_selectors_unique_name_asset_group_tag UNIQUE ("name",asset_group_tag_id,is_default);
 
-
 -- Fix naming inconsistencies for ETAC
 ALTER TABLE IF EXISTS environment_access_control
     RENAME TO environment_targeted_access_control;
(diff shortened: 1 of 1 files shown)


cmd/api/src/database/migration/migrations/v8.3.0.sql
```diff
@@ -120,3 +120,47 @@ WHERE agts.id = duplicate_selectors.id AND duplicate_selectors.rowNumber > 1;
 ALTER TABLE IF EXISTS asset_group_tag_selectors DROP CONSTRAINT IF EXISTS asset_group_tag_selectors_unique_name_asset_group_tag;
 ALTER TABLE IF EXISTS asset_group_tag_selectors ADD CONSTRAINT asset_group_tag_selectors_unique_name_asset_group_tag UNIQUE ("name",asset_group_tag_id,is_default);
 
+-- Update RO-DC default selector within Tier Zero to use the correct attribute name
```
sirisjo, line 136: **question** (non-blocking): for my own understanding, why are we re-inserting the entire selector into the database here, rather than updating the cypher of the existing selector?

sirisjo: Oh I see, it only inserts it if it doesn't already exist. Nice!

ALCooper12: Actually you're correct. I was looking into other ways to reduce all that, but that was the only way I saw fit at the moment. Hopefully one of our SQL experts @computator has some suggestions for it too :eyes:

cmd/api/src/database/migration/migrations/v8.3.0.sql
```diff
@@ -120,3 +120,47 @@ WHERE agts.id = duplicate_selectors.id AND duplicate_selectors.rowNumber > 1;
...
+WITH src_data AS (
```
mistahj67: This selector already exists everywhere, so this should be an update statement just updating the cypher where name = 'Read-Only DCs' and `is_default` = true

mistahj67 (approved):
Tested locally, works. 🚀

PR #1992, opened 2025-10-22
feat: BED-5237 certifications

Holocraft:
## Description

This feature adds the Certification view for Privilege Zones.

## Motivation and Context

Resolves [<BED-5237>](https://specterops.atlassian.net/browse/BED-5237)

## How Has This Been Tested?

Unit tests were added as well as manual testing

## Screenshots (optional):
Screenshot 2025-10-22 at 9 25 10 AM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * New Certifications section under Privilege Zones with list, details panel, filters, and confirm dialog
  * Bulk certify and revoke members with optional notes; actions persist

* **UI Improvements**
  * Certification status icon added
  * Enhanced dropdown selector behavior and styling (variants, tooltips, alignment, caret)
  * New advanced filter dialog (type, approver, date range) for certifications

* **Tests**
  * Added comprehensive tests for dropdown and certification flows

### DIFF
--- a/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
@@ -185,7 +185,7 @@ const FilterDialog: FC<FilterDialogProps> = ({ filters, onApplyFilters }) => {
                                                                 <SelectItem
                                                                     key={user.id}
                                                                     value={user.email_address || user.id}>
-                                                                    {user.email_address || user.principal_name}
+                                                                    {user.email_address}
                                                                 </SelectItem>
                                                             );
                                                         })}


packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,296 @@
...
+                                                    <SelectValue placeholder='Select Object Type' />
+                                                </SelectTrigger>
+                                            </FormControl>
+                                            <SelectPortal>
+                                                <SelectContent>
+                                                    {uniqueObjects.map((obj) => (
+                                                        <SelectItem key={obj.primary_kind} value={obj.primary_kind}>
+                                                            {obj.primary_kind}
+                                                        </SelectItem>
+                                                    ))}
```
urangel: Let's use `ActiveDirectoryNodeKind` and `AzureNodeKind` from graphSchema.ts for this so that the filter options are independent from the results set.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,296 @@
...
+                <DialogContent>
+                    <Form {...form}>
+                        <form className='flex flex-col gap-4'>
+                            <DialogTitle className='flex justify-between items-center'>
+                                <span className='text-xl'>Additional Filters</span>
+                                <Button
+                                    variant='text'
+                                    onClick={() => {
+                                        form.reset({});
+                                        setFilters(defaultValues);
```
urangel: Using `setFilters` here will apply the filters before clicking the confirm button on the filter dialog. I believe clearing the selected values within the dialog and applying the filters should be independent actions.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,296 @@
...
+                            <VisuallyHidden asChild>
+                                <DialogDescription>Filter Privilege Zone Certifications</DialogDescription>
+                            </VisuallyHidden>
+                            <FormField
+                                name='objectType'
+                                control={form.control}
+                                render={({ field }) => (
+                                    <FormItem>
+                                        <FormLabel>Object Type</FormLabel>
+                                        <Select onValueChange={field.onChange} value={field.value || ''}>
```
urangel:
```suggestion
                                        <Select onValueChange={field.onChange} value={field.value}>
```

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,296 @@
...
+                                    </FormItem>
+                                )}
+                            />
+                            <FormField
+                                name='approvedBy'
+                                control={form.control}
+                                render={({ field }) => (
+                                    <FormItem>
+                                        <FormLabel>Approved By</FormLabel>
+                                        <Select onValueChange={field.onChange} value={field.value || ''}>
```
urangel:
```suggestion
                                        <Select onValueChange={field.onChange} value={field.value}>
```

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+        }),
+        columnHelper.accessor('name', {
+            header: 'Member Name',
+            cell: (info) => <div className='min-w-0 w-[150px] truncate'>{info.getValue()}</div>,
+            size: 150,
+        }),
+        columnHelper.accessor('domainName', {
+            header: 'Domain',
+            cell: (info) => <div className='min-w-0 w-[150px] truncate'>{info.getValue()}</div>,
+        }),
```
urangel: We should rename this to Environment instead of Domain.

packages/javascript/bh-shared-ui/src/components/DropdownSelector/DropdownSelector.tsx
```diff
@@ -14,91 +14,110 @@
...
+                                                    {option.value}
+                                                </span>
+                                            </TooltipTrigger>
+                                            <TooltipPortal>
+                                                <TooltipContent side='left' className='dark:bg-neutral-dark-5 border-0'>
+                                                    <span className='uppercase'>{option.value}</span>
+                                                </TooltipContent>
+                                            </TooltipPortal>
+                                        </TooltipRoot>
+                                    </TooltipProvider>
```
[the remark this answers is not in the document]
benwaples: @Holocraft This suggestion seems reasonable IMO! But I havent tested it and Im not sure of any side effects

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+        columnHelper.accessor('zoneName', {
+            header: 'Zone',
+            cell: (info) => <div className='min-w-0 w-[150px] truncate'>{info.getValue()}</div>,
+        }),
+        columnHelper.accessor('date', {
+            header: 'First Seen',
+            cell: (info) => <div className='text-left'>{info.getValue()}</div>,
```
urangel, line 241: The date for this column should be in the format yyyy-MM-dd

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
```diff
@@ -0,0 +1,231 @@
...
+    const PAGE_SIZE = 15;
+
+    const useAssetGroupTagsCertificationsQuery = (filters?: AssetGroupTagCertificationParams) => {
+        return useInfiniteQuery<{
+            count: number;
+            data: { members: AssetGroupTagCertificationRecord[] };
+            limit: number;
+            skip: number;
+        }>({
+            queryKey: ['certifications', filters],
```
benwaples: a bit nit picky: it looks like we reuse this below to invalidate the query when needed? if thats the case, perhaps a common function/util would help us in the future not have to remember to change both!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
```diff
@@ -0,0 +1,231 @@
...
+                `zone-management_update-certification_success`,
+                {
+                    anchorOrigin: { vertical: 'top', horizontal: 'right' },
+                }
+            );
+            setSelectedRows([]);
+            queryClient.invalidateQueries({ queryKey: ['certifications', filters] });
+        },
+        onError: (error: any) => {
+            console.log(error);
```
benwaples: Perhaps a wee console.error instead?

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+                  if (filters.endDate && DateTime.fromISO(item.created_at) > DateTime.fromISO(filters.endDate))
+                      return false;
+                  if (search) {
+                      const query = search.toLowerCase();
+                      return (
+                          item.name?.toLowerCase().includes(query) || item.primary_kind?.toLowerCase().includes(query)
+                      );
+                  }
+                  return true;
+              })
```
urangel: These filters should be sent to the API instead of applying them here. That way we can get a total count for the items with the applied filters so that the scroll loading logic works as intended.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+            size: 65,
+        }),
+        columnHelper.accessor('primary_kind', {
+            header: 'Type',
+            cell: (info) => <div className='text-primary dark:text-secondary-variant-2'>{info.getValue()}</div>,
+            size: 100,
+        }),
+        columnHelper.accessor('name', {
+            header: 'Member Name',
+            cell: (info) => <div className='min-w-0 w-[150px] truncate'>{info.getValue()}</div>,
```
urangel: We can remove the widths for these columns and still have the truncation. That'll give the text more room to display before the ellipsis comes in.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,296 @@
...
+                            <DialogActions>
+                                <DialogClose asChild>
+                                    <Button variant='text'>Cancel</Button>
+                                </DialogClose>
+                                <DialogClose asChild>
+                                    <Button variant='text' onClick={handleConfirm}>
+                                        Confirm
+                                    </Button>
+                                </DialogClose>
+                            </DialogActions>
```
urangel, line 312: Wrapping the Confirm button in <DialogClose> makes it so that the date validation logic is not enforced before closing the filter dialog. DialogClose should be removed and the handleConfirm should handle the close of the dialog but only if the validation logic checks out.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,306 @@
...
+                        onChange={(event) => toggleRow(event.target.checked, info.row.original.id)}
+                    />
+                </div>
+            ),
+            size: 65,
+        }),
+        columnHelper.accessor('primary_kind', {
+            header: 'Type',
+            cell: (info) => <div className='text-primary dark:text-secondary-variant-2'>{info.getValue()}</div>,
+            size: 100,
```
urangel: It looks like we want the node icon here instead of text.

packages/javascript/js-client-library/src/client.ts
```diff
@@ -276,6 +278,38 @@ class BHEAPIClient {
...
+        );
+    };
+
+    searchAssetGroupTagsCertifications = (limit: number, skip: number, body: any) =>
+        this.baseClient.post<AssetGroupTagsCertification>(`/api/v2/asset-group-tags/certifications`, body, {
+            params: {
+                skip,
+                limit,
+            },
+        });
```
urangel: Since there isn't currently a search endpoint for certifications is this handler essentially the same as `updateAssetGroupTagCertification` ? We might only need the latter until a search endpoint is created.

packages/javascript/bh-shared-ui/src/hooks/useAssetGroupTags/useAssetGroupTags.tsx
```diff
@@ -233,6 +235,29 @@ export const getAssetGroupTagMembers = (
...
+                PAGE_SIZE,
+                sortOrder === 'asc' ? 'name' : '-name',
+                environments,
+                { signal }
+            );
+            return response.data.data.members;
+        },
+        enabled: !!tagId,
+    });
+};
```
urangel: It doesn't look like `useAssetGroupTagMembers` is being used anywhere. We can remove it if so.

packages/javascript/bh-shared-ui/src/hooks/useAssetGroupTags/useAssetGroupTags.tsx
```diff
@@ -79,6 +79,8 @@ export const privilegeZonesKeys = {
     members: () => [...privilegeZonesKeys.all, 'members'] as const,
     membersByTag: (tagId: string | number, sortOrder: SortOrder, environments: string[] = []) =>
         [...privilegeZonesKeys.members(), 'tag', tagId, sortOrder, ...environments] as const,
+    membersByTagList: (tagId: string | number, sortOrder: SortOrder, environments: string[] = []) =>
+        [...privilegeZonesKeys.members(), 'tag', tagId, 'list', sortOrder, ...environments] as const,
```
urangel: Since `useAssetGroupTagMembers` is the only function that uses this query key generator and `useAssetGroupTagMembers` isn't being used we should be able to remove both of these.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
urangel: fixed [here](https://github.com/SpecterOps/BloodHound/pull/1992/commits/60742fee3080851543278b0511439762d9c0d8dd) so marking this as resolved

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
```diff
@@ -0,0 +1,238 @@
...
+                        setDropdownSelection={setDropdownSelection}
+                    />
+                </div>
+                <div className='basis-1/3'>
+                    <div className='w-[400px] max-w-[400px]'>
+                        {selectedNode ? (
+                            <EntityInfoPanel DataTable={EntityInfoDataTable} selectedNode={selectedNode} />
+                        ) : (
+                            <EntityInfoPanel DataTable={EntityInfoDataTable} selectedNode={null} />
+                        )}
```
urangel: The `selectedNode` information should be kept track of in the URL state so that one can use the forward/back browser functionality as expected. 

Currently, because the `selectedNode` state is only internal to the component, the URL history will still have entries for what entity panel section was expanded/collapsed but the panel will not open with that entity and it looks like no-op back/forward navigation.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,331 @@
...
+    };
+
+    const columnHelper = createColumnHelper<any>();
+
+    const columns = [
+        columnHelper.display({
+            id: 'bulk-certify',
+            header: () => (
+                <div className='pl-8'>
+                    <input
```
urangel: We should use the Checkbox component from DoodleUI so that these checkboxes match our design system.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { CustomRangeError, END_DATE, LuxonFormat, START_DATE } from '../../../utils';
+
+type FilterFormValues = Partial<ExtendedCertificationFilters> & {
+    'start-date'?: string;
+    'end-date'?: string;
+};
+
+interface FilterDialogProps {
+    filters: Partial<ExtendedCertificationFilters>;
+    setFilters: (filters: ExtendedCertificationFilters) => void;
```
urangel: `setFilters` can be removed from this interface now that it is not being passed in

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
Holocraft: It looks like the truncation will only apply if there are defined widths

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
```diff
@@ -0,0 +1,241 @@
...
+    const [dropdownSelection, setDropdownSelection] = useState('Pending');
+    const [certifyAction, setCertifyAction] = useState<typeof CertificationRevoked | typeof CertificationManual>(
+        CertificationManual
+    );
+
+    const queryClient = useQueryClient();
+
+    const { addNotification } = useNotifications();
+
+    const PAGE_SIZE = 15;
```
mistahj67: can we bump page size to 50 / 100?

slokie-so:
Reviewing the mock, are we able to get the selectors list on the object?
Screenshot 2025-10-22 182140

Code exists in the details page on the end of the object panel

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,293 @@
...
+}
+
+const defaultValues: FilterFormValues = { objectType: '', approvedBy: '', 'start-date': '', 'end-date': '' };
+
+const toDate = DateTime.local().toJSDate();
+const fromDate = DateTime.fromJSDate(toDate).minus({ years: 1 }).toJSDate();
+
+//TODO: we need to consolidate this into one universal shared component but separating in the interest of time
+const FilterDialog: FC<FilterDialogProps> = ({ filters, onApplyFilters }) => {
+    const bloodHoundUsersQuery = useBloodHoundUsers();
```
benwaples: Looks like we need to swap useBloodHoundUsers with useGetUsersMinmal!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
```diff
@@ -0,0 +1,286 @@
...
+                            onClick={() => showDialog(CertificationManual)}
+                            disabled={dropdownSelection === 'Automatic Certification'}>
+                            Certify
+                        </Button>
+                        <Button
+                            variant='secondary'
+                            onClick={() => showDialog(CertificationRevoked)}
+                            disabled={dropdownSelection === 'Automatic Certification'}>
+                            Revoke
+                        </Button>
```
benwaples: Subjectively I prefer action buttons to be disabled if I cant perform the action, and we do that in some parts of the app but not others. This seems like an easy enough check in this case `disabled={isAutomaticCertification || selectedRows.length === 0}`, but again this is my subjective opinion and could deviate from existing patterns in PZ

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
```diff
@@ -0,0 +1,286 @@
...
+                        selectedRows={selectedRows}
+                        setSelectedRows={setSelectedRows}
+                        dropdownSelection={dropdownSelection}
+                        setDropdownSelection={setDropdownSelection}
+                    />
+                </div>
+                <div className='basis-1/3'>
+                    <div className='w-[400px] max-w-[400px]'>
+                        {memberQuery.data && selectedNode ? (
+                            <EntityInfoPanel
```
benwaples, line 274: We dont need to block on this comment, but I wanted to bring attention to it! The entity panel can really extend the page if an accordion with lots of entries is opened. I would suggest restricting the panel height to only extend as far as the table!

packages/javascript/bh-shared-ui/src/components/EntityInfo/EntityInfoDataTableList.test.tsx
```diff
@@ -65,8 +66,8 @@ const EntityInfoContentWithProvider = ({
     nodeType: EntityKinds | string;
     databaseId?: string;
     additionalTables?: {
-        sectionProps: EntityInfoDataTableProps;
-        TableComponent: React.FC<EntityInfoDataTableProps>;
+        sectionProps: any;
+        TableComponent: React.FC<any>;
```
mistahj67, line 70: Future @urangel I didn't have a better solution here. I think there might be a way to have the child props be generically defined here but due to time, popped these to any 😢

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,281 @@
...
+    applyAdvancedFilters,
+    selectedRows,
+    setSelectedRows,
+    dropdownSelection,
+    setDropdownSelection,
+}) => {
+    const scrollRef = useRef<HTMLDivElement>(null);
+
+    const { data: availableEnvironments = [] } = useAvailableEnvironments();
+    const { data: assetGroupTags = [] } = useAssetGroupTags();
```
benwaples, line 98: If the empty array is used as the initial data for `availableEnvironment` and `assetGroupTags`, there is also a config for react-query called `initialData` that we could use! 

I bring this up because Ive seen problems in the past related to the top level component rerendering and effectively redefining the fallback which causes dependency arrays listening to changes to that value to retrigger :/ And I notice these to values used in a dep arrays bellow. Ive also seen it cause infinite loops if the query never returns data but I dont think that would be the case here.

It would mean that we would need to plumb optional query options through `useAssetGroupTags`, but we have a pattern established for that:
```
export const useAssetGroupTags = (options?: GenericQueryOptions<AssetGroupTag[]>) => {
    // stuff that already exists
    return useQuery({
        // stuff that already exists
        ...options,
    });
};
```

benwaples: Looking at this and we would need to update the GenericQueryOptions to allow `readonly string[]` for queryKeys (the last param in the `UseQueryOptions` generic)

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
Holocraft: Fixed!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
Holocraft: Might have to punt this one!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/Certification.tsx
Holocraft: Punting!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,286 @@
...
+        } else {
+            setSelectedRows([]);
+        }
+    };
+
+    const toggleRow = (checked: boolean, selectedId: number) => {
+        const newSelectedRows = checked
+            ? [...selectedRows, selectedId]
+            : selectedRows.filter((rowId: number) => rowId !== selectedId);
+        setSelectedRows(newSelectedRows);
```
benwaples: id recommend using the set state callback to make sure the state and ui are always in sync!
```suggestion
        setSelectedRows(prev => {
        const newSelectedRows =  checked
            ? [...prev, selectedId]
            : prev.filter((rowId: number) => rowId !== selectedId);
            
        if (newSelectedRows...) ...
            
        return newSelectedRows
        })
```

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/CertificationTable.tsx
```diff
@@ -0,0 +1,286 @@
...
+                    fetchNextPage();
+                }
+            }
+        },
+        [fetchNextPage, isFetching, totalFetched, count]
+    );
+
+    useEffect(() => {
+        fetchMoreOnBottomReached(scrollRef.current);
+    }, [fetchMoreOnBottomReached]);
```
benwaples, line 150: Just curious, in what cases does this useEffect run?

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Certification/FilterDialog.tsx
```diff
@@ -0,0 +1,321 @@
...
+                                                </FormControl>
+                                                <SelectPortal>
+                                                    <SelectContent>
+                                                        {bloodHoundUsersQuery.data?.map((user) => {
+                                                            if (!user.email_address) return null;
+                                                            return (
+                                                                <SelectItem
+                                                                    key={user.id}
+                                                                    value={user.email_address || user.id}>
+                                                                    {user.email_address || user.principal_name}
```
benwaples: I think we can safely remove the `|| user.principal_name` fallback because we have the check above!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PrivilegeZones.tsx
```diff
@@ -95,7 +102,7 @@ const PrivilegeZones: FC = () => {
         );
     }
 
-    const tabValue = isHistoryPage ? historyPath : tagType;
+    const tabValue = isCertificationsPage ? certificationsPath : isHistoryPage ? historyPath : tagType;
```
benwaples, line 105: also not a blocker but nested terinarys are hard to parse as a human and Id recommend breaking out these into separate variables or even a switch statement

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PrivilegeZones.tsx
```diff
@@ -112,6 +119,11 @@ const PrivilegeZones: FC = () => {
                         value={tabValue}
                         className={cn('w-full mt-4', { hidden: location.pathname.includes(savePath) })}
                         onValueChange={(value) => {
+                            if (value === certificationsPath) {
+                                return navigate(`/${privilegeZonesPath}/${certificationsPath}`, {
+                                    discardQueryParams: true,
```
benwaples, line 124: this will also remove the selected environmentId that gets used on various pages outside of CE. Is that desired? Or are there specific params that we want to remove that we can target directly? Or add the environmentId to the url string as a query param?

Holocraft: Certifications page will not have any query params used.

benwaples: never mind looks like the history path behaves the same! we can disregard for now

PR #2012, opened 2025-10-28
Refactor: BED-6624 dynamic edit button

LucasParraF:
## Description

On Privilege Zones, the Edit button located at the right side of the screen should dynamically change depending on the view the customer is. If the customer is in "Zones", the button should read "Edit Zone". If the customer is in "Label", the button should read "Edit Label". If a "selector" is selected, the button should read "Edit Rule".

## Motivation and Context

Solves BED-6624 https://specterops.atlassian.net/browse/BED-6624

## How Has This Been Tested?

Tested locally and updated current test.

## Screenshots (optional):

https://github.com/user-attachments/assets/76350773-fb18-4a42-b06b-e2e71647bc7d

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Refactor**
  * Replaced the previous link/button edit UI in Privilege Zones with a unified Edit button rendered in the header that shows contextual labels ("Zone", "Label", "Rule") based on current context.

* **Tests**
  * Added tests covering the new Edit button across Zones, Labels, and Selectors/Rules contexts.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PZEditButton/PZEditButton.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PZEditButton/PZEditButton.tsx
@@ -59,7 +59,7 @@ export const PZEditButton: FC<{
     const titleSuffix = suffix(zoneId, labelId, selectorId);
 
     return (
-        <div className='flex flex-col gap-4 basis-1/3 w-[6.75rem]'>
+        <div className='flex flex-col gap-4 w-[6.75rem]'>
             {showEditButton && (
                 <Button asChild={!!saveLink} variant={'secondary'} disabled={!saveLink}>
                     <AppLink data-testid='privilege-zones_edit-button' to={saveLink || ''}>


packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/Details.tsx
```diff
@@ -115,6 +116,9 @@ const Details: FC = () => {
     const showEditButton = !getEditButtonState(memberId, selectorsQuery, zonesQuery, labelsQuery);
     const saveLink = getSavePath(zoneId, labelId, selectorId);
 
+    // Edit button title changes depending on tab/context
+    const editButtonTitle = selectorId ? 'Rule' : isZonePage ? 'Zone' : 'Label';
```
specter-flq: (Partially copied from BHE) Although I feel this is simple enough to have as a nested ternary, and that it works, I do have something that pops to mind.

I feel that isZonePage is not the ultimate truth in determining if it should be Zone or Label. It happens to work in this case but its not like a isDisabled to determine if to set a class or not. In this case isLabelPage would be the ultimate truth. So feel like this maybe not as descriptive or as scalable as it could be. Although again it works and its simple.

Maybe a solution can be to incorporate all into a function that has each variable that decides the value and returns "Edit {{value}}"

Finally maybe another level of abstraction, considering we do the same in BHE ( an Edit is repeated ), is to maybe have a hook there that determines de Edit Title and is used in both places. ( this could be overkill but figured I would mention it )

urangel: I like the thought in your final remark especially since like you mention the same is done in BHE. A small hook or component would be nice to encapsulate and reuse this logic.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/Details.test.tsx
```diff
@@ -205,4 +204,18 @@ describe('Details', async () => {
...
+    it('renders edit zones button when on Zones page', async () => {
+        setupComponent({ zoneId: '1' });
+
+        expect(await screen.findByText(/Edit Zone/)).toBeInTheDocument();
+    });
+
+    it('renders edit label button when on Labels page', async () => {
+        setupComponent({ labelId: '2' });
+
+        longWait(async () => {
```
urangel: I need to remove this `longWait` utility function as it is not good practice. Let's refactor this test so that we reach for something else instead.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/EditInfoPanelButton/EditInfoButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+    Label = 'Label',
+}
+
+const EditInfoButton: FC<{
+    labelId?: string | undefined;
+    selectorId?: string | undefined;
+    zoneId?: string | undefined;
+    showEditButton: boolean;
+    saveLink?: string | undefined;
+}> = ({ labelId, selectorId, zoneId, showEditButton, saveLink }) => {
```
mistahj67: Thanks for moving this into its own component. Really clean. If you used the PZPathParams hook, then passing the ids can be cleaned up as well as pull the `saveLink` function from `Details.tsx` as it's only used in the one spot. Then you only have a single prop `showEditButton` 🧼

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/EditInfoPanelButton/EditInfoButton.tsx
```diff
@@ -0,0 +1,54 @@
...
+    let titleAppend: TitleAppend | undefined;
```
mistahj67: nit: "append" feels weird here for this variable. It's a verb used as a noun. What do you think of using `TitleSuffix` or `title = 'Edit'` and then in the if / else block just actually append it with `+=`

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/Details.tsx
```diff
@@ -124,12 +123,13 @@ const Details: FC = () => {
...
-                            asChild={showEditButton || !saveLink}
-                            variant={'secondary'}
-                            disabled={showEditButton || !saveLink}>
-                            <AppLink to={saveLink || ''}>Edit</AppLink>
-                        </Button>
+                        <EditInfoButton
+                            labelId={labelId}
+                            selectorId={selectorId}
+                            zoneId={zoneId}
+                            showEditButton={showEditButton}
```
mistahj67: You can call the getEditButtonState directly and eliminate another variable 🤷 

```suggestion
                            showEditButton={!getEditButtonState(memberId, selectorsQuery, zonesQuery, labelsQuery)}
```

LucasParraF:
Hey guys! Thanks for all the feedback! I implemented the suggested changes. Please let me know if you have any other suggestions!

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PZEditButton/PZEditButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+enum TitleSuffix {
+    Selector = 'Rule',
+    Zone = 'Zone',
+    Label = 'Label',
+}
+
+export const getSavePath = (
+    zoneId: string | undefined,
+    labelId: string | undefined,
+    selectorId: string | undefined
```
specter-flq, line 32: Seems like zoneId,labelId and selectorId here and below as args in line 44 have string | undefined types, maybe we can create a `type PZId = string | undefined `and reuse it. 

Even better, but maybe out of scope, would be perfect if we can abstract this even more to maybe the usePZPathParams hook and type it through out to maybe something more specific like the actual values as types. I see alot of 'labels' and 'zones' strings that can maybe be organized into types and referenced everywhere else including here.

mistahj67: I like the idea but I think this PR has had enough fun for one day, might be worth tossing a service continuity ticket up to research!

specter-flq: Sounds fair, I will go ahead and write up the ticket.

PR #2015, opened 2025-10-29
BED-6418: Completeness Metrics Update

maffkipp:
## Description
This PR updates our rules for determining which users or computers to consider "active" when calculating Local Group Completeness and Session Completeness metrics. Both should now use the following criteria:

- user or computer must have the `enabled` property set to `true`
- user or computer must have a `lastlogontimestamp` property within the last 14 days

## Motivation and Context

Resolves BED-6418

Previously, we were not checking the `enabled` property and we had a expiration window of 90 days, which is too long to provide helpful results and makes this metric appear lower than needed.

## How Has This Been Tested?
- Updated testing harness to check the new behavior works as expected

## Types of changes

## Checklist:
  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated activity detection logic to measure system activity based on logon timestamps instead of password change dates for improved accuracy.
  * Reduced activity threshold from 90 to 14 days to better identify inactive systems.
  * Enhanced completeness reports to exclude disabled systems and require valid logon data.


ALCooper12:
Just wondering, but why doesn't  CComputerB(Base Computer) and CGroup(Base Group) not have `enabled: true` or `enabled: false` like the other nodes on the completenesssharness.svg file?

maffkipp:
@ALCooper12 CComputerA and CComputerB are both intended as different kinds of negative cases and should not be included in the calculation for Local Group Completeness, since they do not meet the requirements for both the `enabled` property and the `lastlogontimestamp` property. Similarly, we don't check those properties for Group nodes during these calculations and so don't need to set them for CGroup.

ALCooper12 (approved):
Thanks for clarifying! LGTM 👍🏾

packages/go/analysis/ad/queries.go
```diff
@@ -1526,20 +1527,20 @@ func FetchLocalGroupCompleteness(tx graph.Transaction, domainSIDs ...string) (fl
 	})); err != nil {
 		return completeness, err
 	} else {
-		mostRecentPasswordLastSetTime := time.Unix(0, 0)
+		mostRecentLogonTimestamp := time.Unix(0, 0)
 
 		for _, computer := range computers {
-			if passwordLastSet, err := computer.Properties.Get(common.PasswordLastSet.String()).Time(); err != nil {
+			if lastLogonTimestamp, err := computer.Properties.Get(ad.LastLogonTimestamp.String()).Time(); err != nil {
```
urangel, line 1533: Is there any particular handling that needs to be done for possible the "never" (-1) or "unknown" (0) values of `lastlogontimestamp`?  https://github.com/SpecterOps/BloodHound/blob/main/packages/javascript/bh-shared-ui/src/utils/entityInfoDisplay.ts#L206

maffkipp: hmm that is a good question! This brings our local group completeness calculation in line with the way `FetchUserSessionCompleteness` has always worked, and so we have not historically had any special handling for those cases. It may be that since a value of `0` or `-1` will always be "before" the lower bound of our active threshold, we are just throwing out those particular computers on line 1546 (which makes sense to me that we would not want to include those in our completeness metrics).

urangel (approved):

PR #2020, opened 2025-10-31
BED-6295: ETAC filter group asset combo node

ktstrader:
## Description

Adding ETAC filtering to the GET {{baseURL}}/api/v2/asset-groups/1/combo-node endpoint
The filtering here is in the case where an environment_id isn't specified in the query parameters when making the call.

## Motivation and Context

Apart of the ETAC initiative
Resolves: [BED-6295](https://specterops.atlassian.net/browse/BED-6295)
HS test: [PQE-310](https://specterops.atlassian.net/browse/PQE-310)

## How Has This Been Tested?

Tested locally with ad_sampledata.json
Added unit tests

## Screenshots (optional):
See screenshots at https://github.com/SpecterOps/bloodhound-enterprise/pull/742

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-6295]: https://specterops.atlassian.net/browse/BED-6295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PQE-310]: https://specterops.atlassian.net/browse/PQE-310?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Asset group queries now enforce Environment Targeted Access Control (ETAC), restricting visible nodes and memberships to allowed environments.
* **Refactor**
  * Graph construction and membership expansion consistently propagate ETAC filters so environment-aware filtering is applied end-to-end.
* **API**
  * Query endpoints updated to accept environment-filter input to control ETAC-aware responses.
* **Tests**
  * Integration tests and mocks updated to verify environment-based filtering.

### DIFF
--- a/cmd/api/src/queries/graph_integration_test.go
+++ b/cmd/api/src/queries/graph_integration_test.go
@@ -330,7 +330,8 @@ func TestGetAssetGroupComboNodeETACFilter(t *testing.T) {
 	testContext.SetupActiveDirectory()
 	testContext.DatabaseTest(func(harness integration.HarnessDetails, db graph.Database) {
 
-		environmentB, _ := harness.AssetGroupComboNodeHarness.GroupB.Properties.Get(ad.DomainSID.String()).String()
+		environmentB, err := harness.AssetGroupComboNodeHarness.GroupB.Properties.Get(ad.DomainSID.String()).String()
+		require.Nil(t, err)
 
 		graphQuery := queries.NewGraphQuery(db, cache.Cache{}, config.Configuration{})
 		comboNode, err := graphQuery.GetAssetGroupComboNode(context.Background(), "", ad.AdminTierZero, []string{environmentB})


packages/go/analysis/analysis.go
```diff
@@ -127,6 +128,19 @@ func ExpandGroupMembershipPaths(tx graph.Transaction, candidates graph.NodeSet)
 				BranchQuery: func() graph.Criteria {
 					return query.Kind(query.Relationship(), ad.MemberOf)
 				},
+				PathFilter: func(ctx *ops.TraversalContext, segment *graph.PathSegment) bool {
```
mvlipka, line 131: Having a comment here as to what this is doing might be helpful
Once the ETAC feature flag is removed, that ETAC comment will be removed and it won't be evident that this is specific to ETAC

packages/go/analysis/analysis.go
```diff
@@ -127,6 +128,19 @@ func ExpandGroupMembershipPaths(tx graph.Transaction, candidates graph.NodeSet)
...
+						return true
+					}
+					if domainSid, err := segment.Node.Properties.Get(ad.DomainSID.String()).String(); err != nil {
+						return false
+					} else if !slices.Contains(etacList, domainSid) {
+						return false
+					} else {
+						return true
+					}
+				},
```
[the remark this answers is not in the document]
ktstrader: In the beginning if statement we check for `if candidate.Kinds.ContainsOneOf(ad.Group)`  so I though that we are primarily looking at nodes that are in active directory.

cmd/api/src/queries/graph.go
```diff
@@ -188,6 +193,15 @@ func (s *GraphQuery) GetAssetGroupComboNode(ctx context.Context, owningObjectID
...
 			}
 
+			// ETAC feature flag
+			// Filtering out environments user does not have access to
+			if etacEnabled && !user.AllEnvironments {
+				filters = append(filters, query.Or(
+					query.In(query.NodeProperty(string(ad.DomainSID)), etacAllowedList),
+					query.In(query.NodeProperty(string(azure.TenantID)), etacAllowedList),
+				))
+			}
```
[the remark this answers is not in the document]
ktstrader: The `if len(etacAllowedList) == 0` is a good catch but I think I'll just add that to external if statement.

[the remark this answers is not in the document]
ktstrader: We actually want to leave `&& len(etacAllowedList) > 0` out, this is because if a user has `all_environments = false` and they not assigned to an environment, then it will send back an empty list.

packages/go/analysis/analysis.go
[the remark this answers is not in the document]
ktstrader: Azure Group membership expansion is not in the scope for this PR. Also this is a deprecated endpoint and we are just covering our bases with ETAC.

cmd/api/src/queries/graph.go
```diff
@@ -178,8 +178,13 @@ func NewGraphQuery(graphDB graph.Database, cache cache.Cache, cfg config.Configu
...
 }
 
-func (s *GraphQuery) GetAssetGroupComboNode(ctx context.Context, owningObjectID string, assetGroupTag string) (map[string]any, error) {
+func (s *GraphQuery) GetAssetGroupComboNode(ctx context.Context, owningObjectID string, assetGroupTag string, etacEnabled bool, user model.User) (map[string]any, error) {
 	var graphData = map[string]any{}
+	etacAllowedList := make([]string, 0, len(user.EnvironmentTargetedAccessControl))
+
+	for _, envAccess := range user.EnvironmentTargetedAccessControl {
+		etacAllowedList = append(etacAllowedList, envAccess.EnvironmentID)
+	}
```
bsheth711: We can use the helper function `ExtractEnvironmentIDsFromUser` here

cmd/api/src/queries/graph.go
```diff
@@ -132,7 +132,7 @@ func BuildEntityQueryParams(request *http.Request, queryName string, pathDelegat
 }
 
 type Graph interface {
-	GetAssetGroupComboNode(ctx context.Context, owningObjectID string, assetGroupTag string) (map[string]any, error)
+	GetAssetGroupComboNode(ctx context.Context, owningObjectID string, assetGroupTag string, etacEnabled bool, user model.User) (map[string]any, error)
```
bsheth711: Let's make the interface instead like: 
```suggestion
	GetAssetGroupComboNode(ctx context.Context, owningObjectID string, assetGroupTag string, environmentsFilter []string) (map[string]any, error)
```
This prevents bleed of ETAC into our interface and reduces arguments. Additionally, nil or a nil slice can easily be passed in if there is no need for environments filtering

packages/go/analysis/analysis.go
```diff
@@ -116,7 +117,7 @@ func FetchEdgeByStartAndEnd(ctx context.Context, graphDB graph.Database, start,
 	})
 }
 
-func ExpandGroupMembershipPaths(tx graph.Transaction, candidates graph.NodeSet) (graph.PathSet, error) {
+func ExpandGroupMembershipPaths(tx graph.Transaction, candidates graph.NodeSet, etacEnabled bool, allEnv bool, etacList []string) (graph.PathSet, error) {
```
bsheth711: Same thing here, we can change the signature to be more like:

```suggestion
func ExpandGroupMembershipPaths(tx graph.Transaction, candidates graph.NodeSet, environmentsFilter []string) (graph.PathSet, error) {
```

cmd/api/src/queries/graph.go
ktstrader: What gets sent in the case that for the user `all_environments` is true? Is the `user.EnvironmentTargetedAccessControl` list populated with all the environments?

cmd/api/src/queries/graph_integration_test.go
```diff
@@ -310,7 +310,7 @@ func TestGetAssetGroupComboNode(t *testing.T) {
 	testContext.SetupActiveDirectory()
 	testContext.DatabaseTest(func(harness integration.HarnessDetails, db graph.Database) {
 		graphQuery := queries.NewGraphQuery(db, cache.Cache{}, config.Configuration{})
-		comboNode, err := graphQuery.GetAssetGroupComboNode(context.Background(), "", ad.AdminTierZero)
+		comboNode, err := graphQuery.GetAssetGroupComboNode(context.Background(), "", ad.AdminTierZero, nil)
```
mvlipka, line 313: We should get another test in here to verify the filter is working correctly when passing in the environment list

bsheth711:
Code wise LGTM, thank you for addressing my feedback

PR #2039, opened 2025-11-10
fix(Doodle): address many Doodle UI build issues

TheNando:
Issues addressed:
* `it.todo`
* Vite HMR version breaking change
* External deps in manual chunks
* Misc type issues

### DIFF
--- a/packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
+++ b/packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
@@ -16,9 +16,11 @@
 import { faEllipsis } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Tooltip } from '@mui/material';
-import { createColumnHelper, DataTable } from 'doodle-ui';
+import { DataTable, createColumnHelper, type ColumnDef } from 'doodle-ui';
 import isEmpty from 'lodash/isEmpty';
 import { useCallback, useMemo, useState } from 'react';
+import ExploreTableDataCell from './ExploreTableDataCell';
+import ExploreTableHeaderCell from './ExploreTableHeaderCell';
 import {
     compareForExploreTableSort,
     getExploreTableData,
@@ -26,8 +28,6 @@ import {
     type ExploreTableProps,
     type MungedTableRowWithId,
 } from './explore-table-utils';
-import ExploreTableDataCell from './ExploreTableDataCell';
-import ExploreTableHeaderCell from './ExploreTableHeaderCell';
 
 const columnHelper = createColumnHelper<MungedTableRowWithId>();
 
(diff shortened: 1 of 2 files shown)


cmd/ui/src/views/Explore/ExploreSearch/ExploreSearch.test.tsx
```diff
@@ -133,7 +133,7 @@ describe('ExploreSearch rendering per tab', async () => {
         expect(screen.getByRole('button', { name: /run/i })).toBeInTheDocument();
     });
     // To do: Work on this when TW css classes are applied in test environment
-    it.todo('should hide/expand search widget when user clicks minus/plus button', async () => {
+    it.skip('should hide/expand search widget when user clicks minus/plus button', async () => {
```
TheNando, line 136: it.todo is being used as placeholder test, but .todo only expects the first test name param, however, both the name and test implementation are provided causing an TS argument error. Switched to .skip instead

cmd/ui/package.json
```diff
@@ -64,6 +64,7 @@
     },
     "devDependencies": {
         "@faker-js/faker": "^9.0.3",
+        "@rollup/pluginutils": "^5.3.0",
```
TheNando, line 67: Doodle was using a newer version of a Vite plugin which provides HMR which conflicted with the 4.2.1 version. Because of the `^`, the package manager was resolving to the later version 4.7.0 which  depends are react-fresh 0.17.0 which introduces a breaking change. Explicitly setting the older version prevents this.

packages/javascript/bh-shared-ui/src/components/ExploreTable/useExploreTableRowsAndColumns.tsx
```diff
@@ -40,12 +40,26 @@ type UseExploreTableRowsAndColumnsProps = Pick<ExploreTableProps, 'onKebabMenuCl
     exploreTableData: ReturnType<typeof getExploreTableData>;
 };
 
+/*
+  TypeScript can't infer a portable type for hook's return value as it included
+  @tanstack/table-core types via doodle-ui. Explicitly importing ColumnDef from
+  doodle-ui and creating a named return type makes this portable and resolvable
+  without a direct reference to the internal @tanstack/table-core module
+*/
+type UseExploreTableRowsAndColumnsReturn = {
```
TheNando, line 49: Comment explains issue. ^

packages/javascript/bh-shared-ui/src/components/UpsertSSOProviders/SSOProviderConfigForm.tsx
```diff
@@ -129,7 +129,7 @@ const SSOProviderConfigForm: FC<{
                                 const output = parseInt(e.target.value as string, 10);
                                 field.onChange(isNaN(output) ? 1 : output);
                             }}
-                            value={isNaN(field.value) ? '' : field.value.toString()}
+                            value={!field.value || isNaN(field.value) ? '' : field.value.toString()}
```
TheNando, line 132: `isNaN` expects a number type but field.value may be undefined.

packages/javascript/doodle-ui/vite.config.ts
```diff
@@ -39,7 +39,9 @@ export default defineConfig({
             external: ['react', 'react-dom', 'react/jsx-runtime', 'tailwindcss'],
             output: {
                 manualChunks: {
-                    vendor: Object.keys(packageJson.dependencies),
+                    vendor: Object.keys(packageJson.dependencies).filter(
```
TheNando, line 42: These deps were listed as external just a few lines up. This tells Vite to handles them differently and exclude from output. However, they were also being passed into the manual vendor chunks for optimization. This causes a contradiction that Vite doesn't know how to handle.

package.json
```diff
@@ -31,15 +31,17 @@
     "typescript": "^5.1.6"
   },
   "resolutions": {
+    "@testing-library/jest-dom": "^5.16.5",
```
TheNando, line 34: After updating most deps to match with Doodle's newer versions, `react-window` was not resolving to require React v19 and we are still on 18. React Window is only used in bh-shared-ui, not doodle-ui, so we did not see this issue until merging Doodle into the monorepo. These `@types/react[-dom]`resolutions fix the issue. The rest are also alpha sorted.

PR #2061, opened 2025-11-14
BED-6717: OpenGraph nodes are searchable on the explore page

sirisjo:
## Description

Currently, OpenGraph nodes are only discoverable via cypher queries on the explore page. This change enables users to search by `name` or `objectid` in the search box, and OpenGraph results will be included in the search results. 

This change also includes a new feature flag, `opengraph_search`, to gate this behavior.

## Motivation and Context

Resolves BED-6717

Resolves [BED-6886](https://specterops.atlassian.net/browse/BED-6886) (endpoint now supports opengraph nodes)

This change enables users to interact with their OpenGraph data in an additional way -- via the search box! 

## How Has This Been Tested?

- [x] Tests added
- [x] Existing tests updated / migrated off of the deprecated testing framework
- [x] Testing functionality locally in the UI

## Screenshots (optional):
`Bob` is one of my opengraph nodes -- as you can see, querying `b` pulls this node up, and when I click on the node, I can see all of its properties. 

https://github.com/user-attachments/assets/c48da53c-723d-4be9-ac8f-a04571ef9a82

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Search now matches by name or object ID.
  * New OpenGraph Search feature flag added (defaults disabled).

* **Improvements**
  * Consolidated and improved search result aggregation and open-graph inclusion behavior.
  * Adjusted default kind parsing when no kinds are provided.

* **Tests**
  * Expanded unit and integration tests and added fixtures and data-driven search scenarios.

* **Chores**
  * Database migration added to register the new feature flag.

[BED-6886]: https://specterops.atlassian.net/browse/BED-6886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cmd/api/src/database/migration/migrations/v8.5.0.sql
+++ b/cmd/api/src/database/migration/migrations/v8.5.0.sql
@@ -1,3 +1,18 @@
+-- Copyright 2025 Specter Ops, Inc.
+--
+-- Licensed under the Apache License, Version 2.0
+-- you may not use this file except in compliance with the License.
+-- You may obtain a copy of the License at
+--
+--     http://www.apache.org/licenses/LICENSE-2.0
+--
+-- Unless required by applicable law or agreed to in writing, software
+-- distributed under the License is distributed on an "AS IS" BASIS,
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- SPDX-License-Identifier: Apache-2.0
 -- OpenGraph Search feature flag
 INSERT INTO feature_flags (created_at, updated_at, key, name, description, enabled, user_updatable)
 VALUES (current_timestamp,


cmd/api/src/queries/graph_integration_test.go
```diff
@@ -39,125 +42,158 @@ import (
...
-			require.Equal(t, expectedUser.Name, userWanted)
-		})
+	base, err := generic.LoadGraphFromFile(os.DirFS(testSuite.WorkDir), "base.json")
+	require.NoError(t, err)
+
+	err = generic.WriteGraphToDatabase(testSuite.GraphDB, &base)
+	require.NoError(t, err)
+	return testSuite
 }
+func TestSearchNodesByNameOrObjectId_ExactMatch(t *testing.T) {
```
superlinkx: Due to all of the queries being read only and using the same backing data, we could change all these independent tests into subtests instead.

packages/go/analysis/analysis.go
```diff
@@ -72,7 +72,7 @@ func ParseKind(rawKind string) (graph.Kind, error) {
 
 func ParseKinds(rawKinds ...string) (graph.Kinds, error) {
```
superlinkx, line 73: What a strange artifact that the Kind parsing code lives in analysis. I took a quick look at where it and the underlying ParseKind are used, and it doesn't appear that this should cause any harm. It's virtually unused by analysis (ParseKind is used by `ein` and that's the closest any of this gets to being actually analysis related). I don't see a great place to put this outside of analysis, especially since some tests rely on it, so I'm more documenting this here as something weird but that is currently benign.

cmd/api/src/queries/queries_integration_test.go
```diff
@@ -0,0 +1,149 @@
...
+		DriverName:                "pgx",
+		Host:                      environmentMap["host"],
+		Port:                      environmentMap["port"],
+		User:                      environmentMap["user"],
+		Password:                  environmentMap["password"],
+		Database:                  environmentMap["dbname"],
+		Options:                   "sslmode=disable",
+		ForceTerminateConnections: true,
+	}
+}
```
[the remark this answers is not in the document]
superlinkx: I'm fine with this given it's only used in controlled test environments where we are very selective about how we configure. Main risk would be devs changing their own local config, but they'll be able to figure it out reasonably and there's fallbacks for getting them to a valid state.

superlinkx (dismissed):
We had a miss in required behaviors. We'll need to feature flag this piece as it's not going to function properly until we fix the graph side. Feature flag requirements are being added asap to unblock this PR.

cmd/api/src/api/v2/search.go
```diff
@@ -43,15 +45,26 @@ func (s Resources) SearchHandler(response http.ResponseWriter, request *http.Req
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, "Invalid search parameter", request), response)
 	} else if skip, limit, _, err := utils.GetPageParamsForGraphQuery(context.Background(), queryParams); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("Invalid query parameter: %v", err), request), response)
-	} else if nodeKinds, err := analysis.ParseKinds(nodeTypes...); err != nil {
+	} else if openGraphSearchFeatureFlag, err := s.DB.GetFlagByKey(request.Context(), appcfg.FeatureOpenGraphSearch); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else if nodeKinds, err := getNodeKinds(context.Background(), openGraphSearchFeatureFlag.Enabled, nodeTypes...); err != nil {
```
kpowderly: we're not actually doing anything with the context here when calling getNodeKinds so should we get rid of it?

sirisjo: Good catch, will update!

cmd/api/src/queries/graph.go
```diff
@@ -545,43 +588,44 @@ func nodesToSearchResult(nodes ...*graph.Node) []model.SearchResult {
...
-						query.Kind(query.Node(), kind),
-						query.Or(
-							query.StringStartsWith(query.NodeProperty(common.Name.String()), strings.ToUpper(searchValue)),
-							query.StringStartsWith(query.NodeProperty(common.ObjectID.String()), strings.ToUpper(searchValue)),
-						),
-					)
-				}
-			})); err != nil {
-				return err
+func searchExactOrFuzzyMatchedNodes(ctx context.Context, s *GraphQuery, kind graph.Kind, searchValue string, searchType SearchType, nodes graph.NodeSet) (graph.NodeSet, error) {
```
kpowderly: You can kick an extra parameter by using GraphQuery as a receiver

```
func (s *GraphQuery) searchExactOrFuzzyMatchedNodes(ctx context.Context, kind graph.Kind, searchValue string, searchType SearchType, nodes graph.NodeSet) (graph.NodeSet, error) {
	if err := s.Graph.ReadTransaction(ctx, func(tx graph.Transaction) error {
		if fetchedNodes, err := ops.FetchNodeSet(tx.Nodes().Filterf(func() graph.Criteria {
			if searchType == SearchTypeExact {
				return query.And(createNodeSearchGraphCriteria(kind, strings.ToUpper(searchValue), false)...)
			} else {
				return query.And(createNodeStartsWithSearchGraphCriteria(kind, strings.ToUpper(searchValue))...)
			}
		})); err != nil {
			return err
		} else {
			nodes.AddSet(fetchedNodes)
			return nil
		}
	}); err != nil {
		return nil, err
	}
	return nodes, nil
}
```

cmd/api/src/queries/queries_integration_test.go
```diff
@@ -0,0 +1,149 @@
+// Copyright 2025 Specter Ops, Inc.
```
kpowderly, line 1: Since this integration test setup should serve for the whole package, I think we want this to be like `packagename_integration_test.go` e.g., `queries_integration_test.go`

cmd/api/src/api/v2/pathfinding.go
```diff
@@ -168,8 +169,9 @@ func (s *Resources) GetSearchResult(response http.ResponseWriter, request *http.
 				searchType = strings.ToLower(searchTypeParameters[0])
 			}
 		}
-
-		if nodes, err := s.GraphQuery.SearchByNameOrObjectID(request.Context(), searchValue, searchType); err != nil {
+		if openGraphSearchFeatureFlag, err := s.DB.GetFlagByKey(request.Context(), appcfg.FeatureOpenGraphSearch); err != nil {
+			api.HandleDatabaseError(request, response, err)
+		} else if nodes, err := s.GraphQuery.SearchByNameOrObjectID(request.Context(), openGraphSearchFeatureFlag.Enabled, searchValue, searchType); err != nil {
```
ddlees, line 174: Let's be sure to link BED-6720 in the PR description.

cmd/api/src/queries/graph.go
```diff
@@ -337,34 +362,52 @@ func formatSearchResults(exactResults []model.SearchResult, fuzzyResults []model
 	return searchResults[skip:end]
 }
 
-func (s *GraphQuery) SearchNodesByName(ctx context.Context, nodeKinds graph.Kinds, name string, skip int, limit int) ([]model.SearchResult, error) {
+type NodeSearchResults struct {
+	ExactResults []model.SearchResult
+	FuzzyResults []model.SearchResult
+}
+
+func (s *GraphQuery) SearchNodesByNameOrObjectId(ctx context.Context, nodeKinds graph.Kinds, nameOrObjectId string, skip int, limit int) ([]model.SearchResult, error) {
```
ddlees, line 370: NOT FOR THIS PR: for searching across two fields we should consider a trigram index for `name` and `objectid`. Not sure how this would work in Neo4J but for Postgres:

``` pgsql
CREATE INDEX node_objectid_name_trgm_idx ON nodes USING GIN (CONCAT_WS(' ', objectid, name) gin_trgm_ops)

SELECT *
FROM nodes
WHERE CONCAT_WS(' ', objectid, name) % 'some parameterized search string';
```

Consider capturing as a research spike for backlog.

sirisjo: Spike ticket created! https://specterops.atlassian.net/browse/BED-6882

PR #2091, opened 2025-11-20
feat: OG DB - List/Query Extension Entries BED-6778

AD7ZJ:
## Description

Add a database handler to list extensions with filtering, sorting, and pagination parameters. 

## Motivation and Context

Resolves https://specterops.atlassian.net/browse/BED-6778

## How Has This Been Tested?

A unit test has been added that verifies functionality of the new function. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Enhanced graph schema extension querying with advanced filtering, sorting, and pagination support.

* **Tests**
  * Added comprehensive integration tests for graph schema extension retrieval with various filter and sort scenarios.

* **Chores**
  * Improved internal error handling and SQL query construction utilities.


cmd/api/src/database/graphschema.go
```diff
@@ -78,3 +79,69 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
...
+	var extensionSqlFilterStr string
+	if extensionSqlFilter.SQLString != "" {
+		extensionSqlFilterStr = " WHERE " + extensionSqlFilter.SQLString
+	}
+
+	if len(sort) == 0 {
+		sort = append(sort, model.SortItem{Column: "id", Direction: model.AscendingSortDirection})
+	}
+
+	var sortColumns []string
```
kpowderly: Can we move this var and the vars below to the top level block? I think the reason we group vars at the top is for organization / it's easy to track where the vars are actually initialized.

cmd/api/src/database/graphschema_test.go
```diff
@@ -76,3 +76,96 @@ func TestDatabase_CreateAndGetGraphSchemaExtensions(t *testing.T) {
 	require.Error(t, err)
 	require.Equal(t, "entity not found", err.Error())
 }
+
+func TestDatabase_GetGraphSchemaExtensionsFilteredAndPaginated(t *testing.T) {
```
kpowderly: Can you update this file to be `graphschema_integration_test.go` since they're integration tests.

Then, I think we'll also want to have unit tests to cover the error cases! I know we're crunch time here so if we were told to only do integration, that's fine! 

Here's where we're missing code coverage

Image

kpowderly: Also, I think for now, we're not running integration tests in parallel (that rely on the database) because we're hitting max connection errors. There's a bug ticket to resolve that in the future.

cmd/api/src/database/graphschema.go
```diff
@@ -78,3 +79,69 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
 
 	return extension, nil
 }
+
+// GetGraphSchemaExtensionsFilteredAndPaginated gets all the rows from the extensions table that match the given SQLFilter. It returns a slice of GraphSchemaExtension structs
+// populated with the data, as well as an integer giving the total number of rows returned by the query (excluding any given pagination)
+func (s *BloodhoundDB) GetGraphSchemaExtensionsFilteredAndPaginated(ctx context.Context, extensionSqlFilter model.SQLFilter, sort model.Sort, skip, limit int) (model.GraphSchemaExtensions, int, error) {
```
kpowderly: Resolve this if you disagree but I feel like GetGraphSchemaExtensionsFilteredAndPaginated sounds like the results come pre-filtered/paginated. Would this make more sense if it was simply `GetGraphSchemaExtensions`? I think most `GET` endpoints support filtering/pagination so it just felt a little unclear.

cmd/api/src/database/graphschema.go
```diff
@@ -78,3 +79,69 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
...
+		extensionSqlFilterStr = " WHERE " + extensionSqlFilter.SQLString
+	}
+
+	if len(sort) == 0 {
+		sort = append(sort, model.SortItem{Column: "id", Direction: model.AscendingSortDirection})
+	}
+
+	var sortColumns []string
+	for _, item := range sort {
+		dirString := "ASC"
```
kpowderly: Will these directionals ASC and DESC span across all of the endpoints?

I'm just curious if we should create a helper func in `filter.go` that stringify's these values like...

```
type SortDirection int

func (s SortDirection) String() string {
    switch s {
    case AscendingSortDirection:
        return "ASC"
    case DescendingSortDirection:
        return "DESC"
    default:
        return ""
    }
}
```

It does look like you're defaulting on ASC though so you'd have to account for a fat-fingered sort value... I was just wondering if we could avoid some of these magic strings

cmd/api/src/database/graphschema.go
AD7ZJ: I think that's a good idea, this is a bit of a mouthful!

cmd/api/src/database/graphschema.go
AD7ZJ: I'm torn on this, to me it helps to see the variable declaration right where it's used so I know what I'm dealing with, especially when its use is constrained to this spot in the code.

cmd/api/src/database/graphschema.go
AD7ZJ: Yeah this pattern is used all over and will probably continue to be used more going forward.  I like your idea of that helper function!

cmd/api/src/database/graphschema_test.go
AD7ZJ: It's easy to test the first error, but not sure how to easily hit the second one. I added a test case and I'll rename the file, and get rid of t.Parallel() for now. Ty!

cmd/api/src/database/graphschema.go
```diff
@@ -27,6 +27,7 @@ import (
 type OpenGraphSchema interface {
 	CreateGraphSchemaExtension(ctx context.Context, name string, displayName string, version string) (model.GraphSchemaExtension, error)
 	GetGraphSchemaExtensionById(ctx context.Context, extensionId int32) (model.GraphSchemaExtension, error)
+	GetGraphSchemaExtensions(ctx context.Context, extensionSqlFilter model.SQLFilter, sort model.Sort, skip, limit int) (model.GraphSchemaExtensions, int, error)
```
LawsonWillard: Thoughts on ListGraphSchemaExtensions? We use List in the API layer for functions that return paginated/filtered results but we don't seem to use this prefix in the db layer so not sure if we want to introduce it here yet.

cmd/api/src/database/graphschema.go
```diff
@@ -88,6 +89,66 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
 	return extension, nil
 }
 
+// GetGraphSchemaExtensions gets all the rows from the extensions table that match the given SQLFilter. It returns a slice of GraphSchemaExtension structs
+// populated with the data, as well as an integer giving the total number of rows returned by the query (excluding any given pagination)
+func (s *BloodhoundDB) GetGraphSchemaExtensions(ctx context.Context, extensionSqlFilter model.SQLFilter, sort model.Sort, skip, limit int) (model.GraphSchemaExtensions, int, error) {
```
LawsonWillard: We've got a BuildSQLSort function but its housed in the API layer. Not sure if its worth refactoring to use it here, or push the sort back up into the API layer. I think the long term goal is to build the sort here in the DB layer but as of now, most functions are doing so in the API layer. 

https://github.com/SpecterOps/BloodHound/blob/95438edb9994ad995a8265c827bcc4e7101d215d/cmd/api/src/api/sort.go#L104

AD7ZJ: Well, I copied this pattern from PZ stuff so it's already used all over the place... my thought is just stick with it for now.

LawsonWillard (approved):

cmd/api/src/database/graphschema.go
AD7ZJ: I looked through the DB interface and most of them use `GetXyz` so I think let's leave it as get. The ones that take filter and sort parameters, there seems to be a trend of `GetAllXyz`

cmd/api/src/database/graphschema.go
superlinkx: Our guidelines say to hoist to the top unless we need to keep it localized to prevent unnecessary allocations. I'd also like to suggest that this actually needs to be pre-allocated, we should generally avoid not allocating slices. Use a reasonable starting size to avoid unnecessary allocation (probably 8 would be reasonable for sort). If you do that, then leaving it near the for loop makes sense to me

cmd/api/src/database/graphschema.go
```diff
@@ -94,6 +95,66 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
...
```
superlinkx: SQLFilter is a sin we need to put a stop to, we've just been too stretched thin to start fixing it for old stuff. I'd much rather see us start to create DB CRUD that has an explicit structure for what it expects to be able to filter/sort on and force the service layer to provide the translation. I _HATE_ SQL in the API and service layer and would like us to start avoiding it with prejudice.

Not required to merge, but I'd highly suggest we build things right while we're here. Nothing depends on this interface yet, we have carte blanche to use whatever practices provide the best development experience/follow industry practices

cmd/api/src/database/graphschema.go
```diff
@@ -94,6 +95,111 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
...
+			dirString := "ASC"
+			if !item.SortAscending {
+				dirString = "DESC"
+			}
+			sortColumns = append(sortColumns, item.Column+" "+dirString)
+		}
+		orderSqlStr = "ORDER BY " + strings.Join(sortColumns, ", ")
+	}
+	return orderSqlStr
+}
```
AD7ZJ: These should probably go somewhere else in the database package, maybe in helper.go?

cmd/api/src/model/filter.go
```diff
@@ -492,6 +492,17 @@ const (
...
+	switch s {
+	case AscendingSortDirection:
+		return "ASC"
+	case DescendingSortDirection:
+		return "DESC"
+	default:
+		return ""
+	}
+}
+
```
superlinkx: This appears to be vestigial now

cmd/api/src/database/graphschema.go
superlinkx: The helpers should not be public right now. Moving them to `helper.go` is probably fine for now.

cmd/api/src/database/graphschema.go
```diff
@@ -94,6 +95,111 @@ func (s *BloodhoundDB) GetGraphSchemaExtensionById(ctx context.Context, extensio
...
+		skipLimitString string
+		totalRowCount   int
+		orderSQL        string
+	)
+
+	extensionSqlFilterStr, extensionSqlFilterParams := DBFilterToSql(extensionFilters)
+
+	// if no sort specified, default to ID so pagination is consistent
+	if len(sorts) == 0 {
+		sorts = append(sorts, DBSort{Column: "id", SortAscending: true})
```
LawsonWillard: Not a blocker but may be nice to add this into the DB sort function and take a DBSort struct as an optional  second parameter which gets appended if column is not empty and the len of sort is 0.

cmd/api/src/model/filter.go
AD7ZJ: Yep agree, will t-can!

cmd/api/src/database/graphschema.go
AD7ZJ: I couldn't come up with a clean way to pass that into buildSqlSort(), let's just leave it as-is for now.

AD7ZJ:
> Read through `buildSQLFilter` like 4 times and still a little confused... I'm just going to trust that it's behaving properly

It is pretty much a copy/paste from BuildSQLFilter() in [model/filter.go](https://github.com/SpecterOps/BloodHound/blob/65e755824d2f822d39612b6cdc62e031502d00d7/cmd/api/src/model/filter.go#L201C1-L201C93). 
But yeah I agree it's hard to wrap your brain around these!

PR #2092, opened 2025-11-20
feat(api, ui): BED-6058/BED-6043 - Added explore_toggleable ETAC Parameter

mvlipka:
## Description

This adds a new ExploreEnabled param for ETAC which will enable and disable the Explore Page from appearing in the frontend, and hide the option to toggle this user property when creating or modifying a user

This also fixed an issue where creating/updating an admin and not sending the `AllEnvironments` property would cause the admin to be created without AllEnvironments set to true

## Motivation and Context

Resolves: BED-6058 and BED-6043

## How Has This Been Tested?

**Testing with explore_toggleable param disabled**
On a regular user:

Setting explore_enabled: false should not change the user's value
image

Setting explore_enabled when updating the user to be an admin:
Also testing not sending all_environments in the payload sets all_environments to true on the admin:
image

**Testing with explore_toggleable param enabled**
image

Setting explore_enabled to false on a user:
image

Setting explore_enabled to false on an admin:
image

Setting explore_enabled to true on an admin:
image

I manually set explore_enabled to `false` in the db to circumvent the check, then sent an update user request that did not contain explore_enabled, which should default to false for an admin:
image

Testing updating a user to have all_environments set to true but explore_enabled to false
image

**Testing New Parameter**
Toggling the param to false
image

Toggling the param to true
image

**Testing UI explore toggle**
With explore toggleable and user does have access to explore:
image

With explore toggleable and user does NOT have access to explore:
image

When all environments is set to true, you cannot enable or disable explore access:
image

When explore access is granted, all environments may not be selected
image

With explore_toggleable param set to false you may not enable or disable explore access on a user:
image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/v2/auth/auth_test.go
+++ b/cmd/api/src/api/v2/auth/auth_test.go
@@ -3100,7 +3100,7 @@ func TestManagementResource_UpdateUser_ETAC(t *testing.T) {
 				mockDB.EXPECT().GetConfigurationParameter(gomock.Any(), appcfg.EnvironmentTargetedAccessControlKey).Return(appcfg.Parameter{
 					Key: "key",
 					Value: types.JSONBObject{
-						Object: &appcfg.ScheduledAnalysisParameter{Enabled: false, RRule: "rule"},
+						Object: &appcfg.EnvironmentTargetedAccessControlParameters{ExploreToggleable: true},
 					},
 				}, nil)
 				mockDB.EXPECT().UpdateUser(gomock.Any(), gomock.Any()).Return(nil)


sothisislife101:
@mvlipka pardon if this was already discussed. "explore_enabled" and "explore_togglable" are very generic and don't make it clear that there's a specific context that these are relevant. Did we consider being more explicit with these names, like "etac_explore_enabled"?

mvlipka:
> @mvlipka pardon if this was already discussed. "explore_enabled" and "explore_togglable" are very generic and don't make it clear that there's a specific context that these are relevant. Did we consider being more explicit with these names, like "etac_explore_enabled"?

`explore_toggleable` is currently part of the `auth.environment_targeted_access_control` parameter, which should be enough of a delineation.

That's a good point on `explore_enabled`, though. In its current iteration, these features are all behind the etac feature flag, but once this feature is available to the general audience, we will be removing feature flag checking.
I've been considering moving both `AllEnvironments` and `ExploreEnabled` to a separate table after we get the MVP out, which should help with the delineation there

EDIT: Thank you for the feedback as well! Definitely something we've been considering

bsheth711:
Tickets rejected, closing PR

Issue #2096, opened 2025-11-22
Bug: dagrejs can't be fetched without SSH credentials


42LoCo42:
## Description:

The `cmd/ui` package depends on `git@github.com:BloodHoundAD/dagre.git`.
Build systems which don't use the yarn cache included in this repository, but instead fetch yarn dependencies by themselves, will fail at dagrejs when no valid SSH credentials can be found, such as when the build is running in a sandbox.

A build system that runs into this issue is Nix, for which I'm currently packaging BloodHound in https://github.com/NixOS/nixpkgs/pull/461387.

### Are you intending to fix this bug?

Yes.

## Component(s) Affected:

- Other (tooling, documentation, etc.)

## Steps to Reproduce:

1. Try to build BloodHound using Nix
2. Nix fails when fetching the dependencies

## Expected Behavior:

All dependencies, including dagrejs, will be fetched correctly.

## Actual Behavior:

Yarn is unable to fetch dagrejs from within the build sandbox.

## Screenshots/Code Snippets/Sample Files:

<img width="957" height="497" alt="Image" src="https://github.com/user-attachments/assets/d7aa75cc-4ee1-4f0b-9207-894d935fc107" />

## Environment Information:

BloodHound: 8.3.1

OS: NixOS 25.11 (Xantusia)

Node.js: 22.20.0

## Potential Solution:

Dagrejs should be fetched using an HTTPS URL instead.

## Contributor Checklist:

- [x] I have searched the issue tracker to ensure this bug hasn't been reported before or is not already being addressed.
- [x] I have provided clear steps to reproduce the issue.
- [x] I have included relevant environment information details.
- [x] I have attached necessary supporting documents.
- [ ] I have checked that any JSON files I am attempting to upload to BloodHound are valid.

wes-mil:
Hi @42LoCo42, thanks for reporting this! 

I'm not familiar with the process behind building in a sandbox for something like Nix. Would it be possible to do a git override during the build?

In the past, I've used something like this to get around SSH credentials.

```bash
git config --global url."https://github".insteadOf ssh://git@github
git config --global url."https://github.com/".insteadOf git@github.com:
```

42LoCo42:
[...]

PR #2097, opened 2025-11-22
chore(cmd/ui): fetch dagrejs via HTTPS

42LoCo42:
Build systems which don't use the yarn cache can't fetch git dependencies via ssh if no valid credentials are found, e.g. when the build is run in a sandbox.

## Description

This PR replaces the SSH URL for dagrejs in `cmd/ui/package.json` with an equivalent HTTPS URL,
thus allowing build systems to fetch dagrejs without requiring access to SSH credentials.

## Motivation and Context

Resolves https://github.com/SpecterOps/BloodHound/issues/2096

This change is required to build BloodHound with build systems that don't use the yarn cache included in this repository and also run builds in a sandbox with no access to SSH credentials.

## How Has This Been Tested?

I'm packaging BloodHound for Nix in https://github.com/NixOS/nixpkgs/pull/461387.
There, I currently use a [patch](https://github.com/42LoCo42/nixpkgs/blob/add-bloodhound-ce/pkgs/by-name/bl/bloodhound-ce/fetch-dagrejs-via-https.patch) roughly equivalent to this PR in order to fix the build issue. With it, the BloodHound package builds successfully.

On a system with Nix installed & [Flakes](https://wiki.nixos.org/wiki/Flakes) enabled, the package can be built like this:
```sh
nix build github:42loco42/nixpkgs/add-bloodhound-ce#bloodhound-ce
```

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Chores**
  * Updated project dependency configuration for improved build environment compatibility.


42LoCo42:

wes-mil:
@42LoCo42 @SuperSandro2000  Had a quick question for one of you! 

I suspect this comes from my lack of understanding of how the nix build system works, but I'm curious why our yarn cache can't be used? 

We have a cached version of the dagrejs asset [here](https://github.com/SpecterOps/BloodHound/blob/main/.yarn/cache/dagrejs-git%40github.com-b365243ec9-7c93c58f7f.zip), so we usually do not need to clone these assets.

SuperSandro2000:
> I suspect this comes from my lack of understanding of how the nix build system works, but I'm curious why our yarn cache can't be used?

We try to build as much as possible from source to fully trust the build result. The build sandbox has no internet connectivity and we use a special fetcher before who does yarn install and then copies the files into place that normal yarn commands work as expected. Those kinds of custom cache things just don't work for us. Also the cache seems to be 2 years old and I have my doubts that it works as expected in all environments. Frankly, this is the first time I am seeing something like this.

Cloning repos with ssh is also a major problem for git submodules where https must also be always used.

wes-mil:
Ended up taking a different approach with #2228

PR #2106, opened 2025-11-25
refactor: additional navigation for edit zone on Privilege Zones BED-6628

LucasParraF:
## Description

On Privilege Zones, some changes where made on UI to improve wording and Information displayed when Creating/Editing zones or labels 

## Motivation and Context

Fixes BED-6628 

## How Has This Been Tested?

Tested Locally and created new test 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added a tag-limits hook that reports remaining zones and labels.

* **UI/UX Updates**
  * Moved Delete action into the header; action bar now shows only Cancel and Save.
  * Displays remaining availability in the tag form with a "Contact sales" link when applicable.

* **Chores**
  * Added public exports for the new hook and exposed configuration keys.
  * Export surface updated to include an additional view helper.
  * Mock factory updated to accept tag type.

* **Tests**
  * Added unit tests covering zone/label limit scenarios.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
@@ -415,7 +415,7 @@ export const TagForm: FC = () => {
                                         onClick={() => {
                                             setDeleteDialogOpen(true);
                                         }}>
-                                        <span className=''>
+                                        <span>
                                             <FontAwesomeIcon icon={faTrashCan} className='mr-2' />
                                             {`Delete ${tagTypeDisplay}`}
                                         </span>
@@ -425,7 +425,7 @@ export const TagForm: FC = () => {
                             {/* Checks if Certification is truthy since it is only available on BHE and we want to display this message only on BHE */}
                             {Certification && (
                                 <CardDescription className='pb-3 pl-3'>
-                                    Currently there are <span className='font-bold'>({remainingZonesOrLabels})</span>{' '}
+                                    Currently there are (<span className='font-bold'>{remainingZonesOrLabels}</span>){' '}
                                     available {tagTypePlural}.{' '}
                                     <Link href='https://support.bloodhoundenterprise.io/hc/en-us/requests/new'>
                                         Contact sales


packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
```diff
@@ -397,9 +404,34 @@ export const TagForm: FC = () => {
...
+                                        }}>
+                                        <span className=''>
+                                            <FontAwesomeIcon icon={faTrashCan} className='mr-2' />
+                                            {`Delete ${tagTypeDisplay}`}
+                                        </span>
+                                    </Button>
+                                )}
+                            </div>
+                            <CardDescription className='pb-3 pl-3'>
+                                Currently there are {remainingZonesOrLabels} available {locationString}.
```
urangel: We have a constant coming in with this value already defined in this component. We can remove defining `locationString` on line 105 and use `tagTypePlural` instead.
```suggestion
                                Currently there are {remainingZonesOrLabels} available {tagTypePlural}.
```

packages/javascript/bh-shared-ui/src/mocks/factories/privilegeZones.ts
```diff
@@ -47,6 +48,26 @@ export const createAssetGroupTag = (tagId: number = 0, name?: string): AssetGrou
     };
 };
 
+export const createLabelGroupTag = (tagId: number = 0, name?: string): AssetGroupTag => {
```
urangel: An `AssetGroupTag` is either a Zone or a Label so this function would be better named `createLabelTag`. That being said, I think we should instead make the original function more flexible as the only thing that needs to change is the value for the `type` property. We can update `createAssetGroupTag` to allow passing a type so that we can use it to make labels.

That could look something like this:

```typescript
export const createAssetGroupTag = (
    tagId: number = 0,
    type: AssetGroupTagType = AssetGroupTagTypeZone,
    name?: string
): AssetGroupTag => {
    return {
        id: tagId,
        name: name ? name : `Tier-${tagId - 1}`,
        kind_id: faker.datatype.number(),
        glyph: null,
        type,
        position: tagId,
        description: faker.random.words(1000),
        created_at: faker.date.past().toISOString(),
        created_by: faker.internet.email(),
        updated_at: faker.date.past().toISOString(),
        updated_by: faker.internet.email(),
        deleted_at: faker.date.past().toISOString(),
        deleted_by: faker.internet.email(),
        require_certify: faker.datatype.boolean(),
        analysis_enabled: faker.datatype.boolean(),
    };
};
```

Then in your test instead of using `createLabelGroupTag()` you could do `createAssetGroupTag(1, AssetGroupTagTypeLabel)`

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
```diff
@@ -397,9 +403,34 @@ export const TagForm: FC = () => {
...
+                                )}
+                            </div>
+                            <CardDescription className='pb-3 pl-3'>
+                                Currently there are {remainingZonesOrLabels} available {tagTypePlural}.
+                                <Link href='https://support.bloodhoundenterprise.io/hc/en-us/requests/new'>
+                                    {' '}
+                                    Contact account management
+                                </Link>{' '}
+                                to increase the limit
+                            </CardDescription>
```
urangel: I have a couple of requests for this. Let's move the whitespace as show below:
```suggestion
                            <CardDescription className='pb-3 pl-3'>
                                Currently there are {remainingZonesOrLabels} available {tagTypePlural}.{' '}
                                <Link href='https://support.bloodhoundenterprise.io/hc/en-us/requests/new'>
                                    Contact account management
                                </Link>{' '}
                                to increase the limit
                            </CardDescription>
```

The other request is that we should not show this message in BHCE since this is enterprise specific messaging.

packages/javascript/bh-shared-ui/src/hooks/useTagLimits/useTagLimits.ts
```diff
@@ -0,0 +1,48 @@
...
+import { useGetConfiguration } from '../useConfiguration';
+
+const zoneReducer = (acc: number, tag: AssetGroupTag) => {
+    if (tag.type === AssetGroupTagTypeZone) return acc + 1;
+    else return acc;
+};
+
+export const useTagLimits = () => {
+    const tagsQuery = useAssetGroupTags();
+    const { data } = useGetConfiguration();
```
specter-flq, line 28: nit-ish: maybe give data a more descriptive name { data: `somethingThatDescribes`} to give more context of what it is as the hook is useGetConfiguration which also isn't very descriptive.

packages/javascript/bh-shared-ui/src/hooks/useTagLimits/useTagLimits.ts
```diff
@@ -0,0 +1,48 @@
...
+    const config = parseTieringConfiguration(data);
+
+    if (tagsQuery.isLoading || tagsQuery.isError || !config || !tagsQuery.isSuccess)
+        return {
+            zoneLimitReached: true,
+            labelLimitReached: true,
+            remainingZonesAvailable: 0,
+            remainingLabelsAvailable: 0,
+        };
```
specter-flq: nit-ish: This is fine, but I in a codebase standards document we created I remember someone added that for multiline we should add brackets. [Codebase standards](https://specterops.atlassian.net/wiki/spaces/BE/pages/1187905621/UI+Codebase+Standards)

 ```
 if (tagsQuery.isLoading || tagsQuery.isError || !config || !tagsQuery.isSuccess) {
        return {
            zoneLimitReached: true,
            labelLimitReached: true,
            remainingZonesAvailable: 0,
            remainingLabelsAvailable: 0,
        };
    }
```

File: packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx [hunk 6]
```diff
@@ -397,9 +403,36 @@ export const TagForm: FC = () => {
...
+                                            setDeleteDialogOpen(true);
+                                        }}>
+                                        <span className=''>
+                                            <FontAwesomeIcon icon={faTrashCan} className='mr-2' />
+                                            {`Delete ${tagTypeDisplay}`}
+                                        </span>
+                                    </Button>
+                                )}
+                            </div>
+                            {Certification && (
```
specter-flq, line 426: Maybe add a comment on why certification is being used here

specter-flq: Also I understand that this is a way to determine if BHE or BHCE to show this text but it feels a bit of context for the code block. I feel it should be something like isLimitsEnabled or something like that which relates specifically to what we want.

LucasParraF: I see your point here. Although certification is a feature that is only present at BHE and making it truthy means that we are only showing on BHE, maybe there could be an improvement on how we handle this. I think a comment could be a good way around so if anybody else comes and see this piece of the code can get some context on how it works and would be easier to understand. @urangel what would be your insight on this?

urangel: Yea what @specter-flq is saying makes sense that it is not immediately apparent why this would be tied together. A comment here is definitely a good addition to explain why. 

Something else you could do instead of using the Certification component existence would be to use the SalesMessage component existence to pivot off of. It's still a little indirect since it wouldn't be the SalesMessage component that we are rendering but at least rendering this block and rendering the SalesMessage are more in-line with their intents.

specter-flq: Agreed that would be a good option. Not sure if it would be overkill or its intended use but i wonder if we should leverage the new product_edition field in the version endpoint that returns which product we are in and have it available in a hook or something.

specter-flq: going to post here just in case, but maybe create a ticket to address this.

PR #2108, opened 2025-11-25
fix(BED-6066): resolves docker build

benwaples:
## Description

TARGETING FEATURE BRANCH

This PR addresses the docker build issues on the current feature branch that introduces DoodleUI to the monorepo

## Motivation and Context

Partially-Resolves: BED-6066

Adding doodle caused more sub dependencies to change versions and in this case required python for some compilation. This changeset adds resolutions for those sub-dependencies, and also adds doodle-ui to the docker build. It also has some quality of life changes to the just file so that yarn commands run against doodle-ui as well.

## How Has This Been Tested?
Manually tested locally

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/.github/workflows/build-ui.yml
+++ b/.github/workflows/build-ui.yml
@@ -46,21 +46,9 @@ jobs:
         run: |
           cd cmd/ui && yarn
 
-      - name: Format js-client-library with Prettier
+      - name: Format yarn projects
         run: |
-          cd packages/javascript/js-client-library && yarn format
-
-      - name: Format bh-shared-ui with Prettier
-        run: |
-          cd packages/bh-shared-ui/ && yarn format
-
-      - name: Format doodle-ui with Prettier
-        run: |
-          cd packages/doodle-ui/ && yarn format
-
-      - name: Format cmd-ui with Prettier
-        run: |
-          cd cmd/ui/ && yarn format
+          yarn format
 
       - name: Run Build
         run: |


.github/workflows/build-ui.yml
```diff
@@ -48,7 +48,19 @@ jobs:
...
+        run: |
+          cd packages/bh-shared-ui/ && yarn format
+
+      - name: Format doodle-ui with Prettier
+        run: |
+          cd packages/doodle-ui/ && yarn format
+
+      - name: Format cmd-ui with Prettier
+        run: |
+          cd cmd/ui/ && yarn format
```
benwaples: The name of this step didnt match what was being executed so I broke it out into separate steps.

package.json
```diff
@@ -42,6 +42,10 @@
     "react": "18.2.0",
     "sha.js": "^2.4.12",
     "tar-fs": ">=2.1.3 < 3.0.0",
-    "typescript": "5.1.x"
+    "typescript": "5.1.x",
+    "swagger-client": "3.29.3",
+    "@swagger-api/apidom-reference": "1.0.0-alpha.9",
+    "@swagger-api/apidom-parser-adapter-json": "1.0.0-alpha.9",
+    "@swagger-api/apidom-parser-adapter-yaml-1-2": "1.0.0-alpha.9"
```
benwaples, line 49: resolves the need to have python to be able to build the UI

superlinkx: I don't see any concerns with these dependencies

superlinkx:
Beagle is not long for this world and _shouldn't_ be getting run anywhere anymore (CI pipelines have been updated to use St Bernard instead). Is there a particular place where you were seeing beagle activating?

superlinkx (changes requested):
Blocking until we figure out what to do with Beagle changes

benwaples:
@superlinkx yay i was going to hit you up about that! I didnt see beagle activating, I updated those files because they referenced `bh-shared-ui` and Ive been searching for everything that referenced it so I can account for doodle-ui appropriately. So I figured add the change and then hit you up!

Ill remove all changes to /beagle! Thanks for weighing in!

superlinkx:
> @superlinkx yay i was going to hit you up about that! I didnt see beagle activating, I updated those files because they referenced `bh-shared-ui` and Ive been searching for everything that referenced it so I can account for doodle-ui appropriately. So I figured add the change and then hit you up!
> 
> Ill remove all changes to /beagle! Thanks for weighing in!

The equivalent location for St Bernard is `yarn-workspaces.json`. As long as that's up to date, you'll be good to go

package.json
benwaples: Sick ty! FWIW these are the versions on mainline right now. My understanding is that by adding `doodle-ui` to the workspace changes how yarn resolves dependencies, even if there isnt a connection to the newly added deps. I do suspect these sub deps are getting bumped for _some_ reason by adding storybook!

.github/workflows/build-ui.yml
superlinkx: St Bernard runs formatting for all yarn workspaces when it runs code generation. Would it be helpful to expose code formatting as a single command rather than having to maintain this list?

superlinkx (approved):

.github/workflows/build-ui.yml
benwaples: Hmm that is a good point, maintaining this list seems pointless and we should have another way to do this. St Bernard generate is handy, but I would prefer to isolate the step to just formatting so this doesnt fail for an unexpected reason. 

Im going to first try a yarn format from the root of the directory and see if that does what I think it should do.

superlinkx: Let me know how that goes, I think exposing the internal formatting command that St Bernard has would otherwise help us out here, it just needs to be wired up, the code already exists to support `generate`

benwaples: yarn format from the root of the project does indeed work as intended. Thank you again for the suggestion!

PR #2114, opened 2025-12-01
fix(ApiExplorer): Modified the predicate filter files to use type string. BED-6051

RaymondLaubert:
## Description

Modified the predicate filters to use type string instead of integer and boolean. This enabled the predicates to be read correctly instead of throwing a validation error when entering a value such as, "eq:7".

## Motivation and Context

Resolves BED-6051

The change is required to correctly enable the use of predicates in fields that traditionally are looking for an integer or boolean.

## How Has This Been Tested?

Manually tested making requests using the API Explorer, using with and without predicates.

## Screenshots (optional):

image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Changes**
  * API filter parameters for integer and boolean fields now accept string-formatted predicates (e.g., "eq:7", "gt:3", "eq:true") instead of native integer/boolean values.
  * Examples added to the API schema to illustrate the new string predicate formats for filtering.

### DIFF
--- a/packages/go/openapi/src/schemas/api.params.predicate.filter.boolean.yaml
+++ b/packages/go/openapi/src/schemas/api.params.predicate.filter.boolean.yaml
@@ -15,5 +15,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 type: string
+example: eq:true
 description: |
   Filter results by column boolean value. Valid filter predicates are `eq`, `neq`.
--- a/packages/go/openapi/doc/openapi.json
+++ b/packages/go/openapi/doc/openapi.json
@@ -18469,6 +18469,10 @@
       },
       "api.params.predicate.filter.integer": {
         "type": "string",
+        "examples": {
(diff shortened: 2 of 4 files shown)


packages/go/openapi/src/schemas/api.params.predicate.filter.boolean.yaml
```diff
@@ -14,6 +14,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-type: boolean
+type: string
 description: |
   Filter results by column boolean value. Valid filter predicates are `eq`, `neq`.
```
wes-mil, line 20: I think for all of these, it might be helpful to add examples of what these would look like. Here's some docs on that:
https://swagger.io/docs/specification/v3_0/adding-examples/

Can just add one or two to each, then regenerate with `just genspec`

wes-mil (approved):

PR #2138, opened 2025-12-05
chore: Drop unnecessary error logs to warn - BED-6663

Flake85:
## Description

The below message was a question post-release to validate tenant issues, and highlights an active challenge. There are several ERROR statements that need to be dropped to WARN to properly indicate their impact. This most notably occurs during post-processing from things like ADCS where missing field values prevent confirmation of the existence of an escalation. 

The most common issue is for “property not found” warnings: https://logs.bloodhound-ad.net/app/r/s/1NBBp

## Motivation and Context

Resolves BED-6663

Move `property not found` errors to `WARN` level

## How Has This Been Tested?

This deals with error logging and will be tested once merged

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when expected properties are missing: these cases now produce warnings and processing continues instead of failing, reducing interruptions across validation and analysis flows.
  * Enhanced logging for these warnings to include clearer contextual information for easier troubleshooting.

### DIFF
--- a/packages/go/analysis/ad/esc3.go
+++ b/packages/go/analysis/ad/esc3.go
@@ -352,7 +352,6 @@ func isStartCertTemplateValidESC3(template *graph.Node) bool {
 					attr.Error(err),
 				)
 			}
-			return false
 		}
 		return authorizedSignatures <= 0
 	}


packages/go/analysis/ad/esc3.go
```diff
@@ -295,16 +295,28 @@ func EnrollOnBehalfOfVersionOne(tx graph.Transaction, versionOneCertTemplates []
 
 func isStartCertTemplateValidESC3(template *graph.Node) bool {
 	if reqManagerApproval, err := template.Properties.Get(ad.RequiresManagerApproval.String()).Bool(); err != nil {
-		slog.Error(fmt.Sprintf("Error getting reqmanagerapproval for certtemplate %d: %v", template.ID, err))
+		if errors.Is(err, graph.ErrPropertyNotFound) {
+			slog.Warn(fmt.Sprintf("Did not get reqmanagerapproval for certtemplate %d: %v", template.ID, err))
+		} else {
+			slog.Error(fmt.Sprintf("Error getting reqmanagerapproval for certtemplate %d: %v", template.ID, err))
+		}
 	} else if reqManagerApproval {
```
superlinkx, line 313: These could all be unnested by using the same trick from your `graph.go` changes:

```suggestion
	if reqManagerApproval, err := template.Properties.Get(ad.RequiresManagerApproval.String()).Bool(); errors.Is(err, graph.ErrPropertyNotFound) {
		slog.Warn(fmt.Sprintf("Did not get reqmanagerapproval for certtemplate %d: %v", template.ID, err))
	} else if err != nil {
		slog.Error(fmt.Sprintf("Error getting reqmanagerapproval for certtemplate %d: %v", template.ID, err))
	} else if reqManagerApproval {
```

cmd/api/src/queries/graph.go
```diff
@@ -1069,14 +1069,20 @@ func fromGraphNodes(nodes graph.NodeSet) []model.PagedNodeListEntry {
 			props     = node.Properties
 		)
 
-		if objectId, err := props.Get(common.ObjectID.String()).String(); err != nil {
+		if objectId, err := props.Get(common.ObjectID.String()).String(); errors.Is(err, graph.ErrPropertyNotFound) {
+			slog.Warn(fmt.Sprintf("Did not get objectid for %d: %v", node.ID, err))
```
superlinkx: Since we're touching these old logs, it would be nice to remove the `fmt.Sprintf` hacks for both the net new logs and the existing logs they duplicate (doesn't necessarily need to be expanded to the entire files here, keep scope small for now).

Proper slog usage should be more like the following:

```go
slog.Warn("Node missing objectid", slog.Int("node_id", node.ID), attr.Error(err))
```

Note that we want to avoid string interpolation and give a reasonably detailed description of the warning/error without the context, and then separate attributes for structured logging (attr.Error(err) is designed to uniformly create the appropriate attribute for errors, so it only takes the error as an input).

packages/go/analysis/ad/esc3.go
```diff
@@ -295,21 +296,62 @@ func EnrollOnBehalfOfVersionOne(tx graph.Transaction, versionOneCertTemplates []
...
-			return true
+			slog.Error(
+				"Error getting schemaversion for certtemplate",
+				slog.Int("node_id", int(template.ID)),
+				attr.Error(err),
+			)
+		}
+	}
+	
+	if schemaVersion == 1 {
```
irshadaj, line 335: This block needs to be gated behind the error check. If there's an error, we still proceed to evaluate `schemaversion` and return based on that value, which strays from prior behavior.

packages/go/analysis/ad/esc3.go
```diff
@@ -295,21 +296,62 @@ func EnrollOnBehalfOfVersionOne(tx graph.Transaction, versionOneCertTemplates []
 
 func isStartCertTemplateValidESC3(template *graph.Node) bool {
 	if reqManagerApproval, err := template.Properties.Get(ad.RequiresManagerApproval.String()).Bool(); err != nil {
-		slog.Error(fmt.Sprintf("Error getting reqmanagerapproval for certtemplate %d: %v", template.ID, err))
+		if errors.Is(err, graph.ErrPropertyNotFound) {
```
irshadaj, line 299: return logic changed here too

PR #2142, opened 2025-12-08
chore: env severity updates BED-6602

urangel:
## Description

Downstream type updates to `Environment`.
Refactor `useSelectedTag` to return a callback to avoid always calling both hooks.

## Motivation and Context

Resolves BED-6602

## How Has This Been Tested?

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Selected tag resolution now uses placeholder and empty tag lists to avoid undefined/null results (removed highest-privilege fallback).
  * Abbreviated-number formatting now safely handles very large/unsafe integers.

* **New Features**
  * Environment data includes exposure metrics (percent and associated tag).
  * Added a public selected-tag hook and hygiene-tag metadata (name, id, glyph).

* **Refactoring**
  * Privilege Zones Details view now provides a default export alongside named exports.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/utils/numberFormatting.ts
+++ b/packages/javascript/bh-shared-ui/src/utils/numberFormatting.ts
@@ -15,10 +15,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export const abbreviatedNumber = (num: number, fractionDigits: number = 1) => {
+    // Exit early in case number in response is larger than the max safe integer to avoid doing math on it and getting erronious numbers
+    if (!Number.isSafeInteger(num)) return '>9Q';
+
     if (num < 1000) {
         // If the number is less than 1000, no abbreviation needed
         return num.toString();
     }
+
     const abbreviations = ['', 'K', 'M', 'B', 'T'];
     const log1000 = Math.floor(Math.log10(Math.abs(num)) / 3); // appropriate abbreviation index
 


packages/javascript/bh-shared-ui/src/hooks/useSelectedTag.ts
```diff
@@ -50,8 +50,8 @@ export const useSelectedTagPathParams = () => {
 };
 
 export const useSelectedTag = (fromPathParams = false) => {
-    const pathTag = useSelectedTagPathParams();
-    const queryTag = useSelectedTagQueryParams();
+    const pathTag = useSelectedTagPathParams;
+    const queryTag = useSelectedTagQueryParams;
 
     return fromPathParams ? pathTag : queryTag;
```
specter-flq, line 55: Curious as to why not something like:

` return fromPathParams ? useSelectedTagPathParams() : useSelectedTagQueryParams();`

That way we don't have to change all the other files to have the hook be a Self Invoking function and be in tune with what code rabbit is suggestion which is that it returns the values and not just another function, which is what is maybe expected.

specter-flq: plus was getting a TS error with the Self Invoking functions

urangel: The motivation was to avoid running both hooks. With the updates we return the callback that is needed and only run that one.

specter-flq: Yup i get that part and with running the below it only runs one hook. I console logged to confirm. Works exactly the same as the callback but with less changes and without the need of self invoking functions off a hook.

return fromPathParams ? useSelectedTagPathParams() : useSelectedTagQueryParams();

urangel: This suggestion breaks the rules of hooks

> React Hook "useSelectedTagQueryParams" is called conditionally. React Hooks must be called in the exact same order in every component render. (eslint react-hooks/rules-of-hooks)

maffkipp: I think the current change also technically violate the rule of hooks, since React can't guarantee anymore which hook will run on each render; I think in this case its just obfuscating that from the linter.

The way I have done it in the past has been the original pattern, where you just invoke both and then return the implementation you want to use. If you're trying to correct for some performance issues from running both, maybe two possible solutions would be:
- Combine these two hooks into a single hook that only conditionally computes the expensive operation
- In the components where this hook is called conditionally, instead conditionally render one of two wrapper components that each call one of these hooks and pass along the data to their children

urangel: noooo Wes get outta here with your reason. It was cumbersome mocking both for tests -__-

specter-flq: @urangel true true sorry about that the linter played a dirty one on me. Should have picked up the conditional restriction on hooks. 

Good suggestions @maffkipp,  I agree and feel that the solution should come from/or in the logic of the hook itself and the invocation of the hook should stay as commonly used and maybe stay away from the conditionals and self invoking as we dont really see a lot of that.

urangel: I've put this back to how it was originally so that we are calling both hooks

mistahj67 (approved):

Issue #2133, opened 2025-12-04
Bug:​ 'An error occurred when adding node'


StrafeNDestroy:
## Description:

*Received an error of " An error occurred when adding node"

### Are you intending to fix this bug?

NO

## Component(s) Affected:

- UI
- API
- Neo4j
- PostgreSQL
- Data Collector (SharpHound, AzureHound)
- Other (tooling, documentation, etc.)

## Steps to Reproduce:

Selecting a node and attempted to mark as owned. 

## Expected Behavior:
mark as owned, show up in labels and have the skull to show owned. 

## Actual Behavior:

"'An error occurred when adding node" and not addition to the marked as owned label.

## Screenshots/Code Snippets/Sample Files:

<img width="866" height="870" alt="Image" src="https://github.com/user-attachments/assets/296c24fb-ef1b-4f11-ac06-bfea56fe6137" />

<img width="834" height="616" alt="Image" src="https://github.com/user-attachments/assets/dd6fde25-7060-42f9-a609-18ec757df801" />

## Environment Information:

BloodHound: ​ 8.4.1

Collector: [SharpHound version / AzureHound version]

OS: Linux kali 6.16.8+kali-amd64 #1 SMP PREEMPT_DYNAMIC Kali 6.16.8-1kali1 (2025-09-24) x86_64 GNU/Linux

Browser: Mozilla Firefox 140.5.0esr

Node.js (if UI related: [Node.js version]

Go (if API related): [Go version]

Database (if persistence related): [Neo4j version / PostgreSQL version]

Docker: Docker version 27.5.1+dfsg4, build cab968b3

## Additional Information:

*Any additional context or information that might be helpful in understanding and diagnosing the issue.*

## Potential Solution (optional):

*If you have any ideas about what might be causing the issue or how it could be fixed, you can share them here.*

## Related Issues:

*If you've found related issues in the project's issue tracker, mention them here.*

## Contributor Checklist:

- [x] I have searched the issue tracker to ensure this bug hasn't been reported before or is not already being addressed.
- [x] I have provided clear steps to reproduce the issue.
- [x] I have included relevant environment information details.
- [x] I have attached necessary supporting documents.
- [x] I have checked that any JSON files I am attempting to upload to BloodHound are valid.

Byxs20:
I am also experiencing this issue

StephenHinck:
Thank you for reporting this - I popped an internal ticket so we can get it resolved!

[...]

PR #2150, opened 2025-12-09
feat(PZ): Right Click -> Add to Owned broken with default certification configuration BED-6951

ALCooper12:
## Description

- Created separate return values for zones and labels(owned) within the `createAssetGroupTagSelectorMutation` function

## Motivation and Context

This PR addresses:
- BED-6951
Closes: #2133 

- There was a bug that only allowed users to be able to right-click members and add new selectors(rules) to zones without also being able to do the same for Owned or labels. Now you can properly do both actions

## How Has This Been Tested?
- Locally within my BHE/BHCE env

## Screenshots (optional):
Screenshot 2025-12-08 at 2 56 02 PM
Screenshot 2025-12-08 at 3 10 32 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/ContextMenuPrivilegeZonesEnabled.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/ContextMenuPrivilegeZonesEnabled.tsx
@@ -16,7 +16,11 @@
 
 import { Dialog } from '@bloodhoundenterprise/doodleui';
 import { Menu, MenuItem } from '@mui/material';
-import { AssetGroupTagSelectorAutoCertifySeedsOnly, SeedTypeObjectId } from 'js-client-library';
+import {
+    AssetGroupTagSelectorAutoCertifySeedsOnly,
+    AssetGroupTagSelectorAutoCertifyType,
+    SeedTypeObjectId,
+} from 'js-client-library';
 import { FC, useState } from 'react';
 import { useMutation } from 'react-query';
 import {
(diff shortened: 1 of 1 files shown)


packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/ContextMenuPrivilegeZonesEnabled.tsx
```diff
@@ -47,8 +47,20 @@ const ContextMenu: FC<{
 
     const createAssetGroupTagSelectorMutation = useMutation({
         mutationFn: ({ assetGroupId, node }: { assetGroupId: string | number; node: NodeResponse }) => {
+            if (node.label) {
```
mistahj67: What is the node.label field defining here?

mistahj67 (changes requested):
Adding blocking comment because I forgot to submit the last comment as my review!

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/ContextMenuPrivilegeZonesEnabled.tsx
```diff
@@ -47,8 +47,20 @@ const ContextMenu: FC<{
 
     const createAssetGroupTagSelectorMutation = useMutation({
```
urangel, line 52: It may be better to create two different `handleAddNode` functions. One that passes an auto certify value and gets used as a prop for the tier zero AssetGroupMenuItem and one that doesn't pass an auto certify value that gets used as a prop for the owned AssetGroupMenuItem

urangel (approved):
ty!

PR #2154, opened 2025-12-09
refactor(ContextMenu): re-org asset group item logic BED-6970

TheNando:
## Description

This is just house keeping in prep for a story where a new edge context menu item will be adding. This PR cleans up asset group context menu logic.

## Motivation and Context

Resolves BED-6970

## How Has This Been Tested?

Unit tests have been updated

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Unified Add/Remove asset-group menu with "Tier Zero" and "Owned" actions, explicit loading/unavailable states, and optional confirmation flow when adding.
* **Bug Fixes**
  * More reliable availability indicators and consistent navigation after removing from a group.
* **Public API**
  * Component API expanded to payload-driven props and new helpers to detect owned/tier-zero contexts.
* **Tests**
  * Expanded coverage for add/remove flows, loading/error/permission states, and search-driven scenarios.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
@@ -93,7 +93,7 @@ export const AssetGroupMenuItem: FC<{
         onSettled: closeDialog,
     });
 
-    const createSelectorAction = showConfirmationOnAdd ? () => openDialog() : () => createRule();
+    const createRuleAction = showConfirmationOnAdd ? () => openDialog() : () => createRule();
     const hasPermission = checkPermission(Permission.GRAPH_DB_WRITE);
 
     // Is the selected node already a member of tier zero or owned?
@@ -130,7 +130,7 @@ export const AssetGroupMenuItem: FC<{
 
     return (
         <>
-            <MenuItem onClick={createSelectorAction}>Add to {assetGroupTag.name}</MenuItem>
+            <MenuItem onClick={createRuleAction}>Add to {assetGroupTag.name}</MenuItem>
 
             {showConfirmationOnAdd && (
                 <ConfirmNodeChangesDialog


packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.test.tsx
```diff
@@ -14,16 +14,73 @@
...
 import { setupServer } from 'msw/node';
-import { useState } from 'react';
-import { act, render, screen } from '../../../test-utils';
-import AssetGroupMenuItem from './AssetGroupMenuItemPrivilegeZonesEnabled';
+import { createAuthStateWithPermissions } from '../../../mocks';
+import { render, screen } from '../../../test-utils';
+import { Permission, apiClient } from '../../../utils';
+import { AssetGroupMenuItem } from './AssetGroupMenuItemPrivilegeZonesEnabled';
+
+const assetGroupTagsResponse = {
```
TheNando: The ContextMenu component was feeding many props to this component. Many of those values were only be used in this component, so they were relocated to be sourced within the component instead of from props. This required updating the mocks for the unit tests.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.test.tsx
```diff
@@ -35,178 +92,118 @@ const getEntityInfoTestProps = () => ({
     },
 });
 
-const ROUTE_WITH_SELECTED_ITEM_PARAM = `?selectedItem=${getEntityInfoTestProps().entityinfo.selectedNode.id}&searchType=node&primarySearch=${getEntityInfoTestProps().entityinfo.selectedNode.id}`;
+const server = setupServer(
```
TheNando, line 87: This was hoisted to the module scope to reduce repetition.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.test.tsx
```diff
@@ -35,178 +92,118 @@ const getEntityInfoTestProps = () => ({
...
+                data: createAuthStateWithPermissions([Permission.GRAPH_DB_WRITE]).user,
+            })
+        );
+    }),
+    rest.post('/api/v2/asset-group-tags/:tagId/selectors', (req, res, ctx) => {
+        return res(ctx.json({}));
+    })
+);
 
-const AssetGroupMenuItemWithDialog = (props: any) => {
```
TheNando, line 40: The original design had a Dialog wrapping the ContextMenu (parent component), but the dialog was only used for these conditional asset group menu items. Both components were refactored so that the Dialog is only present within the asset group items directly, removing the need for this wrapper.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.test.tsx
```diff
@@ -35,178 +92,118 @@ const getEntityInfoTestProps = () => ({
...
+            const mutateSpy = vi.spyOn(apiClient, 'createAssetGroupTagSelector');
+
+            render(<AssetGroupMenuItem assetGroupType='tierZero' showConfirmationOnAdd={true} />, {
+                route: ROUTE_WITH_SELECTED_ITEM_PARAM,
+            });
 
             const user = userEvent.setup();
 
-            const addToHighValueButton = screen.getByRole('menuitem', { name: /add to high value/i });
+            const addToHighValueButton = await screen.findByRole('menuitem', { name: /Add to Tier Zero/i });
```
TheNando: `await` + `findByRole` ensures that all the TanStack queries are resolved before running assertions.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.test.tsx
```diff
@@ -35,178 +92,118 @@ const getEntityInfoTestProps = () => ({
...
             expect(confirmationDialog).toBeVisible();
 
-            const applyButton = screen.getByRole('button', { name: /ok/i });
+            const applyButton = screen.getByRole('button', { name: /Ok/i });
+
             await user.click(applyButton);
 
-            expect(testOnAddNode).toHaveBeenCalledTimes(1);
-            expect(testOnAddNode).toHaveBeenCalledWith(tierZeroAssetGroup.id);
+            expect(mutateSpy).toHaveBeenCalled();
```
TheNando: Method to execute on confirmation is no longer provided. Instead, the mutation hook was relocated to within the component. The api call the mutation executes is now spied on instead.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,149 @@
...
     DialogTitle,
 } from '@bloodhoundenterprise/doodleui';
 import { MenuItem } from '@mui/material';
-import { FC } from 'react';
+import { AssetGroupTagSelectorAutoCertifySeedsOnly, SeedTypeObjectId } from 'js-client-library';
+import { FC, useState } from 'react';
+import { useMutation } from 'react-query';
 import { Link } from 'react-router-dom';
 
-const AssetGroupMenuItem: FC<{
```
TheNando, line 29: Majority of these props were values only needed within this component. Refactored so all that data is now sourced within the component. It may be easier to assess this as a whole new file rather than as a diff.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,149 @@
...
+                        </Button>
+                    </DialogActions>
+                </DialogContent>
+            </DialogPortal>
+        </Dialog>
+    );
+};
+
+const TIER_ZERO_POSITION = 1;
+const OWNED_TYPE = 3;
```
urangel: Let's use `AssetGroupTagTypeOwned` from the js-client-library instead of this new OWNED_TYPE constant please

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,149 @@
...
+                            Ok
+                        </Button>
+                    </DialogActions>
+                </DialogContent>
+            </DialogPortal>
+        </Dialog>
+    );
+};
+
+const TIER_ZERO_POSITION = 1;
```
urangel: Might be worth exporting `HighestPrivilegePosition` from useAssetGroupTags.tsx and using that here instead.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
urangel: nice

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,153 @@
...
+                        </Button>
+                    </DialogActions>
+                </DialogContent>
+            </DialogPortal>
+        </Dialog>
+    );
+};
+
+export const AssetGroupMenuItem: FC<{
+    assetGroupType: 'tierZero' | 'owned';
```
mistahj67: While this is the state of things today, PZ has taken a lot of pains to remove as much hardcode as possible with references to these names as technically the architecture by GA should support renaming of even this tags. With that in mind, what do you think of keeping the function handler prop for `onAddNode` and then pass that in while keeping this relatively how you have refactored it? 

That would keep it having that versatility while keeping props at a minimum?

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,153 @@
...
+    const [dialogOpen, setDialogOpen] = useState(false);
+    const { addNotification } = useNotifications();
+    const getAssetGroupTagsQuery = useTagsQuery();
+    const { selectedItemQuery } = useExploreSelectedItem();
+    const { checkPermission } = usePermissions();
+
+    const tierZeroAssetGroup = getAssetGroupTagsQuery.data?.find(
+        (value) => value.position === HighestPrivilegePosition
+    );
+    const ownedAssetGroup = getAssetGroupTagsQuery.data?.find((value) => value.type === AssetGroupTagTypeOwned);
```
mistahj67: Unsure if this will remain after my other comment but there are prebuilt hooks that will determine this for you in the useAssetGroupTags hook. While naming things is hard, PZ has tried to differentiate from the old system by using the `Tag` Suffix so in this case something like ownedAssetGroupTag which would help anyone entering this space or searching for AGT related logic

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,153 @@
...
+    const closeDialog = () => setDialogOpen(false);
+    const openDialog = () => setDialogOpen(true);
+
+    const { mutate: createSelector, isLoading: isMutationLoading } = useMutation({
+        mutationFn: () => {
+            const assetGroupId = assetGroup?.id ?? '';
+            const node = selectedItemQuery.data as NodeResponse;
+            return apiClient.createAssetGroupTagSelector(assetGroupId, {
+                name: node.label ?? node.objectId,
+                auto_certify: AssetGroupTagSelectorAutoCertifySeedsOnly,
```
mistahj67: Heads up: This was actually bugged for the owned case, you can see the fix here: https://github.com/SpecterOps/BloodHound/pull/2150 but that went into hotfix / staging and not main. Hopefully my suggestion above will make this easy but regardless of the direction you go, the support for not having `auto_certify` in the payload for the owned case is a requirement for this feature to function correctly

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
mistahj67: To add to my concern and give more context, there's already a concept of asset group tag type which is 'zone' or 'label' and this conflates that definition as well.

TheNando: This is great feedback, as I'm still learning much of the domain knowledge. I'd definitely like address these concerns.

Regarding the hard coded string, it's worth note that in this case it is used as a const type rather than a value, which is effectively an enum without the overhead of TypeScripts dumb enum implementation. The value is only compared directly within the module for branching rather than compared to any data-sourced values.

I totally agree with your argument from a semantics perspective that assetGroupType is not an appropriate name and I'll resolve this. I think I will try moving the mutation into its own module (since the onAddNode is literally the mutation hook's callback). Effectively, we just need a way to distinguish which mode the component will be acting in.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
```diff
@@ -16,95 +16,153 @@
...
+    }
+
+    // Show an error state if queries failed
+    if (getAssetGroupTagsQuery.isError || selectedItemQuery.isError) {
+        return <MenuItem disabled>Unavailable</MenuItem>;
+    }
+
+    // If selected node is already a member, navigate to the asset group details page for removal
+    if (isCurrentMember) {
+        const path = assetGroupType === 'tierZero' ? zonesPath : labelsPath;
```
maffkipp: I think I share some concerns with Trevor that moving all of this logic inside the MenuItem component makes it a lot less flexible if asset groups get renamed or if we add more in the future, and we'll have to end up unwinding some of this at that point

specter-flq: Yeah agreed on this sentiment, didn't see it before but now that you mention it makes sense.

packages/javascript/bh-shared-ui/src/views/Explore/ContextMenu/AssetGroupMenuItemPrivilegeZonesEnabled.tsx
mistahj67: Appreciate the background re: typescript enums, while I understand it's type safe from mispellings / use, it still feels like the opposite direction than the goal of PZ in terms of pushing for a more generic / expandable feature set and that's where my concern really lies in this refactor choice.

TheNando: I gotcha. I'll go ahead and find a better way.

packages/javascript/bh-shared-ui/src/utils/node.ts
```diff
@@ -0,0 +1,39 @@
...
+import { isNode, ItemResponse } from '../hooks';
+import { detailsPath, labelsPath, privilegeZonesPath, zonesPath } from '../routes';
+
+export function getOwnedObjectAssetGroupTagRemoveNodePath(assetGroupTag?: AssetGroupTag) {
+    return `/${privilegeZonesPath}/${labelsPath}/${assetGroupTag?.id}/${detailsPath}`;
+}
+
+export function getTierZeroAssetGroupTagRemoveNodePath(assetGroupTag?: AssetGroupTag) {
+    return `/${privilegeZonesPath}/${zonesPath}/${assetGroupTag?.id}/${detailsPath}`;
+}
```
urangel: There are equivalent functions provided from `usePZPathParams` that I would recommend using instead so as to keep this logic consolidated. See: `tagDetailsLink`

mistahj67: I wonder if these should go into the pzPathParams or if below should just reference the details routes with the respective type? @urangel what do you think?

packages/javascript/bh-shared-ui/src/hooks/useAssetGroupTags/useAssetGroupTags.tsx
```diff
@@ -490,6 +495,13 @@ export const useLabels = () => {
     return tagsQuery.data?.filter((tag) => labelTypes.includes(tag.type));
 };
 
+export const useOwnedTag = () => {
```
urangel: Would it be possible to leverage the `select` option with `useTagsQuery` so that we return the whole query result? Something like this is what I am thinking https://tkdodo.eu/blog/react-query-data-transformations#3-using-the-select-option

packages/javascript/bh-shared-ui/src/utils/node.ts
urangel: I agree with the latter and have suggested such [here](https://github.com/SpecterOps/BloodHound/pull/2154#discussion_r2620231757)

packages/javascript/bh-shared-ui/src/hooks/useAssetGroupTags/useAssetGroupTags.tsx
TheNando: Possibly. This relies on the useAssetGroupTags query, so I could potentially add an optional partial query params argument. Five queries in this module make use of that query, so it could become volatile. I just based my pattern on what the sibling hook useHighestPrivilegeTagId (for tier zero) was doing.

Honestly, I think this module is a bit bloated and could use some love, but it felt out of scope. What do you think is the best course?

urangel: Yea the scope creep on that is a bit large. No worries, I wanted to ask in case it was an easy swap 👍

PR #2160, opened 2025-12-11
feat: Objects Accordion for Details redesign BED-6844

urangel:
## Description

A redesigned Objects display is implemented for the Details page redesign.

## Motivation and Context

Resolves BED-6844

## How Has This Been Tested?
Render unit test added to the new component added.

## Screenshots (optional):

image

image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Objects accordion with per-type counts, sortable headers, kind-aware lists, and direct navigation to object details; Details view now shows dynamic object counts and zone-based navigation.

* **UX / Bug Fixes**
  * Simplified loading skeletons, improved list height/responsive layout, added SearchBar and site badge, and renamed visible labels from "tier-*" to "tag-*" for consistency.

* **API / Data**
  * Member queries accept kind filtering and rule member counts are surfaced so totals update dynamically.

* **Tests**
  * New and updated tests covering Objects accordion rendering, sorting, navigation, and renamed labels.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PrivilegeZones.tsx
+++ b/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PrivilegeZones.tsx
@@ -50,8 +50,8 @@ import { TagTabValue } from './Details/utils';
 import { PrivilegeZonesContext } from './PrivilegeZonesContext';
 
 // TODO: these will be swapped when all work is ready for the details redesign
-const Details = React.lazy(() => import('./DetailsV2'));
-// const Details = React.lazy(() => import('./Details'));
+// const Details = React.lazy(() => import('./DetailsV2'));
+const Details = React.lazy(() => import('./Details'));
 const Save = React.lazy(() => import('./Save'));
 const History = React.lazy(() => import('./History'));
 


packages/javascript/bh-shared-ui/src/views/PrivilegeZones/PrivilegeZones.tsx
```diff
@@ -46,7 +46,9 @@ import { cn, useAppNavigate } from '../../utils';
 import DefaultRoot from './DefaultRoot';
 import { PrivilegeZonesContext } from './PrivilegeZonesContext';
 
-const Details = React.lazy(() => import('./Details'));
+// TODO: these will be swapped when all work is ready for the details redesign
+const Details = React.lazy(() => import('./DetailsV2'));
```
urangel: Note to self: swap this back before merging to main

packages/javascript/js-client-library/src/client.ts
```diff
@@ -394,6 +397,7 @@ class BHEAPIClient {
                 params: {
                     ...options?.params,
                     environments,
+                    kind,
```
mistahj67: The endpoint query param field will be `primary_kind` I believe

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/TagList.tsx
```diff
@@ -73,7 +85,7 @@ export const TagList: FC<TagListProps> = ({ title, listQuery, selected, onSelect
             <ul>
                 {listQuery.isLoading ? (
                     itemSkeletons.map((skeleton, index) => {
-                        return skeleton(title, index);
+                        return skeleton(index);
```
maffkipp: As long as we are editing this, it probably makes more sense to just render the three components and ditching the `itemSkeletons` array. like:
```
<>
    <ItemSkeleton />
    <ItemSkeleton />
    <ItemSkeleton />
</>
```

packages/javascript/js-client-library/src/client.ts
urangel: Has this already been implemented? Do we need the "primary" prefix?

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/TagList.tsx
urangel: I've made this change but I'm curious as to the motivation. Is there something detrimental with the array map or is that a preference? What if I wanted to render 20 ItemSkeleton?

packages/javascript/js-client-library/src/client.ts
mistahj67: It has not but the rest of the asset group tag endpoints that filter by primary_kind use primary_kind already

urangel: I'll catch the sailing ship then! Thanks for the eyes

KillahDillah:
Ticket shows icons next to each user/node - but they do not render locally.
Screenshot 2025-12-18 at 12 46 37 PM

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/TagList.tsx
maffkipp: From my understanding it doesn't align with React's component model, since `ItemSkeleton` takes a key param instead of a props object and since it is being called as a function instead of invoked with JSX like `<ItemSkeleton />`. Keys in react I believe also need to be set on the element at the array site, and passing them in as a prop will not work.

From what I can find, the simplest way to repeat an element an arbitrary amount of times would be with something like:
```
Array.from({ length: 20 }).map((_, index) => (
  <ItemSkeleton key={index} />
))
```

urangel:
> Ticket shows icons next to each user/node - but they do not render locally. Screenshot 2025-12-18 at 12 46 37 PM

That's true the list items themselves are not rendering the icons. The mock is out of date but ticket specifies that the icons are only needed in the headers in the text of the requirements.

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/TagList.tsx
urangel: ah yea I've seen that array from approach before. Cool thanks for the response and feedback! ❤️

mistahj67 (changes requested):
Local testing looked good and code LGTM just a couple 3 things for feedback:

1. I think hiding the sorting until uncollapsed will make the page feel less "busy"
2. Overflowing the text with toolitp instead of horizontal scrollbar to drop our total scrollbars down
3. Not making the initial members page 0 call until actually uncollapsed to save on spamming the api

mistahj67 (approved):
Thanks for your patience as always good sir! 🚀

PR #2169, opened 2025-12-15
docs: BloodHound RFC 4: OpenGraph Extension Fundamental Requirements BED-7034

superlinkx:
## Description

This RFC proposal covers the design space and requirements for how BloodHound will need to handle the upcoming OpenGraph extensions feature. This PR is meant to be a place for the OpenGraph teams to discuss and refine this RFC as design completes the solidification process. Once all technical leaders of the OpenGraph teams have approved this RFC, it will be considered final and merged for our records.

## Motivation and Context

Resolves BED-7034

Time for a more mature design process for our most complex projects

## How Has This Been Tested?

Documentation only, no testing only approvals

## Types of changes

- Docs

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Documentation**
  * Added a new RFC documenting the OpenGraph Extension Framework: modular extension architecture, namespace-based attribute prefixing and manifest declarations, validation rules and scenarios, customization handling, Kinds table and environment/principal relationships, plus example schemas and diagrams illustrating structure and usage.

### DIFF
--- a/rfc/bh-rfc-4.md
+++ b/rfc/bh-rfc-4.md
@@ -131,7 +131,7 @@ erDiagram
     schema_node_kinds ||--|| kinds : "kind_id FK"
     schema_relationship_kinds ||--|| kinds : "kind_id FK"
     schema_environments ||--|| kinds : "environment_kind_id FK"
-    schema_environments ||--|| kinds : "source_kind_id FK"
+    schema_environments ||--|| source_kinds : "source_kind_id FK"
     schema_environments ||--o{ schema_environments_principal_kinds : "environment_id FK"
     schema_environments_principal_kinds ||--|| kinds : "principal_kind FK"
 
@@ -174,6 +174,11 @@ erDiagram
         int environment_id
         int principal_kind
     }
+    
+    source_kinds {
+        int id
+        text name
+    }
 ```
 
 ## 9. Environments and Principal Kinds
@@ -199,3 +204,86 @@ Environments define the security boundary of a user's model (e.g., Domain in Act
 1. Ensure the specified `environmentKind` exists.
 2. Ensure the specified `sourceKind` exists (create if it doesn't, reactivate if it does).
 3. Ensure all `principalKinds` exist.
+
+## 11. Example Upload Artifacts
+### Extension JSON example
+
+```json
+{
+  "schema": {
+    "name": "AzureHound",
+    "version": "1.0.0",
+    "namespace": "AZ"
+  },
+  "node_kinds": [
+    {
+      "name": "AZ_Tenant",
+      "display_name": "Azure Tenant",
+      "description": "An Azure tenant environment",
+      "is_display_kind": "true",
+      "icon": "cloud",
+      "color": "0xFF00FF",
+    },{
+      "name": "AZ_Device",
+      "display_name": "Azure Device",
+      "description": "An Azure device",
+      "is_display_kind": "true",
+      "icon": "desktop",
+      "color": "0x0000FF",
+    },
+    {
+      "name": "AZ_User",
+      "display_name": "Azure User",
+      "description": "An Azure user account",
+      "is_display_kind": "true",
+      "icon": "user",
+      "color": "0x00FF00",
+    },
+    {
+      "name": "AZ_Group",
+      "display_name": "Azure Group",
+      "description": "An Azure security or distribution group",
+      "is_display_kind": "true",
+      "icon": "group",
+      "color": "0xFF0000",
+    }
+  ],
+  "relationship_kinds": [
+    {
+      "name": "AZ_MemberOf",
+      "description": "User or computer is a member of a group",
+      "is_traversable": false,
+    },
+    {
+      "name": "AZ_HasSession",
+      "description": "User has an active session on a device",
+      "is_traversable": true,
+    }
+  ],
+  "environments": [
+    {
+      "environment_kind": "AZ_Tenant",
+      "source_kind": "AZBase",
+      "principal_kinds": [
+      	"AZ_User",
+      	"AZ_Group",
+      	"AZ_Device"
+      ]
+    }
+  ],
+  "relationship_findings": [
+    {
+      "name": "AZ_GenericAll",
+      "display_name": "Generic All Access (Azure)",
+      "environment_kind": "AZ_Tenant",
+      "relationship_kind": "AZ_HasSession",
+      "remediation": {
+        "short_description": "Principal has excessive permissions on target",
+        "long_description": "This finding indicates that a principal has been granted GenericAll permissions on a target object, providing complete control including read, write, delete, and permission modification rights.",
+        "short_remediation": "Remove the GenericAll permission from the principal's access control list",
+        "long_remediation": "# Remove Generic All Permissions\n\n## Steps to Remediate:\n\n1. **Identify the affected object**\n2. **Remove excessive permissions**\n3. **Grant minimal required permissions**"
+      }
+    }
+  ]
+}
+```


rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,199 @@
...
+  },
+  "node_kinds": [
+    {
+      "symbol": "GH_User",
+      "representation": "Github User",
+      "icon": "user",
+      "color": "#00FF00"
+    }
+  ],
+  "environments": [
```
brandonshearin, line 80: Is this a valid use case? Can a (theoretical) github extension specify details for an AD environment here? If valid, can you articulate that use case.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,199 @@
...
+
+```json
+{
+  "node_kinds": [
+    { "symbol": "User" }
+  ]
+}
+```
+
+The above schema is invalid because `User` should be prefixed (e.g., `GH_User`).
```
brandonshearin: Consider padding this explanation a bit.

e.g. "the above schema is invalid because the name User is already taken by the builtin active directory extension. to express a user-like kind in your domain, it should be prefixed gh_user."

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,199 @@
...
+
+## 7. Handling Customization
```
brandonshearin, line 117: this section is vague. what kind of customization are we talking about? node icons and colors?

rfc/bh-rfc-4.md
superlinkx: Nope, that's bad copy paste from the original, thanks for catching that

rfc/bh-rfc-4.md
superlinkx: It's not actually because User is already taken. It could be Foo and would still be invalid. I'll make it more clear that it should be prefixed with the extension's declared namespace

rfc/bh-rfc-4.md
superlinkx: It's somewhat intentionally vague (this RFC is meant to cover the underlying philosophy of extensions, including future additions we haven't explored yet). But I can add the icon/color examples in, that will help give better guidance than the way this ended up

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+audiences: |
+    - BloodHound OpenGraph Extension Authors
+    - BHE Team
+---
+
+# OpenGraph Extension Fundamental Requirements
+
+## 1. Overview
+
+This RFC introduces **OpenGraph Extensions**, a framework for modularly extending BloodHound’s data models. It mandates that all extension-defined attributes (e.g. node kinds, edge kinds, properties) use a unique namespace prefix to prevent conflicts and ensure traceability.
```
craigw-SO, line 17: This section is really short, and is very focused on the namespace requirement. I think it would be good to expand on this section. Happy to help articulate more here.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+
+## 2. Motivation & Goals
+
+### 2.1 Extensibility
+OpenGraph Extensions enable BloodHound to support modular, community-driven data model extensions. Without a standardized approach to namespacing, multiple extensions could define conflicting attribute names, leading to data integrity issues and ambiguous ownership of schema elements.
```
craigw-SO, line 22: Again this is more focused on the namespacing, and I think there's a lot more to highlight about the modular and extensible nature.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+- **Clarity of Ownership** - Trace attributes back to their defining extension via prefixes.
+
+## 3. Considerations
+
+### 3.1 Impact on Existing Systems
+
+SharpHound and AzureHound will initially bypass namespace validation to avoid breaking changes. A future migration tool will:
+
+- Add namespaces retroactively (e.g. `AD_` for SharpHound, `EAD_` for Azure/EntraHound).
+- No longer bundle these extensions but allow for easy installation.
```
craigw-SO, line 40: I don't think this is accurate. While AD and AZ will be migrated to the OG/extension platform, they will come pre-installed on any build.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+
+- **No Direct Risks** - Namespacing is a structural constraint, not a security feature.
+- **Data Integrity** - Prefixes ensure attribute uniqueness, ensuring schemas have only one source of truth.
+
+### 3.3 Drawbacks & Alternatives
+
+#### 3.3.1 Increased Verbosity
+
+- **Drawback** - Increased verbosity in attribute names (e.g. `GH_User` vs. `User`).
+- **Alternative** - Global registry for attribute names (rejected due to centralization).
```
craigw-SO, line 52: More specifically, I think we rejected due to the complexity involved with orchestrating and maintaining such a centrally organized system.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+
+#### 3.3.2 Multiple Extensions Covering the Same Technology
+
+- **Drawback** - Multiple extensions cannot simply cover the same technology using the same types (e.g. GitHub).
+- **Mitigation**:
+    - Extensions should be focused on their own domain.
+    - Namespacing prevents conflicts. For hybrid paths between technologies, references will be allowed to types in other namespaces.
```
craigw-SO, line 59: Something that I don't think we've captured as a drawback is that we cannot support cross-extension augmentation, which may make it either:
1. Difficult to get additional info added to an existing extension, or
2. Authors will create additional nodes and edges, potentially adding bloat and duplicative objects to the graph

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+      "representation": "Github User",
+      "icon": "user",
+      "color": "#00FF00"
+    }
+  ],
+  "environments": [
+    {
+      "environmentKind": "GH_Organization",
+      "sourceKind": "GHBase",
+      "principalKinds": ["GH_User"]
```
craigw-SO, line 84: We should either consider reusing a snipped from the example JSON schema OR at bare minimum ensure that the keys are consistent within this document. Looks like we're using different casing and names.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+    }
+  ]
+}
+```
+
+## 5. Attribute Prefixing Rules
+
+- **Format** - `namespace` + `_` + `<type>` (e.g., `GH_User`)
+- **Required** - All extension-defined attributes (e.g., `GH_User`, `GH_MemberOf`).
+- **Exempt** - References to attributes not defined by the extension (e.g., `"GHBase"` is a source kind).
```
craigw-SO, line 94: Shouldn't this just be a subsection of #4. Namespacing?

craigw-SO, line 94: Would should be clear on what exactly doesn't require it. So source_kind, but does environment_kind require it?

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+
+## 6. Validation
+
+Reject extensions if:
+
+1. Namespace conflicts with an existing extension.
```
craigw-SO, line 100: We should be clear that we're specifically referring to the namespace key in particular, which is what we rely on to establish/align to the namespace.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+2. Any attribute definition (not reference) lacks the required prefix.
```
craigw-SO, line 101: Realistically, if an attribute is defined with a namespace key, then the reference will also necessarily require it for the reference to be valid. Maybe there's a better way to clarify this?

rfc/bh-rfc-4.md
craigw-SO: We could also clarify that we are trying to facilitate and promote the users ability to choose and manage within their own environment. It's not just about including customization options within the schema declaration, but also once that schema is imported.

I think it would be good to call out that extensions define _defaults_ to be used as a baseline, but users are allowed to further customize within their own instance, and user customization should not be overridden.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+}
+```
+
+The above schema is invalid because `User` should be prefixed with the extension's declared namespace (e.g. `GH_User`).
+
+## 7. Handling Customization
+
+Extensions may include default customization for their attributes in the manifest. These will be stored with the extension but will not overwrite existing user-customized definitions. Examples of customization include node icons and colors, which are currently defined in the `custom_node_kinds` table. Extensions can declare icons and colors, but they will only be written to the actual `custom_node_kinds` table if they do not already exist.
+
+## 8. Kinds Table Handling
```
craigw-SO, line 121: I think we would benefit from clarifying the distinction between the `kinds` system and the schema system we're implementing, as well as specifying how they should relate and interact.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+
+Extensions should use junction tables when creating relationships with tables owned by DAWGS (e.g., kinds table). Modifying these tables directly is discouraged for performance and reliability reasons.
+
+```mermaid
+erDiagram
+    schema_extensions ||--o{ schema_node_kinds : "extension"
+    schema_extensions ||--o{ schema_relationship_kinds : "extension"
+    schema_extensions ||--o{ schema_environments : "extension"
```
craigw-SO, line 129: Should these be shown as "extension_id FK" like we show with the other relationships? Are they not FK'd?

rfc/bh-rfc-4.md
craigw-SO: Actually, as I read further, I think it would be wise to combine #8. Kinds Table Handling and #9. Environments and Principal Kinds all under a broader Kinds and Schemas section.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+authors: |
+    [Alyx Holms](aholms@specterops.io)
+status: DRAFT
+created: 2025-12-15
+audiences: |
+    - BloodHound OpenGraph Extension Authors
+    - BHE Team
+---
+
+# OpenGraph Extension Fundamental Requirements
```
craigw-SO, line 13: As I read through and was contemplating the various sections and how they were organized, I was realizing: I think it might be helpful to separate what are actual requirements (namespacing, validation, modularity, etc), from some of our design considerations (kinds vs schemas, table relations, sample JSON, etc).

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+    }
+  ]
+}
+```
+
+## 10. Validation Rules for Environments
+
+1. Ensure the specified `environmentKind` exists.
+2. Ensure the specified `sourceKind` exists (create if it doesn't, reactivate if it does).
+3. Ensure all `principalKinds` exist.
```
craigw-SO, line 206: Need consistency in these keys' cases (and check for spelling, too).

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+    }
+  ],
+  "environments": [
+    {
+      "environment_kind": "AZ_Tenant",
+      "source_kind": "AZBase",
+      "principal_kinds": [
+      	"AZ_User",
+      	"AZ_Group",
+      	"AZ_Device"
```
craigw-SO, line 270: Here's something we haven't thought about. Do we need to explicitly declare the environment as a "principal" as well, or will analysis already consider it as such? I.e. does it need to be self-referred in this list?

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+{
+  "schema": {
+    "name": "AzureHound",
+    "version": "1.0.0",
+    "namespace": "AZ"
+  },
+  "node_kinds": [
+    {
+      "name": "AZ_Tenant",
+      "display_name": "Azure Tenant",
```
craigw-SO, line 221: I think I saw that we settled on "name" (instead of "symbol") and "display_name" (instead of "name" or "representation"). Is that correct?

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+
+**Example Environment Schema**:
+
+```json
+{
+  "environments": [
+    {
+      "environmentKind": "GH_Organization",
+      "sourceKind": "GHBase",
+      "principalKinds": ["GH_User"]
```
craigw-SO, line 196: Are principalKinds (or principal_kinds) optional? Can that field be omitted if there are none? Or is the field required with an empty array?

MichaelGrafnetter:
Community vs. Enterprise: Hi @superlinkx , do I understand it correctly that the only difference between BHE and BHCE extensions would be the findings section in the JSON?

craigw-SO:
> Community vs. Enterprise: Hi @superlinkx , do I understand it correctly that the only difference between BHE and BHCE extensions would be the findings section in the JSON?

Hi @MichaelGrafnetter, I believe technically there is no difference – as in you could potentially use one payload for both. However, because we want to protect our IP, we should leave out findings and the remediation plans for BHCE extension bundles.

Either way, findings/remediations wouldn't be displayed in BHCE. But we should definitely omit so we don't expose them.

superlinkx:
> Community vs. Enterprise: Hi @superlinkx , do I understand it correctly that the only difference between BHE and BHCE extensions would be the findings section in the JSON?

Correct, currently that's the only difference. Craig is also correct that we will allow findings to be uploaded and may allow them in the database, but there won't be a way to interact with them for now. Extensions are therefore responsible for keeping findings data private if the findings are considered an enterprise only feature of the extension

MichaelGrafnetter:
Hi, what does `"is_display_kind": "true"` mean, please? The JSON also defines `"source_kind": "AZBase"`. Should't it be `AZ_Base` instead of `AZBase`, to follow the naming convention? Or should we rather drop the underscores entirely?

MichaelGrafnetter:
Have we changed the color definition from `#FF0000` to `0xFF0000` intentionally? The former format has a nice preview support in VS Code, GitHub, and other JSON authoring tools.

rfc/bh-rfc-4.md
```diff
@@ -0,0 +1,289 @@
...
+    "namespace": "AZ"
+  },
+  "node_kinds": [
+    {
+      "name": "AZ_Tenant",
+      "display_name": "Azure Tenant",
+      "description": "An Azure tenant environment",
+      "is_display_kind": "true",
+      "icon": "cloud",
+      "color": "0xFF00FF",
```
craigw-SO, line 225: Research folks noticed that in some places we specify "#FF0000" and others we use "0xFF0000". I'm pretty sure that we use "#FF0000" everywhere, correct? If so, we should make it consistent in this example.

craigw-SO:
> Hi, what does `"is_display_kind": "true"` mean, please? The JSON also defines `"source_kind": "AZBase"`. Should't it be `AZ_Base` instead of `AZBase`, to follow the naming convention? Or should we rather drop the underscores entirely?

@MichaelGrafnetter – `"is_display_kind": "true"` will be used to identify which kind is used for the custom node icon. Nodes may have multiple kinds, but there should only be one "display kind" to pull an icon for. I don't think that it's wired up for use just yet, but anticipate that it will be soon, so we want authors to include it now in preparation for its use.

PR #2194, opened 2025-12-19
feat(OpenGraph): Foreign Key node and edge schema tables to DAWGS kind table - BED-7067

LawsonWillard:
## Description

- Adds foreign keys to the schema node and edge kind tables so that they are linked to the DAWGS kind table. 
- Updates the Create Schema Node and Create Schema Edge kind functions to create entries in DAWGS kind table using a CTE. 
- Refactor Update Node/Edge Schema kind function to not allow names to be updatable. 
- Updated corresponding integration tests 

## Motivation and Context
Resolves: BED-7067

To align with other OpenGraphSchema tables, we need to add FKs to the SchemaNodeKind and SchemaEdgeKind tables.

## How Has This Been Tested?

- Updated integration tests to ensure functions run as expected

## Types of changes
- Database Migrations

## Checklist:
  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Public retrieval for graph edge kinds now supports filtering, sorting, pagination, and total counts.

* **Improvements**
  * Node and edge kind identity now uses a centralized kind linkage for more consistent schema mapping.

* **Behavior Changes**
  * Kind names are now managed via the centralized kind system; direct name updates on kinds are no longer applied and existing names are preserved.

* **Tests**
  * Integration tests updated to reflect name-preservation behavior.

### DIFF
--- a/cmd/api/src/database/graphschema.go
+++ b/cmd/api/src/database/graphschema.go
@@ -503,9 +503,8 @@ func (s *BloodhoundDB) GetGraphSchemaEdgeKindsWithSchemaName(ctx context.Context
 			return nil, 0, CheckError(result)
 		} else {
 			if limit > 0 || skip > 0 {
-				countSqlStr := fmt.Sprintf(`SELECT COUNT(*) FROM %s edge JOIN %s schema ON edge.schema_extension_id = schema.id %s`,
-					model.GraphSchemaEdgeKind{}.TableName(),
-					model.GraphSchemaExtension{}.TableName(),
+				countSqlStr := fmt.Sprintf(`SELECT COUNT(*) FROM %s edge JOIN %s schema ON edge.schema_extension_id = schema.id JOIN %s k ON edge.kind_id = k.id %s`,
+					model.GraphSchemaEdgeKind{}.TableName(), model.GraphSchemaExtension{}.TableName(), kindTable,
 					filterAndPagination.WhereClause)
 				if countResult := s.db.WithContext(ctx).Raw(countSqlStr, filterAndPagination.Filter.params...).Scan(&totalRowCount); countResult.Error != nil {
 					return model.GraphSchemaEdgeKindsWithNamedSchema{}, 0, CheckError(countResult)


cmd/api/src/database/migration/migrations/v8.5.0.sql
```diff
@@ -39,11 +38,11 @@ CREATE TABLE IF NOT EXISTS schema_extensions (
     PRIMARY KEY (id)
 );
 
--- OpenGraph schema_node_kinds -  stores node kinds for open graph extensions
+-- OpenGraph schema_node_kinds -  stores node kinds for open graph extensions. This FK's to the DAWGS kind table directly.
 CREATE TABLE IF NOT EXISTS schema_node_kinds (
-    id SERIAL PRIMARY KEY ,
+    id SMALLINT PRIMARY KEY REFERENCES kind (id) ON DELETE CASCADE,
```
wes-mil: I don't think we want to be tying schema node kinds ID to the kind tables id? This means we can only have 1 schema_node_kind attached to a kind

LawsonWillard: I may be wrong but I think we can only ever have 1 schema node kind linked to a kind at any given time due to name uniqueness.

LawsonWillard: I see benefit for hybrid environments though being able to point to the same node kind

wes-mil: Oh yeah you're right... I still don't love tying the IDs in this table to the IDs of another table. I think that the name FK takes care of this

wes-mil: If we want to guarantee uniqueness for kinds tied to extension items, we might need to create a look aside table because nodes and edges have different unique constraints

LawsonWillard: Uniqueness should be guaranteed during insert since we use a CTE expression to transactionally insert into the kinds table and then the schema kind tables. 

all praise to @mistahj67 for helping me with that

wes-mil: Are you sure about that? If a `schema_node_kind` exists with a kind `Test`, and we insert a `schema_edge_kind` with a kind `Test` it should appropriately DO UPDATE and return id, name. Then since the `schema_node_kind` unique tag is separate from the `schema_edge_kind` unique tag, I don't think it would reject

LawsonWillard: Ohh I see what you're saying. You're definitely right, i think both the schema edge kind and node kinds could/would then point to the same DWAGS kind. Thank you for pointing that out!!

cmd/api/src/database/graphschema.go
```diff
@@ -349,6 +357,7 @@ func (s *BloodhoundDB) GetGraphSchemaPropertyById(ctx context.Context, extension
 	return extensionProperty, nil
 }
 
+// UpdateGraphSchemaProperty - updates a row in the schema_properties table based on the provided id. It will return an error if the target schema edge kind does not exist or if any of the updates violate the schema constraints.
```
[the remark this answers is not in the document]
wes-mil: the rabbit has a point

cmd/api/src/database/migration/migrations/v8.5.0.sql
wes-mil: I feel like I would still prefer if we didn't make the primary key of this table to also be a foreign key of another table. I personally would rather if we just had a `SERIAL` ID here and then replaced the `name` field with `kind_id` foreign key. I'm a little scared about data integrity on the table level even though we are ensuring it with the postgres function. 

Is there a reason you're preferring to make the primary key of this table also a foreign key? Perhaps @superlinkx could weigh in here as well

superlinkx: https://github.com/SpecterOps/BloodHound/blob/1bc5b8630aa41f4a002e6603f66fd1388cf9cedf/rfc/bh-rfc-4.md#8-kinds-table-handling

Our current diagram shows that we should have a serial primary key, and then foreign key the kind_id as a separate field. Please use the diagram above as a reference. If you believe anything in the diagram is incorrect, let's get it sorted out.

LawsonWillard: Sorry, I was going with the path of least code change from our original design and missed the design doc. Not opposed to whats in the doc and will update it to conform.

cmd/api/src/database/graphschema.go
```diff
@@ -195,22 +197,32 @@ func (s *BloodhoundDB) DeleteGraphSchemaExtension(ctx context.Context, extension
...
-			VALUES (?, ?, ?, ?, ?, ?, ?)
-			RETURNING id, name, schema_extension_id, display_name, description, is_display_kind, icon, icon_color, created_at, updated_at, deleted_at`,
-		schemaNodeKind.TableName()),
-		name, extensionId, displayName, description, isDisplayKind, icon, iconColor).Scan(&schemaNodeKind); result.Error != nil {
-		if strings.Contains(result.Error.Error(), DuplicateKeyValueErrorString) {
-			return model.GraphSchemaNodeKind{}, fmt.Errorf("%w: %v", ErrDuplicateSchemaNodeKindName, result.Error)
+	var schemaNodeKind = model.GraphSchemaNodeKind{}
+	if db, err := s.db.DB(); err != nil {
+		return schemaNodeKind, err
+	} else if row := db.QueryRowContext(ctx, `
```
wes-mil: Do we need to be fetching the db like this? I don't see a problem with it, it just breaks our standard a bit

cmd/api/src/database/graphschema.go
```diff
@@ -221,16 +233,33 @@ func (s *BloodhoundDB) GetGraphSchemaNodeKinds(ctx context.Context, filters mode
...
+	}
+
+	for idx, _ := range sort {
+		if sort[idx].Column == "name" {
+			sort[idx].Column = fmt.Sprintf("%s.%s", kindTable, sort[idx].Column)
+		} else {
+			sort[idx].Column = fmt.Sprintf("%s.%s", model.GraphSchemaNodeKind{}.TableName(), sort[idx].Column)
+		}
+	}
+
```
wes-mil: Could you explain a little what's going on here? Just a bit difficult to follow logically.

Perhaps a comment could be helpful too

cmd/api/src/database/graphschema.go
```diff
@@ -386,39 +430,69 @@ func (s *BloodhoundDB) DeleteGraphSchemaProperty(ctx context.Context, propertyID
...
-	if result := s.db.WithContext(ctx).Raw(fmt.Sprintf(`
-	INSERT INTO %s (name, schema_extension_id, description, is_traversable)
-    VALUES (?, ?, ?, ?)
-    RETURNING id, name, schema_extension_id, description, is_traversable, created_at, updated_at, deleted_at`, schemaEdgeKind.TableName()),
-		name, schemaExtensionId, description, isTraversable).Scan(&schemaEdgeKind); result.Error != nil {
-		if strings.Contains(result.Error.Error(), DuplicateKeyValueErrorString) {
-			return schemaEdgeKind, fmt.Errorf("%w: %v", ErrDuplicateSchemaEdgeKindName, result.Error)
+	if db, err := s.db.DB(); err != nil {
+		return model.GraphSchemaEdgeKind{}, err
+	} else if row := db.QueryRowContext(ctx, `
```
wes-mil: Same as above

cmd/api/src/database/graphschema.go
```diff
@@ -386,39 +430,69 @@ func (s *BloodhoundDB) DeleteGraphSchemaProperty(ctx context.Context, propertyID
...
+		}
+	}
+
+	for idx, _ := range sort {
+		if sort[idx].Column == "name" {
+			sort[idx].Column = fmt.Sprintf("%s.%s", kindTable, sort[idx].Column)
+		} else {
+			sort[idx].Column = fmt.Sprintf("%s.%s", model.GraphSchemaEdgeKind{}.TableName(), sort[idx].Column)
+		}
+	}
```
wes-mil: Same as above

cmd/api/src/database/migration/migrations/v8.5.0.sql
```diff
@@ -184,3 +182,96 @@ $$
...
+            ( INSERT INTO kind (name) VALUES (edge_kind_name) ON CONFLICT (name) DO UPDATE SET name = edge_kind_name RETURNING id, name)
+        INSERT INTO schema_edge_kinds (kind_id, schema_extension_id, description, is_traversable)
+    SELECT id,
+           edge_kind_schema_extension_id,
+           edge_kind_description,
+           edge_kind_is_traversable
+    FROM dawgs_kinds
+    RETURNING id, schema_extension_id, edge_kind_name, description, is_traversable, created_at, updated_at, deleted_at;
+END
+$$ LANGUAGE plpgsql;
```
wes-mil: I think a comment explaining the SQL trickery here would be helpful

cmd/api/src/database/migration/migrations/v8.5.0.sql
```diff
@@ -184,3 +182,96 @@ $$
...
+    SELECT dk.id,
+           node_kind_schema_extension_id,
+           node_kind_display_name,
+           node_kind_description,
+           node_kind_is_display_kind,
+           node_kind_icon,
+           node_kind_icon_color
+    FROM dawgs_kinds dk
+    RETURNING id, schema_extension_id, node_kind_name, display_name, description, is_display_kind,
+        icon, icon_color, created_at, updated_at, deleted_at;
```
wes-mil: Same as above

cmd/api/src/database/graphschema.go
LawsonWillard: I had trouble with GORM's scan function since the returned columns dont match the struct field names. I ended up pulling the SQL DB out so we can scan the row directly into the struct based on return position instead of letting it infer.

cmd/api/src/database/graphschema.go
LawsonWillard: Since the Kind and Schema Kind tables share identical column names (id), we need to add table identifiers to the filtering/sorting columns so we don't return an ambiguous column error. 

I can add this as a comment

cmd/api/src/database/migration/migrations/v8.5.0.sql
LawsonWillard: Yeah I can add to the function doc string a better explanation as to whats happening.

cmd/api/src/database/graphschema.go
LawsonWillard: I had trouble with GORM's scan function since the returned columns dont match the struct field names. I ended up pulling the SQL DB out so we can scan the row directly into the struct based on return position instead of letting it infer.

GORM doesnt provide the option to do this so I had to interact with the SQL DB obj directly.

AD7ZJ (approved):
Lot going on in there, it looks good as far as I can tell! I don't know if Wes' method using triggers is less trouble but I think either way gets us equivalent integrity.

wes-mil (approved):
I'm approving it in its current design assuming we choose to stick with this design, but not explicitly approving the design itself. 

I personally would prefer the supertable DB design keyed by name, but I'll let you and the senior engineers make that call

cmd/api/src/database/migration/migrations/v8.5.0.sql
```diff
@@ -184,3 +182,107 @@ $$
...
+
+-- upsert_schema_edge_kind - atomically upserts an edge kind into both the DAWGS kind and schema_edge_kinds tables.
+-- This function addresses the edge case where both a schema_edge_kind and schema_node_kind can point to same DAWGS kind.
+-- First, the function locks the schema_node_kinds and schema_edge_kinds tables so that another insert won't add a kind during this function.
+-- Second, the function checks the schema_node_kinds table to ensure the kind isnt already defined there.
+-- Assuming the kind is unique, it will then insert the kind into the DAWGS table. Since the DAWGS table is append-only
+-- the ON CONFLICT DO UPDATE is used to ensure we get the id and name back even if it already exists (without an error).
+-- Following this, we insert into the schema_edge_kinds table and return the values needed to fill out a model.GraphSchemaEdgeKind.
+-- The return table columns must be unique otherwise Postgres will return an error for ambiguous column names.
+CREATE OR REPLACE FUNCTION upsert_schema_edge_kind(edge_kind_name TEXT, edge_kind_schema_extension_id INT,
```
superlinkx: I'm not sure we need to go to these lengths to validate kinds like this. I'd like to do a design review with the tech lead on the project before we accept this. Putting a temporary hold until that discussion can happen.

superlinkx (changes requested):

superlinkx:
Using requested changes to block until we can have a design review

cmd/api/src/database/parameters_test.go
```diff
@@ -176,7 +176,7 @@ func TestParameters_GetAGTParameter(t *testing.T) {
 	require.Equal(t, appcfg.AGTParameters{
 		DAWGsWorkerLimit:     6,
 		SelectorWorkerLimit:  7,
-		ExpansionWorkerLimit: 7,
+		ExpansionWorkerLimit: 3,
```
wes-mil, line 179: Was this on purpose?

LawsonWillard: Yeah, not sure whats going on with it but it was causing integration tests to always fail. Going to reach out to @mistahj67 for confirmation on what should happen.

mistahj67: Not sure how this passed in CI the first time, thanks for fixing!

cmd/api/src/database/graphschema.go
```diff
@@ -403,27 +469,57 @@ func (s *BloodhoundDB) CreateGraphSchemaEdgeKind(ctx context.Context, name strin
...
+				Direction: sortItem.Direction,
+				Column:    fmt.Sprintf("%s.%s", "k", sort[idx].Column),
+			}
+		} else {
+			tableIdentifiedSort[idx] = model.SortItem{
+				Direction: sortItem.Direction,
+				Column:    fmt.Sprintf("%s.%s", "ek", sort[idx].Column),
+			}
+		}
+	}
```
superlinkx: Remove as there's no longer ambiguity with `name` between the tables

cmd/api/src/database/graphschema.go
```diff
@@ -477,18 +573,29 @@ func (s *BloodhoundDB) GetGraphSchemaEdgeKindsWithSchemaName(ctx context.Context
 func (s *BloodhoundDB) GetGraphSchemaEdgeKindById(ctx context.Context, schemaEdgeKindId int32) (model.GraphSchemaEdgeKind, error) {
 	var schemaEdgeKind model.GraphSchemaEdgeKind
 	return schemaEdgeKind, CheckError(s.db.WithContext(ctx).Raw(fmt.Sprintf(`
```
superlinkx: Nit: Avoid doing work in a return that mutates another return value, this pattern has made life harder in other parts of the app and adds unnecessary cognitive load

superlinkx (changes requested):
Minor cleanup requested, then this looks ready to merge

cmd/api/src/database/graphschema.go
```diff
@@ -227,18 +239,18 @@ func (s *BloodhoundDB) GetGraphSchemaNodeKinds(ctx context.Context, filters mode
...
+									JOIN %s k ON nk.kind_id = k.id 
+									%s %s %s`,
+			model.GraphSchemaNodeKind{}.TableName(), kindTable, filterAndPagination.WhereClause, filterAndPagination.OrderSql, filterAndPagination.SkipLimit)
 		if result := s.db.WithContext(ctx).Raw(sqlStr, filterAndPagination.Filter.params...).Scan(&schemaNodeKinds); result.Error != nil {
 			return nil, 0, CheckError(result)
 		} else {
 			if limit > 0 || skip > 0 {
-				countSqlStr := fmt.Sprintf(`SELECT COUNT(*) FROM %s %s`, model.GraphSchemaNodeKind{}.TableName(), filterAndPagination.WhereClause)
+				countSqlStr := fmt.Sprintf(`SELECT COUNT(*) FROM %s nk JOIN %s k ON nk.kind_id = k.id %s`,
+					model.GraphSchemaNodeKind{}.TableName(), kindTable, filterAndPagination.WhereClause)
```
wes-mil, line 253: do we need to be joining for this count query? is this to enable the where clause?

LawsonWillard: Yeah, if the WHERE clause filters on name that would break the count query

PR #2215, opened 2025-12-31
fix: environment selector filter bug BED-6874

Holocraft:
## Description

This fixes the environment selector filter bug.

## Motivation and Context

When going to PZ details and using the environment selector to select an environment, the total object count updates. But when you select "All environments", the count doesn't update.
Resolves: [BED-6874](https://specterops.atlassian.net/browse/BED-6874)

## How Has This Been Tested?
Updated unit tests and tested manually

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs
- [ x] I have followed proper test practices

[BED-6874]: https://specterops.atlassian.net/browse/BED-6874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Release Notes

* **New Features**
  * Environment aggregation is now supported in Privilege Zones, providing more flexible environment management capabilities
  * Environment filtering behavior is now configurable, allowing customized data collection approaches
  * Results are returned in deterministic order for consistent, predictable behavior across operations

### DIFF
--- a/packages/javascript/bh-shared-ui/src/hooks/useAvailableEnvironments/useAvailableEnvironments.tsx
+++ b/packages/javascript/bh-shared-ui/src/hooks/useAvailableEnvironments/useAvailableEnvironments.tsx
@@ -59,7 +59,7 @@ export const useSelectedEnvironment = (
     };
 
     const setEnvironmentAggregation = (aggregation: EnvironmentQueryParams['environmentAggregation']) => {
-        setEnvironmentParams({ environmentId: null, environmentAggregation: aggregation });
+        setEnvironmentParams({ environmentAggregation: aggregation });
     };
 
     return {
--- a/packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
+++ b/packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
@@ -22,11 +22,12 @@ import { useMatchingPaths } from './useMatchingPaths';
 export const getEnvironmentAggregationIds = (
     environmentAggregation: EnvironmentAggregation,
     environments: Environment[],
-    collectedOnly: boolean
+    collectedOnly: boolean = true
 ) => {
     const aggregationIds: string[] = [];
 
     environments.forEach((environment) => {
+        //check if we're including the collected count
         if (
             (!collectedOnly || environment.collected) &&
             (environmentAggregation === 'all' || environment.type === environmentAggregation)
@@ -41,7 +42,7 @@ export const getEnvironmentAggregationIds = (
 
 export const useEnvironmentIdList = (
     ENVIRONMENT_AGGREGATION_SUPPORTED_ROUTES: (string | PathPattern)[],
-    collectedOnly: boolean
+    collectedOnly: boolean = true
 ): Environment['id'][] => {
     const { data: availableEnvironments } = useAvailableEnvironments();
     const { environment, environmentAggregation } = useSelectedEnvironment();


packages/javascript/bh-shared-ui/src/routes/index.tsx
```diff
@@ -55,6 +57,10 @@ export const ROUTE_PZ_LABEL_CREATE_RULE = `/${labelsPath}/:labelId/${rulesPath}/
 export const ROUTE_PZ_ZONE_UPDATE_RULE = `/${zonesPath}/:zoneId/${rulesPath}/:ruleId/${savePath}`;
 export const ROUTE_PZ_LABEL_UPDATE_RULE = `/${labelsPath}/:labelId/${rulesPath}/:ruleId/${savePath}`;
 
+export const ENVIRONMENT_AGGREGATION_SUPPORTED_ROUTES: (string | PathPattern)[] = [
```
urangel, line 60: Thanks for making this a constant!

packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
```diff
@@ -39,14 +40,19 @@ export const getEnvironmentAggregationIds = (
 };
 
 export const useEnvironmentIdList = (
-    ENVIRONMENT_AGGREGATION_SUPPORTED_ROUTES: (string | PathPattern)[]
+    ENVIRONMENT_AGGREGATION_SUPPORTED_ROUTES: (string | PathPattern)[],
+    collectedOnly: boolean
```
urangel: What is the use case for the `collectedOnly` boolean?

packages/javascript/bh-shared-ui/src/hooks/useAvailableEnvironments/useAvailableEnvironments.tsx
```diff
@@ -59,7 +59,7 @@ export const useSelectedEnvironment = (
     };
 
     const setEnvironmentAggregation = (aggregation: EnvironmentQueryParams['environmentAggregation']) => {
-        setEnvironmentParams({ environmentAggregation: aggregation });
+        setEnvironmentParams({ environmentId: null, environmentAggregation: aggregation });
```
urangel: I wonder if there is a reason that this does not currently null the `environmentId`.

benwaples: Great question! The way the `setEnvironmentParams` works will only remove a param if specifically set to an "empty value": https://github.com/SpecterOps/BloodHound/blob/main/packages/javascript/bh-shared-ui/src/utils/searchParams/searchParams.ts#L42

packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
```diff
@@ -21,13 +21,14 @@ import { useMatchingPaths } from './useMatchingPaths';
...
     environmentAggregation: EnvironmentAggregation,
-    environments: Environment[]
+    environments: Environment[],
+    collectedOnly: boolean
 ) => {
     const aggregationIds: string[] = [];
 
     environments.forEach((environment) => {
         if (
-            environment.collected &&
```
specter-flq, line 30: Maybe we could add a comment to explain what this conditional is checking for

packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
```diff
@@ -39,14 +40,19 @@ export const getEnvironmentAggregationIds = (
 };
 
 export const useEnvironmentIdList = (
-    ENVIRONMENT_AGGREGATION_SUPPORTED_ROUTES: (string | PathPattern)[]
+    ENVIRONMENT_AGGREGATION_SUPPORTED_ROUTES: (string | PathPattern)[],
```
specter-flq, line 44: Im not 100% sure on this but i will throw it out there, if we are passing the constant as an argument and the argument is actually expecting the constant's value can we just import it directly in the file? If not should we then change the name of the argument to something that expects to be of different values. ( sorry if its confusing, we can talk through it if needed )

packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
specter-flq: +1 to the question and maybe to have a more descriptive name to it, not sure it is super clear what it does at first glance, but I'm also a bit out of context ( and its still the first week. of january :) )

benwaples: It appears this hook is used a handful of times in BHE as well! To avoid conflicts after merging I think we either need to add a fallback to this param (cleanest/easiest IMO), or we have to update the call sights in BHE and add the extra param!

Holocraft: So the posture page and findings will include the collected count which we don't want to count on PZ. Adding this, we can specify which views will include the collected counts.

packages/javascript/bh-shared-ui/src/hooks/useAvailableEnvironments/useAvailableEnvironments.tsx
benwaples: Im curious if there is a way that avoids removing the `environtmentId` from params so we can maintain that piece of state? It appears that even if the environmentId existed params, `useEnvironmentIdList` would return all environment ids included in the selected `environmentAggregation` value, including the one thats getting removed.

Holocraft: I think we can remove setting the `environmentId` to null and keep it in the params. I just tested it and it's working locally.

packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
Holocraft: I can set it to default to true so we have a fallback.

packages/javascript/bh-shared-ui/src/hooks/useEnvironmentIdList.tsx
Holocraft: Yeah we are importing the constant into each file we're using it in. Check out the [Details.tsx](https://github.com/SpecterOps/BloodHound/blob/3d6ffd61ea7390ae0cc34e90b93a4fd3e44258df/packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Details/Details.tsx#L72) file for an example.

specter-flq: yeah i get that part but does it change by feature or view ? or is it always the same across the app? because if its going to be always the same why not just import it once here and not pass it as an argument

mistahj67: It is different on BHE pages

specter-flq: Yup that makes sense totally missed that apologies and thanks for the clarification

Holocraft: Oh I see what you mean. I didn't implement this but it looks like the `useMatchingPaths` hook is changing the `isEnvironmentAggregationSupportedPage` boolean based on the query params. Since the query params are changing, this is why?

specter-flq: That makes sense as well thanks man. But i meant more on the constants usage but its what Trevor mentioned. I just missed it.

PR #2221, opened 2026-01-06
refactor: AuthExtensions to Support Bearer Authentication Validation - BED-7073

mykeelium:
## Description
The changes in this PR update the Authentication Logic to allow the extension of logic while validating a bearer token in the AuthMiddleware. In addition, some previously extended logic was unified into a struct that is now at the API layer rather than the Database layer. 

## Motivation and Context

Resolves BED-7073

## How Has This Been Tested?

This has been tested by creating an instance and extending this logic here to be able to test the extended functionality in order to authenticate using AD FS. 

## Types of changes

- 
## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Optional bearer-token authentication for clients (feature-flagged)
  * Pluggable authentication extensions and explicit bearer-token validation path
  * Client configuration schema extended with issuer address and override fields
  * Configuration helper to produce root URL host string

* **Validation**
  * New conditional validator enforcing required fields when a specific auth type is used

* **Tests**
  * Expanded authentication tests covering token parsing, signature and request validation

### DIFF
--- a/cmd/api/src/api/auth_test.go
+++ b/cmd/api/src/api/auth_test.go
@@ -48,9 +48,6 @@ import (
 
 func Test_NewRequestSignature(t *testing.T) {
 	t.Run("returns error on context timeout", func(t *testing.T) {
-		ctrl := gomock.NewController(t)
-		defer ctrl.Finish()
-
 		req, err := http.NewRequest(http.MethodGet, "http://teapotsrus.dev", nil)
 		require.NoError(t, err)
 
@@ -65,9 +62,6 @@ func Test_NewRequestSignature(t *testing.T) {
 	})
 
 	t.Run("returns error on empty hmac signature", func(t *testing.T) {
-		ctrl := gomock.NewController(t)
-		defer ctrl.Finish()
-
 		req, err := http.NewRequest(http.MethodGet, "http://teapotsrus.dev", nil)
 		require.NoError(t, err)
 
@@ -81,9 +75,6 @@ func Test_NewRequestSignature(t *testing.T) {
 	})
 
 	t.Run("success", func(t *testing.T) {
-		ctrl := gomock.NewController(t)
-		defer ctrl.Finish()
-
 		req, err := http.NewRequest(http.MethodGet, "http://teapotsrus.dev", nil)
 		require.NoError(t, err)
 
@@ -272,11 +263,10 @@ func TestValidateRequestSignature(t *testing.T) {
 		mockDB := dbMocks.NewMockDatabase(ctrl)
 		mockAuthExtensions := apimocks.NewMockAuthExtensions(ctrl)
 
-		tempDir := t.TempDir()
 		cfg := config.Configuration{
-			WorkDir: tempDir,
+			WorkDir: t.TempDir(),
 		}
-		os.Mkdir(cfg.TempDirectory(), 0755)
+		os.Mkdir(cfg.TempDirectory(), 0o755)
 
 		authenticator := api.NewAuthenticator(cfg, mockDB, mockAuthExtensions)
 
@@ -306,7 +296,7 @@ func TestValidateRequestSignature(t *testing.T) {
 
 		// Closing the body should remove the tmp file
 		req.Body.Close()
-		tmpFiles, err = os.ReadDir(tempDir)
+		tmpFiles, err = os.ReadDir(cfg.TempDirectory())
 		assert.NoError(t, err)
 		assert.Len(t, slicesext.Filter(tmpFiles, func(file fs.DirEntry) bool {
 			return strings.HasPrefix(file.Name(), "bh-request-")


cmd/api/src/api/auth.go
```diff
@@ -523,21 +583,3 @@ func (s authenticator) ValidateSession(ctx context.Context, jwtTokenString strin
 		return authContext, nil
 	}
 }
-
-type LoginRequest struct {
```
mykeelium, line 527: Moved to model layer due to a circular dependency with api+mock+test

cmd/api/src/config/config.go
```diff
@@ -365,3 +365,7 @@ func generateCollectorManifest(collectorDir string) (CollectorManifest, error) {
 		Versions: collectorVersions,
 	}, nil
 }
+
+func (s Configuration) GetRootURLHost() string {
+	return fmt.Sprintf("%s://%s", s.RootURL.Scheme, s.RootURL.Hostname())
+}
```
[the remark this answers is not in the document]
juggernot325: I looked into this and we don't perform validation on any of the values put into the configuration file. I do think we should start doing that at some point, but that's out of scope for this PR.

cmd/api/src/api/auth.go
```diff
@@ -17,6 +17,7 @@
 package api
 
 //go:generate go run go.uber.org/mock/mockgen -copyright_file=../../../../LICENSE.header -destination=./mocks/authenticator.go -package=mocks . Authenticator
+//go:generate go run go.uber.org/mock/mockgen -copyright_file=../../../../LICENSE.header -destination=./mocks/authExtensions.go -package=mocks . AuthExtensions
```
superlinkx: This extra generate line is going to slow down generation and create an additional mock file. It would be better to use the existing generate and append AuthExtensions to it:
```go
//go:generate go run go.uber.org/mock/mockgen -copyright_file=../../../../LICENSE.header -destination=./mocks/authenticator.go -package=mocks . Authenticator,AuthExtensions
```

cmd/api/src/api/auth.go
```diff
@@ -460,24 +515,29 @@ func (s authenticator) CreateSession(ctx context.Context, user model.User, authP
 	}
 }
 
-func (s authenticator) jwtSigningKey(token *jwt.Token) (any, error) {
-	return s.cfg.Crypto.JWT.SigningKeyBytes()
+func (s AuthenticatorBase) ValidateBearerToken(ctx context.Context, jwtToken string) (auth.Context, error) {
+	if claims, err := s.authExtensions.ParseClaimsAndVerifySignature(ctx, jwtToken); err != nil {
+		return auth.Context{}, err
+	} else if authContext, err := s.authExtensions.InitContextFromClaims(ctx, claims); err != nil {
+		slog.ErrorContext(ctx, fmt.Sprintf("Error initializing auth context from claims: %v", err))
```
superlinkx, line 539: When writing new slog statements, we need to make sure we're using structured syntax (the historical reason we have so many Sprintf's is an artifact of converting old logs to slog quickly, and we're still working on cleaning them up).

For this statement, I'd suggest:

```go
slog.ErrorContext(ctx, "Error initializing auth context from claims", attr.Error(err))
```

cmd/api/src/api/auth.go
superlinkx: This is unfortunate, as they really do belong in the view layer. However, I attempted to solve another way and without refactor to fix older sins, this is too much effort to solve correctly right now.

cmd/api/src/api/auth.go
```diff
@@ -460,24 +515,29 @@ func (s authenticator) CreateSession(ctx context.Context, user model.User, authP
...
+		// The above logic is currently used to determine if the token is created from BloodHound. If nil, it was created by BloodHound.
+		slog.DebugContext(ctx, "No owner claim found for token, defaulting to BloodHound provided token")
+		if authContext, err = s.ValidateSession(ctx, claims.ID); err != nil {
+			return auth.Context{}, err
+		} else {
+			return authContext, nil
+		}
+	} else {
+		return authContext, nil
+	}
```
bsheth711, line 551: I don't understand this. This could just be due to my unfamiliarity with auth, but InitContextFromClaims() just returns a new auth.Context{}. The authContext.Owner will always be nil then.

juggernot325: I'm making some assumptions because I didn't write this code. But I'm thinking this is basically just a placeholder function in BHCE to satisfy the `AuthExtensions` interface. It's not intended for BHCE clients to auth via Bearer token, because BHCE doesn't have client auth at all. The BHE PR has this logic more fleshed out, which I'm assuming is because that's where this is actually utilized.

mykeelium: That is correct, this is the current logic that is being used to determine if a request comes from a client, if the owner is nil, see [this code here](https://github.com/SpecterOps/BloodHound/blob/b9e22919a3f13e043009e1b594b87a4eed5eef23/cmd/api/src/api/auth.go#L524) for that check in BHCE. This was done to extend the functionality to find clients specifically for Enterprise. The functionality of this method is largely here to be able to extend the functionality in BHE.

If the owner is nil from this function, then [this function](https://github.com/SpecterOps/BloodHound/blob/b9e22919a3f13e043009e1b594b87a4eed5eef23/cmd/api/src/api/auth.go#L537) is called, which is the current logic for bearer validation in BHCE, [seen here.](https://github.com/SpecterOps/BloodHound/blob/e58a45b7ae275f5eb2c07a87b2d6f2fdfde76032/cmd/api/src/api/auth.go#L467)  Note some logic was moved to refine the session function, for instance, some logic pertaining to jwt claims was moved into [this function here](https://github.com/SpecterOps/BloodHound/blob/b9e22919a3f13e043009e1b594b87a4eed5eef23/cmd/api/src/api/auth.go#L106). 

see [this here ](https://github.com/SpecterOps/bloodhound-enterprise/blob/dcc4bb0e3d6568e85ddf4bcefd87fe55e4d32de1/lib/go/api/v2/resources/auth.go#L165) for the updated functionality that overrides this functionality in BHE.

See [this here](https://github.com/SpecterOps/bloodhound-enterprise/blob/7fbbf3dddbb65f37b42ff297a3df3b7025908d3f/lib/go/auth/auth.go#L12) for an example of a later check being done currently.

bsheth711: I see, I understand now. Thank you for the explanation.

cmd/api/src/api/auth.go
```diff
@@ -460,24 +532,29 @@ func (s authenticator) CreateSession(ctx context.Context, user model.User, authP
...
-		}
-
-		return auth.Context{}, err
-	} else if !token.Valid {
-		slog.InfoContext(ctx, "Token invalid")
-		return auth.Context{}, ErrInvalidAuth
-	} else if sessionID, err := claims.SessionID(); err != nil {
-		slog.InfoContext(ctx, fmt.Sprintf("Session ID %s invalid: %v", claims.Id, err))
+	if sessionID, err := strconv.ParseInt(claimsID, 10, 64); err != nil {
+		slog.InfoContext(ctx, fmt.Sprintf("Session ID %s invalid: %v", claimsID, err))
```
superlinkx, line 557:
```go
slog.InfoContext(ctx, "Session ID is invalid", slog.String("claims_id", claimsID), attr.Error(err))
```

cmd/api/src/api/auth.go
juggernot325: We're going to address these in a future PR. A ticket has been opened to track it.

PR #2147, opened 2025-12-08
feat(dogtags): Adds DogTags functionality for SKU Flagging BED-6650

kpom-specter:
Adds the plumbing and structure for our SKU level flags

## DogTags
Add a sku level flagging system with providers to be implemented in enterprise only

## Motivation and Context

Resolves BED-6650

## Types of changes

- New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Associated an issue: https://github.com/SpecterOps/BloodHound/issues/672
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [ ] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments (GoDocs / JSDocs)
- [ ] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

## Summary by CodeRabbit

* **New Features**
  * Added a typed feature-flag system (bool/string/int) with a no-op provider and a new API endpoint to fetch all flags.

* **Behavior / Chore**
  * Service initializes at startup, is injected into route setup, and logs an initial configuration snapshot (defaults applied).

* **Tests**
  * Unit tests validating default values and retrieval behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #2226, opened 2026-01-07
docs(bh-rfc-6): DogTags (SKU Feature Entitlements) BED-7094

kpom-specter:
## Description
DogTags RFC, to be paired with https://github.com/SpecterOps/BloodHound/pull/2147

## Motivation and Context
BED-7094

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC describing a SKU-based "Dogtags" feature entitlements system with a provider abstraction, runtime provider injection, startup-time loading/caching, and fail-fast behavior on misconfiguration.
  * Documents a new API endpoint to retrieve entitlements as a flat, dot-notated JSON payload, guidance for adding SKU flags, and testing recommendations.

### DIFF
--- a/rfc/bh-rfc-6.md
+++ b/rfc/bh-rfc-6.md
@@ -79,6 +79,7 @@ Dogtags are exposed via `GET /api/v2/dog-tags`. The response is always a flat ma
 ```json
 {
     "data": {
+        "auth.environment_targeted_access_control": false,
         "privilege_zones.tier_limit": 3,
         "privilege_zones.label_limit": 10,
         "privilege_zones.multi_tier_analysis": false
@@ -118,12 +119,25 @@ if s.DogTags.GetFlagAsBool(dogtags.MY_NEW_FEATURE) {
 
 The service handles defaults, so you never need to check errors.
 
+### 5.4 Migrating Existing Flags
+
+When moving a configuration value to dogtags, all previous references must be deprecated and removed. The old configuration endpoint, any backend code reading from the parameters table, and any UI components consuming the old values should be updated to use the dogtags service instead. Leaving stale references creates a confusing split where the displayed value may not match what the application actually uses.
+
 ## 6. Testing
 
-Use `NoopProvider` for tests - it always returns errors, forcing the service to use defaults. For testing specific flag values, create a mock provider.
+For tests that don't care about specific flag values, use `NewDefaultService()` which returns defaults for everything:
 
 ```go
-dogtagsService := dogtags.NewDefaultService() // uses NoopProvider
+dogtagsService := dogtags.NewDefaultService()
+```
+
+For tests that need specific flag values, use `NewTestService` with `TestOverrides`. Values not in the overrides map fall back to defaults:
+
+```go
+dogtagsService := dogtags.NewTestService(dogtags.TestOverrides{
+    Bools: map[dogtags.BoolDogTag]bool{dogtags.PZ_MULTI_TIER_ANALYSIS: true},
+    Ints:  map[dogtags.IntDogTag]int64{dogtags.PZ_TIER_LIMIT: 5, dogtags.PZ_LABEL_LIMIT: 3},
+})
 ```
 
 ## 7. File Locations


rfc/bh-rfc-6.md
```diff
@@ -0,0 +1,135 @@
...
+    GetFlagAsInt(key IntDogTag) int64
+    GetAllDogTags() map[string]any
+}
+```
+
+Note the typed keys (`BoolDogTag`, `IntDogTag`, etc.) - this prevents mixing up flag types at compile time.
+
+### 4.3 API Endpoint
+
+Dogtags are exposed via `GET /api/v2/dog-tags`:
```
mistahj67: Can there be a callout that the response here should always maintain a flat map?

rfc/bh-rfc-6.md
```diff
@@ -0,0 +1,149 @@
...
+
+Providers are injected at runtime. The application receives a configured provider at startup and uses it for all entitlement lookups. Provider implementation details are intentionally hidden from the application. The FOSS application uses a no-op provider, meaning default values in the SKU lookup field will apply.
+
+### 3.2 Startup-Time Loading
+
+Entitlements are loaded once at startup and cached. Changes require a restart. This is intentional; Runtime updates add complexity and create less predictable behavior.
+
+### 3.3 Fail-Fast
+
+If a provider can't load its configuration, the application fails to start. Silent fallbacks mask misconfigurations and make debugging painful.
```
urangel, line 36: Were I an end user, my confidence in the application may be negatively impacted if startup fails due to an upstream configuration provider being flakey. Would an end user know to tell when there is a problem with a configuration provider vs the rest of the application code?

I see that the default values are mainly intended for the community edition but would it make sense to fallback to defaults for enterprise as well instead of not starting?

rfc/bh-rfc-6.md
```diff
@@ -0,0 +1,149 @@
...
+
+### 4.1 Provider Interface
+
+Providers implement a simple interface:
+
+```go
+type Provider interface {
+    GetFlagAsBool(key string) (bool, error)
+    GetFlagAsString(key string) (string, error)
+    GetFlagAsInt(key string) (int64, error)
```
urangel, line 54: Dogs like to fetch! 🐶 

```suggestion
    FetchFlagAsBool(key string) (bool, error)
    FetchFlagAsString(key string) (string, error)
    FetchFlagAsInt(key string) (int64, error)
```

Just a thought, no need to run with it 🙂

rfc/bh-rfc-6.md
```diff
@@ -0,0 +1,149 @@
...
+if s.DogTags.GetFlagAsBool(dogtags.MY_NEW_FEATURE) {
+    // feature enabled
+}
+```
+
+The service handles defaults, so you never need to check errors.
+
+### 5.4 Migrating Existing Flags
+
+When moving a configuration value to dogtags, all previous references must be deprecated and removed. The old configuration endpoint, any backend code reading from the parameters table, and any UI components consuming the old values should be updated to use the dogtags service instead. Leaving stale references creates a confusing split where the displayed value may not match what the application actually uses.
```
urangel, line 124: I agree stale references create clutter and confusion though am a bit wary of quickly removing support. Would removing the old configuration endpoint constitute a breaking change in the API? Should we return a deprecated message and show deprecated tags in the UI for those endpoints for a while before removing them. A sort of grace period. Though I suppose if the intent is to keep things simple, keeping track of grace periods is the other direction

rfc/bh-rfc-6.md
kpom-specter: Each provider can fail as gracefully/not as they see fit. For the community provider, it is just a passthrough, no concern. For the other providers, our internal fleet behaves the same. Replicated requires a license at startup. If you don't have the license present, we currently hard crash the app. There's more functionality we could build to extend this behavior and make it more user friendly, but I don't see the use case. We'll be hand holding customers going through this, at least for the forseeable future.

rfc/bh-rfc-6.md
kpom-specter: I would argue `Get*` implies a local lookup, returning something quick and at hand. `Fetch*` implies a lookup, network or IO in a way you might consider making it async. Since the flags are loaded at startup and kept in memory, `Get*` feels slightly more appropriate. 

(I appreciate the pun quite a bit though, I'll admit I missed it initially)

rfc/bh-rfc-6.md
kpom-specter: So right now, the config endpoint returns PZ values. The argument made against that was the API contract is a key/value store, removing a value from that doesn't break our documented contract.  That being said, it may break code expecting on those keys. 

I think that if we are changing the behavior of enterprise/paid functionality, we should be fully in control of the consumers of this. No external consumers should be relying on "If my bloodhound account has this feature" API call, and we should strive to change it decisively and completely. 

There's some opinion/experience/trauma coming through on this for sure, so if I'm off the mark, we can absolutely build in some sort of sunsetting path for these, but you'll have to really sell me on it. :)

rfc/bh-rfc-6.md
urangel: Ah so this is keeping current behavior, gotcha 👍

rfc/bh-rfc-6.md
urangel: Yea that makes sense to me! I appreciate your commitment to sensible nomenclature

rfc/bh-rfc-6.md
urangel: We've historically been stringent on maintaining API contracts so as not to break external integrations which is why I ask. 

I can see the key/value store nature of the data changing not technically breaking things argument and you're right that we don't have specific keys documented. Seems like that should be fine then! Thanks for breaking this down for me 🙇

urangel (approved):
This looks like it should provide a good solution! Falling back to defaults to simplify error handling sounds nice. Thanks for the thought you put into this 🚀

PR #2248, opened 2026-01-14
feat(api): shortest-path ETAC filtering BED-7131

mvlipka:
## Description

We previously wanted to deny access to all Explore Page endpoints if a user had an ETAC control applied to them.
Eventually we were able to filter graphs meaning we could now support Explore Page endpoints.

This PR removes the `RequiresAllEnvironmentAccessMiddleware` middleware from the `shortest-paths` endpoint and allows filtering on nodes.

## Motivation and Context

Resolves: BED-7131

## How Has This Been Tested?

I tested with a user that only has access to the `WRAITH.CORP` domain
image

No `Permission Denied` error message is observed

**TESTING REFACTOR**
Query for paths only in a user's etac list
image

Query with no restriction on user's ETAC list
image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added environment-based ETAC graph filtering to hide inaccessible nodes and mark related edges.

* **Bug Fixes**
  * Shortest-path endpoint access simplified to require only graph-read permission.
  * Improved ETAC-aware error handling and response behavior; node properties are correctly omitted when not requested.

* **Tests**
  * Expanded tests covering ETAC filtering, hidden/visible node behavior, and shortest-path flows.

### DIFF
--- a/cmd/api/src/api/v2/cypherquery.go
+++ b/cmd/api/src/api/v2/cypherquery.go
@@ -113,7 +113,7 @@ func (s Resources) CypherQuery(response http.ResponseWriter, request *http.Reque
 	shouldFilterETAC, err := ShouldFilterForETAC(request.Context(), s.DB, user)
 	if err != nil {
 		slog.Error("Unable to check ETAC filtering")
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, "error filtering ETAC graph", request), response)
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, "error determining if ETAC should filter", request), response)
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/api/src/api/v2/etac.go
```diff
@@ -86,3 +90,89 @@ func ShouldFilterForETAC(ctx context.Context, db database.Database, user model.U
 
 	return true, nil
 }
+
```
mvlipka, line 92: This change is due to moving the `filterETACGraph` method to an independent function and passing in the database interface
Logic is not changed other than accessing the database through the argument rather than through `s.db`

cmd/api/src/api/v2/cypherquery.go
```diff
@@ -178,89 +176,3 @@ func (s Resources) cypherMutation(request *http.Request, preparedQuery queries.P
 	return graphResponse, err
 
 }
-
-// filterETACGraph applies ETAC(Environment-based Access Control) filtering for the CypherQuery endpoint.
-// Nodes that the user does not have access to are replaced with hidden placeholder nodes,
-// and edges connected to hidden nodes are marked as hidden.
-func (s Resources) filterETACGraph(ctx context.Context, graphResponse model.UnifiedGraph, user model.User) (model.UnifiedGraph, error) {
```
mvlipka, line 185: This method was moved to a function, see the other comment

cmd/api/src/api/v2/etac.go
superlinkx: What was the impetus for this change? I don't think it matters too much one way or the other but I am curious

superlinkx (approved):

PR #2253, opened 2026-01-15
feat: Local Group Post-Processing Rework - BED-7147

zinic:
## Description

  * Restructured internal analysis data processing to improve efficiency and maintainability of group membership and access control analysis.
  * Removed deprecated analysis function and introduced improved data structures for local group and remote desktop access analysis.

## Motivation and Context

Resolves BED-7147

Local group processing (DCOM, CanPSRemote, AdminTo, CanRDP) utilized a data structure from the original metatree algorithm to rapidly compute reach: inbound reach being exposure and outbound reach being impact. This data structure was deficient in the amount of memory and CPU required to operate it.

This requires a refactor to address as the core data structure in use is the crux of the problem.

## How Has This Been Tested?

Multiple tests were refactored but assertions remained the same, validating existing behavior with the new datastructures and data fetch methods.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Chores**
  * Updated multiple dependencies for security, crypto, tooling, and performance.

* **Refactor**
  * Consolidated local-group and per-computer RDP data into a unified data-driven pipeline; authentication, certificate, sync, ownership and local-group post-processing now use the new pipeline.
  * Pruning tick no longer triggers pruning actions.

* **New Features**
  * Added processing to create edges from local-group memberships and per-computer RDP/URA relationships.

* **Tests**
  * Integration tests updated to exercise the new data-driven and per-computer flows.


packages/go/analysis/ad/local_groups.go
```diff
@@ -0,0 +1,339 @@
...
+				if err := batch.CreateRelationshipByIDs(nextPost.FromID, nextPost.ToID, nextPost.Kind, relProperties); err != nil {
+					return err
+				}
+
+				stats.AddRelationshipsCreated(nextPost.Kind, 1)
+			}
+
+			return nil
+		}); err != nil {
+			slog.Error("Write Computer CanRDP Post Processed Edge", slog.String("err", err.Error()))
```
superlinkx: There's a new helper that ensures we consistently format errors: `attr.Error(err)`

packages/go/analysis/ad/local_groups.go
```diff
@@ -0,0 +1,339 @@
...
+		}
+	}()
```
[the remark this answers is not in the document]
superlinkx: This seems like a valid concern

packages/go/analysis/ad/local_groups.go
```diff
@@ -0,0 +1,339 @@
...
+
+	for workerID := 0; workerID < runtime.NumCPU()/2+1; workerID++ {
```
superlinkx, line 88: Should we promote this calculation to a shared definition somewhere? Also, prior to go1.26, this can have unintended results in environments that set CPU resources (like containers with CPU limits)

packages/go/analysis/ad/local_groups.go
```diff
@@ -0,0 +1,339 @@
...
+	})
+
+	close(workC)
+	workerWG.Wait()
+
+	close(computerC)
+	computerWG.Wait()
+
+	close(postC)
+	postWG.Wait()
```
superlinkx, line 175: Verified the order of operations here, everything looks like it matches up correctly for closing channels

packages/go/analysis/ad/local_groups.go
```diff
@@ -0,0 +1,339 @@
...
+				if err := batch.CreateRelationshipByIDs(nextPost.FromID, nextPost.ToID, nextPost.Kind, relProperties); err != nil {
+					return err
+				}
+
+				stats.AddRelationshipsCreated(nextPost.Kind, 1)
+			}
+
+			return nil
+		}); err != nil {
+			util.SLogError("Write Computer Local Group Post Processed Edge", err)
```
superlinkx: We should probably use `slog.Error("", attr.Error(err))` to match convention elsewhere in BloodHound

packages/go/analysis/ad/local_groups.go
```diff
@@ -0,0 +1,339 @@
...
+
+		go func(workerID int) {
+			defer fetchWG.Done()
+
+			if err := graphDB.ReadTransaction(ctx, func(tx graph.Transaction) error {
+				for {
+					if computerID, shouldContinue := channels.Receive(ctx, computerC); !shouldContinue {
+						break
+					} else {
+						if localAdminGroup, err := FetchComputerLocalGroupIDBySIDSuffix(tx, graph.ID(computerID), adminGroupSuffix); err == nil {
```
superlinkx: Avoid inverting error path handling. This could be rewritten as `!graph.IsErrNotFound(err) && err != nil` to handle the return case first, then have the happy path in an else. This will ensure others won't get confused by the inversion of the nil check and keeps happy path as the final branch

packages/go/analysis/ad/post.go
```diff
@@ -558,133 +491,274 @@ func FetchRemoteInteractiveLogonRightEntities(tx graph.Transaction, computerId g
...
+// FetchCanRDPComputerData uses the given transaction to fetch all computer-specific data related to the given computer
+// that is required to compute the computer's inbound CanRDP edges.
+func (s *CanRDPData) FetchCanRDPComputerData(tx graph.Transaction, computer graph.ID) (*CanRDPComputerData, error) {
+	computerData := &CanRDPComputerData{
+		CanRDPData: *s,
+		Computer:   computer,
+	}
+
+	if dauGroup, err := FetchComputerLocalGroupByName(tx, computer, "Direct Access Users"); err != nil {
+		if !graph.IsErrNotFound(err) {
```
superlinkx, line 705: The error case could be simplified to `err != nil && !graph.IsErrNotFound(err)`

packages/go/analysis/ad/post.go
```diff
@@ -558,133 +491,274 @@ func FetchRemoteInteractiveLogonRightEntities(tx graph.Transaction, computerId g
...
+	return computerData, nil
+}
+
+func FetchRemoteDesktopUsersBitmapForComputer(computer graph.ID, canRDPData *CanRDPComputerData, rdpGroup graph.ID, skipURA bool) (cardinality.Duplex[uint64], error) {
+	if skipURA {
+		return FetchRemoteDesktopUsersLocalGroupMembership(computer, canRDPData), nil
+	} else {
+		return ProcessRDPWithUra(rdpGroup, computer, canRDPData)
+	}
+}
```
superlinkx: Given the caller must know whether it needs URA or not because of the boolean, should we just let the callers of this function decide which to use directly to remove the indirection?

packages/go/analysis/ad/local_groups.go
zinic: Nice! Will patch that up.

packages/go/analysis/ad/local_groups.go
zinic: > Should we promote this calculation to a shared definition somewhere?

Outside of moving this to a function parameter, I'm unsure where we would source this from without full front to back plumbing through the application configuration.

packages/go/analysis/ad/local_groups.go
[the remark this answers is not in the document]
zinic: I had forgotten this! Patched up with a operation specific context using a cancel function. I'm not super interested in bubbling up the error message as post-processing upstream would simply report it.

packages/go/analysis/ad/local_groups.go
zinic: Patched up

packages/go/analysis/ad/local_groups.go
zinic: This was carry over from the previous logic - patched up!

packages/go/analysis/ad/post.go
zinic: The original logic here treats the DAU group as optional. While we could condense the logic, I'd prefer the assignment to only happen in the happy path.

packages/go/analysis/ad/post.go
zinic: Patched this up

packages/go/analysis/ad/local_groups.go
superlinkx: That's fair, I don't have a good suggestion here, I'm fine to let it be a future cleanup effort if we find it's needed

packages/go/analysis/ad/post.go
superlinkx: Oh right, my mistake. Nested ifs sometimes mess with my head, I see it now

superlinkx (approved):
Thanks for all the fixups, looks good!

PR #2264, opened 2026-01-16
feat(QueryLimits): Support bypassing query limits when prefer:wait=-1 header is provided BED-6991

Useinovski:
## Description

If the feature flag for query-limit bypass is enabled and the incoming request includes the header wait: -1, we should skip all query-limit checks and allow the request to execute without restriction.

We set in the configuration parameter in build.config.json '"disable_timeout_limit" : true' which will allow the user to bypass query limits on all end points.

## Motivation and Context

Resolves BED-6991

To be able to bypass query limits for longer running queries that would be timed out with the default timeout set from the UI.

## How Has This Been Tested?

Tested in Bruno, with all cases of the Prefer header value. When using the UI, we have a default prefer header set by client.ts which will be handled in future PR to be removed (BED-6985). Unit tests have also been created and or modified.

This default value is only applied in the search endpoint via the explore page.

**"disable_timeout_limit" : _true_**

Values tested:         Corresponding Status code:
Prefer : "wait=40"   200/201 - Sets timeout of 40 seconds
Prefer : "wait=0"   200/201 - Bypasses query Limits (current behavior)
Prefer : "wait=-1"   200/201 - Bypasses query limits
Prefer : "wait=-30"   400 - Do not allow for negative values < -1

**"disable_timeout_limit" : _false_**

Values tested:          Corresponding Status code:
Prefer : "wait=40"   200/201 - Sets timeout of 40 seconds
Prefer : "wait=0"   200/201 - Bypasses query Limits (current behavior)
Prefer : "wait=-1"   400 - Failed to bypass limits
Prefer : "wait=-30"   400 - Do not allow for negative values < -1

## Screenshots (optional):
**"disable_timeout_limit" : _true_**
Screenshot 2026-01-15 at 11 56 40 AM
Screenshot 2026-01-15 at 11 57 23 AM
Screenshot 2026-01-15 at 11 57 39 AM
Screenshot 2026-01-15 at 11 57 55 AM
Screenshot 2026-01-16 at 10 32 12 AM

**"disable_timeout_limit" : _false_**
Screenshot 2026-01-15 at 11 50 59 AM
Screenshot 2026-01-15 at 11 52 46 AM
Screenshot 2026-01-15 at 11 53 57 AM
Screenshot 2026-01-15 at 11 54 19 AM
Screenshot 2026-01-16 at 10 30 58 AM

**From UI**
Screenshot 2026-01-16 at 9 52 53 AM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Config-driven optional bypass for Prefer wait values; responses indicate when bypass was applied and report the effective wait.
  * Middleware now ensures per-request IDs, timing, timeout indicators, and relevant headers are consistently added to requests/responses.

* **Bug Fixes**
  * Stricter validation and clearer error responses for invalid or disallowed negative Prefer wait values; normalized timeout behavior.

* **Tests**
  * Added/updated tests covering negative waits, bypass-enabled/disabled flows, and Prefer header parsing/validation.

### DIFF
--- a/cmd/api/src/api/middleware/middleware.go
+++ b/cmd/api/src/api/middleware/middleware.go
@@ -84,7 +84,7 @@ func getScheme(request *http.Request) string {
 }
 
 // RequestWaitDuration is responsible for returning a time.Duration if the Prefer header is specified.
-// When bypassLimitsParam is true and wait=-1 is specified, returns -1 to indicate no timeout.
+// When bypassLimitsParam is false and wait=-1 is specified, returns -1 to indicate no timeout.
 // Returns an error if the header value is invalid or if bypass is requested but not enabled.
 func RequestWaitDuration(request *http.Request, bypassLimitsParam bool) (time.Duration, error) {
 	var (
@@ -146,7 +146,6 @@ func ContextMiddleware(bypassLimitsParam bool) mux.MiddlewareFunc {
 					requestCtx, cancel = context.WithTimeout(request.Context(), requestedWaitDuration)
 					defer cancel()
 				} else if requestedWaitDuration == bypassLimit && canBypassLimits {
-					// Bypassing timeout so no context.WithTimeout call, request can run indefinitely
 					response.Header().Set(headers.PreferenceApplied.String(), "wait=-1; bypass=enabled")
 				}
 
--- a/cmd/api/src/api/middleware/middleware_test.go
+++ b/cmd/api/src/api/middleware/middleware_test.go
@@ -110,7 +110,7 @@ func TestContextMiddlewareBypassEnabled(t *testing.T) {
 		req.Header.Set(headers.Prefer.String(), "wait=-1")
 		req.URL.RawQuery = q.Encode()
 		rr := httptest.NewRecorder()
-		wrapHandler := middleware.ContextMiddleware(true)(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {}))
+		wrapHandler := middleware.ContextMiddleware(false)(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {}))
 		wrapHandler.ServeHTTP(rr, req)
 		require.Equal(t, http.StatusOK, rr.Code)
 	}
@@ -126,7 +126,7 @@ func TestContextMiddlewareBypassDisabled(t *testing.T) {
 		req.Header.Set(headers.Prefer.String(), "wait=-1")
 		req.URL.RawQuery = q.Encode()
 		rr := httptest.NewRecorder()
-		wrapHandler := middleware.ContextMiddleware(false)(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {}))
+		wrapHandler := middleware.ContextMiddleware(true)(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {}))
 		wrapHandler.ServeHTTP(rr, req)
 		require.Equal(t, http.StatusBadRequest, rr.Code)
 	}


cmd/api/src/api/middleware/middleware.go
```diff
@@ -87,11 +87,17 @@ func RequestWaitDuration(request *http.Request) (time.Duration, error) {
...
 	)
+	const bypassLimit = time.Second * time.Duration(-1)
 
 	if preferValue := request.Header.Get(headers.Prefer.String()); len(preferValue) > 0 {
 		if requestedWaitDuration, err = parsePreferHeaderWait(preferValue); err != nil {
 			return 0, err
+		} else if requestedWaitDuration < bypassLimit {
+			return 0, errors.New("incorrect bypass limit value")
+		} else if requestedWaitDuration == bypassLimit && !canBypassLimits {
+			return 0, errors.New("failed to bypass limits")
```
irshadaj: Let's add some detail to the return statement here: `failed to bypass limits: feature disabled`

cmd/api/src/api/middleware/middleware.go
```diff
@@ -132,13 +140,15 @@ func ContextMiddleware(next http.Handler) http.Handler {
...
+			} else if requestedWaitDuration == bypassLimit && canBypassLimits {
+				response.Header().Set(headers.PreferenceApplied.String(), fmt.Sprintf("wait=%.2f; bypass=enabled", requestedWaitDuration.Seconds()))
 			}
 
 			// Insert the bh context
 			requestCtx = ctx.Set(requestCtx, &ctx.Context{
 				StartTime: startTime,
 				RequestID: requestID,
-				Timeout:   requestedWaitDuration,
+				Timeout:   setUserTimeout(requestedWaitDuration, bypassLimit),
```
irshadaj: Setting a zero timeout here will mean the context is already expired. When you want to bypass the timeout, you will instead need to create a new context an empty (NOT zero) `Timeout` field. Here, check out how these dynamics play out:
https://go.dev/play/p/pWTyGNo0R-U

You'll need to use the third block for the case where the limit is being bypassed.

cmd/api/src/api/middleware/middleware.go
```diff
@@ -132,13 +140,15 @@ func ContextMiddleware(next http.Handler) http.Handler {
 
 				requestCtx, cancel = context.WithTimeout(request.Context(), requestedWaitDuration)
 				defer cancel()
+			} else if requestedWaitDuration == bypassLimit && canBypassLimits {
+				response.Header().Set(headers.PreferenceApplied.String(), fmt.Sprintf("wait=%.2f; bypass=enabled", requestedWaitDuration.Seconds()))
```
stephanieslamb: Also, since we aren't using time, this can just be something like:
`response.Header().Set(headers.PreferenceApplied.String(), "wait=-1; bypass=enabled")

cmd/api/src/api/middleware/middleware.go
```diff
@@ -101,9 +107,11 @@ func RequestWaitDuration(request *http.Request) (time.Duration, error) {
 func ContextMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(response http.ResponseWriter, request *http.Request) {
 		var (
-			startTime = time.Now()
-			requestID string
+			startTime       = time.Now()
+			requestID       string
+			canBypassLimits = config.GetLoadedConfig().DisableTimeoutLimit
 		)
+		const bypassLimit = time.Second * time.Duration(-1)
```
stephanieslamb: same as above

cmd/api/src/api/middleware/middleware.go
```diff
@@ -87,11 +87,17 @@ func RequestWaitDuration(request *http.Request) (time.Duration, error) {
 	var (
 		requestedWaitDuration time.Duration
 		err                   error
+		canBypassLimits       = config.GetLoadedConfig().DisableTimeoutLimit
 	)
+	const bypassLimit = time.Second * time.Duration(-1)
```
stephanieslamb, line 95: I like this to avoid magic numbers! But I think since this is just a sentinel value, we dont need to convert to seconds.  It can just be `const bypassLimit = -1`

cmd/api/src/api/middleware/middleware.go
stephanieslamb: I might be wrong but I think we can just remove the setUserTimeout call and this will work as expected. On line 141 we're setting context.WithTimeout(Go context) to whatever the requestedWaitDuration is. If we're bypassing timeouts, we have our sentinel value of -1, or if no timeout is specified it will be 0. Since we're just setting the Timeout field in bhCtx(not Go context), we can store the value directly with no conversions.

cmd/api/src/api/middleware/middleware.go
Useinovski: We set requestedWaitDuration, err = parsePreferHeaderWait(preferValue) and parsePreferHeaderWait will return a time.Duration.
Therefore the if check of requrestedWaitDuration < bypassLimit we get incorrect behavior always resulting in true if we pass in a -1. This should only be true for values less than -1
-1000000000 < -1 is the comparison

I also tried to change the return value we get from parsePreferHeaderWait() to return only a time.Duration and not a time.Second * time.duration() but resulted in getting a 401 "Token Authorization failed" so I thought this was the best approach to not breaking anything else in the application.

cmd/api/src/api/middleware/middleware.go
Useinovski: I believe I agree with Stephanie here. I believe contextWithTimeout() is still responsible for handling the timeout and not the Timeout: field in our struct.

We have a "Context" struct in ctx.go that stores a time.duration but doesn't seem to do anything with it beyond that. I believe for logging purposes? But I don't see anything indicating that the Timeout: field is responsible for handling the timeout.

Useinovski: My logic behind the setUserTimeouts() function was to not pass around negative values in Context for the Timeout: field. I don't believe a negative, 0, or positive time.duration() value will directly affect the functionality. I'm more than happy to revert it though if we think this is overly excessive.

cmd/api/src/api/middleware/middleware.go
irshadaj: Might be worthwhile adding another constant to make that clear.
```const (
bypassLimitFlag=-1
bypassLimit = time.Second * time.Duration(bypassLimitFlag)
)
```

or if you have a better name for it feel free to use something else

cmd/api/src/api/middleware/middleware.go
irshadaj: I did indeed miss the block above this, where we're setting timeout only if requestedWaitDuration>0. I'm good with the changes here. As for `setUserTimeout` you can use go's builtin functions to achieve this without having to declare a separate function.

`time.Duration(max(requestedWaitDuration.Seconds(), 0))` should give you 0 seconds

cmd/api/src/api/middleware/middleware.go
stephanieslamb: So in parsePreferHeaderWait, we can add a check if we are bypassing the limit and just return -1.  Something like this:
```
func parsePreferHeaderWait(value string) (time.Duration, error) {
	if waitStr, hasWaitSpecified := ParseHeaderValues(value)["wait"]; hasWaitSpecified {
		if parsedNumSeconds, err := strconv.Atoi(waitStr); err != nil {
			return 0, err
		} else {
		         if parsedNumSeconds == -1 {
                            return -1, nil
                          }
			return time.Second * time.Duration(parsedNumSeconds), nil
		}
	} else {
		return 0, errors.New("leave field empty or specify with : 'wait=x'")
	}
}
```
Just add a comment that the check is for bypassing query limits.

Useinovski: I like this because it kicks out of the function sooner but we only return a -1 which resolves to -1 nano second. We would still need to multiply by time.Second() to set the requestedWaitDuration in func RequestWaitDuration() so that it is still being handled properly in the rest of the ContextMiddleware Func as well as logging.go

Please let me know if I missed something in my breakdown

stephanieslamb: So basically we are not using the -1 as actual seconds.  It is just a sentinel value.  So here would be the flow:
1. parsePreferHeaderWait returns -1 (the integer, as a sentinel)
2. RequestWaitDuration returns -1 
3. In the middleware, we check: `if requestedWaitDuration == -1` 
4. When it's -1, we don't call context.WithTimeout at all (no timeout enforcement)
5. We store -1 in the bhCtx for logging/tracking
-1 is never being used as an actual time.Duration.  Does this make sense?

PR #2266, opened 2026-01-16
fix(ui): Update lazy loading components to avoid UI crashes BED-7170

maffkipp:
## Description
A particular lazy-loading pattern we were using for pages built entirely in `bh-shared-ui` was causing 429 errors in production when fetching assets that would crash the UI on browser refresh.

This adds support to `bh-shared-ui` for exporting those components individually so we can import them for lazy loading with a path like`import('bh-shared-ui/FileIngest')`. The downside to this approach is needing to add a shim for each component to the top-level `src/` directory so that typescript can correctly associate that export with its definition file.

If this fix seems too intrusive, we could also instead add wrapper component files for each page to each UI app, and then lazy load those wrapper components instead.

## Motivation and Context

Resolves BED-7170

## How Has This Been Tested?
This bug does not occur with our local dev setup; I tested it out locally by building the production image and running it with docker-compose. Detailed testing instructions are in the ticket!

## Types of changes

## Checklist:
  - Assigned myself to this PR
  - Open API docs


superlinkx (approved):
Leaving this as a codeowner unblock, no concerning changes to dependencies. Should still get a proper review of the UI change

packages/javascript/bh-shared-ui/src/views/Users/UsersWithEnvironmentAccessControls.tsx
elikmiller, line 1: Why did this file get deleted?

elikmiller: Nevermind - I can see it was moved to the enterprise repository.

elikmiller (approved):

PR #2314, opened 2026-01-28
fix(PZ): GET /asset-group-tags/:tagId/members source kind filtering is broken BED-7089

ALCooper12:
## Description
- Provided an additional check that checks if the filter(s) provided is a `primary_kind` filter, grabs the current source kinds via the `GetSourceKinds`db method, and then if the `primary_kind` filter provided is of a source kind: only calls the `GetGraphSchemaNodeKinds` method once so that the DAWGS functions used to process the AGT nodes can be efficiently populated and filtered using the results from both db methods above for the `primary_kind` filter beneath the surface
-  Added 2 `ToKind` methods to the graphschema model

## Motivation and Context
This PR addresses:
- BED-7089

- At the moment, the `GET /asset-group-tags/:tagId/members` endpoint works perfectly fine when using the `primary_kind` filter for members/nodes that are NOT source kind nodes. If one were to use: `primary_kind=eq:Base`, the endpoint would return all members instead of just those whose primary kind is its source kind. This change fixes that bug

## How Has This Been Tested?
- Locally within my BHE/BHCE env

## Screenshots (optional):
Screenshot 2026-01-28 at 2 23 53 PM

Screenshot 2026-01-28 at 2 24 05 PM

Screenshot 2026-01-28 at 2 24 41 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Asset group member filtering now correctly handles primary-kind filters that reference source kinds, respects negated primary-kind conditions, and gracefully handles filter-builder errors to avoid applying empty filters.

* **Refactor**
  * Improved graph schema and source-kind handling with lazy loading to build accurate composite filters and reduce redundant lookups.

* **Tests**
  * Tests updated to exercise the source-kind lookup path.


cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -971,17 +983,72 @@ func buildAssetGroupMembersByTagGraphDbFilters(queryFilterMap model.QueryParamet
...
+
+					var sourceKindRef graph.Criteria
+					switch queryFilter.Operator {
+					case model.Equals:
+						sourceKindRef = query.KindIn(query.Node(), graph.StringKind(queryFilter.Value))
+					case model.NotEquals:
+						var filteredSourceKinds []string
+
+						// if the neq: operator is used, we want to filter to the other source kinds, still skipping tag kinds
+						if queryFilter.Operator == model.NotEquals {
```
mistahj67: I think this if block is redundant now
```suggestion
```

cweidenkeller (approved):

PR #2325, opened 2026-01-30
feat(OpenGraph): Add Upsert Extension Endpoint - BED-6723

LawsonWillard:
## Description

Create the Upsert Graph Extension Endpoint 

## Motivation and Context
Resolves: BEd-6723

To support OpenGraph extensions, we need the ability to upload an extension 

## How Has This Been Tested?

- Added new unit tests
- Added new integration tests
- Manually tested upload endpoint

## Types of changes
- Database Migrations

## Checklist:
  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Feature flag added for OpenGraph Extension Management; PUT /api/v2/extensions published to upsert OpenGraph extensions (JSON) and gated by the feature flag plus auth.
  * GET /api/v2/graph-schema/edges reintroduced as a guarded listing endpoint (feature flag + permissions).
* **Improvements**
  * New public request/response shapes and OpenAPI examples for graph extensions.
  * Stronger validation, clearer duplicate-key errors, and automatic graph-kind refresh after upsert.
* **Tests / Chores**
  * Expanded unit/integration tests and DB migration to register the new feature flag.

### DIFF
--- a/cmd/api/src/services/opengraphschema/extension_internal_test.go
+++ b/cmd/api/src/services/opengraphschema/extension_internal_test.go
@@ -412,7 +412,7 @@ func Test_validateGraphSchemaModel(t *testing.T) {
 					},
 				},
 			},
-			wantErr: fmt.Errorf("graph schema environment source kind %s should not be declared as a node or relationship kind", "AD_node_kind_1"),
+			wantErr: fmt.Errorf("graph schema environment source kind %s should not be declared as a node kind", "AD_node_kind_1"),
 		},
 		{
 			name: "fail - environment principal kind missing namespace prefix",


cmd/api/src/database/database_integration_test.go
```diff
@@ -59,14 +64,15 @@ func setupIntegrationTestSuite(t *testing.T) IntegrationTestSuite {
 	err = db.Migrate(ctx)
 	require.NoError(t, err)
 
-	err = db.PopulateExtensionData(ctx)
+	// err = db.PopulateExtensionData(ctx)
 	require.NoError(t, err)
```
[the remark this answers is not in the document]
LawsonWillard: This is intentional for the time being

cmd/api/src/database/kind_integration_test.go
```diff
@@ -73,3 +75,70 @@ func TestGetKindByName(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var (
+				err         error
+				createdKind model.Kind
+				got         model.Kind
+			)
+			createdKind = tt.setup(t)
+			if got, err = testSuite.BHDatabase.GetKindById(testSuite.Context, createdKind.ID); tt.want.err != nil {
+				require.EqualError(t, err, tt.want.err.Error())
```
kpowderly: Can you change all of these `require`'s to `assert` instead? It's the same library.

if `require` fails, it immediately stops the execution of the test, failing immediately. An assert will allow the test to continue executing so we can get all of the failures at once.

I think generally the rule of thumb is to only use require on setting up the test or specifically for things that we're not trying to assert on.

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+				fmt.Sprintf("invalid content-type: %s", request.Header[headers.ContentType.String()])), request), response)
 		return
 	}
 
-	if err := s.OpenGraphSchemaService.UpsertGraphSchemaExtension(ctx, req); err != nil {
-		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("error upserting graph schema extension: %v", err), request), response)
+	if graphExtensionPayload, err = extractExtensionData(request.Body); err != nil {
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
 		return
+	} else if updated, err = s.OpenGraphSchemaService.UpsertOpenGraphExtension(ctx,
```
kpowderly, line 148: If we're no longer updating extensions and we're doing a cascade delete and re-creating the extension, do we really need any of this "updated" logic on if it previously existed? I feel like we should just log that it existed and was deleted for our own tracking purposes but that's it.

It would dramatically simplify this func, UpsertOpenGraphExtension at the service level, and UpsertOpenGraphExtension at the database level.

I think the caller will only care that the new extension was created.

cmd/api/src/database/upsert_schema_extension.go
```diff
@@ -13,53 +13,153 @@
...
+	// Check for an immediate error after beginning the transaction
+	if err = tx.Error; err != nil {
+		return false, err
+	}
+
+	defer func() {
+		tx.Rollback() // rollback is a no-op if the tx has already been committed
+	}()
+
+	if existingGraphExtensions, _, err = bloodhoundDBTransaction.GetGraphSchemaExtensions(ctx,
```
kpowderly: I think it's probably cleanest if we move this into a helper function to make this logic more readable.

I made another comment but I'm hoping we can get rid of some of this `schemaExists` logic since we're deleting and re-creating.

Something along these line...

```
func (s *BloodhoundDB) cleanupExistingExtension(ctx context.Context, tx BloodhoundDB, graphExtensionInput model.GraphExtensionInput) (bool, error) {
	var (
		schemaExists            bool
		existingGraphExtensions model.GraphSchemaExtensions
	)

	// Check if Extension exists filtering by extension name
	existingGraphExtensions, _, err := tx.GetGraphSchemaExtensions(ctx,
		model.Filters{
			"name": []model.Filter{
				{
					Operator:    model.Equals,
					Value:       graphExtensionInput.ExtensionInput.Name,
					SetOperator: model.FilterAnd,
				},
			},
		}, model.Sort{}, 0, 1)
	if err != nil {
		if !errors.Is(err, ErrNotFound) {
			return schemaExists, err
		}
		return schemaExists, nil
	}

	if len(existingGraphExtensions) == 0 {
		return schemaExists, nil
	}

	// Extension exists
	schemaExists = true
	existingGraphExtension := existingGraphExtensions[0]

	// Check if it's a built-in extension (cannot be deleted)
	if existingGraphExtension.IsBuiltin {
		return schemaExists, model.ErrGraphExtensionBuiltIn
	}

	// Delete Existing Extension
	if err = tx.DeleteGraphSchemaExtension(ctx, existingGraphExtension.ID); err != nil {
		return schemaExists, err
	}

	return schemaExists, nil
}
```

Then you can move the call to the helper into the if/else chain and it will flow nicer...

```
if schemaExists, err := s.cleanupExistingExtension(ctx, bloodhoundDBTransaction, graphExtensionInput); err != nil {
		return schemaExists, err
	}
```

kpowderly:
Heads up, we have an Integrate Service Functions jira ticket that we anticipated we would need to do some E2E testing in so don't kill yourself testing every scenario! We can always go back and add some additional testing in.

https://specterops.atlassian.net/browse/BED-6892

cmd/api/src/api/registration/v2.go
```diff
@@ -369,7 +366,10 @@ func NewV2API(resources v2.Resources, routerInst *router.Router) {
 		routerInst.DELETE(fmt.Sprintf("/api/v2/custom-nodes/{%s}", v2.CustomNodeKindParameter), resources.DeleteCustomNodeKind).RequireAuth(),
 
 		// Open Graph Schema
-		routerInst.PUT("/api/v2/extensions", resources.OpenGraphSchemaIngest).RequireAuth(),
+		routerInst.PUT("/api/v2/extensions", resources.OpenGraphSchemaIngest).CheckFeatureFlag(resources.DB, appcfg.FeatureOpenGraphExtensionManagement).RequireAuth(),
 		routerInst.GET("/api/v2/extensions", resources.ListExtensions).RequireAuth(),
```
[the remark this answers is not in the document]
kpowderly: I believe you're referring to context that was given on another PR when the skeleton endpoint was created. This is the ticket that the authentication was behind handled in.

cmd/api/src/services/opengraphschema/extension.go
```diff
@@ -17,69 +17,83 @@ package opengraphschema
...
-			SourceKindName:       finding.SourceKind,
-			RelationshipKindName: finding.RelationshipKind,
-			EnvironmentKindName:  finding.EnvironmentKind,
-			RemediationInput: database.RemediationInput{
-				ShortDescription: finding.Remediation.ShortDescription,
-				LongDescription:  finding.Remediation.LongDescription,
-				ShortRemediation: finding.Remediation.ShortRemediation,
-				LongRemediation:  finding.Remediation.LongRemediation,
-			},
+// validateGraphExtension - Ensures the incoming model.GraphExtensionInput has an extension name, node kinds exist, and
```
kpowderly: Your comment is missing the version validation

If you wanna copy/pasta, up to you!

```
// validateGraphExtension validates the incoming model.GraphExtensionInput for the
// following fields: name, version, namespace.
// It also ensures node kinds exist and there are no duplicate kinds. Additionally, it
// validates that node and edge kinds must be prepended with the extension namespace.
```

cmd/api/src/api/v2/opengraphschema.go
LawsonWillard: I agree and I wouldn’t normally include it but part of the tickets acceptance criteria is a 200 on update and 201 on creation and I’m not sure if there’s another way to do that. I’ll check on Monday to see if we can drop that requirement.

cmd/api/src/database/upsert_schema_extension.go
```diff
@@ -13,53 +13,153 @@
...
+		if schemaFinding, err := s.UpsertFinding(ctx, extensionId, finding.SourceKindName,
+			finding.RelationshipKindName, finding.EnvironmentKindName, finding.Name, finding.DisplayName); err != nil {
+			return fmt.Errorf("failed to upsert finding: %w", err)
+		} else {
+			if err := s.UpsertRemediation(ctx, schemaFinding.ID, finding.RemediationInput.ShortDescription,
+				finding.RemediationInput.LongDescription, finding.RemediationInput.ShortRemediation, finding.RemediationInput.LongRemediation); err != nil {
+				return fmt.Errorf("failed to upsert remediation: %w", err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
LawsonWillard: This will be handled in a separate ticket

cmd/api/src/database/upsert_schema_extension_integration_test.go
```diff
@@ -13,214 +13,1179 @@
...
+		}
+	}
+
+	// Test Findings
+	gotSchemaRelationshipFinding, err = db.GetSchemaRelationshipFindingsBySchemaExtensionId(testContext, gotGraphExtension.ID)
+	if len(want.FindingsInput) > 0 {
+		require.NoError(t, err)
+	}
+	require.Equalf(t, len(want.FindingsInput), len(gotSchemaRelationshipFinding), "mismatched number of findings")
+	for i, finding := range gotSchemaRelationshipFinding {
```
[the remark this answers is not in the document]
LawsonWillard: Db layer defaults to ID if nothing is provided.

cmd/api/src/api/v2/opengraphschema.go
kpowderly: Did you check in if we could drop this requirement?

kpowderly: Did you check in if we could drop this requirement?

LawsonWillard: This is a hard requirement as the endpoint needs to conform to the PUT RFC: https://www.rfc-editor.org/rfc/rfc9110#section-9.3.4

cmd/api/src/database/upsert_schema_extension.go
```diff
@@ -13,53 +13,161 @@
...
+		createdExtension.ID, graphExtensionInput.PropertiesInput); err != nil {
+		return false, fmt.Errorf("failed to upsert properties: %w", err)
+	} else if err = bloodhoundDBTransaction.upsertGraphEnvironments(ctx, createdExtension.ID,
+		graphExtensionInput.EnvironmentsInput); err != nil {
+		return false, err
+	} else if err = bloodhoundDBTransaction.upsertFindingsAndRemediations(ctx, createdExtension.ID,
+		graphExtensionInput.FindingsInput); err != nil {
+		return false, err
+	} else if err = tx.Commit().Error; err != nil {
+		return false, err
```
urangel: Would it be valid to return `schemaExists` instead of `false` for these error conditions?

cmd/api/src/database/graphschema.go
```diff
@@ -931,30 +970,3 @@ func parseFiltersAndPagination(filters model.Filters, sort model.Sort, skip, lim
 	}
 	return filtersAndPagination, nil
 }
-
```
urangel, line 934: Thanks for the cleanup! 🧹

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
-		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	// feature flag is checked as part of middleware
+	if user, isUser := auth.GetUserFromAuthCtx(bhCtx.FromRequest(request).AuthCtx); !isUser {
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, "No associated "+
+			"user found", request), response)
+		return
+	} else if !user.Roles.Has(model.Role{Name: auth.RoleAdministrator}) {
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusForbidden, "user does not "+
+			"have sufficient permissions to create or update an extension", request), response)
+		return
```
urangel: Is this string addition done to prevent the line getting too big? It made me pause for a second lol 😅 

This is subjective but this would read more clearly to me:

```go
		var (
			errorMessage  = "user does not have sufficient permissions to create or update an extension"
			errorResponse = api.BuildErrorResponse(http.StatusForbidden, errorMessage, request)
		)

		api.WriteErrorResponse(ctx, errorResponse, response)
        return
```

LawsonWillard: Yeah haha I can clean it up to read better!

cmd/api/src/database/upsert_schema_extension.go
LawsonWillard: It would but I'm not sure how useful that value would be if there's already an error. I can update them though

cmd/api/src/model/graphschema.go
```diff
@@ -173,3 +182,70 @@ type GraphSchemaRelationshipKindWithNamedSchema struct {
...
+
+type FindingsInput []FindingInput
+type FindingInput struct {
+	Name                 string
+	DisplayName          string
+	SourceKindName       string
+	RelationshipKindName string // edge kind
+	EnvironmentKindName  string
+	RemediationInput     RemediationInput
+}
```
urangel: Do we need a `TargetKindName` on this struct since a relationship is involved?

cmd/api/src/database/upsert_schema_extension.go
urangel: Agreed, it funnels through the same route on error afaict. The break in the pattern made me think there might be something I was missing so mostly a question of minimizing thought overhead.

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -96,6 +274,16 @@ func (s Resources) ListExtensions(response http.ResponseWriter, request *http.Re
...
 		return
 	} else {
-		api.WriteJSONResponse(ctx, ExtensionsResponse{Extensions: extensions}, http.StatusOK, response)
+
+		var extensionsResponse = make([]ExtensionInfo, len(extensions))
+		for i, extension := range extensions {
+			extensionsResponse[i] = ExtensionInfo{
+				ID:      strconv.Itoa(int(extension.ID)),
+				Name:    extension.DisplayName,
+				Version: extension.Version,
```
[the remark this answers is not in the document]
LawsonWillard: This is expected

cmd/api/src/model/graphschema.go
LawsonWillard: I dont think so, I believe its applied to all instances of the relationship in the provided environment regardless of target node kind. 

@kpowderly or @brandonshearin would be better at answering this though.

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -96,6 +274,16 @@ func (s Resources) ListExtensions(response http.ResponseWriter, request *http.Re
...
+			}
+		}
+
+		api.WriteJSONResponse(ctx, ExtensionsResponse{Extensions: extensionsResponse}, http.StatusOK, response)
```
urangel, line 284: This may be out of scope for your work and I'm not sure how sticky we are about changing this endpoint since it already exists but I typically see responses wrapped in metadata fields. I would think that we would reach for `api.WriteBasicResponse` over `api.WriteJSONResponse` for conforming to the [api spec](https://specterops.atlassian.net/wiki/spaces/BE/pages/820517/BloodHound+API+Specification#Request-Response-Payloads).

cmd/api/src/model/graphschema.go
LawsonWillard: Though blanket apply may not be what we want in all scenarios

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+		graphExtensionPayload GraphExtensionPayload
 	)
 
-	var req GraphSchemaExtension
-	if err := json.NewDecoder(request.Body).Decode(&req); err != nil {
-		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	// feature flag is checked as part of middleware
+	if user, isUser := auth.GetUserFromAuthCtx(bhCtx.FromRequest(request).AuthCtx); !isUser {
+		var errMessage = "No associated user found"
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, errMessage, request), response)
```
urangel: Should this status be `http.StatusUnauthorized`?

cmd/api/src/model/graphschema.go
urangel: Ah okay. I'm not familiar with what the generalized finding approach is but I figured I'd ask anyway. Thanks for giving it some thought!

cmd/api/src/api/v2/opengraphschema.go
LawsonWillard: Thats out of the scope of this work however I'm not sure if this endpoint is used yet as there's no OpenApi spec for it. We might be able to change it before GA'ing the feature. Think its worth creating another ticket or just amending it here?

cmd/api/src/api/v2/opengraphschema.go
LawsonWillard: I think so, Unauthorized would be more informative and accurate.

cmd/api/src/api/v2/opengraphschema.go
urangel: Cool cool. Yea let's leave it for now and verify whether this should be wrapped or not before creating a ticket.

kpowderly: This was my work. I can update it in my next ticket! I didn't realize the api spec in Confluence existed and we're adding the open spec at the end for our endpoints. Thanks for the call out!!

LawsonWillard: Just a heads up, I think we are creating but commenting out the API spec

cmd/api/src/model/graphschema.go
craigw-SO: Overall, Findings are tied to edges. Right now, we're loose and not validating what edges connect to which nodes, but eventually we plan to build that validation in. So you wouldn't be able to make a "MemberOf" edge where it isn't possible, for example.

That will supplant the need for making conditionals on the Findings themselves. And beyond that, the Post-Processing/Complex Edges project will facilitate pattern matching for conditionals like that. So Findings simply need to refer to a specific edge.

cmd/api/src/api/v2/opengraphschema.go
AD7ZJ: I was going to suggest using tics ` , these allow multi line strings

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+	if graphExtensionPayload, err = extractExtensionData(request.Body); err != nil {
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
 		return
+	} else if updated, err = s.OpenGraphSchemaService.UpsertOpenGraphExtension(ctx,
+		convertGraphExtensionPayloadToGraphExtension(graphExtensionPayload)); err != nil {
+		switch {
+		case strings.Contains(err.Error(), model.ErrGraphExtensionValidation.Error()) ||
+			strings.Contains(err.Error(), model.ErrGraphExtensionBuiltIn.Error()):
+			api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, err.Error(), request), response)
+		case strings.Contains(err.Error(), model.ErrGraphDBRefreshKinds.Error()): // Open Graph TODO: Do we want to return an error or let it succeed?
```
AD7ZJ: I would think let it fall through, this is not strictly an error related to the upsert operation....

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -96,6 +274,16 @@ func (s Resources) ListExtensions(response http.ResponseWriter, request *http.Re
 		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("error listing graph schema extensions: %v", err), request), response)
 		return
 	} else {
-		api.WriteJSONResponse(ctx, ExtensionsResponse{Extensions: extensions}, http.StatusOK, response)
+
+		var extensionsResponse = make([]ExtensionInfo, len(extensions))
```
AD7ZJ:
```suggestion
		var extensionsResponse = make([]ExtensionInfo, len(extensions))
```

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
```diff
@@ -0,0 +1,135 @@
...
+          example: Computer
+        description:
+          type: string
+          example: Active Directory Computer object
+        is_display_kind:
+          type: boolean
+        icon:
+          type: string
+          example: Desktop
+        icon_color:
```
AD7ZJ:
```suggestion
        color:
```

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
```diff
@@ -0,0 +1,135 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+type: object
+properties:
+  extension:
```
AD7ZJ:
```suggestion
  schema:
```

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
```diff
@@ -0,0 +1,135 @@
...
+          type: string
+          example: MemberOf
+        environment_kind:
+          type: string
+          example: Domain
+        remediation:
+          type: object
+          properties:
+            short_description:
+              type: string
```
AD7ZJ, line 128:
```suggestion
            display_name:
              type: string
            short_description:
              type: string
```

cmd/api/src/database/migration/migrations/v8.6.0.sql
```diff
@@ -80,16 +80,5 @@ DO $$
...
--- OpenGraph Extension Management feature flag
-INSERT INTO feature_flags (created_at, updated_at, key, name, description, enabled, user_updatable)
-VALUES (current_timestamp,
-        current_timestamp,
-        'opengraph_extension_management',
-        'OpenGraph Extension Management',
-        'Enable OpenGraph Extension Management',
-        false,
-        false)
-ON CONFLICT DO NOTHING;
```
craigw-SO: Is there any risk with this being moved to a different migration file, when the current RC has the creation of this feature flag in v8.6? I'm thinking probably not, but wanted to double check.

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
```diff
@@ -125,6 +125,8 @@ properties:
         remediation:
           type: object
           properties:
+            display_name:
+              type: string
```
craigw-SO: I don't think we need "display_name" for the remediation plan, do we? Remediations don't have a distinct title separate from the Finding itself.

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
craigw-SO: Since all nodes, edges, and findings need to include the namespace prefix, we should make sure the examples reflect that throughout.

craigw-SO: Include namespace prefix

craigw-SO: Include namespace prefix

craigw-SO: Include namespace prefixes

craigw-SO: Include namespace prefix

craigw-SO: Include namespace prefix

craigw-SO (changes requested):
So this was one of the things I was contending with for an example JSON payload schema: The most recognizable case is Active Directory BUT we don't want to confuse people because AD & AZ won't follow the new conventions yet until they are officially migrated to the OG extension platform.

We should either include a comment clarifying that this is for illustration purposes only, or consider using a non-AD/AZ example.

cmd/api/src/database/migration/migrations/v8.6.0.sql
LawsonWillard: I'd just thought about that and moved it to a new v8.7.0 migration file.

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
LawsonWillard: Honestly unsure, however display_name was recently added as a column in the schema_remediations table  and is in the example payload from the RFC: https://github.com/SpecterOps/BloodHound/blob/bh-rfc-4/rfc/bh-rfc-4.md. 

Its not currently wired up at the DB layer but there's a ticket to implement that.

LawsonWillard:
> So this was one of the things I was contending with for an example JSON payload schema: The most recognizable case is Active Directory BUT we don't want to confuse people because AD & AZ won't follow the new conventions yet until they are officially migrated to the OG extension platform.
> 
> We should either include a comment clarifying that this is for illustration purposes only, or consider using a non-AD/AZ example.

I think I'll use another more generic example however I think it would be beneficial to also leave in the AD example with a note clarifying this. I want to add some proper documentation around validation / payload conventions but not sure the best place to do so.

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+}
+
+type FindingsPayload struct {
+	Name             string             `json:"name"`
+	DisplayName      string             `json:"display_name"`
+	SourceKind       string             `json:"source_kind"`
+	RelationshipKind string             `json:"relationship_kind"`
+	EnvironmentKind  string             `json:"environment_kind"`
+	Remediation      RemediationPayload `json:"remediation"`
+}
```
craigw-SO: Something we should keep in mind is that these are "relationship findings" specifically, since they are based on an edge. We will need to add "list findings" very soon after. So we should probably make these distinct in the schema, and be explicit about the distinction.

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+
+type RemediationPayload struct {
+	DisplayName      string `json:"display_name"`
```
craigw-SO: Remove "Display Name", I don't think this is needed nor distinct from the Finding's "Display Name"

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+		graphExtension = model.GraphExtensionInput{
+			ExtensionInput: model.ExtensionInput{
+				Name:        payload.GraphSchemaExtension.Name,
+				DisplayName: payload.GraphSchemaExtension.DisplayName,
+				Version:     payload.GraphSchemaExtension.Version,
+				Namespace:   payload.GraphSchemaExtension.Namespace,
+			},
+			NodeKindsInput:         make(model.NodesInput, 0),
+			RelationshipKindsInput: make(model.RelationshipsInput, 0),
+			PropertiesInput:        make(model.PropertiesInput, 0),
```
craigw-SO, line 194: Do we want to comment out properties since we're not actually implementing them?

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+	IsDisplayKind bool   `json:"is_display_kind"` // indicates if this kind should supersede others and be displayed
+	Icon          string `json:"icon"`            // font-awesome icon for the registered node kind
+	IconColor     string `json:"color"`           // icon hex color
+}
+type GraphSchemaPropertiesPayload struct {
+	Name        string `json:"name"`
+	DisplayName string `json:"display_name"`
+	DataType    string `json:"data_type"`
+	Description string `json:"description"`
+}
```
craigw-SO, line 79: Comment out since we're not implementing yet?

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,240 @@ import (
...
+			EnvironmentKindName:  findingPayload.EnvironmentKind,
+			RemediationInput: model.RemediationInput{
+				DisplayName:      findingPayload.Remediation.DisplayName,
+				ShortDescription: findingPayload.Remediation.ShortDescription,
+				LongDescription:  findingPayload.Remediation.LongDescription,
+				ShortRemediation: findingPayload.Remediation.ShortRemediation,
+				LongRemediation:  findingPayload.Remediation.LongRemediation,
+			},
+		})
+	}
```
craigw-SO: Should probably rename to specify this is for relationship findings.

Also, remove Display Name from Remediation.

cmd/api/src/api/v2/opengraphschema_internal_test.go
```diff
@@ -0,0 +1,156 @@
...
+							RelationshipKind: "GraphSchemaEdgeKind_1",
+							EnvironmentKind:  "EnvironmentInput",
+							Remediation: RemediationPayload{
+								ShortDescription: "remediation for Finding_1",
+								LongDescription:  "a remediation for Finding 1",
+								ShortRemediation: "do x",
+								LongRemediation:  "do x but better",
+							},
+						},
+					},
```
craigw-SO: Specify relationship findings

cmd/api/src/api/v2/opengraphschema_internal_test.go
```diff
@@ -0,0 +1,156 @@
...
+						RelationshipKindName: "GraphSchemaEdgeKind_1",
+						EnvironmentKindName:  "EnvironmentInput",
+						RemediationInput: model.RemediationInput{
+							ShortDescription: "remediation for Finding_1",
+							LongDescription:  "a remediation for Finding 1",
+							ShortRemediation: "do x",
+							LongRemediation:  "do x but better",
+						},
+					},
+				},
```
craigw-SO: Specify relationship findings

cmd/api/src/api/v2/opengraphschema_test.go
```diff
@@ -35,148 +40,398 @@ import (
...
+					RelationshipKind: "TEST_GraphSchemaEdgeKind_1",
+					EnvironmentKind:  "TEST_EnvironmentInput",
+					Remediation: v2.RemediationPayload{
+						ShortDescription: "remediation for Finding_1",
+						LongDescription:  "a remediation for Finding 1",
+						ShortRemediation: "do x",
+						LongRemediation:  "do x but better",
+					},
+				},
+			},
```
craigw-SO: Relationship findings

cmd/api/src/api/v2/opengraphschema_test.go
```diff
@@ -35,148 +40,398 @@ import (
...
+					RelationshipKindName: "TEST_GraphSchemaEdgeKind_1",
+					EnvironmentKindName:  "TEST_EnvironmentInput",
+					RemediationInput: model.RemediationInput{
+						ShortDescription: "remediation for Finding_1",
+						LongDescription:  "a remediation for Finding 1",
+						ShortRemediation: "do x",
+						LongRemediation:  "do x but better",
+					},
+				},
+			},
```
craigw-SO: Relationship findings

cmd/api/src/model/graphschema.go
```diff
@@ -185,3 +192,70 @@ type GraphSchemaRelationshipKindWithNamedSchema struct {
...
```
craigw-SO: Realized that we should rename this as it pertains to specifically to relationship findings.

cmd/api/src/services/opengraphschema/extension.go
```diff
@@ -13,73 +13,132 @@
...
+		if _, ok := kinds[finding.RelationshipKindName]; !ok {
+			return fmt.Errorf("graph schema finding relationship kind %s not declared as a relationship kind", finding.RelationshipKindName)
+		}
+		if finding.SourceKindName == "" {
+			return fmt.Errorf("graph schema finding source kind cannot be empty")
+		}
+		if _, ok := kinds[finding.SourceKindName]; ok {
+			return fmt.Errorf("graph schema finding source kind %s should not be declared as a node or relationship kind", finding.SourceKindName)
+		}
 	}
```
craigw-SO: Relationship findings

packages/go/openapi/src/schemas/model.graph-extension.payload.yaml
```diff
@@ -0,0 +1,136 @@
...
+          type: string
+          example: OGE_ConnectedTo
+        environment_kind:
+          type: string
+          example: OGE_Environment
+        remediation:
+          type: object
+          properties:
+            display_name:
+              type: string
```
craigw-SO: Drop display name

cmd/api/src/api/v2/opengraphschema.go
LawsonWillard: Properties are commented out of the payload but I figured to leave the structs in to easily implement it in the future. At a service and DB level nothing needs to change as no data is making it in so properties will just be skipped during upsert.

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -19,62 +19,238 @@ import (
...
+	IsDisplayKind bool   `json:"is_display_kind"` // indicates if this kind should supersede others and be displayed
+	Icon          string `json:"icon"`            // font-awesome icon for the registered node kind
+	IconColor     string `json:"color"`           // icon hex color
+}
+type GraphSchemaPropertiesPayload struct {
+	Name        string `json:"name"`
+	DisplayName string `json:"display_name"`
+	DataType    string `json:"data_type"`
+	Description string `json:"description"`
+}
```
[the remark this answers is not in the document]
LawsonWillard: This is expected

PR #2333, opened 2026-02-03
feat: OpenGraph findings- BED-6903

brandonshearin:
## Description
This PR contains odds and ends bits to support the wire up of datapipe to the opengraph schema definitions that both teams have been chuggin on.

- az_graph_schema incorrectly defined an azure environment with a type of `Tenant`, when `AZTenant` is needed.  @wes-mil Please look over my change to `az_graph_schema.sql`
- refactor `GetEnvironments*` methods on `OpenGraphSchema` interface to use the same shared func with configurable sql filters.
- refactor `GetSchemaRelationshipFinding*` methods on `OpenGraphSchema` interface to use a shared func with configurable sql filters.
- methods to get kind/source_kind by id added 
- in `convertors.go` i capitalize the `environment_id` property on any generic-ingested node. this will keep environment_id's consistent with the uppercase convention we currently maintain for DomainSIDs and TenantIDs

## Motivation and Context
BED-6903
OpenGraph findings

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Filtered environment retrieval and environment-scoped relationship findings.
  * New lookups to fetch kinds and source kinds by ID.

* **Improvements**
  * Migrations and bootstrap now upsert missing kinds on demand instead of failing.
  * Environment handling moved to a single-key uniqueness approach; environment identifiers are normalized (uppercase) where applicable.

* **Tests**
  * Mocks and test helpers updated to reflect the new retrieval and lookup methods.


packages/go/schemagen/generator/sql.go
```diff
@@ -365,8 +365,7 @@ func GenerateADSpecifics(sb io.StringWriter) {
 
 func GenerateAZSpecifics(sb io.StringWriter) {
 	sb.WriteString("\tPERFORM genscript_upsert_source_kind('AZBase');\n")
-	sb.WriteString("\tPERFORM genscript_upsert_kind('Tenant');\n")
```
wes-mil, line 368: Can you remove the upsert `Domain` from `GenerateADSpecifics` as well

wes-mil (commented):
The schema changes look good to me

cmd/api/src/database/graphschema.go
```diff
@@ -56,11 +56,13 @@ type OpenGraphSchema interface {
 	GetEnvironmentByKinds(ctx context.Context, environmentKindId, sourceKindId int32) (model.SchemaEnvironment, error)
 	GetEnvironmentById(ctx context.Context, environmentId int32) (model.SchemaEnvironment, error)
 	GetEnvironments(ctx context.Context) ([]model.SchemaEnvironment, error)
+	GetEnvironmentsFiltered(ctx context.Context, filters model.Filters) ([]model.SchemaEnvironment, error)
```
mistahj67, line 60: nit: naming things is always difficult but I think in these cases extra caution should be utilized as the codebase is riddled with various `getEnvironments` functions most reaching into the graphdb. This goes for the other methods here as well

```suggestion
	GetSchemaEnvironmentsFiltered(ctx context.Context, filters model.Filters) ([]model.SchemaEnvironment, error)
```

cmd/api/src/database/sourcekinds.go
```diff
@@ -127,6 +128,39 @@ func (s *BloodhoundDB) GetSourceKindByName(ctx context.Context, name string) (So
...
+
+	if result.RowsAffected == 0 || raw.ID == 0 {
+		return SourceKind{}, ErrNotFound
+	}
+
+	kind := SourceKind{
+		ID:     raw.ID,
+		Name:   graph.StringKind(raw.Name),
+		Active: raw.Active,
+	}
```
mistahj67: It feels a bit antipattern to force the Name to be a graphKind here. What do you think of instead adding a method to SourceKind such as `ToKind()` that then transforms it and leaves Name as just a string? Similar methods are found in RelationshipKind / NodeKind / and PZ

mistahj67 (commented):
Not blocking, but left a few comments to ponder 🤔

cmd/api/src/database/graphschema.go
```diff
@@ -624,37 +640,44 @@ func (s *BloodhoundDB) GetEnvironments(ctx context.Context) ([]model.SchemaEnvir
...
+	envs, err := s.GetEnvironmentsFiltered(ctx, filters)
+	if err != nil {
+		return model.SchemaEnvironment{}, err
+	}
+	if len(envs) == 0 {
 		return model.SchemaEnvironment{}, ErrNotFound
 	}
 
-	return env, nil
+	return envs[0], nil
```
sirisjo: **suggestion** (non-blocking): Could we use `LIMIT 1` in the SQL string instead of returning the first element in the slice here? Using sql to get the first element feels less brittle to me. Same comment applies for the other functions using `[0]`.

packages/go/schemagen/generator/sql.go
brandonshearin: yep

wes-mil (approved):
Only comment is that you _could_ return the ID from `genscript_upsert_source_kind` and then you could skip having to `SELECT` but tomato tomato

EDIT: Approving specifically the sql generator changes, please get other approvals for the rest of the changes :smile:

cmd/api/src/database/sourcekinds.go
brandonshearin: yes

cmd/api/src/database/graphschema.go
brandonshearin: method was removed

cmd/api/src/database/migration/migrations/v8.7.0.sql
```diff
@@ -0,0 +1,31 @@
...
+DO $$
+    BEGIN
+        IF NOT EXISTS (
+                      SELECT 1
+                      FROM pg_constraint
+                      WHERE conname = 'schema_environments_environment_kind_id_key'
+        ) THEN
+            ALTER TABLE schema_environments
+                ADD CONSTRAINT schema_environments_environment_kind_id_key UNIQUE (environment_kind_id);
+        END IF;
```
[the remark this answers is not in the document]
brandonshearin: this feature hasn't rolled out yet so i think this is safe to ignore

cmd/api/src/database/sourcekinds.go
```diff
@@ -120,7 +125,40 @@ func (s *BloodhoundDB) GetSourceKindByName(ctx context.Context, name string) (So
...
+		return SourceKind{}, result.Error
+	}
+
+	if result.RowsAffected == 0 || raw.ID == 0 {
+		return SourceKind{}, ErrNotFound
+	}
+
+	kind := SourceKind{
+		ID:     raw.ID,
+		Name:   raw.Name,
```
urangel: I like the update to the `SourceKind` struct! Now that `Name` is a string are we able to remove of all of the `rawSourceKind` type definitions and mapping to `SourceKind` by scanning directly to a struct of type `SourceKind`?

brandonshearin: done

urangel: 🙇 🙇 🙇

urangel (approved):

cmd/api/src/database/graphschema.go
```diff
@@ -494,6 +497,24 @@ func (s *BloodhoundDB) GetGraphSchemaRelationshipKinds(ctx context.Context, rela
 	}
 }
 
+// GetTraversableRelationshipKindsByExtensionID returns all traversable relationship kinds for a given schema extension.
+// This is a purpose-built query for the analysis pipeline that needs traversable edges for graph traversal.
+func (s *BloodhoundDB) GetTraversableRelationshipKindsByExtensionID(ctx context.Context, extensionID int32) (model.GraphSchemaRelationshipKinds, error) {
```
LawsonWillard, line 502: Non-blocking

I think this func can be simplified to use filters + GetGraphSchemaRelationshipKinds like environments and findings:

```suggestion
func (s *BloodhoundDB) GetTraversableRelationshipKindsByExtensionID(ctx context.Context, extensionID int32) (model.GraphSchemaRelationshipKinds, error) {
	filters := model.Filters{
		"ek.is_traversable":      []model.Filter{{Operator: model.Equals, Value: model.TrueString}},
		"ek.schema_extension_id": []model.Filter{{Operator: model.Equals, Value: fmt.Sprintf("%d", extensionID)}},
	}

	if relationshipKinds, _, err := s.GetGraphSchemaRelationshipKinds(ctx, filters, model.Sort{}, 0, 0); err != nil {
		return model.GraphSchemaRelationshipKinds{}, err
	} else {
		return relationshipKinds, nil
	}
}

```

Also may be worth adding an integration test or two

packages/go/schemagen/generator/sql.go
```diff
@@ -241,14 +241,16 @@ DECLARE
 BEGIN
 	SELECT id INTO retreived_environment_kind_id FROM kind WHERE name = v_environment_kind_name;
 	IF retreived_environment_kind_id IS NULL THEN
-		RAISE EXCEPTION 'couldn''t find matching kind_id';
```
LawsonWillard, line 244: I think if the exception is removed, the environment kinds must also be declared as node kinds before being created as an environment.

cmd/api/src/database/graphschema.go
brandonshearin: its a really good idea, and its actually the first approach i took. however, check out this comment on the BHE PR

https://github.com/SpecterOps/bloodhound-enterprise/pull/975#discussion_r2785166783

its brittle to define the filter which uses the ek alias outside of the `GetGraphSchemaRelationshipKinds` func. If somebody changed that func, and modified the alias locally, then the calling func would break.

LawsonWillard: Ahhh I see, I was debating breaking out db layer models and having fields correspond to columns to help with this but its too similar to a query builder / was a bit too much to do for the upsert work.

PR #2348, opened 2026-02-05
feat(OpenGraph): Stave off kind table ID exhaustion - BED-7259

LawsonWillard:
## Description

Replace the existing upsert_kind func with a new one to better assist with id exhaustion. 

## Motivation and Context
Resolves: BED-7259

Currently, when performing schema node and relationship upserts, the kind table will begin to unnecessarily exhaust its ID space as an insert that conflicts with an existing kind will always increment even if DO NOTHING or DO UPDATE is provided. 

In order to ensure we don't waste ID space, a new upsert_kind function was created to check if kinds already exist before attempting the insert. 

## How Has This Been Tested?

- Ran integration tests
- TODO: Manual performance testing

## Types of changes
- Database Migrations

## Checklist:
  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced rare insertion conflicts during concurrent operations by adding a retry-backed upsert, lowering chance of runtime failures.

* **Chores**
  * Improved database behavior to avoid identifier exhaustion and preserve performance under high-concurrency scenarios.

### DIFF
--- a/cmd/api/src/database/migration/migrations/v8.7.0.sql
+++ b/cmd/api/src/database/migration/migrations/v8.7.0.sql
@@ -43,7 +43,7 @@ BEGIN
     IF kind_row IS NOT NULL THEN
         RETURN kind_row;
     END IF;
-    -- Insert with retry, handles the edge case where two transactions could add the same kind at the same time
+    -- Insert with retry, handles the race condition where two transactions try to add the same kind at the same time
     FOR i IN 1..5 LOOP
         BEGIN
             INSERT INTO kind (name)


cmd/api/src/database/migration/migrations/v8.7.0.sql
```diff
@@ -24,3 +24,43 @@ VALUES (current_timestamp,
...
+            RETURNING * INTO kind_row;
+            RETURN kind_row;
+        EXCEPTION
+            WHEN unique_violation THEN
+                -- Check if the insert conflict was for same kind
+                SELECT * INTO kind_row FROM kind WHERE name = node_kind_name;
+                IF kind_row IS NOT NULL THEN
+                    RETURN kind_row;
+                END IF;
+            -- Otherwise retry (id collision, another kind was inserted)
```
kpowderly: Since Serial's ID auto-increments in Postgres, it shouldn't be possible for a collision to occur. This function only checks the name. Also, retrying with a collision would always lead to the same outcome.  

The retry logic only makes sense for handling race conditions where another transaction inserts the same kind name between our initial SELECT and INSERT. Should we update the comment to clarify that the retries are specifically for concurrent insert race conditions, not ID collisions?

cmd/api/src/database/migration/migrations/v8.7.0.sql
```diff
@@ -24,3 +24,43 @@ VALUES (current_timestamp,
         false,
         false)
 ON CONFLICT DO NOTHING;
+
+-- upsert_kind is a stop-gap function used in open graph schema node,
+-- relationship and source kind creation.
+-- This function is needed to stave off kind table id exhaustion while maintaining
```
kpowderly, line 30: How does this stave off id exhaustion? Sorry, maybe I'm not following here / don't know background.

cmd/api/src/database/migration/migrations/v8.7.0.sql
LawsonWillard: That "Otherwise retry" comment is out of date, it originally used a MAX + Coalesce + 1 func to get the new id. 

I'll update it to be more accurate

cmd/api/src/database/migration/migrations/v8.7.0.sql
LawsonWillard: The SELECT at the beginning allows us to return the kind row if it already exists without incrementing the kind_id_seq table. It also helps not increment the kind_id_seq further if there's a conflicting transaction trying to insert the same kind. 

Background: The original function used a CTE to insert into the kinds table before inserting into the schema_node/edge tables. Unfortunately Postgres will increase a SERIAL PK on a conflict even if DO NOTHING or DO UPDATE is provided (kind_id column will increment even if the kind already exists).

PR #2372, opened 2026-02-17
feat: Schema Service additional testing & Source Kind Deactivation - BED-6920

kpowderly:
## Description

This ticket was initially created as a placeholder to hook up the Schema Service & add additional testing for anything that was missed along the way. Since the Extension Upsert piece was done after the dependent schemas, it didn't end up being necessary.

Therefore, this ticket was updated to reflect these things:
1. Look for missing code coverage
2. Make sure everything is working as expected
3. De-Activate Source Kinds when deleting

## Motivation and Context

Resolves [BED-6920](https://specterops.atlassian.net/browse/BED-6920)

Source Kind Deletion + Additional Missed Testing

## How Has This Been Tested?

Manual Tests & Integration Tests

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-6920]: https://specterops.atlassian.net/browse/BED-6920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a graph schema extension now collects related source kinds, deletes the extension, and deactivates only those source kinds that are no longer referenced and are not core base kinds. Error reporting around deactivation was improved.
  * Validation timing adjusted to defer duplicate detection and provide clearer conflict messages for environment/relationship source-kind name collisions.

* **Tests**
  * Significantly expanded integration tests for deletion cascades, deactivation/retention semantics, containment-based assertions, pagination/filter errors, and duplicate-name/conflict cases.

### DIFF
--- a/cmd/api/src/database/graphschema_integration_test.go
+++ b/cmd/api/src/database/graphschema_integration_test.go
@@ -2749,7 +2749,7 @@ func TestDatabase_GraphSchemaProperties_CRUD(t *testing.T) {
 
 				// Attempt to Update Property w/ name conflict
 				_, err = testSuite.BHDatabase.UpdateGraphSchemaProperty(testSuite.Context, createdPropertyB)
-				assert.EqualError(t, err, "duplicate graph schema extension property name: ext_prop_1")
+				assert.ErrorIs(t, err, model.ErrDuplicateGraphSchemaExtensionPropertyName)
 			},
 		},
 		{


cmd/api/src/services/opengraphschema/extension.go
```diff
@@ -72,9 +72,6 @@ func validateGraphExtension(graphExtension model.GraphExtensionInput) error {
 		if _, ok := nodeKinds[kind.Name]; ok {
 			return fmt.Errorf("duplicate graph kinds: %s", kind.Name)
 		}
-		if _, ok := relationshipKinds[kind.Name]; ok {
```
kpowderly, line 75: @LawsonWillard just an fyi, I removed this check because the relationshipKinds map isn't populated during the node kinds loop, making this collision check against it unreachable. 

Wanna double check me on this?

cmd/api/src/services/opengraphschema/extension_internal_test.go
```diff
@@ -374,7 +374,7 @@ func Test_validateGraphExtension(t *testing.T) {
 			wantErr: fmt.Errorf("graph schema environment source kind cannot be empty"),
 		},
 		{
-			name: "fail - environment source kind cannot be declared as a node or relationship kind",
+			name: "fail - environment source kind cannot be declared as a node kind",
```
brandonshearin: why are we failing here? 
why do we guard against a source kind being registered in `schema_node_kinds`?

cmd/api/src/services/opengraphschema/extension_internal_test.go
```diff
@@ -830,6 +871,55 @@ func Test_validateGraphExtension(t *testing.T) {
 			},
 			wantErr: fmt.Errorf("graph schema relationship finding source kind %s should not be declared as a node kind", "AD_node_kind_1"),
 		},
+		{
+			name: "fail - relationship finding source kind cannot be declared as a relationship kind",
```
brandonshearin, line 875: i think this business logic is not needed.

look at how we create a relationship finding at the DB layer-- `CreateSchemaRelationshipFinding`. the write path for the relationship finding resource doesn't need an environment kind name. its not a column in the database table. 

there may be confusion because the read path for the resource, like `GetSchemaRelationshipFindingById` adds it to the return object by following the FK. 

removing this business logic from the service layer would be a good idea i think.

thoughts @kpowderly @LawsonWillard

cmd/api/src/services/opengraphschema/extension_internal_test.go
kpowderly: This is just an additional test to what already existed but it's an input validation that a source kinds name cannot be the same as a node/relationship kind name.

@LawsonWillard  We probably want to update the list of validations to include this:

Environment source kind name cannot be empty
Environment source kind name must NOT be a declared node kind
Environment source kind name must NOT be a declared relationship kind

Relationship finding source kind name cannot be empty
Relationship finding source kind name must NOT be a declared node kind
Relationship finding source kind name must NOT be a declared relationship kind

brandonshearin: one invariant we must maintain is that if something is present in the `source_kind` table, it is present in the dawgs `kind` table. 

is this invariant maintained by the code you and lawson are working on.

context: during analysis, we read the whole graph into memory. that query uses source kinds to fetch all nodes. if the query uses a source kind that is not in the dawgs kinds table, things break.

kpowderly: Correct me if I'm wrong but the Source Kind will still exist in both the Source Kind Table and the Kind Table but this validation is simply that Source Kind NAME is "unique" or "different" than a node kind name / relationship kind name.

So to put in in plain speaking, `the source kind name must be different from any node kind name or relationship kind name`.

brandonshearin: coo coo, thanks for putting up with my q's

cmd/api/src/services/opengraphschema/extension.go
LawsonWillard: Yeah that makes sense to me to remove it

cmd/api/src/services/opengraphschema/extension_internal_test.go
LawsonWillard: For additional clarity: a PR was just merged to add source_kinds to the DAWGS kind table during the source_kind registration func. This way it doesnt need to be added as a node or relationship kind to make its way into the DAWGS kinds table. 

@kpowderly those additional tests should already be added but please lmk if not!

cmd/api/src/services/opengraphschema/extension_internal_test.go
LawsonWillard: I think you're right, the original design required each finding to have a source_kind as that was the only way to query for the environment however with your changes to the environments table (dropping the composite unique constraint) we should be able to eliminate it entirely from the findings portion of the payload as all we need is env_kind_name to query for the correct env.

We also could update the upsert environment func to retrieve the existing environment using just its name rather than both source_kind and env_kind_name. 

We also probably want to update the RFC to let extension authors know that its no longer a necessary field.

cmd/api/src/services/opengraphschema/extension_internal_test.go
kpowderly: I have all of the most recent changes. 

The original test was "name: "fail - environment source kind cannot be declared as a node or relationship kind"," and I split them into two tests: 1. source kind cannot be declared node kind 2. source kind cannot be declared relationship kind

cmd/api/src/services/opengraphschema/extension_internal_test.go
kpowderly: **Sorry**, I was getting a bit confused. I guess I was thinking about this in terms of _this_ ticket but I'm kind of tracking now. 

Another ticket was created to capture the following work:
- Source Kind is not unique on Environment
- This means that the RelationshipFindingsPayload no longer needs Source Kind to attach it to an environment
- Findings will no longer need to validate that Source Kind exists.
- Environments still requires a Source Kind

I kind of feel like we need to break this out into a ticket to capture the Source Kind work.

Do y'all know if Source Kinds has any specific validations?

LawsonWillard: I think that new ticket should capture everything needed. I don't _think_ there's any additional source_kind validation that needs to happen atm.

brandonshearin (approved):

cmd/api/src/database/graphschema.go
```diff
@@ -231,12 +234,42 @@ func (s *BloodhoundDB) DeleteGraphSchemaExtension(ctx context.Context, extension
...
 
+		// Get the source_kind_id from this extension's environment BEFORE deleting it
+		if result := tx.Raw(`
+			SELECT source_kind_id
+			FROM schema_environments
+			WHERE schema_extension_id = ?
+			LIMIT 1
+		`, extensionId).Scan(&sourceKindId); result.Error != nil {
+			return fmt.Errorf("failed to get source kind: %w", result.Error)
+		}
```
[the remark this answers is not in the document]
kpowderly: There is a ticket that I am working on next to remove the source_kind_id UNIQUE constraint from the schema_environments table. We will only have an environment_kind_id constaint. However, I did overlook that there can be multiple schema environments on an extension and have since resolved that. Please double check me.

PR #2316, opened 2026-01-29
chore(api): Add context information to analysis logs: BED-7265

StephenHinck:
## Description

Added contextual information to logs identified as common during analysis runs to provide improved context during log review and investigation of issues. Currently, those logs may prevent a user or administrator from understanding where within the process a certain log applies or is related to other steps in the process.

## Motivation and Context

Resolves BED-7265

*Why is this change required? What problem does it solve?*

## How Has This Been Tested?

Validated locally through multiple analysis runs with test data.

## Types of changes

- Chore (a change that does not modify the application functionality)

## Checklist:

- [X] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Added the appropriate labels
  - Associated an issue: https://github.com/SpecterOps/BloodHound/issues/672
  - Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
- [X] I have ensured that related documentation is up-to-date
  - Open API docs
  - Code comments (GoDocs / JSDocs)
- [X] I have followed proper test practices
  - Added/updated tests to cover my changes
  - All new and existing tests passed

## Summary by CodeRabbit

* **Refactor**
  * Enhanced observability across analysis and data-processing with richer structured telemetry and contextual logging.
  * Reworked Active Directory post-processing into a multi-step pipeline with early-error returns and aggregated reporting.
  * Standardized call patterns by replacing variadic relationship arguments with explicit composite types for clearer, consistent invocation.

PR #2384, opened 2026-02-19
chore: Add non-thresholded measure.go support and convert - BED-7453

StephenHinck:
## Description
The overall intent of this changeset is to make the use of the `measureThreshold` filter explicit (where it was previously implicit) so that critical logs can bypass that filter for troubleshooting purposes. In support of this, all functions modified in #2316 were reviewed for likely migration to the new functions which bypass the threshold as the majority of these functions represent critical system activity during analysis.

* rename measure.go functions to properly indicate they utilize a threshold
* Introduce bypass functions to guarantee output for critical logs, regardless of duration
* Add GoDoc documentation to measure.go and attr.go to encourage consistent, expected use

## Motivation and Context

Resolves BED-7453

## How Has This Been Tested?

Validated in local environment to ensure proper build, and to identify any additional functions requiring threshold filtering.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added richer structured-log attributes for better contextual logging.
  * Introduced measurement helpers including threshold-aware variants.

* **Improvements**
  * Switched many handlers, migrations, and background tasks to threshold-aware or log-enabled measurement for more consistent performance tracing.

* **Bug Fixes**
  * Raised slow-query warn/error thresholds to reduce spurious alerts.
  * Downgraded two noisy error logs to warnings for tenant/management lookups.

### DIFF
--- a/cmd/api/src/daemons/datapipe/agt.go
+++ b/cmd/api/src/daemons/datapipe/agt.go
@@ -620,7 +620,7 @@ func SelectNodes(ctx context.Context, db database.Database, agtParameters appcfg
 
 // selectAssetGroupNodes - concurrently selects all nodes for all tags
 func selectAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph.Database) []error {
-	defer measure.ContextLogAndMeasure(
+	defer measure.ContextMeasure(
 		ctx,
 		slog.LevelInfo,
 		"Selecting agt nodes",
@@ -829,7 +829,7 @@ func tagAssetGroupNodesForTag(ctx context.Context, db database.Database, graphDb
 
 // tagAssetGroupNodes - concurrently tags all nodes for all tags
 func tagAssetGroupNodes(ctx context.Context, db database.Database, graphDb graph.Database, additionalFilters ...graph.Criteria) []error {
-	defer measure.ContextLogAndMeasure(
+	defer measure.ContextMeasure(
 		ctx,
 		slog.LevelInfo,
 		"Tagging asset group nodes",
--- a/packages/go/analysis/post.go
+++ b/packages/go/analysis/post.go
@@ -170,7 +170,7 @@ func DeleteTransitEdges(ctx context.Context, db graph.Database, baseKinds graph.
 		operationName   = fmt.Sprintf("Delete %v post-processed relationships", strings.Join(targetRelationships.Strings(), ", "))
 	)
 
-	defer measure.ContextLogAndMeasure(
+	defer measure.ContextMeasure(
 		ctx,
 		slog.LevelInfo,
 		operationName,
@@ -225,7 +225,7 @@ func ClearOrphanedNodes(ctx context.Context, db graph.Database) error {
 	defer measure.ContextLogAndMeasure(
 		ctx,
 		slog.LevelInfo,
-		"Finished deleting orphaned nodes",
+		"Deleting orphaned nodes",
 		attr.Namespace("analysis"),
 		attr.Function("ClearOrphanedNodes"),
 		attr.Scope("process"),


cmd/api/src/database/db.go
```diff
@@ -240,8 +240,8 @@ func (s *BloodhoundDB) Transaction(ctx context.Context, fn func(tx *BloodhoundDB
 func OpenDatabase(connection string) (*gorm.DB, error) {
 	gormConfig := &gorm.Config{
 		Logger: &GormLogAdapter{
-			SlowQueryErrorThreshold: time.Second * 10,
-			SlowQueryWarnThreshold:  time.Second * 1,
+			SlowQueryErrorThreshold: time.Second * 30,
```
StephenHinck, line 243: Discussed change with Alyx. Across the fleet, the very large majority of our previous ERROR level logs were below 30s, and we completely ignore them. Bumping ERROR to >30, while WARN moves to >10 gives better indication of poorly performing queries.

packages/go/bhlog/measure/measure.go
StephenHinck, line 1: The primary changes for consideration/review live here.

packages/go/bhlog/attr/attr.go
StephenHinck, line 1: Added godoc strings to encourage consistent utilization.

packages/go/analysis/azure/post.go
StephenHinck, line 1: Dropped to WARN as these do not break the process, they are handled gracefully.

packages/go/bhlog/attr/attr.go
```diff
@@ -18,18 +18,33 @@ package attr
 
 import "log/slog"
 
+// attr.Error consistently includes an error message via standard logging in the "err" field.
 func Error(value error) slog.Attr {
+	if value == nil {
+		return slog.String("err", "<nil>")
+	}
```
ALCooper12: So I could be wrong, but I believe this addition is unnecessary. I feel like we should only be logging an error whenever error is not nil. This addition could potentially lead to unneeded logging/noise  :sweat_smile:

packages/go/bhlog/attr/attr.go
```diff
@@ -18,18 +18,33 @@ package attr
...
+
 	return slog.String("err", value.Error())
 }
 
+// attr.Namespace consistently includes the namespace for a given log via standard logging in the "namespace" field.
```
ALCooper12: I noticed your comments about a function's purpose have the format: package name.function name. I guess this is more of a nit, but we usually only use the function name when commenting about a function. But def ignore if you prefer this instead!

packages/go/bhlog/attr/attr.go
```diff
@@ -18,18 +18,33 @@ package attr
...
+// Examples of namespaces include "analysis" for functions executing during the overall analysis process, and "dogtags" for license-related logs.
 func Namespace(value string) slog.Attr {
 	return slog.String("namespace", value)
 }
 
+// attr.Scope consistently includes the scope for a given log via standard logging in the "scope" field.
+// Scope was originally created for striating the levels of the overall analysis process, and the following scopes are used:
```
ALCooper12: Also is it possible we could use a different word instead of "striating"? I feel like it's kind of confusing being used in this way

packages/go/bhlog/attr/attr.go
StephenHinck: This was something the 🐇 suggested, as the function would panic if a nil value was provided there. I completely agree with you, and I'm happy to rip it back out if preferred.

packages/go/bhlog/attr/attr.go
StephenHinck: I'll stick to the team's standards here, thank you for the callout!

packages/go/bhlog/attr/attr.go
StephenHinck: Adjusted, let me know if this is more clear!

packages/go/bhlog/attr/attr.go
ALCooper12: Still trying to test if this potential suggested change I have in mind will actually take care of a panic without unnecessary logging. But you can rip it out in the mean time!

packages/go/bhlog/attr/attr.go
ALCooper12: It definitely is. Thanks Stephen!

packages/go/bhlog/attr/attr.go
StephenHinck: Ripped out!

PR #2387, opened 2026-02-19
fix: Update ListExtensions endpoint to return id as int and built_in fields - BED-7461

kpowderly:
## Description

Updated List Extensions endpoint to return
1. ID as an int instead of a string
2. is_builtin field for UI to filter with

I also noticed that we were missing some auth checks.

## Motivation and Context

Resolves [BED-7461](https://specterops.atlassian.net/browse/BED-7461)

The UI needs to filter extensions that cannot be deleted by is_builtin.

## How Has This Been Tested?

Updated unit tests and ran the app locally to hit the ListExtensions endpoint

## Screenshots (optional):

Screenshot 2026-02-19 at 4 35 25 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-7461]: https://specterops.atlassian.net/browse/BED-7461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Extension listing now includes an "is_builtin" flag for each extension
  * Listing extensions now requires authenticated users with administrator role

* **Bug Fixes**
  * Extension IDs returned as numeric values (integers) instead of strings in API responses

* **Tests**
  * Added authorization, authentication and error-handling tests for extension listing, including request ID propagation

### DIFF
--- a/cmd/api/src/api/v2/opengraphschema_test.go
+++ b/cmd/api/src/api/v2/opengraphschema_test.go
@@ -582,6 +582,7 @@ func TestResources_ListExtensions(t *testing.T) {
 						},
 						DisplayName: "Display 1",
 						Version:     "v1.0.0",
+						IsBuiltin:   true,
 					},
 					{
 						Serial: model.Serial{
@@ -602,7 +603,7 @@ func TestResources_ListExtensions(t *testing.T) {
 			expected: expected{
 				responseCode:   http.StatusOK,
 				responseHeader: http.Header{"Content-Type": []string{"application/json"}},
-				responseBody:   `{"data": {"extensions":[{"id":1, "name":"Display 1", "version":"v1.0.0", "is_builtin":false}, {"id":2, "name":"Display 2", "version":"v2.0.0", "is_builtin":false}, {"id":3, "name":"Display 3", "version":"v3.0.0", "is_builtin":false}]}}`,
+				responseBody:   `{"data": {"extensions":[{"id":1, "name":"Display 1", "version":"v1.0.0", "is_builtin":true}, {"id":2, "name":"Display 2", "version":"v2.0.0", "is_builtin":false}, {"id":3, "name":"Display 3", "version":"v3.0.0", "is_builtin":false}]}}`,
 			},
 		},
 	}


cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -262,26 +262,33 @@ type ExtensionsResponse struct {
 }
 
 type ExtensionInfo struct {
-	ID      string `json:"id"`
-	Name    string `json:"name"`
-	Version string `json:"version"`
+	ID        int32  `json:"id"`
```
superlinkx, line 265: Is the ID an int32? Most of our models use int for the id field of an entity so I'm curious if this is meant to be different

cmd/api/src/api/v2/opengraphschema.go
```diff
@@ -262,26 +262,33 @@ type ExtensionsResponse struct {
...
 }
 
 func (s Resources) ListExtensions(response http.ResponseWriter, request *http.Request) {
 	var (
 		ctx = request.Context()
 	)
 
-	if extensions, err := s.OpenGraphSchemaService.ListExtensions(ctx); err != nil {
+	// feature flag is checked as part of middleware
+	if user, isUser := auth.GetUserFromAuthCtx(authctx.FromRequest(request).AuthCtx); !isUser {
```
superlinkx, line 277: Have we not set up a proper permission for working with extensions yet? Feels like this sort of thing should be handled by passing the required permission to the middleware like we do for other permissions

cmd/api/src/api/v2/opengraphschema.go
kpowderly: ID is defined as an int32 on our Serial struct

```
type Serial struct {
	ID int32 `json:"id" gorm:"primaryKey"`

	Basic
}
```

They both serialize to the same so I figured no reason to convert it to an int. Let me know if you think otherwise!

cmd/api/src/api/v2/opengraphschema.go
kpowderly: Honestly, I'm not super familiar with the way we handle authentication yet. The reason I added this is because it's consistent with our OpenGraphSchemaIngest handler method.

cmd/api/src/api/v2/opengraphschema.go
superlinkx: Good enough for me, we'll leave it as is then

cmd/api/src/api/v2/opengraphschema.go
superlinkx: The permissions for an endpoint would be defined in the router wire up (v2.go). I'm not sure if we've added a specific permission for this yet, but if not we should really switch to that and remove these manual administrator checks both for consistency and simplicity. For now if we don't have this wired up, it can go in as-is, but we need to work with the OG project to make sure this gets addressed

PR #2390, opened 2026-02-20
fix: updating docs to include wait header to bypass query limit - BED-6991

stephanieslamb:
## Description

This is to update the documentation for the prefer header to include limitless queries

## Motivation and Context

Resolves BED-6991

This is needed to document how to use limitless queries in the prefer header

## How Has This Been Tested?

No code changes.  Validated docs are updated

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Documentation**
  * Clarified Prefer header behavior: passing `wait=-1` is documented as bypassing all timeout limits when the feature is enabled; the header now accepts `-1` in addition to numeric values.
  * Normalized API documentation text for the RunCypherQuery endpoint (minor punctuation cleanup).
  * No API shapes, parameters, or responses were changed; these are documentation-only updates.

### DIFF
--- a/packages/go/openapi/src/parameters/header.prefer.yaml
+++ b/packages/go/openapi/src/parameters/header.prefer.yaml
@@ -23,4 +23,4 @@ required: false
 schema:
   type: string
   default: "wait=30"
-  pattern: "^wait=[0-9]+$"
+  pattern: "^wait=(-1|[0-9]+)$"
--- a/packages/go/openapi/doc/openapi.json
+++ b/packages/go/openapi/doc/openapi.json
@@ -18216,7 +18216,7 @@
         "schema": {
           "type": "string",
           "default": "wait=30",
-          "pattern": "^wait=[0-9]+$"
+          "pattern": "^wait=(-1|[0-9]+)$"
         }
       },
       "query.created-at": {


packages/go/openapi/src/parameters/header.prefer.yaml
```diff
@@ -16,10 +16,11 @@
 
 name: Prefer
 description: Prefer header, used to specify a custom timeout in seconds using the
-  wait parameter as per RFC7240.
+  wait parameter as per RFC7240. Passing in wait=-1 bypasses all timeout limits
```
Useinovski, line 19: I think it would be good to specify - When the feature is enabled:true limits cannot be bypassed. When feature is enabled:false, limits can be bypassed.

stephanieslamb: Since this is not user updatable, I am not sure this is entirely needed? It might be handy for devs to declare the difference.  But this comment is more for users.  `failed to bypass limits: feature disabled"` is the current error when timeouts are `enabled:true` if someone tries to use the -1 header.  So the thought was to stick to that wording.

PR #2404, opened 2026-02-24
chore(OG): convert schema_relationship_findings to schema_findings BED-7508

mistahj67:
## Description

- Drops the vestigial schema_list_findings table
- Creates a new schema_findings table and populates with the old schema_relationship_table
- Adds schema_findings_subtypes table

## Motivation and Context
Prep work for dynamic OG findings

Resolves BED-7508

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Introduced typed Findings and support for extensible Finding subtypes.

* **Chores**
  * Restructured Findings storage and migrated existing data to the new schema; added indices and updated remediation references to the new Findings.
  * Renamed public APIs and errors to reflect the new SchemaFinding model.

* **Tests**
  * Updated integration and unit tests to use the new SchemaFinding types, APIs, and behaviors.

### DIFF
--- a/cmd/api/src/database/graphschema.go
+++ b/cmd/api/src/database/graphschema.go
@@ -792,7 +792,7 @@ func (s *BloodhoundDB) DeleteEnvironment(ctx context.Context, environmentId int3
 	return nil
 }
 
-// CreateSchemaFinding - creates a new schema relationship finding.
+// CreateSchemaFinding - creates a new schema finding.
 func (s *BloodhoundDB) CreateSchemaFinding(ctx context.Context, findingType model.SchemaFindingType, extensionId int32, kindId int32, environmentId int32, name string, displayName string) (model.SchemaFinding, error) {
 	var finding model.SchemaFinding
 
--- a/cmd/api/src/database/graphschema_integration_test.go
+++ b/cmd/api/src/database/graphschema_integration_test.go
@@ -79,10 +79,10 @@ func createTestRelationshipKind(t *testing.T, testSuite IntegrationTestSuite, na
 	return edgeKind
 }
 
-func createTestFinding(t *testing.T, testSuite IntegrationTestSuite, extensionID int32, kindId int32, environmentID int32, name, displayName string) model.SchemaFinding {
+func createTestFinding(t *testing.T, testSuite IntegrationTestSuite, finding model.SchemaFinding) model.SchemaFinding {
 	t.Helper()
 	// Create New Finding with input arguments
-	finding, err := testSuite.BHDatabase.CreateSchemaFinding(testSuite.Context, model.SchemaFindingTypeRelationship, extensionID, kindId, environmentID, name, displayName)
+	finding, err := testSuite.BHDatabase.CreateSchemaFinding(testSuite.Context, finding.Type, finding.SchemaExtensionId, finding.KindId, finding.EnvironmentId, finding.Name, finding.DisplayName)
 	require.NoError(t, err, "unexpected error occurred when creating finding")
 	return finding
 }
@@ -3518,7 +3518,7 @@ func TestDatabase_Environments_CRUD(t *testing.T) {
 	}
 }
 
-// Graph Schema Relationship Findings may contain dynamically pre-inserted data, meaning the database
+// Graph Schema Findings may contain dynamically pre-inserted data, meaning the database
 // may already contain existing records. These tests should be written to account for said data.
 func TestDatabase_Findings_CRUD(t *testing.T) {
 	// Helper functions to assert on Findings
@@ -3528,6 +3528,7 @@ func TestDatabase_Findings_CRUD(t *testing.T) {
 			found := false
 			for _, finding := range got {
 				if finding.SchemaExtensionId == want.SchemaExtensionId &&
+					finding.Type == want.Type &&
 					finding.KindId == want.KindId &&
 					finding.EnvironmentId == want.EnvironmentId &&
 					finding.Name == want.Name &&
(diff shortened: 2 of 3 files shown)


cmd/api/src/database/migration/migrations/v8.8.0.sql
```diff
@@ -0,0 +1,69 @@
...
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- Drop list findings table, it was unused;
+DROP TABLE IF EXISTS schema_list_findings;
+
+-- Create schema_findings table
+CREATE TABLE IF NOT EXISTS schema_findings (
+  id SERIAL,
+  type INTEGER NOT NULL,
```
kpowderly, line 39: What is the difference between type here and subtype below?

cmd/api/src/database/migration/migrations/v8.8.0.sql
LawsonWillard, line 1: Just a heads up, I think the next release will be v9.0.0

cmd/api/src/database/migration/migrations/v8.8.0.sql
```diff
@@ -0,0 +1,69 @@
...
+  schema_extension_id INTEGER NOT NULL REFERENCES schema_extensions(id) ON DELETE CASCADE,
+  environment_id INTEGER NOT NULL REFERENCES schema_environments(id) ON DELETE CASCADE,
+  kind_id INTEGER NOT NULL REFERENCES kind(id),
```
sirisjo, line 42: **question**: why not `ON DELETE CASCADE` for kinds as well?

cmd/api/src/database/migration/migrations/v8.8.0.sql
```diff
@@ -0,0 +1,69 @@
...
+
+-- Drop schema_relationship_findings
+DROP TABLE IF EXISTS schema_relationship_findings;
+
+-- Add subtypes table to schema_findings
+CREATE TABLE IF NOT EXISTS schema_findings_subtypes (
+  schema_finding_id INTEGER NOT NULL REFERENCES schema_findings(id) ON DELETE CASCADE,
+  subtype TEXT NOT NULL,
+  PRIMARY KEY(schema_finding_id, subtype)
+);
```
sirisjo, line 75: **thought**: I'm wondering if instead of `subtype` being a plain string, we should store the subtypes in a separate table (something super simple like the `kind` table, with just an `id` and a `subtype`) and then reference that here.

cmd/api/src/database/migration/migrations/v8.8.0.sql
kpowderly: Dang, I just merged in v8.8.0 migration too. I checked the releases in Confluence and didn't see anything but I think you're right.

mistahj67: This hasn't been determined yet, will stick to v8.8.0 and can swap if need be

cmd/api/src/database/migration/migrations/v8.8.0.sql
mistahj67: Absolutely great suggestion and one that may ultimately be the route that's taken but that is level of complexity that I didn't want to implement at this time, given the large blast radius of this refactor. It would be fairly straightforward to migrate these out should that be a design concern in the future.

cmd/api/src/database/migration/migrations/v8.8.0.sql
mistahj67: The original `schema_relationship_findings` table didn't have this constraint so I elected to follow the design choice here as well. 

I believe one of the higher assumptions for this architecture is that no kind is ever deleted, and so it'd be a waste to have that here now

cmd/api/src/database/migration/migrations/v8.8.0.sql
mistahj67: type pertains to the locked in `relationship` or `list` type of finding which denotes if it's tied to an `edge` or `node` kind, subtypes are bespoke groupings of findings that are required to hinge certain bespoke logic off of

cmd/api/src/database/migration/migrations/v8.8.0.sql
sirisjo: Can we confirm that OG Kinds will never be deleted? That feels like an assumption made for AD and AZ, but might not hold up for OG.

cmd/api/src/database/graphschema.go
```diff
@@ -792,46 +792,46 @@ func (s *BloodhoundDB) DeleteEnvironment(ctx context.Context, environmentId int3
 	return nil
 }
 
-// CreateSchemaRelationshipFinding - creates a new schema relationship finding.
-func (s *BloodhoundDB) CreateSchemaRelationshipFinding(ctx context.Context, extensionId int32, relationshipKindId int32, environmentId int32, name string, displayName string) (model.SchemaRelationshipFinding, error) {
-	var finding model.SchemaRelationshipFinding
+// CreateSchemaFinding - creates a new schema relationship finding.
```
kpowderly: need to update the comment here

cmd/api/src/database/graphschema_integration_test.go
```diff
@@ -3522,13 +3522,13 @@ func TestDatabase_Environments_CRUD(t *testing.T) {
...
 func TestDatabase_Findings_CRUD(t *testing.T) {
 	// Helper functions to assert on Findings
-	assertContainsFindings := func(t *testing.T, got []model.SchemaRelationshipFinding, expected ...model.SchemaRelationshipFinding) {
+	assertContainsFindings := func(t *testing.T, got []model.SchemaFinding, expected ...model.SchemaFinding) {
 		t.Helper()
 		for _, want := range expected {
 			found := false
 			for _, finding := range got {
 				if finding.SchemaExtensionId == want.SchemaExtensionId &&
-					finding.RelationshipKindId == want.RelationshipKindId &&
```
kpowderly, line 3531: Technically, I think we should probably assert on the new finding type as well

sirisjo (approved):
Ran locally in conjunction with the findings on-start migration to test it out, and everything appears to be populated correctly! 🎉

PR #2420, opened 2026-02-25
feat: Create API Token Expiration Parameter. BED-7448

RaymondLaubert:
## Description

Creates the API token expiration parameter which will allow for expiring API tokens to be enabled/disabled.

## Motivation and Context

Resolves BED-7448

This change resolves the request for API tokens that are able to be expired automatically after a set time frame.

## How Has This Been Tested?

Created new unit tests that tests setting and obtaining the parameter. As well as unit tests that test updating the auth token expiration date, deleting the expired auth tokens, and validation of the signature when using auth tokens that are expired.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Introduced configurable API key expiration with a default 90-day period. Expired keys are now rejected with 401 unauthorized responses.
  * Added configuration option to enable/disable API key expiration for administrators.

* **Tests**
  * Added integration test for API token expiration parameter configuration and retrieval.

### DIFF
--- a/cmd/api/src/api/auth.go
+++ b/cmd/api/src/api/auth.go
@@ -54,7 +54,7 @@ var (
 	ErrUserNotAuthorizedForProvider = errors.New("user not authorized for this provider")
 	ErrInvalidAuthProvider          = errors.New("invalid auth provider")
 	ErrApiKeysDisabled              = errors.New("use of API keys has been disabled")
-	ErrApiKeyExpired                = errors.New("api key has expired")
+	ErrApiKeyExpired                = errors.New("API key has expired")
 )
 
 type LoginRequest struct {
--- a/cmd/api/src/api/v2/app_config.go
+++ b/cmd/api/src/api/v2/app_config.go
@@ -68,7 +68,7 @@ func (s Resources) SetApplicationConfiguration(response http.ResponseWriter, req
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("Configuration parameter %s is not valid.", parameter.Key), request), response)
 	} else if errs := parameter.Validate(); errs != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
-	} else if !checkParamAvailable(request.Context(), parameter, s.DB) {
+	} else if !checkApiKeyExpirationParamAvailable(request.Context(), parameter, s.DB) {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusForbidden, fmt.Sprintf("Configuration parameter %s is not available.", parameter.Key), request), response)
 	} else if err := s.DB.SetConfigurationParameter(request.Context(), parameter); err != nil {
 		api.HandleDatabaseError(request, response, err)
@@ -77,10 +77,9 @@ func (s Resources) SetApplicationConfiguration(response http.ResponseWriter, req
 	}
 }
 
-func checkParamAvailable(ctx context.Context, param appcfg.Parameter, flag appcfg.GetFlagByKeyer) bool {
+func checkApiKeyExpirationParamAvailable(ctx context.Context, param appcfg.Parameter, flag appcfg.GetFlagByKeyer) bool {
 	if param.Key == appcfg.APITokenExpiration {
 		if f, err := flag.GetFlagByKey(ctx, appcfg.FeatureAdcs); err != nil || !f.Enabled {
-			// TODO: Decide if this needs a log line for error path
 			return false
 		}
 	}


cmd/api/src/database/auth.go
```diff
@@ -403,6 +410,42 @@ func (s *BloodhoundDB) UpdateAuthToken(ctx context.Context, authToken model.Auth
...
+	// Create the Audit Details
+	auditDetails := model.AuditData{
+		"table":  "auth_tokens",
+		"column": "expires_at",
+	}
+
+	// Create the Audit Entry
+	auditEntry, err := model.NewAuditEntry(model.AuditLogActionUpdateAuthTokenExpiration, model.AuditLogStatusIntent, auditDetails)
+	if err != nil {
+		return errors.New("error creating update auth token expiration audit entry")
```
superlinkx: Why are we throwing away the error value from the actual call here?

cmd/api/src/model/appcfg/parameter.go
```diff
@@ -579,3 +582,27 @@ func GetEnvironmentTargetedAccessControlParameters(ctx context.Context, service
...
+}
+
+func GetAPITokenExpirationParameter(ctx context.Context, service ParameterService) APITokenExpirationParameter {
+	result := APITokenExpirationParameter{Enabled: false, ExpirationPeriod: 90}
+
+	if cfg, err := service.GetConfigurationParameter(ctx, APITokenExpiration); err != nil {
+		slog.WarnContext(ctx, "Failed to fetch API tokens expiration configuration; returning default values.")
+	} else if err := cfg.Map(&result); err != nil {
+		slog.WarnContext(ctx, "Invalid API tokens expiration configuration supplied, returning default values.",
+			slog.String("invalid_configuration", err.Error()),
```
superlinkx: Errors should only be expressed with `attr.Error(err)` to ensure they're easier to parse in logs

cmd/api/src/database/auth.go
```diff
@@ -403,6 +410,42 @@ func (s *BloodhoundDB) UpdateAuthToken(ctx context.Context, authToken model.Auth
...
+	// Obtain the API Key Expiration Parameter
+	apiExpParam := appcfg.GetAPITokenExpirationParameter(ctx, s)
+
+	// If Enabled Set Expiration Date for Auth Tokens; Unless the Token Expires before the New Expiration
+	// If Disabled Set Expiration Date for Auth Tokens to NULL
+	if apiExpParam.Enabled {
+		expirationDate := time.Now().AddDate(0, 0, apiExpParam.ExpirationPeriod)
+
+		return s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+			result := tx.WithContext(ctx).Exec("UPDATE auth_tokens SET expires_at = ? WHERE expires_at NOT BETWEEN ? AND ?", expirationDate, time.Now(), expirationDate)
```
superlinkx: Subtle bug, if an auth token is already expired, this will cause it to become alive again unless it's been properly deleted from the database

superlinkx: I think what we want is to set it for all where expires_at is greater than expiration date, or null, allowing less than NOW() to remain as they are

cmd/api/src/database/auth.go
```diff
@@ -403,6 +410,42 @@ func (s *BloodhoundDB) UpdateAuthToken(ctx context.Context, authToken model.Auth
...
+			return CheckError(result)
+		})
+	} else {
+		return s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+			result := tx.WithContext(ctx).Exec("UPDATE auth_tokens SET expires_at = NULL")
```
superlinkx, line 447: On the subject of logic, I'm realizing we may not have thought through how to handle the fact that some API keys before this is turned off are effectively "expired", but when we null them out, they'll not only be alive again while the feature is off (which seems mostly expected, but we may want to have product weigh in on user expectation here), but more importantly if the toggle goes back on, they'll become alive for the full duration again.

This isn't so much an engineering problem as a product design problem, so we should talk with the EM+PM to resolve if any changes might be warranted.

cmd/api/src/api/v2/app_config.go
```diff
@@ -69,6 +69,10 @@ func (s Resources) SetApplicationConfiguration(response http.ResponseWriter, req
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
 	} else if err = s.DB.SetConfigurationParameter(request.Context(), parameter); err != nil {
 		api.HandleDatabaseError(request, response, err)
+	} else if parameter.Key == appcfg.APITokenExpiration {
```
superlinkx: Just realize this will prevent us from returning status OK anytime the parameter is APITokenExpiration. Move this into the else block itself so it'll always fallthrough to the statusOK if no errors occur

cmd/api/src/api/auth.go
```diff
@@ -314,6 +315,8 @@ func (s AuthenticatorBase) ValidateRequestSignature(tokenID uuid.UUID, request *
 		return auth.Context{}, http.StatusBadRequest, fmt.Errorf("malformed signature header: %w", err)
 	} else if authToken, err := s.db.GetAuthToken(request.Context(), tokenID); err != nil {
 		return handleAuthDBError(err)
+	} else if authToken.ExpiresAt.Before(time.Now()) {
+		return auth.Context{}, http.StatusUnauthorized, ErrApiKeyExpired
```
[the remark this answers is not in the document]
superlinkx: I missed this in my review, but it's definitely correct. Same thing goes for wiring up collector tokens as well

cmd/api/src/api/v2/app_config.go
RaymondLaubert: I originally had it in the else to prevent exactly this, but ended up fucking up the logic after realizing I couldn't check first and then use `&&` while assigning err to the second half lol. Fixed this though and moved it back into the else.

cmd/api/src/model/appcfg/parameter.go
RaymondLaubert: Changed it to attr.Error() and I do remember you mentioning that for the one we fixed above my mistake.

cmd/api/src/database/auth.go
RaymondLaubert: Fixed, I have no idea what I was doing here.

cmd/api/src/database/auth.go
RaymondLaubert: Already implemented the change, but we talked about that. :)

cmd/api/src/database/auth.go
RaymondLaubert: I believe the decision was to delete tokens that have expired when turning it off or on, I am starting to implement this but think some other deletion instances may be good to add as well. Such as when an expired key is attempted to be used but we can discuss that further and get products input as well.

cmd/api/src/api/v2/app_config.go
```diff
@@ -70,6 +70,13 @@ func (s Resources) SetApplicationConfiguration(response http.ResponseWriter, req
 	} else if err = s.DB.SetConfigurationParameter(request.Context(), parameter); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else {
+		if parameter.Key == appcfg.APITokenExpiration {
+			if err = s.DB.UpdateAuthTokenExpiration(request.Context()); err != nil {
+				api.HandleDatabaseError(request, response, err)
+				return
+			}
+		}
+
```
superlinkx: Looking at how this file works, this may make more sense to be checked in the existing database method, as it's effectively just some additional processing at the database layer already

cmd/api/src/api/auth.go
```diff
@@ -296,6 +298,17 @@ func handleAuthDBError(err error) (auth.Context, int, error) {
 	}
 }
 
+func validateAuthTokenExpiration(nt sql.NullTime) bool {
+	if !nt.Valid {
+		slog.Warn("Auth token expiration value is null, but should have a valid date when API token expiration is enabled.")
+		return false
+	} else if nt.Time.Before(time.Now()) {
+		return false
+	}
```
superlinkx: This can be simplified to `nt.Time.Before(time.Now())`, as a "null" time is going to be the zero value for time.Time, and that will always be before time.Now()*. I don't think warning on null is necessary here, so just returning false if null or otherwise lower than current time seems perfectly reasonable.

As a result, we don't need this helper function at all, as the one place it's being used can simply substitute this function call for `nt.Time.Before(time.Now())`, since it's currently having to negate the false

*Barring some truly crazy bad stuff happening outside the application's control

cmd/api/src/database/migration/migrations/v8.9.0.sql
superlinkx, line 1: This will need to get moved to v9.0.0.sql, since v8.9.0.sql is locked in for release on Monday

cmd/ui/src/components/SigmaChart/GraphEvents.tsx
superlinkx, line 1: I don't think this was meant to be in here

RaymondLaubert: Yeah, I don't know what it is but something in my IDE or something when maybe I'm running `just prepare-for-codereview` is auto-formatting files that I have opened; typically when looking at references or trying to find where things go or start from. Should I just remove the change to this file?

superlinkx: I think it's a bad merge, not entirely sure. If this isn't an intentional change, just taking the copy from `main` should be good enough

cmd/api/src/api/auth.go
RaymondLaubert: Good catch and that makes sense. Removed it and replaced

cmd/api/src/api/v2/app_config.go
RaymondLaubert: Resolved together.

cmd/api/src/database/migration/migrations/v8.9.0.sql
RaymondLaubert: Resolved together.

cmd/api/src/api/v2/app_config.go
```diff
@@ -67,9 +68,22 @@ func (s Resources) SetApplicationConfiguration(response http.ResponseWriter, req
...
+	} else if !checkParamAvailable(request.Context(), parameter, s.DB) {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusForbidden, fmt.Sprintf("Configuration parameter %s is not available.", parameter.Key), request), response)
+	} else if err := s.DB.SetConfigurationParameter(request.Context(), parameter); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else {
 		api.WriteBasicResponse(request.Context(), appConfig, http.StatusOK, response)
 	}
 }
+
+func checkParamAvailable(ctx context.Context, param appcfg.Parameter, flag appcfg.GetFlagByKeyer) bool {
```
irshadaj: Since this function sits in a file that deals with _all_ app configs, we should make the name more specific: `checkAPIKeyExpirationParamAvailable` works better

cmd/api/src/api/v2/app_config.go
```diff
@@ -67,9 +68,22 @@ func (s Resources) SetApplicationConfiguration(response http.ResponseWriter, req
...
+	if param.Key == appcfg.APITokenExpiration {
+		if f, err := flag.GetFlagByKey(ctx, appcfg.FeatureAdcs); err != nil || !f.Enabled {
+			// TODO: Decide if this needs a log line for error path
```
irshadaj: Not needed, we're returning the error from the handler. That should suffice.

cmd/api/src/api/auth.go
```diff
@@ -54,6 +54,7 @@ var (
 	ErrUserNotAuthorizedForProvider = errors.New("user not authorized for this provider")
 	ErrInvalidAuthProvider          = errors.New("invalid auth provider")
 	ErrApiKeysDisabled              = errors.New("use of API keys has been disabled")
+	ErrApiKeyExpired                = errors.New("api key has expired")
```
irshadaj: We start errors in lowercase but API is in uppercase- I'd change this to `expired API key`

RaymondLaubert: I agree but wasn't sure how it was done here. Changed

cmd/api/src/api/v2/app_config.go
RaymondLaubert: Agreed, changed.

cmd/api/src/api/v2/app_config.go
RaymondLaubert: Resolved.

PR #2422, opened 2026-02-25
feat: Support Arbitrary Endpoint Matching for Edge Ingest - BED-7451

zinic:
## Description

Updates OpenGraph Ingest schema to support a `match_by` strategy of *"property"* that then allows an OG implementer to specify an arbitrary property to match against for an edge endpoint. 

## Motivation and Context

Resolves BED-7451

*HYBRID ENVIRONMENTS PATHS*

## How Has This Been Tested?

Existing tests.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added support for property-based matching when resolving entities in relationships, with operators for exact and case-insensitive matching.
  * Enhanced edge definitions to include a properties field and support multiple property matchers.

* **Improvements**
  * Optimized entity resolution performance with caching mechanisms.
  * Updated relationship ingestion workflow for improved efficiency.

* **Tests**
  * Added integration tests for property-based entity resolution and endpoint matching scenarios.

### DIFF
--- a/cmd/api/src/daemons/datapipe/pipeline_integration_test.go
+++ b/cmd/api/src/daemons/datapipe/pipeline_integration_test.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/specterops/bloodhound/cmd/api/src/model"
 	"github.com/specterops/bloodhound/cmd/api/src/services/graphify"
+	"github.com/specterops/bloodhound/cmd/api/src/services/graphify/endpoint"
 	"github.com/specterops/bloodhound/packages/go/lab/generic"
 	"github.com/specterops/dawgs/graph"
 	"github.com/stretchr/testify/require"
@@ -47,8 +48,10 @@ func TestDeleteData_Sourceless(t *testing.T) {
 	)
 
 	defer teardownIntegrationTestSuite(t, &testSuite)
+	ingestCtx := graphify.NewIngestContext(ctx, graphify.WithEndpointResolver(endpoint.NewResolver(testSuite.GraphDB)))
+
 	for _, file := range files {
-		fileData, err := testSuite.GraphifyService.ProcessIngestFile(graphify.NewIngestContext(ctx), model.IngestTask{StoredFileName: file, FileType: model.FileTypeJson})
+		fileData, err := testSuite.GraphifyService.ProcessIngestFile(ingestCtx, model.IngestTask{StoredFileName: file, FileType: model.FileTypeJson})
 		require.NoError(t, err)
 
 		failed := 0
@@ -94,9 +97,10 @@ func TestDeleteData_SourceKinds(t *testing.T) {
 	)
 
 	defer teardownIntegrationTestSuite(t, &testSuite)
+	ingestCtx := graphify.NewIngestContext(ctx, graphify.WithEndpointResolver(endpoint.NewResolver(testSuite.GraphDB)))
 
 	for _, file := range files {
-		fileData, err := testSuite.GraphifyService.ProcessIngestFile(graphify.NewIngestContext(ctx), model.IngestTask{StoredFileName: file, FileType: model.FileTypeJson})
+		fileData, err := testSuite.GraphifyService.ProcessIngestFile(ingestCtx, model.IngestTask{StoredFileName: file, FileType: model.FileTypeJson})
 		require.NoError(t, err)
 
 		failed := 0
@@ -142,9 +146,10 @@ func TestDeleteData_All(t *testing.T) {
 	)
 
 	defer teardownIntegrationTestSuite(t, &testSuite)
+	ingestCtx := graphify.NewIngestContext(ctx, graphify.WithEndpointResolver(endpoint.NewResolver(testSuite.GraphDB)))
 
 	for _, file := range files {
-		fileData, err := testSuite.GraphifyService.ProcessIngestFile(graphify.NewIngestContext(ctx), model.IngestTask{StoredFileName: file, FileType: model.FileTypeJson})
+		fileData, err := testSuite.GraphifyService.ProcessIngestFile(ingestCtx, model.IngestTask{StoredFileName: file, FileType: model.FileTypeJson})
 		require.NoError(t, err)
 
 		failed := 0
--- a/cmd/api/src/services/graphify/ingest_integration_test.go
+++ b/cmd/api/src/services/graphify/ingest_integration_test.go
@@ -97,7 +97,7 @@ func Test_ReadFileForIngest(t *testing.T) {
 
 		testContext.DatabaseTestWithSetup(func(harness *integration.HarnessDetails) error { return nil }, func(harness integration.HarnessDetails, db graph.Database) {
 			_ = db.BatchOperation(testContext.Context(), func(batch graph.Batch) error {
-				ingestContext := graphify.NewIngestContext(testContext.Context(), graphify.WithBatchUpdater(batch))
+				ingestContext := graphify.NewIngestContext(testContext.Context(), graphify.WithBatchUpdater(batch), graphify.WithEndpointResolver(endpoint.NewResolver(testContext.Graph.Database)))
 
 				err := graphify.ReadFileForIngest(ingestContext, invalidReader, readOptions)
 				require.NotNil(t, err)
--- a/cmd/api/src/services/graphify/ingestnodes_integration_test.go
+++ b/cmd/api/src/services/graphify/ingestnodes_integration_test.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/specterops/bloodhound/cmd/api/src/daemons/changelog"
 	"github.com/specterops/bloodhound/cmd/api/src/services/graphify"
+	"github.com/specterops/bloodhound/cmd/api/src/services/graphify/endpoint"
 	"github.com/specterops/bloodhound/packages/go/ein"
 	"github.com/specterops/bloodhound/packages/go/graphschema/common"
 	"github.com/specterops/dawgs/graph"
@@ -64,10 +65,12 @@ func ingestTestNode(t *testing.T, suite IntegrationTestSuite, nodeData testNodeD
 		if withChangelog {
 			ingestCtx = graphify.NewIngestContext(ctx,
 				graphify.WithBatchUpdater(batch),
-				graphify.WithChangeManager(suite.Changelog))
+				graphify.WithChangeManager(suite.Changelog),
+				graphify.WithEndpointResolver(endpoint.NewResolver(suite.GraphDB)))
 		} else {
 			ingestCtx = graphify.NewIngestContext(ctx,
-				graphify.WithBatchUpdater(batch))
+				graphify.WithBatchUpdater(batch),
+				graphify.WithEndpointResolver(endpoint.NewResolver(suite.GraphDB)))
 		}
 
 		return graphify.IngestNode(ingestCtx, nodeData.SourceKind,
@@ -162,7 +165,8 @@ func TestIngestNode(t *testing.T) {
 		err := testSuite.GraphDB.BatchOperation(ctx, func(batch graph.Batch) error {
 			ingestCtx := graphify.NewIngestContext(ctx,
 				graphify.WithBatchUpdater(batch),
-				graphify.WithChangeManager(testSuite.Changelog))
+				graphify.WithChangeManager(testSuite.Changelog),
+				graphify.WithEndpointResolver(endpoint.NewResolver(testSuite.GraphDB)))
 
 			node := ein.IngestibleNode{
 				ObjectID:    nodeData.ObjectID,


cmd/api/src/services/graphify/endpoint/harness/main.go
```diff
@@ -0,0 +1,165 @@
...
+
+	slog.Info("Stopping Metrics HTTPServer")
+
+	if err := StopHTTP(ctx); err != nil {
+		Fatalf(err.Error())
+	}
+}
+
+func main() {
+	Launch("/home/zinic/configs/integration-config-local.json")
```
superlinkx: This needs to be changed to a sane default, and likely configurable

zinic: Wasn't even sure if I wanted to check this in - worth a conversation since this is a performance testing harness and nothing more.

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,217 @@
...
+            "required": ["match_by"]
+          },
+          "then": {
+            "required": ["property_matchers"]
+          },
+          "else": {
+            "required": ["value"]
+          }
+        }
+      ]
```
craigw-SO: I think this is pretty close. We can remove the allOf wrapper, don't need it in this case. We also need to pull the `"required": ["match_by"]` out of the if block. Also probably makes sense to guard against the opposite property type than the one required, I would think?

So something like this I think:
``` json
"required": ["match_by"],
  "if": {
    "properties": { "match_by": { "const": "property" } }
  },
  "then": {
    "required": ["property_matchers"],
    "not": { "required": ["value"] }
  },
  "else": {
    "required": ["value"],
    "not": { "required": ["property_matchers"] }
  }
```

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,217 @@
...
+  "title": "Generic Ingest Edge",
+  "description": "Defines an edge between two nodes in a generic graph ingestion system. Each edge specifies a start and end node using either a unique identifier (id) or a name-based lookup. A kind is required to indicate the relationship type. Optional properties may include custom attributes. You may optionally constrain the start or end node to a specific kind using the kind field inside each reference.",
+  "type": "object",
+  "properties": {
+    "start": {
+      "type": "object",
+      "properties": {
+        "match_by": {
+          "type": "string",
+          "enum": ["id", "name"],
```
craigw-SO: Need to update "name" to "property"

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,217 @@
...
+          }
+        }
+      ]
+    },
+    "end": {
+      "type": "object",
+      "properties": {
+        "match_by": {
+          "type": "string",
+          "enum": ["id", "name"],
```
craigw-SO: Swap "name" to "property"

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,217 @@
...
+    "properties": {
+      "type": ["object", "null"],
+      "description": "A key-value map of edge attributes. Values must not be objects. If a value is an array, it must contain only primitive types (e.g., strings, numbers, booleans) and must be homogeneous (all items must be of the same type).",
+      "additionalProperties": {
+        "type": ["string", "number", "boolean", "array"],
+        "items": {
+          "not": {
+            "type": "object"
+          }
         }
```
craigw-SO: When Jeff was working on docs for us, we realized that the validation did not match the description.

In spirit of no broken windows, we should update this to reflect what was actually intended. Should be applied to both edges AND nodes.

``` json
"oneOf": [
                    { "type": "string" },
                    { "type": "number" },
                    { "type": "boolean" },
                    {
                        "type": "array",
                        "items": { "type": "string" },
                        "minItems": 1
                    },
                    {
                        "type": "array",
                        "items": { "type": "number" },
                        "minItems": 1
                    },
                    {
                        "type": "array",
                        "items": { "type": "boolean" },
                        "minItems": 1
                    }
]
```

See this thread: https://github.com/SpecterOps/bloodhound-docs/pull/109/changes/57dae5a82b8455936f3073c9d1df6c578bef65de#diff-1067a819fd3bee3efdd6572dedadf2f3343b997967942e37a4d60de47f70011d

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,217 @@
...
+              "operator": {
+                "type": "object",
+                "enum": ["equals"]
+              },
+              "value": {
+                "type": ["string", "number", "boolean"]
+              }
             },
-            "required": ["value"]
+            "required": ["name", "operator", "value"]
```
craigw-SO: Remembered one other question I had: is there a particular reason we're using "name" here instead of "key"? There's already specifically a "name" property, so I worry about the context getting mixed up here a bit. Key-Value seems the obvious choice to me, but not sure if there's other thinking going on here.

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,180 @@
...
-                "kind": {
-                    "type": "string",
-                    "description": "Optional kind filter; the referenced node must have this kind."
-                }
-            },
-            "required": ["value"]
+  "title": "Generic Ingest Edge",
+  "description": "Defines an edge between two nodes in a generic graph ingestion system. Each edge specifies a start and end node using either a unique identifier (id) or a name-based lookup. A kind is required to indicate the relationship type. Optional properties may include custom attributes. You may optionally constrain the start or end node to a specific kind using the kind field inside each reference.",
+  "type": "object",
+  "$defs": {
```
craigw-SO, line 5: Beautiful modularity, love it :chefs_kiss

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,180 @@
...
+    "property_map": {
+      "type": ["object", "null"],
+      "description": "A key-value map of edge attributes. Values must not be objects. If a value is an array, it must contain only primitive types (e.g., strings, numbers, booleans) and must be homogeneous (all items must be of the same type).",
+      "additionalProperties": {
+        "type": ["string", "number", "boolean", "array"],
+        "items": {
+          "not": {
```
craigw-SO: Think there's an extraneous "not" here, just needs removed to bring into alignment with nodes.

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,180 @@
...
-                },
-                "kind": {
-                    "type": "string",
-                    "description": "Optional kind filter; the referenced node must have this kind."
-                }
+              "key": {
+                "type": "string"
+              },
+              "operator": {
+                "type": "object",
```
craigw-SO: As far as I know this works fine, but personally curious why "object" and not "string" here? We do similar for match_by but use string there.

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,180 @@
...
+        "value": {
+          "type": "string",
+          "description": "The value used for matching — either an object ID or a name, depending on match_by."
         },
+        "kind": {
+          "type": "string",
+          "description": "Optional kind filter; the referenced node must have this kind."
+        }
+      },
+      "if": {
```
craigw-SO, line 62: I think we're missing `"required": ["match_by"]` right before this

wes-mil:
Have a few things:

[This](https://github.com/SpecterOps/BloodHound/blob/0c31df9c03803902be54e4baf41e2a3f233171e1/cmd/api/src/services/upload/streamdecoder.go#L376-L388) function needs to be removed as the JSON schema now does the array check

It gets used [here](https://github.com/SpecterOps/BloodHound/blob/0c31df9c03803902be54e4baf41e2a3f233171e1/cmd/api/src/services/upload/streamdecoder.go#L463-L469), which also needs to be removed as we don't need to manually check the properties bag any more

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,180 @@
...
+    "kind": {
+      "type": "string",
+      "description": "Edge kind name must contain only alphanumeric characters and underscores.",
+      "pattern": "^[A-Za-z0-9_]+$"
+    },
+    "properties": {
+      "$ref": "#/$defs/property_map"
+    }
+  },
+  "required": ["start", "end", "kind"],
```
craigw-SO: Honest question, do we not require the properties bag as part of ingest? Do we not require a "name" field at minimum within it? I see the "null" type option for the property_map, not sure if that means "properties" is present but empty.

cmd/api/src/services/upload/jsonschema/node.json
```diff
@@ -1,49 +1,61 @@
...
+    },
+    "kinds": {
+      "type": ["array"],
+      "items": { "type": "string" },
+      "minItems": 0,
+      "maxItems": 3,
+      "description": "An array of kind labels for the node. The first element is treated as the node's primary kind and is used to determine which icon to display in the graph UI. This primary kind is only used for visual representation and has no semantic significance for data processing."
+    }
+  },
+  "required": ["id", "kinds"],
```
craigw-SO: Similar to above, do we not require properties?

wes-mil:
Not sure if I'm missing the intent, but the way I did property bag validation in chow was with this:
```JSON
"properties": {
      "type": ["object", "null"],
      "description": "A key-value map of node attributes. Values must not be objects. If a value is an array, it must contain only primitive types (e.g., strings, numbers, booleans) and must be homogeneous (all items must be of the same type).",
      "additionalProperties": {
        "type": ["string", "number", "boolean", "array"],
        "anyOf": [
          { "type": "string" },
          { "type": "number" },
          { "type": "boolean" },
          { "type": "array", "items": { "type": "string" } },
          { "type": "array", "items": { "type": "number" } },
          { "type": "array", "items": { "type": "boolean" } }
        ]
      }
    },
```

cmd/api/src/services/upload/streamdecoder_test.go
```diff
@@ -190,10 +190,17 @@ type testEdge struct {
...
+	Operator string `json:"operator"`
+	Value    any    `json:"value"`
+}
+
 type edgePiece struct {
-	Value   string `json:"value,omitempty"`
-	MatchBy string `json:"match_by,omitempty"`
-	Kind    string `json:"kind,omitempty"`
+	Value            string `json:"value,omitempty"`
+	MatchBy          string `json:"match_by,omitempty"`
```
craigw-SO: Match_by should be required, so we can probably leave off "omitempty"?

packages/go/ein/incoming_models.go
```diff
@@ -381,8 +381,15 @@ type GenericEdge struct {
 	Properties map[string]any
 }
 
+type PropertyMatcher struct {
+	Name     string `json:"name"`
```
craigw-SO: Swap to "key"

packages/go/ein/models.go
```diff
@@ -35,17 +35,43 @@ func NewIngestibleRelationship(source IngestibleEndpoint, target IngestibleEndpo
...
 
 const (
-	MatchByID   IngestMatchStrategy = "id"
-	MatchByName IngestMatchStrategy = "name"
+	MatchByID       IngestMatchStrategy = "id"
+	MatchByName     IngestMatchStrategy = "name"
+	MatchByProperty IngestMatchStrategy = "property"
+
+	OperatorEquals           IngestMatchOperator = "equals"
+	OperatorEqualsIgnoreCase IngestMatchOperator = "equals_ignore_case"
```
craigw-SO, line 50: Did we end up nixing this for now?

packages/go/ein/models.go
```diff
@@ -35,17 +35,43 @@ func NewIngestibleRelationship(source IngestibleEndpoint, target IngestibleEndpo
...
 )
 
+type MatchExpression struct {
+	Property string
```
craigw-SO: Feels weird to call this just property. Would "key" or "property_key" be better?

packages/go/ein/models.go
```diff
@@ -35,17 +35,43 @@ func NewIngestibleRelationship(source IngestibleEndpoint, target IngestibleEndpo
...
+	Operator IngestMatchOperator
+	Value    any
+}
+
+func PropertyMatchersToMatchExpressions(matchers []PropertyMatcher) []MatchExpression {
+	expressions := make([]MatchExpression, len(matchers))
+
+	for idx, matcher := range matchers {
+		expressions[idx] = MatchExpression{
+			Property: matcher.Name,
```
craigw-SO: Swap to key

cmd/api/src/services/upload/streamdecoder_test.go
```diff
@@ -190,10 +190,17 @@ type testEdge struct {
 	Properties map[string]any `json:"properties"`
 }
 
+type propertyMatcher struct {
+	Name     string `json:"name"`
```
craigw-SO: Swap to "key"

cmd/api/src/services/graphify/endpoint/fetch_test.go
```diff
@@ -0,0 +1,138 @@
...
+			input:    nil,
+			expected: []ein.MatchExpression{},
+		},
+		{
+			name: "single element",
+			input: []ein.MatchExpression{
+				{Property: "Name", Operator: "equals", Value: "test"},
+			},
+			expected: []ein.MatchExpression{
+				{Property: "Name", Operator: "equals", Value: "test"},
```
craigw-SO: Swap "name" to "key" here, I think?

cmd/api/src/services/graphify/endpoint/fetch_test.go
```diff
@@ -0,0 +1,138 @@
...
+			name: "same property, different operators (sorted by operator)",
+			input: []ein.MatchExpression{
+				{Property: "Name", Operator: "z_operator", Value: 1},
+				{Property: "Name", Operator: "a_operator", Value: 2},
+				{Property: "Name", Operator: "m_operator", Value: 3},
+			},
+			expected: []ein.MatchExpression{
+				{Property: "Name", Operator: "a_operator", Value: 2},
+				{Property: "Name", Operator: "m_operator", Value: 3},
+				{Property: "Name", Operator: "z_operator", Value: 1},
```
craigw-SO: Swap "name" to "key" I think?

cmd/api/src/services/graphify/endpoint/fetch_test.go
```diff
@@ -0,0 +1,138 @@
...
+				{Property: "Name", Operator: "equals", Value: "Z"},
+				{Property: "ID", Operator: "equals", Value: "100"},
+				{Property: "Name", Operator: "contains", Value: "A"},
+				{Property: "ID", Operator: "contains", Value: "50"},
+			},
+			expected: []ein.MatchExpression{
+				{Property: "ID", Operator: "contains", Value: "50"},
+				{Property: "ID", Operator: "equals", Value: "100"},
+				{Property: "Name", Operator: "contains", Value: "A"},
+				{Property: "Name", Operator: "equals", Value: "Z"},
```
craigw-SO: Swap "name" to "key" I think?

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,180 @@
...
+        "timestamp": "2025-04-16T12:00:00Z",
+        "duration_minutes": 45
+      }
+    },
+    {
+      "start": {
+        "match_by": "property",
+        "property_matchers": [
+          {
+            "name": "prop_1",
```
craigw-SO: Swap "name" to "key"

craigw-SO:
> Have a few things:
> 
> [This](https://github.com/SpecterOps/BloodHound/blob/0c31df9c03803902be54e4baf41e2a3f233171e1/cmd/api/src/services/upload/streamdecoder.go#L376-L388) function needs to be removed as the JSON schema now does the array check
> 
> It gets used [here](https://github.com/SpecterOps/BloodHound/blob/0c31df9c03803902be54e4baf41e2a3f233171e1/cmd/api/src/services/upload/streamdecoder.go#L463-L469), which also needs to be removed as we don't need to manually check the properties bag any more

@wes-mil Does the streamdecoder or the JSON schema check happen first? It looks like the homogeneity test is more verbose, so if that happens first, it might be nice to leave in and get more specific on what failed. I think the JSON schema check just returns a blanket "this failed" without reason for why, right?

cmd/api/src/services/upload/jsonschema/edge.json
wes-mil: I'm fairly sure this is not a valid way to validate an array for our need. In order to validate that all the elements are the same throughout the array, you need to do an "anyOf" on the whole property, not just the items like I have here: https://github.com/SpecterOps/BloodHound/pull/2422#issuecomment-4013021452

wes-mil:
@craigw-SO The schema validation happens first. The way this works, this is actually going to cause us to have a lot of really verbose errors that aren't super useful

wes-mil:
This is what we will end up getting back from schema validation:
image

craigw-SO:
@wes-mil John did a similar type validation using oneOf instead of anyOf. Though one difference I'm noticing is that his specifies no empty arrays and yours does not. Should we be allowing empty arrays?

wes-mil:
The problem with `oneOf` is that it then has to check every line in the list, `anyOf` stops processing after you find one valid match. Since all of our options are fully distinct, they have the same result but `anyOf` is faster. I allowed empty arrays because I figured there could be a reason for it, but I'm not married to it

cmd/api/src/services/upload/jsonschema/edge.json
craigw-SO: Coordinated w/ Wes and agree that this section needs rewritten.

If we don't want to allow for empty arrays, then:
```json
"additionalProperties": {
	"anyOf": [
		{ "type": "string" },
		{ "type": "number" },
		{ "type": "boolean" },
		{
			"type": "array",
			"minItems": 1,
			"anyOf": [
				{ "items": { "type": "string" } },
				{ "items": { "type": "number" } },
				{ "items": { "type": "boolean" } }
			]
		}
	]
}
```

If we want to allow for empty arrays, then just remove `minItems`.

Tested against a couple validators to confirmed.

wes-mil: You can still include `"type": ["string", "number", "boolean", "array"],`, it will simply check that the type passes both of these

cmd/api/src/services/upload/jsonschema/node.json
zinic: Uh... I think so? Will change this.

cmd/api/src/services/upload/streamdecoder_test.go
zinic: `omitempty` here is to make malformed payloads for testing

packages/go/ein/models.go
zinic: I'm just not advertising it in the contract but I don't want to loose the machinery since it's critical for `name` matching backwards compat.

cmd/api/src/services/upload/jsonschema/edge.json
zinic: I'll just update this to include the properties as required

cmd/api/src/services/upload/jsonschema/edge.json
zinic: Will update

cmd/api/src/services/upload/jsonschema/edge.json
zinic: Pulling this in

cmd/api/src/services/upload/jsonschema/edge.json
zinic: Will update

cmd/api/src/services/upload/jsonschema/edge.json
zinic: Will update

cmd/api/src/daemons/datapipe/datapipe_integration_test.go
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-//go:build integration
+////go:build integration
```
wes-mil: Does this break the go build?

cmd/api/src/services/upload/jsonschema/edge.json
```diff
@@ -1,124 +1,195 @@
...
+          {
+            "type": "array",
+            "minItems": 1,
+            "anyOf": [
+              { "items": { "type": "string" } },
+              { "items": { "type": "number" } },
+              { "items": { "type": "boolean" } }
+            ]
+          }
+        ]
```
wes-mil, line 22: Would much prefer this to be `anyOf` instead of `oneOf` because `oneOf` required that the JSON Schema checks against every property everytime to ensure uniqueness. That would be a pretty significant performance degrade, and we already know that there's no potential overlap from these. I also still feel that we should permit empty arrays

cmd/api/src/services/upload/jsonschema/node.json
```diff
@@ -1,49 +1,61 @@
...
+        "items": {
+          "oneOf": [
+            { "type": "string" },
+            { "type": "number" },
+            { "type": "boolean" },
+            { "type": "array", "items": { "type": "string" }, "minItems": 1 },
+            { "type": "array", "items": { "type": "number" }, "minItems": 1 },
+            { "type": "array", "items": { "type": "boolean" }, "minItems": 1 }
+          ]
         }
```
wes-mil: This is the broken version of the additionalProperties matching

cmd/api/src/services/graphify/ingestrelationships_integration_test.go
```diff
@@ -704,36 +709,6 @@ func Test_ResolveRelationships(t *testing.T) {
...
-					require.Empty(t, updates)
-
-					return nil
-				})
-
-				require.Nil(t, err)
-
-			})
-	})
-
```
wes-mil, line 736: Why was this test removed?

cmd/api/src/services/graphify/endpoint/resolver_test.go
```diff
@@ -0,0 +1,48 @@
...
+	// Mock out the read transaction calls
+	mockGraph.EXPECT().ReadTransaction(ctx, gomock.Any()).Return(nil).AnyTimes()
+
+	// Ensure that start and done are reentrant-safe
+	resolver.Start(ctx, 4)
+	resolver.Start(ctx, 4)
+
+	assert.NoError(t, resolver.Done())
+	assert.NoError(t, resolver.Done())
+}
```
wes-mil, line 48: Pretty confused what's going on here

zinic: Tests reenterant behavior so we don't get weird panics if someone wires up start and done incorrectly.

cmd/api/src/daemons/datapipe/datapipe_integration_test.go
zinic: my debugging - fixing

cmd/api/src/services/upload/jsonschema/node.json
zinic: patching up

cmd/api/src/services/graphify/ingestrelationships_integration_test.go
zinic: whoops - restoring and updating

PR #2452, opened 2026-03-02
fix(OpenGraph): Custom Icons should be added to the custom_node_kinds table on schema upsert BED-7500

sirisjo:
## Description

We want to make sure custom icons are added to the `custom_node_kinds` table when an OpenGraph schema is upserted. The icons should also be upserted -- create a new icon if the icon doesn't exist, or edit it if it does. 

A new, nullable `schema_node_kind_id` column was added to the `custom_node_kinds` table to track icons that are upserted via OpenGraph ingest. Any time an icon with a `schema_node_kind_id` is updated, the matching record in `schema_node_kinds` is also updated, to keep them in sync.

## Motivation and Context

Resolves BED-7500

Solves the issue of needing to use the `custom-nodes` API to add icons for OpenGraph kinds. Now, if the icon definitions are included in the schema extension file, they are added to the `custom_node_kinds` table and will show up across the application.

## How Has This Been Tested?

- integration tests added
- tested by adding a kind via the `custom-nodes` API, and then editing it by uploading a schema extension that included updates to that kind, and confirming the icon was updated ✅ 
- tested by uploading a schema extension with multiple kinds and confirming they were all added to the `custom_node_kinds` table ✅ 
- tested the existing `custom-nodes` API and confirmed the endpoints all still function as expected, even with the addition of the new `id` column on the backend ✅ 
## Screenshots (optional):

Screenshot 2026-03-02 at 15 56 10

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Link custom node icons to graph schema node kinds for improved visual representation.
  * Add color validation for custom icons using hex format to ensure visual consistency.
  * Support icon updates for node kinds within graph extension workflows.

* **Tests**
  * Added comprehensive integration tests for custom node operations including create, retrieve, update, and delete scenarios.

### DIFF
--- a/cmd/api/src/database/graphschema.go
+++ b/cmd/api/src/database/graphschema.go
@@ -306,7 +306,7 @@ func (s *BloodhoundDB) CreateGraphSchemaNodeKind(ctx context.Context, name strin
 		FROM dawgs_kind dk
 		RETURNING id, kind_id, schema_extension_id, display_name, description, is_display_kind, icon, icon_color, created_at, updated_at, deleted_at
 	)
-	SELECT isn.id, isn.kind_id, isn.schema_extension_id, dk.name, isn.display_name, isn.description, isn.is_display_kind, isn.icon, isn.icon_color, isn.created_at, isn.updated_at, isn.deleted_at
+	SELECT isn.id, isn.schema_extension_id, dk.name, isn.display_name, isn.description, isn.is_display_kind, isn.icon, isn.icon_color, isn.created_at, isn.updated_at, isn.deleted_at
 	FROM inserted_schema_node isn
 	JOIN dawgs_kind dk ON isn.kind_id = dk.id;`, name, extensionId, displayName, description,
 		isDisplayKind, icon, iconColor).Scan(&schemaNodeKind); result.Error != nil {


cmd/api/src/api/v2/azure.go
```diff
@@ -427,7 +427,7 @@ func (s *Resources) GetAZEntity(response http.ResponseWriter, request *http.Requ
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, fmt.Sprintf("query parameter %s is required", objectIDQueryParameterName), request), response)
 	} else if azKind, err := azEntityParamToKind(entityType); err != nil {
 		slog.WarnContext(request.Context(), "Could not determine AZ type from entityType request var",
-			slog.String("entityType", entityType),
+			slog.String("entity_type", entityType),
```
sirisjo, line 430: **note**: this is an unrelated PFC change.

cmd/api/src/api/v2/customnode.go
```diff
@@ -33,8 +32,6 @@ const (
 	CustomNodeKindParameter = "kind_name"
 )
 
-var validColorString = regexp.MustCompile("^#([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$")
```
sirisjo, line 36: **note**: this was moved to the database layer so that it can be used in validating color string during opengraph ingest as well. I don't really like putting it in the database layer but wasn't sure where else it could go.

packages/javascript/bh-shared-ui/src/views/UserProfile/UserProfile.test.tsx
sirisjo, line 1: **note**: This is an unrelated PFC change

packages/javascript/bh-shared-ui/src/views/Users/UserActionsMenu.test.tsx
sirisjo, line 1: **note**: this is an unrelated PFC change

cmd/api/src/api/v2/customnode.go
LawsonWillard: Moving it to the model layer will work for the time being. It should probably live in a service layer somewhere but we'll need to wait until we are ready to add that for this endpoint.

cmd/api/src/database/upsert_schema_extension.go
```diff
@@ -169,3 +209,31 @@ func (s *BloodhoundDB) upsertFindingsAndRemediations(ctx context.Context, extens
...
+	}
+	return existingIconMap, nil
+}
+
+// parseIconDefinitionFromNodeKinds creates a CustomNodeKind that will be used in either Creating or Updating a custom node kind in the custom_node_kinds tables.
+func parseIconDefinitionFromNodeKind(nodeKind model.GraphSchemaNodeKind) (model.CustomNodeKind, error) {
+	var customNodeKind model.CustomNodeKind
+	if !nodeKind.IsDisplayKind {
+		return customNodeKind, fmt.Errorf("kind %s is not a display kind", nodeKind.Name)
+	} else if nodeKind.Icon == "" || nodeKind.IconColor == "" || !ValidColorStringRegex.MatchString(nodeKind.IconColor) {
```
LawsonWillard: I'm thinking `!ValidColorStringRegex.MatchString(nodeKind.IconColor)` should be moved to the validation func in the service layer, return a 400 if icon_color is not empty and doesnt pass the regex check.

sirisjo: The ticket requirements state that `If icon and color field values are NOT provided, do nothing`, which I interpreted as "don't fail the request if validation fails" -- but let me double check this requirement

sirisjo: Ok I got clarification on this, you're right! We should validate the icon colors in the service layer and return an error if the color is not a valid hex code 👍

cmd/api/src/database/customnode.go
```diff
@@ -88,13 +88,21 @@ func (s *BloodhoundDB) UpdateCustomNodeKind(ctx context.Context, customNodeKind
 	)
 
 	err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
-		result := tx.Raw(fmt.Sprintf("UPDATE %s SET config = ?, updated_at = NOW() WHERE kind_name = ? RETURNING id;", customNodeKindTable), customNodeKind.Config, customNodeKind.KindName).
-			Scan(&customNodeKind.ID)
-		if result.RowsAffected == 0 {
+		bhdb := NewBloodhoundDB(tx, s.idResolver)
```
kpowderly, line 91: why create a new bloodhound db here instead of just referencing the UpdateGraphSchemaNodeKindIconById with the receiver like `s.UpdateGraphSchemaNodeKindIconById`

sirisjo: I believe we have to do it this way to keep it inside of the same transaction.

cmd/api/src/model/customnode.go
```diff
@@ -77,3 +81,16 @@ func (s *CustomNodeKindConfig) Scan(value interface{}) error {
 func (s CustomNodeKindConfig) Value() (driver.Value, error) {
 	return json.Marshal(s)
 }
+
+func (s CustomNodeKindConfig) Validate() error {
+	if s.Icon.Type != "font-awesome" {
```
kpowderly: not PR blocking but wouldn't hurt to declare this as a constant as well

cmd/api/src/model/graphschema.go
```diff
@@ -97,7 +97,7 @@ type GraphSchemaNodeKinds []GraphSchemaNodeKind
 // GraphSchemaNodeKind - represents a node kind for an extension
 type GraphSchemaNodeKind struct {
 	Serial
-
+	KindId            int // indicates the DAWGS kind ID for this schema kind
```
kpowderly: I don't see this being used anywhere. These public models shouldn't need it?

sirisjo: Good catch, this must have been left over from a previous iteration 👍

cmd/api/src/database/graphschema.go
```diff
@@ -305,7 +306,7 @@ func (s *BloodhoundDB) CreateGraphSchemaNodeKind(ctx context.Context, name strin
 		FROM dawgs_kind dk
 		RETURNING id, kind_id, schema_extension_id, display_name, description, is_display_kind, icon, icon_color, created_at, updated_at, deleted_at
 	)
-	SELECT isn.id, isn.schema_extension_id, dk.name, isn.display_name, isn.description, isn.is_display_kind, isn.icon, isn.icon_color, isn.created_at, isn.updated_at, isn.deleted_at
+	SELECT isn.id, isn.kind_id, isn.schema_extension_id, dk.name, isn.display_name, isn.description, isn.is_display_kind, isn.icon, isn.icon_color, isn.created_at, isn.updated_at, isn.deleted_at
```
LawsonWillard: I think this is left over from a previous iteration and can be removed

sirisjo: yes, good catch!

PR #2651, opened 2026-04-14
chore: exclude mockServiceWorker.js from production builds BED-7993

elikmiller:
## Description

Excludes `mockServiceWorker.js` from production builds while keeping it available for development mode.

MSW (Mock Service Worker) requires the service worker file to be served from the public directory to function properly in development. However, this file should not be included in production builds as it's only needed for local mocking during development.

This PR adds a custom Vite plugin that automatically removes `mockServiceWorker.js` from the build output after Vite copies public assets to the dist folder.

## Motivation and Context

Resolves BED-7993

The mockServiceWorker.js file was being included in production builds, adding unnecessary development-only code to the production bundle. This change ensures that MSW artifacts are excluded from production while maintaining full functionality in development mode.

## How Has This Been Tested?

- Built the BHCE UI with `yarn build`
- Verified build output and dist folder contents (to ensure mockServiceWorker.js is not present)

## Screenshots (optional):

N/A

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Chores**
  * Remove mock service worker from production builds to streamline output and avoid shipping test artifacts.
  * Ensure build reliably locates the output directory for consistent deployment artifacts.
  * Add a build-only step that detects and removes unwanted mock assets during production packaging.

### DIFF
--- a/cmd/ui/vite.config.ts
+++ b/cmd/ui/vite.config.ts
@@ -16,8 +16,9 @@
 
 /// <reference types="vitest" />
 import react from '@vitejs/plugin-react';
+import fs from 'fs';
 import path from 'path';
-import { defineConfig, loadEnv, searchForWorkspaceRoot } from 'vite';
+import { defineConfig, loadEnv, Plugin, searchForWorkspaceRoot } from 'vite';
 import glsl from 'vite-plugin-glsl';
 
 // https://vitejs.dev/config/
@@ -25,7 +26,7 @@ export default defineConfig(({ mode }) => {
     const env = loadEnv(mode, process.cwd(), '');
 
     return {
-        plugins: [react(), glsl()],
+        plugins: [react(), glsl(), excludeMockServiceWorker()],
         resolve: {
             alias: {
                 src: path.resolve(__dirname, './src'),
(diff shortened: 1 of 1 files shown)


cmd/ui/vite.config.ts
```diff
@@ -110,3 +111,23 @@ export default defineConfig(({ mode }) => {
...
+ * Exclude mockServiceWorker.js from production builds
+ * MSW service worker should only be available in development mode
+ */
+function excludeMockServiceWorker(): Plugin {
+    return {
+        name: 'exclude-mock-service-worker',
+        apply: 'build', // Only apply during build (production)
+        closeBundle() {
+            // Remove mockServiceWorker.js from the output directory after build
+            const outDir = process.env.BUILD_PATH || './dist';
```
utafrali: The plugin reads `process.env.BUILD_PATH` to determine the output directory, but the rest of the config resolves it via Vite's `loadEnv` (line 26) which also reads `.env` files. If `BUILD_PATH` is defined in a `.env` file rather than as a system environment variable, `process.env.BUILD_PATH` will be `undefined` here and the plugin will silently look in `./dist` while the actual build output went somewhere else — meaning `mockServiceWorker.js` would not be deleted.

The idiomatic fix is to read the resolved config via the `configResolved` hook, which is always consistent with what Vite itself uses:

```ts
function excludeMockServiceWorker(): Plugin {
    let outDir: string;
    return {
        name: 'exclude-mock-service-worker',
        apply: 'build',
        configResolved(config) {
            outDir = config.build.outDir;
        },
        closeBundle() {
            const mockServiceWorkerPath = path.resolve(__dirname, outDir, 'mockServiceWorker.js');
            if (fs.existsSync(mockServiceWorkerPath)) {
                fs.unlinkSync(mockServiceWorkerPath);
            }
        },
    };
}
```

This removes the duplicated fallback logic and keeps the plugin in sync with the Vite config automatically.

utafrali (approved):
The custom Vite plugin is a clean, well-scoped solution and the `closeBundle` + `apply: 'build'` combination is the right approach. There is one real bug: the plugin resolves the output directory from `process.env.BUILD_PATH` while the build config uses Vite's `loadEnv`, so any `BUILD_PATH` set in a `.env` file would be invisible to the plugin, causing a silent no-op deletion.

zinic (approved):
Upon testing this drops the file on disk but when I run the service I can no longer request it from the API. This appears to be working as intended.

PR #2655, opened 2026-04-15
chore: added PDF conditional BED-7359

LucasParraF:
## Description

Added a isPDF conditional to the DashboardCard component so styling can be applied independently to DashboardCardPDF without affecting other instances.

## Motivation and Context

Resolves BED-7359

Tested locally

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Style**
  * Enhanced Badge component structure to enable improved CSS customization of icon and label elements through dedicated styling classes.


packages/javascript/doodle-ui/src/components/Badge/Badge.tsx
```diff
@@ -38,8 +39,12 @@ const Badge = React.forwardRef<HTMLDivElement, BadgeProps>(
...
                     backgroundColor: backgroundColor,
                 }}>
-                {icon && <span style={{ color }}>{icon}</span>}
-                {label}
+                {icon && (
+                    <span className={isPdf ? 'badge-icon' : ''} style={{ color }}>
+                        {icon}
+                    </span>
+                )}
+                <span className={isPdf ? 'badge-label' : ''}>{label}</span>
```
dcairnsspecterops: I don't think we need to exclude this css class in the non-pdf case. The classes themselves seem harmless to me, since they are unique enough and unlikely to cause collisions and because we do not actually use traditional css files in our app, so a conflict is in any case very unlikely.

However, if we do want to exclude them, I would want to call the prop 'includeCssClasses', rather than reference the PDF, since the PDF detail refers to a specific use case as a consumer, when it's possible consumers may have a non-pdf related reason to include these class names..

dcairnsspecterops (changes requested):

dcairnsspecterops (approved):
Looks good to me!

PR #2663, opened 2026-04-16
fix: added more ciphers BED-8029

emmanuelrobles:
## Description

Added more ciphers, all the ciphers added are TLS 1.2 and 1.3 from [here](https://developers.cloudflare.com/ssl/edge-certificates/additional-options/cipher-suites/supported-cipher-suites/)

## Motivation and Context

Resolves BED-8029

## How Has This Been Tested?

ran the api locally

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs


cmd/api/src/daemons/api/toolapi/api.go
```diff
@@ -118,6 +118,13 @@ func NewDaemon[DBType database.Database](ctx context.Context, connections bootst
 					tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 					tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
 					tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+					tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+					tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_AES_128_GCM_SHA256,
```
mykeelium, line 125: This is not my subject matter, but the tls1.3 configurations here do not work, or at least are ignored. I am not sure if we should add a comment or something about that here to maybe save us some headaches in the future.

https://github.com/golang/go/blob/e990d4f8e43447b60c0ab4931d818e81aae26398/src/crypto/tls/common.go#L716

https://pkg.go.dev/crypto/tls#Config

cmd/api/src/daemons/api/bhapi/api.go
```diff
@@ -49,6 +49,13 @@ func NewDaemon(cfg config.Configuration, handler http.Handler) Daemon {
 					tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 					tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
 					tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+					tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,
+					tls.TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,
+					tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+					tls.TLS_AES_128_GCM_SHA256,
```
lrfalslev, line 56: I think these last 3 TLS 1.3 suites might not be needed. As they should be implemented by default. It sounds like go doesn't allow any configuration of TLS 1.3 https://github.com/golang/go/issues/54072#issuecomment-2873789387

Image

cmd/api/src/daemons/api/toolapi/api.go
```diff
@@ -118,6 +118,13 @@ func NewDaemon[DBType database.Database](ctx context.Context, connections bootst
 					tls.TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,
 					tls.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
```
mykeelium, line 119: question: I know that this is not the scope of this PR, but it seems that this CipherSuite is considered insecure from to Go Team's standards. Is there a reason why we are keeping it around?

https://github.com/golang/go/blob/e990d4f8e43447b60c0ab4931d818e81aae26398/src/crypto/tls/cipher_suites.go#L94

mykeelium (approved):
Few questions listed above, ultimately not a hard block for the hotfix, but I do think some documentation in the code might be helpful for next time.

PR #2686, opened 2026-04-21
feat: add GET /display-kinds to enrich graph UI BED-7925

mistahj67:
## Description

Extends `/kinds` with array for `node_kinds` and `edge_kinds`
Adds `/display-kinds` with `node_kinds` and `edge_kinds` map to provide icon / display name information to be used by the UI

## Motivation and Context

Resolves BED-7925

## How Has This Been Tested?

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

### DIFF
--- a/cmd/api/src/api/v2/kinds.go
+++ b/cmd/api/src/api/v2/kinds.go
@@ -23,25 +23,115 @@ import (
 
 	"github.com/specterops/bloodhound/cmd/api/src/api"
 	"github.com/specterops/bloodhound/cmd/api/src/model"
+	"github.com/specterops/bloodhound/packages/go/graphschema"
+	"github.com/specterops/bloodhound/packages/go/graphschema/common"
 	"github.com/specterops/dawgs/graph"
 )
 
+func isExtendedNodeKind(kind graph.Kind) bool {
+	return strings.HasPrefix(kind.String(), model.AssetGroupTagKindPrefix) || kind.Is(graph.StringKind("Meta"), graph.StringKind("MetaDetails"), common.MigrationData)
+}
+
 type ListKindsResponse struct {
-	Kinds graph.Kinds `json:"kinds"`
+	Kinds     graph.Kinds `json:"kinds"`
+	NodeKinds graph.Kinds `json:"node_kinds"`
+	EdgeKinds graph.Kinds `json:"edge_kinds"`
 }
 
 // ListKinds returns all node kinds, edge kinds, and tier tags present in the system.
 // It is a comprehensive view of the various kinds the graph currently recognizes.
 func (s Resources) ListKinds(response http.ResponseWriter, request *http.Request) {
+	var (
+		resp = ListKindsResponse{}
+		err  error
+	)
+
+	if resp.Kinds, err = s.Graph.FetchKinds(request.Context()); err != nil {
+		api.HandleDatabaseError(request, response, err)
+	} else {
+		// Alpha sort kind list
+		slices.SortFunc(resp.Kinds, func(a, b graph.Kind) int {
+			return strings.Compare(a.String(), b.String())
+		})
+
+		// This gets both custom node kinds (schemaless) and schema based node kinds
+		// Note: currently this is not an exhaustive list until all the custom-node sync work is complete
+		if displayKinds, err := s.DB.GetPrimaryDisplayKinds(request.Context()); err != nil {
+			api.HandleDatabaseError(request, response, err)
+		} else {
+			// Add nodes and edge kinds for UI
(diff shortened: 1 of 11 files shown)


cmd/api/src/api/v2/kinds.go
```diff
@@ -23,25 +23,115 @@ import (
...
-	Kinds graph.Kinds `json:"kinds"`
+	Kinds     graph.Kinds `json:"kinds"`
+	NodeKinds graph.Kinds `json:"node_kinds"`
+	EdgeKinds graph.Kinds `json:"edge_kinds"`
 }
 
 // ListKinds returns all node kinds, edge kinds, and tier tags present in the system.
 // It is a comprehensive view of the various kinds the graph currently recognizes.
 func (s Resources) ListKinds(response http.ResponseWriter, request *http.Request) {
+	var (
```
kpowderly, line 44: Sorry (not sorry) but these endpoints need tests...

cmd/api/src/api/v2/kinds.go
```diff
@@ -23,25 +23,115 @@ import (
...
+		resp = ListKindsResponse{}
+		err  error
+	)
+
+	if resp.Kinds, err = s.Graph.FetchKinds(request.Context()); err != nil {
```
kpowderly, line 49: Can we declare a var for `request.Context()` since we're using it in more than one place?

sirisjo, line 49: **thought** (non-blocking): it might be nice to put some of the duplicated logic here into a shared helper function

cmd/api/src/api/v2/kinds.go
```diff
@@ -23,25 +23,115 @@ import (
...
+		api.HandleDatabaseError(request, response, err)
+	} else {
+		// Alpha sort kind list
+		slices.SortFunc(resp.Kinds, func(a, b graph.Kind) int {
+			return strings.Compare(a.String(), b.String())
+		})
+
+		// This gets both custom node kinds (schemaless) and schema based node kinds
+		// Note: currently this is not an exhaustive list until all the custom-node sync work is complete
```
kpowderly, line 58: I'm slightly hesitant to add a comment like this unless there's a jira addressing to remove said dead comment once it's complete.

I do think it's a good call out but maybe we should address the effect vs the cause? `Only kinds that have been synced to the display kinds table will be included in this response.`

cmd/api/src/api/v2/kinds.go
```diff
@@ -23,25 +23,115 @@ import (
...
-		// Alpha sort
+		// Alpha sort kind list
 		slices.SortFunc(kinds, func(a, b graph.Kind) int {
 			return strings.Compare(a.String(), b.String())
 		})
 
-		api.WriteBasicResponse(request.Context(), ListKindsResponse{Kinds: kinds}, http.StatusOK, response)
+		// This gets both custom node kinds (schemaless) and schema based node kinds
+		// Note: currently this is not an exhaustive list until all the custom-node sync work is complete
+		if displayKinds, err := s.DB.GetPrimaryDisplayKinds(request.Context()); err != nil {
```
kpowderly, line 98: same dealio here with making `request.Context` a var since we're using it multiple times

packages/go/graphschema/primarykind.go
```diff
@@ -71,16 +71,19 @@ type DisplayNodeIcon struct {
 }
 
 type DisplayKind struct {
-	Name string
-	Icon DisplayNodeIcon
+	Name                 string           `json:"name"`
+	DisplayName          string           `json:"display_name"`
```
kpowderly, line 75: If DisplayName is only meaningful for Extensions Nodes/Edges, why utilize this field at all if Name / Display Name will always be the same? Can the UI just fallback?

packages/go/openapi/src/paths/graph.kinds.yaml
kpowderly, line 22: I am actually so annoying, but can we update this to maybe say something like "Lists all node kinds, edge kinds, and tier tags" Sorrrry.

I think we're in a bit of a bad habit of just making the summary and the description the same thing for our api spec.

cmd/api/src/api/v2/kinds.go
mistahj67: they result in different shapes so I wasn't sure, but I can take a crack at it

sirisjo: I think even just de-duping the calls to `FetchKinds`, sort, and then `GetPrimaryDisplayKinds` would help improve readability / maintain cleanliness.

mistahj67:
Going a different direction. Closing this now

PR #2732, opened 2026-05-04
chore: automatically bundle latest collector versions in docker image BED-7545

lrfalslev:
## Description

BHCE should automatically fetch the latest collector versions to bundle into image.
Just recipes and cd workflow call github /releases/latest endpoints to fetch latest version, and use resolved version to fetch and bundle latest collector artifacts.

## Motivation and Context

Resolves [BED-7545](https://specterops.atlassian.net/browse/BED-7545)

This removes the need for manual collector version bumps in dockerfile during release week. Once collectors are released the latest versions will be pulled automatically on next run.

## How Has This Been Tested?

[test workflow](https://github.com/SpecterOps/BloodHound/actions/runs/25679069683) ran with automatic collector resolution.
just recipes resolve latest versions and docker commands succeed.

## Screenshots (optional):
image
Screenshot_2026-05-11_12-35-35

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-7545]: https://specterops.atlassian.net/browse/BED-7545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Chores**
  * Docker build configuration now requires explicit collector version parameters to ensure reproducible builds across environments.
  * Added checksum verification for all downloaded collector artifacts to validate integrity during the build process.
  * Optimized collector asset packaging for improved build consistency and maintainability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SpecterOps/BloodHound/pull/2732?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


.github/workflows/cd.yml
mykeelium, line 1: note (non-blocking): This work is great and it will definitely save us a lot of time in releases. However, it may be beneficial to have optional arguments on this workflow so we can specify the version of SharpHound and AzureHound, that if they are passed the resolve does not run. It might be beneficial to clarify with James if we have ever needed a build with an older version of the collectors, but if we do, something like that would save us from having to make larger code changes.

dockerfiles/bloodhound.Dockerfile
mykeelium, line 1: nit: While I do like the fix here for removing this variable from having to update it all the time, and I personally primarily use just, if anyone uses the docker engine directly or compose through docker-compose they will run into issues. I know the naive way is just to move this same problem up, but it may be worthwhile to ensure users who are reliant on the docker way of doing things do not need to go through just.

justfile
mykeelium, line 1: nit: we may want to rethink the structure of the justfile here. As I understand it, and correct me if I am wrong. This makes it so either the environment variables SHARPHOUND_VERSION and AZUREHOUND_VERSION must be set, or internet is required to resolve the latest version prior to building the images, which could be cached, to many common workflows. This is partly dependent on the reliance just wrapper here.

justfile
```diff
@@ -278,6 +299,50 @@ init wipe="":
...
+  fi
+  docker compose --profile "{{profile}}" -f docker-compose.dev.yml {{ARGS}}
+
+# Returns true if args may build images (e.g. matches "build", "up", or "--build") false otherwise
+_needs-collector-versions *ARGS:
+  #!/usr/bin/env bash
+  set -euo pipefail
+  args="$*"
+  case " $args " in
+    *" build "*|*" --build "*|*" up "*)
```
mykeelium: nit: this might need to be adjusted. With the current setup `just bh-dev up --no-build` will still build.

dockerfiles/bloodhound.Dockerfile
```diff
@@ -19,43 +19,63 @@
...
+FROM version-validator AS hound-builder
+ARG SHARPHOUND_VERSION
+ARG AZUREHOUND_VERSION
 RUN apk --no-cache add p7zip
-RUN mkdir -p /tmp/sharphound /tmp/azurehound
-
-ADD https://github.com/SpecterOps/SharpHound/releases/download/${SHARPHOUND_VERSION}/SharpHound_${SHARPHOUND_VERSION}_windows_x86.zip /tmp/sharphound/sharphound-${SHARPHOUND_VERSION}.zip
-ADD https://github.com/SpecterOps/SharpHound/releases/download/${SHARPHOUND_VERSION}/SharpHound_${SHARPHOUND_VERSION}_windows_x86.zip.sha256 /tmp/sharphound/sharphound-${SHARPHOUND_VERSION}.zip.sha256
-
-ADD https://github.com/SpecterOps/AzureHound/releases/download/${AZUREHOUND_VERSION}/AzureHound_${AZUREHOUND_VERSION}_darwin_amd64.zip \
```
mykeelium, line 38: question: I see a move away from ADD to RUN here. Is there a reasoning for this? Since these are releases, the content shouldn't necessarily change for the same tag, but I am curious why the change since we would lose that checking.

lrfalslev: This is just a holdover from an earlier iteration where versions were written to a .env and needed copied for each command so I consolidated all the downloads under one command to avoid copying the env for each one. I'll move them back to ADD

dockerfiles/bloodhound.Dockerfile
lrfalslev: yeah building the images outside of just or the pipeline requires manually supplying collector version build args. 

Earlier I had the dockerfiles resolve the collector versions, but was calling the api in a RUN and had issues with invalidating the cached layers. But ADD should result in a cache miss when the checksum changes so I can try keeping all collector version work in the dockerfile.

mykeelium: If you run into issues, we can get with the api team on other possible solutions. Please feel free to reach out.

.github/workflows/cd.yml
lrfalslev: workflow is unchanged now. If needed you can supply build args to specify older collector version(s). `SHARPHOUND_VERSION=v2.11.0 AZUREHOUND_VERSION=v2.11.0 just bh-dev build --no-cache`

dockerfiles/bloodhound.Dockerfile
lrfalslev: Fetch versions from releases/latest github api with ADD which invalidates the cache on checksum change, but since the api returns metadata (like download_count) it changes more often than we need. 

By splitting api calls into independent stages we can copy version files that only contain the collector versions so that if the actual tag of the release changed it rebuilds the collector packaging layers, but if the version stays the same we can reuse the cached layer.

dockerfiles/bloodhound.Dockerfile
lrfalslev: Using wget inside of RUN so we can reference the collector-versions output by the version-validator stage. This should be fine since our release artifacts should be immutable, and new release versions will force the stage to be re-evaluated with the new urls.

PR #2754, opened 2026-05-07
fix: reset File Ingest Log pagination when rows per page changes BED-8109

elikmiller:
## Description
`FileIngestTable` in `bh-shared-ui` tracks `page` independently from `rowsPerPage`. When the user paginates to a non-zero page and then changes Rows per page, the API call keeps the stale `skip = rowsPerPage * page` value. If the new (`skip`, `limit`) window falls past the end of the result set, the table renders empty even though matching jobs exist.

## Motivation and Context
Resolves BED-8109

## How Has This Been Tested?
- Added two integration tests under a new `pagination reset` describe block in
  `FileIngestTable.test.tsx`

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Pagination now resets to the first page when rows-per-page is changed.
  * Applying a status filter resets pagination to the first page and correctly applies the filter.

* **Tests**
  * Added tests validating pagination reset when filters are applied and when page size changes.
  * Tests simulate server responses and verify the table displays the expected number of rows after interactions.


urangel (approved):
lgtm 🚀

PR #2786, opened 2026-05-18
fix(BL): Saved Queries Uploads Not Using Ingest File Checks BED-7040

ALCooper12:
## Description
- Updated helper functions inside `streamdecoder.go`, `opengraphschema.go`, and `saved_queries.go` in order to utilize BOM normalization when uploading individual or zip files containing JSON with UTF-8 BOM
- Added 2 more unit test cases for the ImportSavedQueries endpoint

## Motivation and Context
This PR addresses:
- BED-7040

- At the moment, whenever someone tries to upload a custom query stored in a JSON file with UTF-8 BOM or a ZIP file containing JSON with UTF-8 BOM an error message is displayed. The changes introduced add another form of file checking/validation that utilizes BOM normalization so that uploading individual or zip files containing JSON with UTF-8 BOM are accepted with no error messages displayed 

## How Has This Been Tested?
- Added more unit tests and tested via Bruno

## Screenshots (optional):
- `PUT /api/v2/extensions`  
[dog-park-schema_utf8bom.json](https://github.com/user-attachments/files/28977050/dog-park-schema_utf8bom.json)

Screenshot 2026-06-15 at 1 18 22 PM

Now confirm with  `GET /api/v2/extensions`

Screenshot 2026-06-15 at 1 20 39 PM

- `POST /api/v2/saved-queries/import`
[return_nodes_utf8bom.json](https://github.com/user-attachments/files/28977077/return_nodes_utf8bom.json)

Screenshot 2026-06-15 at 2 32 50 PM

- `POST /api/v2/file-upload/{id}`

In order to test the endpoint above, you have to hit its other 2 connected endpoints. And so you will hit them in this order: file-upload/start, file-upload{job-id}, then file-upload/end
Screenshot 2026-06-15 at 2 04 01 PM

[dog_park_utf8bom.json](https://github.com/user-attachments/files/28977098/dog_park_utf8bom.json)

Screenshot 2026-06-15 at 2 16 45 PM

Keep in mind that this header is also REQUIRED above: `Content-Type   application/json`

And finally
Screenshot 2026-06-15 at 2 19 21 PM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Release Notes

* **Bug Fixes**
  * Saved query imports now properly handle JSON files with UTF-8 BOM encoding, including files within ZIP archives.
  * OpenGraph schema ingestion improved for files with UTF-8 BOM.
  * Enhanced error messages to better distinguish between encoding normalization and parsing failures during import.


LawsonWillard (approved):
Saved queries changes look good to me! Giving it a ✅ but may be worth waiting for someone else whose closer to the requirements to give it another approval before merging.

cmd/api/src/config/config.go
```diff
@@ -280,11 +283,16 @@ func WriteConfigurationFile(path string, config Configuration) error {
 	return nil
 }
 
+// check to see if BOM normalization is needed for config files generated for the user
 func ParseConfiguration(content []byte) (Configuration, error) {
```
brandonshearin, line 294: I didn't see this code path called out in the acceptance criteria. 

What feature does this code path support? are we sure we need/want to add bom logic to this branch? it may be unneccessary.

brandonshearin (approved):
I didn't know what BOM encoding was, but after some quick research this morning, it seems like adding a BOM to a JSON file is an outdated practice, and I'm not sure what has gone into the decision to support it on our end. It seems that just rejecting a JSON file with a BOM would be a defensible choice for us, and maybe even preferred, because it would be less to maintain on our side.

However, from a code perspective, this looks good -- and the tests you added to `saved_queries_test.go` are excellent.

superlinkx:
> I didn't know what BOM encoding was, but after some quick research this morning, it seems like adding a BOM to a JSON file is an outdated practice, and I'm not sure what has gone into the decision to support it on our end. It seems that just rejecting a JSON file with a BOM would be a defensible choice for us, and maybe even preferred, because it would be less to maintain on our side.
> 
> However, from a code perspective, this looks good -- and the tests you added to `saved_queries_test.go` are excellent.

Technically this is correct, and is also why the Go stdlib doesn't support BOM encoding on JSON by default. However, there are several JSON implementations (looking at you Python) that use BOM encoding by default. As a result, we need to support it or we'll eventually get bug reports because inevitably someone will build an integration that uses BOM and their users will get frustrated.

While supporting BOM isn't my favorite, we've run into this problem twice over the course of this application's lifespan so it's best to keep supporting it anytime we accept arbitrary json bundles. This also lines up with the philosophy of "be conservative with what you emit, be liberal with what you accept" that shaped the internet itself.

superlinkx (approved):

cmd/api/src/config/config.go
superlinkx: I misunderstood the concern around BOM here. As far as I'm aware, we don't have any current issues with BOM in user config.json files, so I agree we should drop it from this changeset as I don't see a clear problem this will solve.

PR #2801, opened 2026-05-20
feat: adding Storage, FileService, and FileServiceResolver - BED-8314

mykeelium:
## Description

This change sets up and builds out the implementation of a centralized file service and storage mechanism that can be used in future updated. 

## Motivation and Context

Resolves BED-8314

## How Has This Been Tested?

This change does not add any functionality to current workflows. Tests were added and previous tests pass. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added a sandboxed local filesystem storage backend with atomic writes, safe directory handling, listing, copy/move semantics, and content-type detection.
  * Added backend-agnostic file service abstractions and a resolver for managing named file services.

* **Bug Fixes**
  * Validate upload validation callback presence and return an error if missing.

* **Tests**
  * Added extensive tests and mocks covering storage, file services, path-safety, cancellation, and error cases.

* **Chores**
  * Bumped AWS SDK-related dependency patch versions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/SpecterOps/BloodHound/pull/2801?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cmd/api/src/services/storage/storage.go
```diff
@@ -0,0 +1,314 @@
...
+	List(ctx context.Context, name string, options ListOptions) ([]FileInfo, error)
+
+	// Copy duplicates an object.
+	Copy(ctx context.Context, srcName, dstName string, options WriteOptions) error
+
+	// Move moves an object. Is done by a copy and a delete.
+	Move(ctx context.Context, srcName, dstName string, options WriteOptions) error
+}
+
+// FileService serves as an abstraction to hanlde files with different storage backends. This functions
```
wes-mil: hanlde

wes-mil (approved):

cmd/api/src/services/storage/storage.go
mykeelium: corrected

cmd/api/src/services/graphify/tasks.go
```diff
@@ -49,122 +44,11 @@ func (s *GraphifyService) clearFileTask(ingestTask model.IngestTask) {
...
-	}
-}
-
 // ProcessIngestFile reads the files at the path supplied, and returns the total number of files in the
 // archive, the number of files that failed to ingest as JSON, and an error
 func (s *GraphifyService) ProcessIngestFile(ic *IngestContext, task model.IngestTask) ([]IngestFileData, error) {
 	// Try to pre-process the file. If any of them fail, stop processing and return the error
-	if fileData, err := s.extractIngestFiles(task.StoredFileName, task.OriginalFileName, task.FileType); err != nil {
+	if fileData, err := ExtractIngestFiles(s.ctx, s.cfg, task.StoredFileName, task.OriginalFileName, task.FileType, "bh"); err != nil {
 		return []IngestFileData{}, err
```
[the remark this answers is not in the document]
mykeelium: not in this changeset. plus was existing functionality on a refactor PR. If desired I can add a change for this.

cmd/api/src/services/storage/file_service_resolver.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+func NewFileServiceResolver(services map[storage.FileServiceName]storage.FileService) (FileServiceResolver, error) {
+	var (
+		serviceName    storage.FileServiceName
+		fileService    storage.FileService
+		copiedServices = make(map[storage.FileServiceName]storage.FileService, len(services))
+	)
+
+	for serviceName, fileService = range services {
```
mistahj67: scoping the vars for loop only is probably preferable here
```suggestion
	for serviceName, fileService := range services {
```

cmd/api/src/services/storage/file_service_resolver.go
```diff
@@ -0,0 +1,135 @@
...
+	// Resolve returns a filestorage.FileService interface if a filestorage.FileService is found with the given name.
+	// Otherwise, an error is returned.
+	Resolve(name storage.FileServiceName) (storage.FileService, error)
+}
+
+type fileServiceResolver struct {
+	services map[storage.FileServiceName]storage.FileService
+}
+
+func NewFileServiceResolver(services map[storage.FileServiceName]storage.FileService) (FileServiceResolver, error) {
```
mistahj67: nit: consider creating a formal type for `map[storage.FileServiceName]storage.FileService` as it is used quite extensively

cmd/api/src/services/storage/fileserviceresolver.go
```diff
@@ -0,0 +1,135 @@
...
+
+	closeOpened := func() {
+		for _, store := range openedStores {
+			_ = store.Close()
+		}
+	}
+
+	workStore, err := storage.NewLocalStore(cfg.WorkDir)
+	if err != nil {
+		return nil, err
```
mistahj67: while not "needed" I'd probably include the closeOpened here to guard against future changes, it'll just no-op while the list is empty but if any are added above, this won't be missed

cmd/api/src/services/storage/file_service_resolver.go
```diff
@@ -0,0 +1,135 @@
...
+	if !found {
+		return nil, fmt.Errorf("%w: %s", storage.ErrFileServiceNotFound, name)
+	}
+
+	return fileService, nil
+}
+
+// NewDefaultFileServices creates the file services that should be considered default with
+// BloodHound. Additional FileServices can still be created prior to creating a Resolver.
+func NewDefaultFileServices(cfg config.Configuration) (map[storage.FileServiceName]storage.FileService, error) {
```
mistahj67: could use some cleanup I think on this function, what that looks like i'm not entirely sure but I trust you

packages/go/storage/localstore.go
```diff
@@ -0,0 +1,442 @@
...
+	"sync"
+)
+
+var ErrIsDirectory = errors.New("is a directory")
+
+// ctxReader wraps an io.Reader so that context cancellation is observed between
+// reads. io.Copy calls Read in a loop and each call checks ctx.Err() before delegating.
+type ctxReader struct {
+	ctx context.Context
+	r   io.Reader
```
mistahj67: nit: lets use a variable here to be descriptive please

packages/go/storage/localstore.go
```diff
@@ -0,0 +1,442 @@
...
+}
+
+func (s *ctxReader) Read(p []byte) (int, error) {
```
mistahj67: nit: please swap p with descriptive variable name for maintainability

packages/go/storage/localstore.go
```diff
@@ -0,0 +1,442 @@
...
+// within the root, including through symlinks. Passing a path with a leading slash
+// will cause an error.
+type LocalStore struct {
+	root      *os.Root
+	closeOnce sync.Once
+	closeErr  error
+}
+
+func NewLocalStore(root string) (*LocalStore, error) {
+	r, err := os.OpenRoot(root)
```
mistahj67: Same feedback, please be descriptive with the variable names

packages/go/storage/localstore.go
```diff
@@ -0,0 +1,442 @@
...
+		return nil, err
+	}
+	listName := name
+	if strings.TrimSpace(listName) == "" || listName == "/" {
+		listName = "."
+	}
+	fsys := s.root.FS()
+	if options.Recursive {
+		out := []FileInfo{}
+		err := fs.WalkDir(fsys, listName, func(p string, d fs.DirEntry, err error) error {
```
mistahj67:
```suggestion
		err := fs.WalkDir(fsys, listName, func(path string, dir fs.DirEntry, err error) error {
```

packages/go/storage/s3.go
```diff
@@ -0,0 +1,510 @@
...
+		return "", errors.New("path escapes storage root")
+	}
+
+	return cleaned, nil
+}
+
+func s3DetectContentType(name string) string {
+	ext := path.Ext(name)
+	if ext == "" {
+		return "application/octet-stream"
```
mistahj67: please consider the mediatypes constant here

packages/go/storage/s3.go
```diff
@@ -0,0 +1,510 @@
...
+	name = strings.TrimSpace(name)
+	name = strings.TrimPrefix(name, "/")
+	name = strings.ReplaceAll(name, "\\", "/")
+
+	cleaned := path.Clean(name)
+	if cleaned == "." || cleaned == "" {
+		return "", errors.New("invalid empty path")
+	}
+	if cleaned == ".." || strings.HasPrefix(cleaned, "../") {
+		return "", errors.New("path escapes storage root")
```
mistahj67: Please verify that this covers escape potential accurately

packages/go/storage/storage.go
```diff
@@ -0,0 +1,222 @@
...
+	Exists(ctx context.Context, name string) (bool, error)
+
+	// List returns a list of files in a given path.
+	List(ctx context.Context, name string, options ListOptions) ([]FileInfo, error)
+
+	// Copy duplicates an object.
+	Copy(ctx context.Context, srcName, dstName string, options WriteOptions) error
+
+	// Move moves an object. Is done by a copy and a delete.
+	Move(ctx context.Context, srcName, dstName string, options WriteOptions) error
```
mistahj67, line 90: Should this consider a formal "Close" method that can be elected to no-op in storage drivers that do not need it?

mistahj67 (commented):
couple of nits, mostly looks good to me, still need to test locally!

cmd/api/src/services/storage/file_service_resolver.go
mykeelium: good catch, added.

cmd/api/src/services/storage/file_service_resolver.go
mykeelium: brilliant, added

cmd/api/src/services/storage/file_service_resolver.go
mykeelium: An update was pushed for this, please let me know if you find it more appealing. I do believe this structure will help in a later ticket when configuration comes into play on whether to use s3 vs localstore

packages/go/storage/localstore.go
mykeelium: updated to something more understandable.

packages/go/storage/localstore.go
mykeelium: updated to something more understandable.

packages/go/storage/localstore.go
mykeelium: Updated, however, I also had to change the input param, because I am not sure it was clear any more. Please let me know what you think of these updates.

packages/go/storage/localstore.go
mykeelium: I do not think this suggestion is the way to go. "path" would shadow the path package that is being imported, and dir is not descriptive, since it is an entry, it could be either a dir or a file. 

I chose to go with `entryPath` and `entry`. Please let me know if you would prefer something else.

packages/go/storage/s3.go
mykeelium: updated

packages/go/storage/s3.go
mykeelium: S3 does not have path traversal, that being said I did to a bit of hardening here because they do recommend not to have periods as a part of the key name. See here for documentation: [object key guiddlines.](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html#object-key-guidelines)

Please check the updated form and let me know if you see any issues. Thank you

packages/go/storage/storage.go
mykeelium: I am not sure how I feel about this. It is not something that is broad across implementations of the interface, particularly since s3 does not own the client. In addition, I think it could impart confusing information to people using the s3 store thinking that closing it was freeing up resources when it does not. 

The localstores should only be created in one place, and the close is a best practice in this case, and I do not think it would be a concern if it werent there for this. That being said, this is the location that technically owns the localstore creation, so it make sense for it call the close here. I am not sure a public close is the best option for this interface.

cmd/api/src/services/storage/fileserviceresolver.go
mykeelium: Due to the altered structure of NewDefaultFileServices, I do not believe this is a concern, because new services should just have to add to the table and it should be handled for them.

mistahj67 (approved):
Thanks for the patience! 🚀

mykeelium:
> Thanks for the patience! 🚀

Thank you for the effort! this was a lot. That is not lost on me <3

PR #2821, opened 2026-05-26
chore: Migrate cross-product lookups to use bitmask isntead of CommutativeDuplex - BED-8362

StephenHinck:
## Description

Migrates `CalculateCrossProductNodeSets` to utilize a bitmap instead of `CommutativeDuplex` for performance optimization. In pathological cases, this resulted in a _98% reduction_ in post-processing time for ADCS.

## Motivation and Context

Resolves BED-8362

## How Has This Been Tested?

Existing test suites are unchanged. Additionally, performed manual validation utilizing several local datasets.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Refactor**
  * Optimized cross-product computation by switching to an aggregated set representation for membership checks.
  * Replaced per-item iteration with bulk set intersections and unions, reducing iteration overhead and CPU work.
  * Improved reference-set handling and remainder merging for lower runtime and memory usage, yielding faster and more predictable set operations.

### DIFF
--- a/packages/go/analysis/ad/ad.go
+++ b/packages/go/analysis/ad/ad.go
@@ -607,10 +607,10 @@ func CalculateCrossProductNodeSets(localGroupData *LocalGroupData, nodeSlices ..
 	}
 
 	// This means that len(firstDegreeSets) must be greater than or equal to 2 i.e. we have at least two nodesets (unrolled) without Auth. Users/Everyone
-	// Materialize the check set as a single bitmap by Or-ing all component bitmaps within each set, then And-ing across sets.
+	// Materialize the check set as a single bitmap from the union of all component bitmaps (`Or`) within each set, then intersecting (`And`) across sets.
 	materializedCheckSet := cardinality.NewBitmap64()
-	for _, bm := range unrolledSets[1] {
-		materializedCheckSet.Or(bm)
+	for _, set := range unrolledSets[1] {
+		materializedCheckSet.Or(set)
 	}
 
 	for _, unrolledSet := range unrolledSets[2:] {
--- a/packages/go/analysis/ad/ad.go
+++ b/packages/go/analysis/ad/ad.go
@@ -607,9 +607,7 @@ func CalculateCrossProductNodeSets(localGroupData *LocalGroupData, nodeSlices ..
 	}
 
 	// This means that len(firstDegreeSets) must be greater than or equal to 2 i.e. we have at least two nodesets (unrolled) without Auth. Users/Everyone
-	// Materialize the check set as a single bitmap by Or-ing all component bitmaps within each set,
-	// then And-ing across sets. This replaces the lazy CommutativeDuplexes approach which required
-	// per-element binary searches and was the dominant bottleneck (86.9% of CPU in profiling).
+	// Materialize the check set as a single bitmap by Or-ing all component bitmaps within each set, then And-ing across sets.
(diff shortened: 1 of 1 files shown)


packages/go/analysis/ad/ad.go
```diff
@@ -610,15 +607,23 @@ func CalculateCrossProductNodeSets(localGroupData *LocalGroupData, nodeSlices ..
 	}
 
 	// This means that len(firstDegreeSets) must be greater than or equal to 2 i.e. we have at least two nodesets (unrolled) without Auth. Users/Everyone
-	checkSet.Or(cardinality.CommutativeOr(unrolledSets[1]...))
```
urangel, line 613: I would like to understand the motivation behind https://github.com/SpecterOps/BloodHound/pull/2364 before moving forward with this changeset. The commutative work was done fairly recently seemingly in response to observed bottlenecks that this changeset will undo. 

Perhaps we need to have the code in place for both and pivot based on the graph shape instead of always doing or the other.

urangel: After reviewing more closely I see that the contains check is where we currently have an excess of computation since we have to check every bitmap in the and slice for a commutative duplex. That combined with contains checks being done three times inside of different loops is way more processing than up front intersection that then benefits from direct contains checking with no additional looping. Nice

urangel (approved):

packages/go/analysis/ad/ad.go
```diff
@@ -610,15 +607,23 @@ func CalculateCrossProductNodeSets(localGroupData *LocalGroupData, nodeSlices ..
...
+	// Materialize the check set as a single bitmap by Or-ing all component bitmaps within each set, then And-ing across sets.
+	materializedCheckSet := cardinality.NewBitmap64()
+	for _, bm := range unrolledSets[1] {
```
urangel: nit: The changeset is good but would appreciate updating `bm` to `bitmap` or `set` in this loop and the nested loop below for a bit more clarity

packages/go/analysis/ad/ad.go
```diff
@@ -610,15 +607,23 @@ func CalculateCrossProductNodeSets(localGroupData *LocalGroupData, nodeSlices ..
 	}
 
 	// This means that len(firstDegreeSets) must be greater than or equal to 2 i.e. we have at least two nodesets (unrolled) without Auth. Users/Everyone
-	checkSet.Or(cardinality.CommutativeOr(unrolledSets[1]...))
+	// Materialize the check set as a single bitmap by Or-ing all component bitmaps within each set, then And-ing across sets.
```
urangel: nit: While Or and And should be understood given the context, the bitmap methods are named this way already so I suggest updating the comment here to spell out union and intersection so that it adds more insight than seeing the code itself. 
```suggestion
	// Materialize the check set as a single bitmap from the union of all component bitmaps (`Or`) within each set, then intersecting (`And`) across sets.
```

packages/go/analysis/ad/ad.go
StephenHinck: I moved to `set` since we had `unrolledSets` as well.

packages/go/analysis/ad/ad.go
StephenHinck: Good call-out. Fixed for clarity!

PR #2823, opened 2026-05-26
chore: Batch fetch Owns/WriteOwner - BED-8364

StephenHinck:
## Description

When processing OwnsRaw and WriteOwnerRaw relationships, the code needs to look up the target node of each relationship to check whether the owning domain enforces BlockOwnerImplicitRights. Previously, this was done by calling ops.FetchNode individually inside the loop — one database query per relationship. In environments with many ownership relationships, this created an N+1 query pattern that scaled poorly.

This change collects all the unique target node IDs upfront and fetches them in a single bulk query before entering the loop. The loop then does a simple map lookup instead of a database call for each relationship. The actual edge-creation logic and all the enforcement checks (domain SID lookup, computer-derived check, admin group membership) are unchanged — only the way target nodes are retrieved is different.

A small helper function collectUniqueEndIDs was added to deduplicate the relationship end IDs before the bulk fetch.

The same pattern is applied to both postOwnsEdges and postWriteOwnerEdges since they have identical structure.

In tested datasets, this change yielded approximately 50% reduction in post-processing duration for these edges.

## Motivation and Context

Resolves BED-8364

## How Has This Been Tested?

No changes have been made to existing test suites. Additionally, validated locally with several datasets.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Performance**
  * Improved Active Directory relationship processing by bulk-fetching and de-duplicating referenced targets, reducing repeated lookups and speeding analysis.

* **Bug Fixes / Behavior**
  * Missing referenced targets are now logged and skipped.
  * Error-handling changes may cause some relationships to be submitted in cases that were previously skipped.

### DIFF
--- a/packages/go/analysis/ad/owns.go
+++ b/packages/go/analysis/ad/owns.go
@@ -102,7 +102,7 @@ func postOwnsEdges(ctx context.Context, db graph.Database, sink *post.FilteredRe
 			return err
 		} else {
 			// When anyEnforced, batch-fetch all target nodes upfront instead of one query per relationship.
-			var targetNodes graph.NodeSet
+			var targetInfoMap map[graph.ID]ownsTargetInfo
 			if anyEnforced {
 				endIDs := collectUniqueEndIDs(relationships)
 				if fetched, err := ops.FetchNodeSet(tx.Nodes().Filterf(func() graph.Criteria {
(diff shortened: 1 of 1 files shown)


urangel (approved):
This looks to produce the same results while minimizing db hits. This is at the expense of holding the targetNodes graph.NodeSet in memory. Assuming the memory cost is manageable the stated duration improvement seems favorable.

PR #2829, opened 2026-05-27
chore: upsert any custom-node-kinds entries that are missing from kind table BED-7922

AD7ZJ:
## Description

This scans through the custom_node_kinds table and upserts to the kind table any kinds that are missing. It also gets rid of the custom_node_kind name column and makes a new kind_id column FK'ed  to the kind table. 

## Motivation and Context

Resolves https://specterops.atlassian.net/browse/BED-7922

## How Has This Been Tested?

A migration integration test has been created. 

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved custom node kind behavior so listed and displayed kinds consistently map to stored underlying kinds, avoiding missing/invalid references.
  * Updated custom node kind create/update responses and persistence to use stable kind identifiers.

* **Chores**
  * Simplified custom node kind retrieval and validation so the complete custom kind set is used.
  * Updated the custom-node-kind schema/migration flow to upsert and backfill via `kind_id`, including cleanup of reserved `Tag_` kinds.

* **Tests**
  * Updated unit/integration tests and mocks to match the new retrieval behavior and `kindId` response field, including added migration coverage.


cmd/api/src/database/migration/migrations/20260526065858_v9_upsert_kind_from_custom_node_kind.sql
Useinovski, line 1: nit: There is nothing to catch the file naming of migration files quite yet so I would kindly ask if you could change the naming convention of the this file :) Only missing the v9 prefix to the name.
timestamp_file_name.sql -> timestamp_v9_file_name.sql

cmd/api/src/database/migration/migrations/20260526065858_upsert_kind_from_custom_node_kind.sql
```diff
@@ -0,0 +1,111 @@
...
+ALTER TABLE custom_node_kinds
+    ADD CONSTRAINT custom_node_kinds_kind_name_key UNIQUE (kind_name);
+
+-- Drop the FK and unique constraints on kind_id, then drop the column.
+ALTER TABLE custom_node_kinds DROP CONSTRAINT IF EXISTS fk_custom_node_kinds_kind_id;
+ALTER TABLE custom_node_kinds DROP CONSTRAINT IF EXISTS custom_node_kinds_kind_id_key;
+ALTER TABLE custom_node_kinds DROP COLUMN kind_id;
+
+-- Finally, drop the function we updated in the Up migration, which will be re-created in the previous migration's Up function.
+DROP FUNCTION IF EXISTS ensure_stubbed_custom_node_kind_for_ingest(TEXT, JSONB);
```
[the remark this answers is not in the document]
AD7ZJ: You're right, fixed. Good :rabbit:

cmd/api/src/database/migration/migrations/20260526065858_v9_upsert_kind_from_custom_node_kind.sql
AD7ZJ: Thank you! fixed!

cmd/api/src/database/customnode.go
```diff
@@ -157,8 +184,10 @@ func (s *BloodhoundDB) DeleteCustomNodeKind(ctx context.Context, kindName string
 	)
 
 	err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
-		if err := tx.Raw(fmt.Sprintf("DELETE FROM %s WHERE kind_name = ? RETURNING id, config;", model.CustomNodeKind{}.TableName()), kindName).
-			Row().Scan(&customNodeKind.ID, &customNodeKind.Config); errors.Is(err, sql.ErrNoRows) {
+		if err := tx.Raw(
```
urangel, line 185: Since we are upserting custom node kinds to the kind table would it make sense to delete them from the kinds table here as well? We would need to check that the kind is not part of a current extension for that.

cmd/api/src/model/customnode.go
```diff
@@ -42,8 +42,9 @@ func (s CustomNodeKinds) AuditData() AuditData {
 
 type CustomNodeKind struct {
 	ID               int32                `json:"id"`
-	KindName         string               `json:"kindName"`
-	SchemaNodeKindId *int32               `json:"-"` // SchemaNodeKindId is nullable
+	KindId           int16                `json:"-"`        // FK to kind.id; stored column
+	KindName         string               `json:"kindName"` // joined from kind.name; not a stored column
```
urangel: This cleans up kind handling so much! 😌

cmd/api/src/database/migration/migrations/20260526065858_v9_upsert_kind_from_custom_node_kind.sql
```diff
@@ -0,0 +1,139 @@
...
+-- kind table. The WHERE NOT EXISTS filter avoids using up kind IDs (a SMALLSERIAL sequence) 
+-- as would happen with ON CONFLICT DO NOTHING.
+INSERT INTO kind (name)
+SELECT kind_name FROM custom_node_kinds cnk
+WHERE NOT EXISTS (
+    SELECT 1 FROM kind WHERE kind.name = cnk.kind_name
+);
+
+-- Add the new kind_id FK column (SMALLINT matches kind.id SMALLSERIAL).
+ALTER TABLE custom_node_kinds ADD COLUMN IF NOT EXISTS kind_id SMALLINT;
```
mistahj67: Due to problems in the past, this should just be an integer

cmd/api/src/model/customnode.go
```diff
@@ -42,8 +42,9 @@ func (s CustomNodeKinds) AuditData() AuditData {
 
 type CustomNodeKind struct {
 	ID               int32                `json:"id"`
-	KindName         string               `json:"kindName"`
-	SchemaNodeKindId *int32               `json:"-"` // SchemaNodeKindId is nullable
+	KindId           int16                `json:"-"`        // FK to kind.id; stored column
```
mistahj67: I suspect we can add this to the response without a breaking change

cmd/api/src/database/customnode.go
mistahj67: I actually think the opposite is true and this endpoint / handler might need to be deprecated.....

mistahj67 (changes requested):

cmd/api/src/database/migration/migrations/20260526065858_v9_upsert_kind_from_custom_node_kind.sql
AD7ZJ: Ok! Changed

cmd/api/src/model/customnode.go
AD7ZJ: Good idea!

cmd/api/src/database/customnode.go
AD7ZJ: I marked the handler as deprecated and removed the DeleteCustomNode() logic.

urangel (approved):
I am unclear whether deprecating this endpoint is a palatable direction to take so will let @mistahj67 follow up on that request. Assuming we do want to go that route, the changes look good to me!

cmd/api/src/database/customnode.go
mistahj67, line 174: I think this method can be deleted entirely now that the endpoint is going to be deprecated.

cmd/api/src/database/migration/migrations/20260526065858_v9_upsert_kind_from_custom_node_kind.sql
```diff
@@ -0,0 +1,139 @@
...
+SET kind_id = kind.id
+FROM kind
+WHERE kind.name = custom_node_kinds.kind_name;
+
+-- remove any rows that could not be resolved to a kind_id. This should never happen but we'll kill them just in case so the following steps don't fail due to nulls in kind_id. 
+DELETE FROM custom_node_kinds WHERE kind_id IS NULL;
+
+-- Enforce NOT NULL and add the FK constraint
+ALTER TABLE custom_node_kinds ALTER COLUMN kind_id SET NOT NULL;
+ALTER TABLE custom_node_kinds
```
mistahj67: If migrations need to remain idempotent, constraints need a begin wrapper

cmd/api/src/database/graphschema.go
mistahj67: Now that schemanodekinds that are display kinds are all present on custom_node_kinds, I think even this can be removed and it'll just be custom node kind query right?

cmd/api/src/database/upsert_schema_extension.go
```diff
@@ -143,7 +143,15 @@ func (s *BloodhoundDB) nodeKindReconcileConfig(extensionId int32) reconcileConfi
...
 		delete: func(ctx context.Context, existing model.GraphSchemaNodeKind) error {
-			return s.DeleteGraphSchemaNodeKind(ctx, existing.ID)
+			if err := s.DeleteGraphSchemaNodeKind(ctx, existing.ID); err != nil {
+				return err
+			}
+			// keep custom_node_kinds the source of truth for display node kinds.
+			// Node kinds that were not display kinds never had a row, so a missing row could happen and wouldn't be an error
+			if err := s.DeleteCustomNodeKind(ctx, existing.Name); err != nil && !errors.Is(err, ErrNotFound) {
+				return err
+			}
```
mistahj67: I'm wondering if these shouldn't be deleted much like the kind themselves as data may still exist despite not being in the schema anymore. 🤔

cmd/api/src/database/customnode.go
AD7ZJ: It's being used in the upsertCustomIcons() function, I think we need to keep custom_node_kinds in sync if an extension deletes one of its node kinds or changes one from a display kind to not a display kind.

cmd/api/src/database/migration/migrations/20260526065858_v9_upsert_kind_from_custom_node_kind.sql
AD7ZJ: Good call, ty!

cmd/api/src/database/graphschema.go
AD7ZJ: I think so, yes. I removed it.

cmd/api/src/database/upsert_schema_extension.go
AD7ZJ: That's a fair point, I am not real sure either. Since we are kinda using this table as the "display node kind discriminator" for the kind table, which is append only, that's a data point for not deleting them...

mistahj67 (approved):
Code looks good to me. Tested locally worked! 🚀 

Thanks for your patience and diligence attacking this nuanced task!

PR #2838, opened 2026-05-28
fix: editing saved query while another is selected - BED-7759

Flake85:
bug: editing saved query while another is selected - BED-7759

## Description
If you navigate to your Saved Queries and select one, but then you edit a separate query while the other is selected, it shows the wrong query in the edit box. Not the one you chose for editing. An unfortunate side-effect of this is that you may save over the wrong query. After fixing this issue, I found another bug where if the user edited the cypher query, selected a different query and re-selected the edited query, the original unedited query would show instead of the new query.

I changed the way editQuery works by passing the query string to it and setting the the query being edited to the selected query to fix the original issue. I changed the way the cache updates were being done by removing the api call since this seemed to be the root of the cause.

## Motivation and Context

Resolves BED-7759

Tenants could potentially edit and save over a query that they didn't intend to edit. Also, selecting a different query and re-selecting an edited query could show the original un-edited query.

## How Has This Been Tested?

I used my local environment to test this by creating at least two saved cypher queries. I selected "query a" and then without selecting "query b", clicked to edit "query b" to make an edit. After saving the edit, I then selected "query a" (make sure to confirm the query didn't change) and then re-selected "query b" to make sure the edit was saved properly. I also hard refreshed the browser to confirm both queries were what they should be. I also added a test for checking if the save query dialog renders the selected query when another query is different than the active search query.

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Save dialog now shows the selected query text when editing (no longer empty).
  * Edit action validates inputs and only proceeds when both id and query are present.

* **Improvements**
  * Faster, memoized filtering for prebuilt and saved queries for snappier UI filtering.
  * More efficient saved-queries cache updates so creates/edits/deletes reflect immediately without full refresh.

* **Tests**
  * Added test to verify the editor displays the correct query when editing.


packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/SavedQueries/CommonSearches.tsx
```diff
@@ -54,61 +54,23 @@ const CommonSearches = ({
...
-            onPerformCypherSearch(query);
-        }
-    };
+    const isFiltered = searchTerm.length > 2 || platform !== '' || categoryFilter.length > 0 || source !== '';
 
-    const handleDeleteQuery = (id: number) => {
-        setQueryId(id);
-        setOpen(true);
-    };
+    const filteredList = useMemo<QueryListSection[]>(() => {
```
jvacca-specterops, line 69: Nice!  This is a nice improvement.

packages/javascript/bh-shared-ui/src/hooks/useSavedQueries.ts
```diff
@@ -136,26 +136,38 @@ export const useSavedQueries = (
 export const useCreateSavedQuery = () => {
     const queryClient = useQueryClient();
     return useMutation(createSavedQuery, {
-        onSuccess: () => {
-            queryClient.invalidateQueries(savedQueryKeys.all);
+        onSuccess: async (createdQuery) => {
```
jvacca-specterops, line 139: This is a nice pattern.  Saves us a server call for each of these.

jvacca-specterops (approved):
Great work!

PR #2841, opened 2026-05-29
fix: Last Analysis Run At not updating on every analysis BED-8136

sirisjo:
## Description

We decided to change the `last_analysis_run_at` field to update every time analysis runs, instead of only using it to track scheduled analysis in BHE. This required a re-work of how scheduled analysis works, but since the `datapipe_status` table lives in both BHE and BHCE, some changes had to be made here as well. 

We also want to make sure we are updating the `next_scheduled_analysis_at` timestamp every time we set the `analysis.scheduled` parameter. Because this handler lives in BHCE, we needed to do some interesting things here with the database interface (since there is no concept of scheduled analysis in BHCE). The real implementation lives on the BHE side, but must be called from BHCE. 

## Motivation and Context

Resolves BED-8136

Updates `last_analysis_run_at` every time analysis starts.

## How Has This Been Tested?

Tested manually by running analysis and ensuring the timestamp is updated. Also added some more unit & integration tests.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Validates scheduled-analysis recurrence rules, persists configuration, and stores the next scheduled run time
  * Records analysis start timestamps and exposes last/next analysis times in API responses
* **Documentation**
  * OpenAPI schema updated to include last/next analysis timestamp fields
* **Tests**
  * Expanded coverage for scheduled-analysis handlers, configuration updates, datapipe status, and timestamp behavior
* **Chores**
  * Database migration adds a next-scheduled-analysis timestamp column

### DIFF
--- a/cmd/api/src/api/tools/analysis_schedule_test.go
+++ b/cmd/api/src/api/tools/analysis_schedule_test.go
@@ -24,13 +24,13 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
-	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/specterops/bloodhound/cmd/api/src/api/tools"
 	"github.com/specterops/bloodhound/cmd/api/src/ctx"
 	"github.com/specterops/bloodhound/cmd/api/src/database/mocks"
 	"github.com/specterops/bloodhound/cmd/api/src/database/types"
+	"github.com/specterops/bloodhound/cmd/api/src/database/types/null"
 	"github.com/specterops/bloodhound/cmd/api/src/model/appcfg"
 	"github.com/specterops/bloodhound/packages/go/headers"
 	"github.com/specterops/bloodhound/packages/go/mediatypes"
@@ -179,7 +179,7 @@ func TestToolContainer_SetScheduledAnalysisConfiguration(t *testing.T) {
 		reqBody, _ := json.Marshal(scheduledAnalysisRequest)
 
 		mockDB.EXPECT().SetConfigurationParameter(gomock.Any(), parameterConfig).Return(nil)
-		mockDB.EXPECT().SetNextScheduledAnalysisStartTime(gomock.Any(), time.Time{}).Return(nil)
+		mockDB.EXPECT().SetNextScheduledAnalysisStartTime(gomock.Any(), null.Time{}).Return(nil)
 
 		if req, err := http.NewRequestWithContext(requestCtx, http.MethodPut, endpoint, bytes.NewBuffer(reqBody)); err != nil {
 			t.Fatal(err)
--- a/cmd/api/src/api/tools/analysis_schedule_test.go
+++ b/cmd/api/src/api/tools/analysis_schedule_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/specterops/bloodhound/cmd/api/src/api/tools"
@@ -40,7 +41,7 @@ import (
 func TestToolContainer_GetScheduledAnalysisConfiguration(t *testing.T) {
 	endpoint := "/analysis/schedule"
 
-	t.Run("happy path", func(t *testing.T) {
+	t.Run("success getting scheduled analysis", func(t *testing.T) {
 		var (
 			ctrl     = gomock.NewController(t)
 			mockDB   = mocks.NewMockDatabase(ctrl)
@@ -108,7 +109,7 @@ func TestToolContainer_SetScheduledAnalysisConfiguration(t *testing.T) {
 	endpoint := "/analysis/schedule"
 	validRRule := "FREQ=DAILY;INTERVAL=1;DTSTART=20230101T100000Z"
 
-	t.Run("happy path", func(t *testing.T) {
+	t.Run("success enabling scheduled analysis", func(t *testing.T) {
 		var (
 			ctrl     = gomock.NewController(t)
 			mockDB   = mocks.NewMockDatabase(ctrl)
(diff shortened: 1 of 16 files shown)


cmd/api/src/database/datapipestatus.go
```diff
@@ -18,32 +18,49 @@ package database
...
 	"github.com/specterops/bloodhound/cmd/api/src/model"
 )
 
 type DatapipeStatusData interface {
+	SetLastAnalysisStartTime(ctx context.Context) error
 	UpdateLastAnalysisCompleteTime(ctx context.Context) error
 	SetDatapipeStatus(ctx context.Context, status model.DatapipeStatus) error
 	GetDatapipeStatus(ctx context.Context) (model.DatapipeStatusWrapper, error)
+	SetNextScheduledAnalysisStartTime(ctx context.Context, time time.Time) error
+}
```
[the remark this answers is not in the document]
sirisjo: this is intentional, scheduled analysis is only a feature in Enterprise.

cmd/api/src/api/v2/app_config.go
```diff
@@ -58,25 +63,43 @@ func (s Resources) GetApplicationConfigurations(response http.ResponseWriter, re
...
 	} else if !checkApiKeyExpirationParamAvailable(request.Context(), parameter, s.DB) {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusForbidden, fmt.Sprintf("Configuration parameter %s is not available.", parameter.Key), request), response)
-	} else if err := s.DB.SetConfigurationParameter(request.Context(), parameter); err != nil {
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusForbidden, fmt.Sprintf("Configuration parameter %s is not available.", parameter.Key), request), response)
+	} else if err := s.DB.SetConfigurationParameter(ctx, parameter); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else {
+		if parameter.Key == appcfg.ScheduledAnalysis {
+			updateNextScheduledAnalysisStartTime(ctx, s.DB, parameter)
+		}
```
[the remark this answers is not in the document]
sirisjo: the RRULE is validated -- the code calls parameter.Validate() which then calls the RRULE validator.

cmd/api/src/api/tools/analysis_schedule_test.go
```diff
@@ -20,20 +20,139 @@ import (
...
 	"github.com/specterops/bloodhound/packages/go/mediatypes"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
-func TestToolContainer_GetScheduledAnalysisConfiguration_Errors(t *testing.T) {
+func TestToolContainer_GetScheduledAnalysisConfiguration(t *testing.T) {
+	endpoint := "/analysis/schedule"
+
+	t.Run("happy path", func(t *testing.T) {
```
mistahj67: nit: do you mind adding some more explanation here?

cmd/api/src/api/tools/analysis_schedule_test.go
```diff
@@ -20,20 +20,139 @@ import (
...
+			require.Contains(t, response.Body.String(), "error retrieving configuration data")
+		}
+	})
+}
+
+func TestToolContainer_SetScheduledAnalysisConfiguration(t *testing.T) {
+	endpoint := "/analysis/schedule"
+	validRRule := "FREQ=DAILY;INTERVAL=1;DTSTART=20230101T100000Z"
+
+	t.Run("happy path", func(t *testing.T) {
```
mistahj67: Same feedback here, because the string is surfaced in CI, it's helpful to be prescriptive here

cmd/api/src/api/v2/app_config.go
```diff
@@ -58,25 +63,43 @@ func (s Resources) GetApplicationConfigurations(response http.ResponseWriter, re
...
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
 	} else if !checkApiKeyExpirationParamAvailable(request.Context(), parameter, s.DB) {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusForbidden, fmt.Sprintf("Configuration parameter %s is not available.", parameter.Key), request), response)
-	} else if err := s.DB.SetConfigurationParameter(request.Context(), parameter); err != nil {
+		api.WriteErrorResponse(ctx, api.BuildErrorResponse(http.StatusForbidden, fmt.Sprintf("Configuration parameter %s is not available.", parameter.Key), request), response)
+	} else if err := s.DB.SetConfigurationParameter(ctx, parameter); err != nil {
 		api.HandleDatabaseError(request, response, err)
 	} else {
+		if parameter.Key == appcfg.ScheduledAnalysis {
+			updateNextScheduledAnalysisStartTime(ctx, s.DB, parameter)
```
mistahj67, line 84: Should this surface an error here at all if updating the next scheduled start fails?

cmd/api/src/daemons/datapipe/datapipe_integration_test.go
```diff
@@ -115,13 +115,16 @@ func setupIntegrationTestSuite(t *testing.T, fixturesPath string) IntegrationTes
 
 	cl := changelog.NewChangelog(graphDB, db, changelog.DefaultOptions())
 
+	cache, err := cache.NewCache(cache.Config{MaxSize: 1})
```
mistahj67, line 118: Was this change intentional? 🤔

packages/go/openapi/src/paths/datapipe.datapipe.status.yaml
```diff
@@ -43,6 +43,12 @@ get:
                   last_complete_analysis_at:
                     type: string
                     format: date-time
+                  last_analysis_run_at: 
+                    type: string
+                    format: date-time
+                  next_scheduled_analysis_at:
```
mistahj67, line 49: random thought: does open api support description at this level? might be worth tossing in a "this field is BHE only"

cmd/api/src/api/v2/app_config.go
sirisjo: I went the route of only logging an error because at this point, the app config value has been successfully set. returning an error may mislead the consumer into thinking the config value failed to be set. Since the `nextScheduledAnalysis` time will self-correct if `null`, I don't think it's a huuuuge deal to only log an error instead of surfacing.

cmd/api/src/daemons/datapipe/datapipe_integration_test.go
sirisjo: Yes! The new tests I added go down a codepath where, if the cache is `nil`, the tests will error out with a `nil pointer` exception.

packages/go/openapi/src/paths/datapipe.datapipe.status.yaml
sirisjo: Yes! I can add this!

mistahj67 (approved):
Code looks clean 🧼 and pulled and verified working as intended! 🚀

PR #2859, opened 2026-06-03
feat(Playwright): add a11y regression testing - BED-7788

TheNando:
## Description

Add accessibility regression testing suite via Playwright

## Motivation and Context

Resolves BED-7788

## How Has This Been Tested?

Tested by manually running new testing suite locally

```
yarn test:a11y
yarn report:a11y
```

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to run accessibility checks and build/serve Allure reports; introduced a shared accessibility testing package.

* **Tests**
  * Automated WCAG A/AA accessibility tests added for multiple UI pages with themed auth snapshots and artifact capture.

* **Documentation**
  * New README documenting shared accessibility testing utilities, conventions, and how to add a11y specs.

* **Chores**
  * Example env vars added for accessibility runs; test artifacts directory added to ignore list.

* **Bug Fixes**
  * Dev startup now tolerates local test-service registration failures and continues rendering.


cmd/ui/src/App.tsx
```diff
@@ -104,7 +104,7 @@ export const Inner: React.FC = () => {
             </Helmet>
             <div className={classes.applicationContainer} id='app-root'>
                 {showNavBar && <MainNav mainNavData={mainNavData} />}
-                <div className='bg-neutral-1 grow overflow-y-auto overflow-x-hidden'>
+                <div id='content-wrapper' className='bg-neutral-1 grow overflow-y-auto overflow-x-hidden'>
```
TheNando, line 108: This adds an easy and repeatable selector to only capture the page component content, omitting the Nav menu, which may be tested in isolation. This reduces overhead for Axe Core. This is purely for makeAxeBuilder analyze targeting. User-perspective selectors (byRole, byLabel, etc) are still used to gauge loading and viewable content state.

cmd/ui/tests/a11y/login.a11y.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { expect, expectNoAccessibilityViolations, test } from 'bh-playwright-testing';
+
+test.describe('Login page accessibility', () => {
```
TheNando, line 19: An example of an exceptional test. All other tests are pre-authed. But to test login we need to first wipe out the auth.

cmd/ui/tests/a11y/no-data-dialog.a11y.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import { expectNoAccessibilityViolations, test } from '../fixtures';
+
+// The exact cypher query issued by `useGraphHasData`. The shared a11y fixture stubs this query
+// with a populated payload by default; this spec installs a higher-priority handler that returns
+// an empty payload so the "No Data Available" upload dialog renders and can be scanned.
+const GRAPH_HAS_DATA_QUERY = 'MATCH (A) WHERE NOT A:MigrationData RETURN A LIMIT 1';
+
+test.describe('No Data Available dialog accessibility', () => {
+    test('upload dialog has no detectable WCAG A/AA violations', async ({ page, makeAxeBuilder }, testInfo) => {
```
TheNando, line 25: Another exceptional test. Cypher graph state is stubbed to ensure "No Data" dialog is not present. To test this dialog, we change the stub to be explicitly empty.

cmd/ui/.env.example
```diff
@@ -1 +1,6 @@
 TARGET_PROXY_URL="http://localhost:8080"
+
+A11Y_TEST_URL="http://127.0.0.1:3000"
```
TheNando, line 3: I need to double check that the README explicitly mentions that adding these .env vars is required for authing test sessions.

cmd/ui/tsconfig.json
```diff
@@ -23,24 +19,15 @@
         "noFallthroughCasesInSwitch": true,
         "declaration": true,
         "declarationMap": true,
-        "baseUrl": ".",
```
TheNando, line 26: The only real change here is removing the baseUrl, which is a deprecated practice. The rest is just whitespace.

packages/go/stbernard/command/license/internal/cmd.go
```diff
@@ -51,6 +51,7 @@ func Run(env environment.Environment, args Args) error {
 			"justfile",
 			filepath.Join("cmd", "api", "src", "api", "static", "assets"),
 			filepath.Join("cmd", "api", "src", "cmd", "testidp", "samlidp"),
+			filepath.Join("cmd", "ui", "playwright"),
```
TheNando, line 54: This ensures that `just prepare-for-codereview` ignores any of the generated a11y test artifacts.

cmd/ui/playwright.a11y.config.ts
```diff
@@ -0,0 +1,120 @@
...
+                use: {
+                    ...device,
+                    storageState: authStorageStateFor(theme),
+                    theme,
+                },
+                dependencies: ['setup'],
+            }))
+        ),
+    ],
+    ...(shouldRunWebServer ? { webServer } : {}),
```
dcairnsspecterops: Why not just `webServer: shouldRunWebServer ? webServer : null`, or the like?

TheNando: Yeah. Good call. This feels unnecessary. The only difference I can even think of is that the key won't be added, rather than the key being present but explicitly undefined (instead of null, to match type). Not really an advantage of any sort. I'll update.

packages/javascript/bh-playwright-testing/src/axe.ts
```diff
@@ -0,0 +1,173 @@
...
+        .map((node) => {
+            const target = node.target.join(' ');
+            const failureSummary = node.failureSummary ? `\n  ${node.failureSummary}` : '';
+
+            return `- \`${target}\`${failureSummary}`;
+        })
+        .join('\n');
+
+    return `### ${violation.id} (${violation.impact ?? 'unknown impact'})
+${violation.help}
```
Holocraft, line 116: nit: looks like indentation is off here.

packages/javascript/bh-playwright-testing/src/auth.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+// Login once and snapshot `storageState` for both light and dark themes.
+//
+// Capturing both snapshots from a single session avoids the parallel-login race where two
+// setups as the same user would invalidate each other's session. This relies on the BH
+// shared UI shell (login form labels "Email Address" / "Password", the LOGIN submit button,
+// the `global_nav-dark-mode` toggle, and the `persistedState` localStorage key written by
+// the global store's throttled subscriber) being identical across consumers.
+export async function loginAndSnapshotThemes(opts: LoginAndSnapshotThemesOptions): Promise<void> {
```
urangel, line 57: We can reduce some of the complexity here by only persisting the auth token since the difference between the two persisted files is a toggle on the darkMode boolean. 

A helper that produces a certain local storage state similar to what is done in the login spec would allow for more flexibility.

packages/javascript/bh-playwright-testing/src/stubs/graph-has-data.ts
```diff
@@ -0,0 +1,48 @@
...
+// higher-priority `page.route` for the same URL — Playwright runs handlers in LIFO order,
+// so a test-local handler wins for the cases it cares about.
+export async function installGraphHasDataStub(page: Page): Promise<void> {
+    await page.route('**/api/v2/graphs/cypher', async (route) => {
+        if (route.request().method() !== 'POST') {
+            return route.fallback();
+        }
+        return route.fulfill({
+            json: {
+                data: {
```
urangel, line 33: There are some mock factory utilities in the ui projects that produce similar shapes of data. It would be nice to share and reuse some of these for related ideas to be coupled and maintained together. I could see us making a different package in packages/javascript specifically for this so that both unit tests and integration tests can consume them. 

That would be too much for adding here but I wanted to voice that so it's on our minds.

urangel (approved):
Great work! Looking forward to having this run in CI

packages/javascript/bh-playwright-testing/src/axe.ts
TheNando: That should be ok. It's a template literal string. aka. Multiline backtick string with interpolation. If I indent here, it will indent in the output report as well. One way around that would be to construct the string line by line, but this should be ok I think.

packages/javascript/bh-playwright-testing/src/stubs/graph-has-data.ts
TheNando: This is a good consideration. My main concern is dependency shape. `bh-playwright-testing` is a small, mostly agnostic helper package. Pulling in a dep from `bh-shared-ui` (where our mocks are) would introduce React, js-client-library, msw, doodle-ui...the whole stack. This could especially get tricky when we try to point Playwright at Doodle directly to do component testing.

One solution we could consider would be to generate the mocks at the app level (at `cmd/ui`) and pass those values in with `installGraphHasDataStub`. I'll try that out and see how it goes. The other alternative that comes to mind would changing this playwright lib to be a generalized testing lib, and move all the mocks and fixtures in here.

What do you think?

urangel: True, dep shape could get tricky. I was describing having a sibling package in packages/javascript, maybe packages/javascript/mocks, though we probably need to isolate types better in advance. Having response types generated from the open api spec would help get us there.

Passing in the mock sounds like a good solution for this case but no need to make any changes unless you want. My comment was a general observation and thought about similar patterns and maintenance but I don't have a solid idea of what that should look like 😄

PR #2870, opened 2026-06-09
chore: proposal for Audit Logging Middleware BHE-8674

superlinkx:
## Description

An RFC proposal for how we handle audit logging going forward to resolve known design limitations in light of both the API refactor and customer pressure to audit log every user action.

## Motivation and Context

Resolves BED-8674

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Documentation**
  * Added an RFC for default HTTP-layer audit logging via middleware, including breaking audit action naming, dual-write transition with deduplication guidance, and `source` tagging.
  * Documented security behavior (sensitive-field redaction, no request-body logging), anonymous handling, outcome mapping, and scoped exclusions (e.g., health).
  * Added design docs for an audit middleware reference, module structure/public API guidance, and an audit logs partitioning migration/retention procedure.

### DIFF
--- a/rfc/bh-rfc-7.md
+++ b/rfc/bh-rfc-7.md
@@ -151,29 +151,38 @@ This is the mechanism for **preserving an existing action string across a refact
 
 The enrichment hook also covers the model-level `Fields` that `AuditData()` provides today. Handlers may attach structured detail (affected entity IDs, before/after diffs) to `bhctx`; the middleware drains it into `Fields` on the result row. Generic coverage (actor, action, status, IP) is automatic everywhere; high-value endpoints keep the same rich detail they emit today.
 
-## 6. Audit Log Slice
+## 6. Audit Module
 
-Both the middleware and individual handlers need to write audit rows. Rather than have each re-implement actor resolution, intent/result pairing, and persistence, this RFC introduces an **audit log slice** in the new `server/` architecture (`server/audit/`). It follows the standard four-layer slice pattern (`appdb → services → handlers`) and exposes a small service that is injected wherever audit rows are written.
+Both the middleware and individual handlers need to write audit rows. Rather than have each re-implement actor resolution, intent/result pairing, and persistence, this RFC introduces an **audit module** in the BHCE server architecture (`bhce/server/audit/`). It follows the module isolation pattern established by `server/clients/` and `server/kennel/`, using `internal/` packages for encapsulation and exposing a small public API for audit recording.
 
 ### 6.1 Responsibilities
 
-The slice is the single chokepoint for writing to `audit_logs`:
+The audit module is the single chokepoint for writing to `audit_logs`:
 
--   Owns `audit_logs` persistence in its `appdb` layer, including the partitioned-table writes described in [Section 7](#7-retention-and-table-stability).
--   Owns the domain `Entry` type in its `services` layer. The three database status values (`intent`, `success`, `failure`) are internal to the appdb layer; callers express outcome through which `Recorder` method they call.
--   Exposes a `Recorder` interface that hides `commit_id` generation and intent/result bookkeeping behind a small surface, so the middleware and handlers never construct `model.AuditLog` rows by hand.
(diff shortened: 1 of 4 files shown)


rfc/bh-rfc-7.md
```diff
@@ -0,0 +1,384 @@
...
+
+```
+server/audit/
+├── audit.go        # Register(): builds appdb → services, starts the worker, returns the Recorder
+├── appdb/          # audit_logs writes + partition DDL; defines its own pgxQuerier
+├── services/       # Entry, Status, CommitID, Contribution, Recorder, Database interface, the worker
+└── handlers/       # OPTIONAL — only if the read endpoint (Section 6.4.5) is migrated here
+```
+
+-   **`services`** owns the domain types (`Entry`, `CommitID`, `Contribution`), the public `Recorder` surface, and the consumer-defined `Database` interface (the only methods the service calls). The buffered result-writer worker lives here. `Status` is an internal constant set used only by `appdb` to write the correct `status` column value; it is not part of the public interface.
```
mistahj67: If status is an internal constant, is the new definition not a breaking change for any of the original audit logs?

rfc/bh-rfc-7.md
```diff
@@ -0,0 +1,384 @@
...
+
+This is the mechanism for **preserving an existing action string across a refactor**: when an endpoint that previously emitted a typed `AuditLogAction` (e.g. `"CreateUser"`) is moved into the `server/` slice architecture, its handler records that same action through the audit slice, so the database value does not change. The existing `AuditLogAction` constants are not removed or deprecated by this RFC; they remain available for handlers that choose to keep using them.
+
+### 5.3 Model-Level Fields
+
+The enrichment hook also covers the model-level `Fields` that `AuditData()` provides today. Handlers may attach structured detail (affected entity IDs, before/after diffs) to `bhctx`; the middleware drains it into `Fields` on the result row. Generic coverage (actor, action, status, IP) is automatic everywhere; high-value endpoints keep the same rich detail they emit today.
```
mistahj67, line 219: I'm not following exactly what this means, handlers will now be shoving data into the ctx that gets passed around? Is this the right place for that data?

rfc/bh-rfc-7.md
```diff
@@ -0,0 +1,384 @@
...
+-   **Consistent attribution** - Actor, source IP, request ID, and timing derive from one place (`bhctx` + `IdentityResolver`) rather than being re-derived at each call site.
+-   **Bounded table growth** - Monthly `created_at` partitioning with `DROP PARTITION` retention closes an existing operational gap regardless of this proposal's other details.
+
+## 3. Considerations
+
+### 3.1 Impact on Existing Systems
+
+#### 3.1.1 Action Data Contract
+
+The most significant impact of this proposal is a **breaking change to the `Action` field** of `audit_logs`. Today, `Action` is one of approximately 43 typed string constants (e.g., `"CreateUser"`, `"ExportAllRisks"`). The proposed default changes this to a derived string of the form `"METHOD /path/template"` (e.g., `"POST /api/v2/users"`).
```
mistahj67, line 33: Is there any concern with having a space in the action name? I think it might be worth either swapping space to `-` or no space and just METHOD/path/template (e.g. GET/api/v2/etc)

rfc/bh-rfc-7.md
mistahj67: Discussed offline. Can disregard

rfc/bh-rfc-7.md
superlinkx: It shouldn't be, we have the list of constants and what they map to today, we should replicate them but I don't want to expose those constants to consumers of the interface in the new setup.

rfc/bh-rfc-7.md
superlinkx: This is out of scope of the RFC imo. We should verify during implementation if a space is going to be a problem, but it shouldn't be and I'd prefer it to stay as-is for readability

AD7ZJ (approved):
I've read through this and I don't see any issues. I like the migration plan and the ability to auto-generate action values going forward. Let's give it a shot!

PR #2892, opened 2026-06-15
refactor: optimize az descendant queries BED-4867

urangel:
## Description

Refactors are made to the azure entity descendant relationship data retrieval query logic for improving efficiency and query duration.

## Motivation and Context

Resolves BED-4867

Our current methods for obtaining entity descendant relationship data often result in timeout errors in modestly sized tenants. This issue is exacerbated when attempting to view this information through the entity panel in the UI as a series of requests are executed through a promise.all call that hinges on all of the requests succeeding and otherwise fails to show any information even if only one of the requests fails. 

The logic has been updated to pivot based on source and target kinds in order to route queries more effectively. Our modeled AZContains relationships include those that are only ever 1 hop away (like AZTenant to AZDevice) and those that may be several hops away (AZTenant to AZLogicApp will at minimum cross an AZSubscription and AZResourceGroup).

We were previously traversing any AZContains relationships then returning paths that had the target kind in the path. This means that we would follow AZContains to deep terminals for any target kind when most if not all traversals could be eliminated up front.

The logic has been updated to first pull the set of terminals by kind and tenant id then traversing backwards to the source for relationships that are nested. Direct relationships simply do an edge filter rather than a traversal.

Finally, the UI is updated to use promise.allSettled over promise.all so that successful nested sections are still displayed and functional even if a sibling section fails.

## How Has This Been Tested?
API integration tests have been added. The [azure example data](https://bloodhound.specterops.io/get-started/quickstart/ce-ingest-sample-data#azure) was uploaded and various descendant entity panels were inspected (the AZTenant descendants section was of particular focus).

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Release Notes

* **New Features**
  * Improved Azure “related entities” rendering to better respect source context across graph and list views.

* **Performance Improvements**
  * Faster, tenant-scoped descendant traversal and more consistent descendant counting for entity detail panels.

* **Bug Fixes**
  * Fixed graphed table count aggregation across multiple section requests, including partial failure scenarios.
  * Updated the related-entities API to return a generic error if display-kind lookup fails.

* **Tests**
  * Expanded Azure integration tests for direct vs non-direct descendant paths and nested listings.


packages/go/analysis/azure/queries.go
```diff
@@ -788,31 +788,163 @@ func FetchKeyVaultReaderCounts(tx graph.Transaction, keyVault *graph.Node) (KeyV
...
+	azure.Tenant: {
+		azure.App,
+		azure.Device,
+		azure.Group,
+		azure.Role,
+		azure.ServicePrincipal,
+		azure.Subscription,
+		azure.User,
+	},
+	azure.ManagementGroup: {
```
AD7ZJ: According to this, `ManagementGroups` can be nested. But isDirectDescendent() then uses this data to skip searching nested paths and assumes there can be only a single hop between the source and end type. Do you think this is a problem, that we might miss nested ManagementGroups?

Equating valid end kinds with kinds that can only be a single hop away seems confusing and maybe not correct. Perhaps rename this map to something like `containsSingleHopKinds`?

urangel: Thanks for catching this! You're right. I've removed the ManagementGroup key from the map so that these do a traversal instead.

AD7ZJ (approved):
Looks good to me, working nicely on my local setup.

PR #2895, opened 2026-06-17
feat(logging): add query parameter logging for /api/v2 requests - BED-8563

cami-specter:
## Description
Add query parameter logging for /api/v2 requests for easier Elastic/Kibana lookups. This logs the raw string and makes it easier to filter and search in Elastic without parsing the full request_uri field. 

- adds query_parameters slog attribute in LoggingMiddleware
- only logs for /api/v2 endpoints with non-empty query strings
- ignores SSO callback endpoint and a few others (see link for [full audit](https://specterops.atlassian.net/browse/BED-8722 )). Even if params are already being logged via `request_uri`, it is probably better not to add them to the new `query_parameters` field. 
- adds table-driven tests

Note:
After some deliberation, decided not to add nested logging due to potential Elastic index mapping problems. (If we were to add a large list of parameters (in the hundreds) to our Elastic daily indexes mapping, that could cause issues – indexes are already > 1000). 

## Motivation and Context

Resolves BED-8563. 

There's a general benefit to having query_parameters as a dedicated field since it makes it easier to filter and aggregate on query param usage across most endpoints. 

Specifically, we were looking at non-default limit and skip parameters queries for the api/v2/search endpoint

## How Has This Been Tested?

- Added unit tests (table-driven) covering non-/api/v2 paths, empty query strings, ignoring a specific endpoint, etc.
- Manually tested output with EnableAPILogging enabled, hit various /api/v2 endpoints, verified logs

## Screenshots (optional):

log output showing query_parameters field (/api/v2 request with a single parameter)

pretty print of the above log output (disabling text logging)

log output showing query_parameters field (/api/v2 request with multiple parameters)

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Improvements

* Request logs now omit query-parameter details for login-support, SSO callback, and SAML authentication endpoints.
* Query parameters continue to be logged for other API v2 requests, including repeated parameters and complete query strings.

## Tests

* Added coverage for query-parameter logging and exclusion behavior across API, SSO, login-support, and SAML URL patterns.

### DIFF
--- a/cmd/api/src/api/middleware/logging.go
+++ b/cmd/api/src/api/middleware/logging.go
@@ -118,6 +118,19 @@ func setSignedRequestFields(request *http.Request, logAttrs *[]slog.Attr) {
 	}
 }
 
+func isSSOCallbackPath(requestPath string) bool {
+	pathSegments := strings.Split(strings.Trim(requestPath, "/"), "/")
+
+	if len(pathSegments) < 5 {
+		return false
+	}
+
+	return pathSegments[0] == "api" &&
+		pathSegments[1] == "v2" &&
+		pathSegments[2] == "sso" &&
+		pathSegments[4] == "callback"
+}
+
 // LoggingMiddleware is a middleware func that outputs a log for each request-response lifecycle. It includes timestamped
 // information organized into fields suitable for searching or parsing.
 func LoggingMiddleware(_ auth.IdentityResolver, bypassLimitsParam bool) func(http.Handler) http.Handler {
@@ -193,7 +206,8 @@ func LoggingMiddleware(_ auth.IdentityResolver, bypassLimitsParam bool) func(htt
 			)
 
 			// Add logging of query parameters for /api/v2 endpoints
-			if strings.HasPrefix(request.URL.Path, "/api/v2") && request.URL.RawQuery != "" {
+			// Ignore SSO callback path logging
+			if strings.HasPrefix(request.URL.Path, "/api/v2/") && request.URL.RawQuery != "" && !isSSOCallbackPath(request.URL.Path) {
 				logAttrs = append(logAttrs, slog.String("query_parameters", request.URL.RawQuery))
 			}
 		})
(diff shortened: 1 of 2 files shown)


superlinkx (approved):
This shows a lot of attention to detail and will make logging far more valuable

cmd/api/src/api/middleware/logging.go
```diff
@@ -117,9 +118,22 @@ func setSignedRequestFields(request *http.Request, logAttrs *[]slog.Attr) {
...
+
+	if len(pathSegments) < 5 {
+		return false
+	}
+
+	return pathSegments[0] == "api" &&
+		pathSegments[1] == "v2" &&
+		pathSegments[2] == "sso" &&
+		pathSegments[4] == "callback"
+}
```
mistahj67: Wanted to note that there are old but still maintained due to still in use saml endpoints that likely also should be ignored and those are "versionless" 

Further if sso is a concern, there is the `sso/{id}/metadata` endpoint as well. Might be worth considering a way to flag these in the route layer to be skipped 🤔 vs maintaining them here

mistahj67 (changes requested):
Blocked to verify intent

mistahj67 (approved):
approving to remove block. discussed offline. will likely benefit from a longer audit into logging queries. thank you for your patience and waiting 🙇

cami-specter:
Full audit on logging queries is now complete: 
https://specterops.atlassian.net/browse/BED-8722 (audit linked in ticket)

Adds two additional exclusions: the login/support already pointed out by TJ + a guard against a deprecated path (v2 acs as v1 is already excluded). See details in the audit.

cmd/api/src/api/middleware/logging.go
cami-specter: Function has been modified to add a few additional paths, thanks for bringing up one of them! See https://specterops.atlassian.net/browse/BED-8722 for full list of considered endpoints. 

v1 is already excluded where the helper is called, but added a preventive guard for a deprecated path anyway. 
As for the architecture, that's an interesting suggestion that I looked into. It is more involved however, although we could consider that in the future.

PR #2926, opened 2026-06-25
fix(ui): preserve deeplink redirect after login - BED-5483

### LINKED ISSUE #2806: Bug: UI/UX Feedback: Unauthenticated deeplinks do not redirect
## Description:

Using a deeplink to navigate to an unauthenticated BloodHound instance redirects to the login page but does not redirect back to the deeplink after 

Claude's suggestion below

## Component(s) Affected:

- UI

## Steps to Reproduce:

1. Click on a BH deeplink for an instance you are not yet logged into
2. Log in
3. You will be presented with a blank graph rather than the deeplink

## Expected Behavior:

1. Click on a BH deeplink for an instance you are not yet logged into
2. Log in
3. You will be presented with query / data from the cypher embedded in the deeplink

## Actual Behavior:

1. Click on a BH deeplink for an instance you are not yet logged into
2. Log in
3. You will be presented with a blank graph rather than the deeplink

N/A

## Environment Information:

BloodHound: v9.1.0

## Additional Information:

N/A

- Capture works — cmd/ui/src/components/AuthenticatedRoute/AuthenticatedRoute.tsx:29 correctly stashes the original location into React Router's state.from when bouncing an unauthenticated user to /login:
 `return <Navigate to={ROUTE_LOGIN} state={{ from: location }} />;`
- Consumption is missing — cmd/ui/src/views/Login.tsx:76 hardcodes ROUTE_HOME and never reads location.state.from:
`if (authState.sessionToken !== null && authState.user !== null) return <Navigate to={ROUTE_HOME} replace />;`

So the original deeplink is captured, then silently dropped on successful login. A repo-wide search shows nothing else       consumes state.from in the auth flow — it's dead data.
Suggested fix                                                                                                              
  In cmd/ui/src/views/Login.tsx, read useLocation().state.from and prefer it over ROUTE_HOME:                                
```
  import { Navigate, useLocation } from 'react-router-dom';
// ...
const Login: React.FC = () => {
const location = useLocation();
const from = (location.state as { from?: Location } | null)?.from;
const redirectTo = from ? `${from.pathname}${from.search ?? ''}${from.hash ?? ''}` : ROUTE_HOME;
// ...
if (authState.sessionToken !== null && authState.user !== null)
return <Navigate to={redirectTo} replace />;
```
Notes on edge cases:                                                                                                                                                                                                                                      - Safety: state.from is React-Router in-memory state, not URL-controlled, so there is no open-redirect risk (unlike a ?redirect= query param). No need to validate it's same-origin.
  - SSO path (Login.tsx:69-71): window.location.assign('/api/v2/sso/{slug}/login') does a full page navigation, which wipes router state. The from won't survive the round-trip. If you want SSO deeplinks to also work, you'd need to either (a) stash   from in sessionStorage before calling window.location.assign, then read it back when the SSO callback lands on /login, or
  - (b) pass it through the backend SSO flow as a query param (requires backend changes). I'd scope this PR to the password/OTP flow first and file SSO as a follow-up.
  - Already-logged-in users hitting /login directly will correctly fall back to ROUTE_HOME since location.state will be null.
```
## Related Issues:

N/A

## Contributor Checklist:

### PR DESCRIPTION
**Pull request by ZyphenSVC**
## Description

This change fixes navigating BloodHound utilizing deeplinks when not logged in.

`AuthenticatedRoute` preserves the original route in React Router state when redirecting unauthenticated users to `/login`, but `Login.tsx` always redirects authenticated users to `ROUTE_HOME` instead of the deeplink route. 

This change changes the state location and redirects users back to their original destination after successful authentication. If no prior route exists, it would go back to `ROUTE_HOME` as usual.

## Motivation and Context

Resolves #2806

Users opening a Bloodhound deeplink while unauthenticated should be returned to that deeplink after logging in. This change fixes that, letting users land on the change query or graph instead of the default blank view.

## How Has This Been Tested?

1. Started local development environment with `just bh-dev`.
2. Imported sample dataset.
3. Created a functioning path between two users and copied deeplink.
4. Opened the exact same deeplink when authenticated, then logged out and tried while unauthenticated.
5. Confirmed that authenticated view redirected back to original deeplink.
6. Confirmed that unauthenticated view redirected back to `/login` and then redirected back to original deeplink after authenticating.

## Screenshots (optional):

N/A

## Types of changes

## Checklist:

  - [x] Assigned myself to this PR
  - [x] Added the appropriate labels
  - [x] Associated an issue: #2806 
  - [x] Read the Contributing guide: https://github.com/SpecterOps/BloodHound/wiki/Contributing
  - [ ] Open API docs
  - [ ] Code comments (GoDocs / JSDocs)
  - [x] Added/updated tests to cover my changes
  - [x] All new and existing tests passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login redirect behavior so already signed-in users return to their original destination instead of always landing on the home page.
  * Preserves the full target path, including query and hash details, when available.

### DIFF
--- a/cmd/ui/src/views/Login.tsx
+++ b/cmd/ui/src/views/Login.tsx
@@ -19,12 +19,20 @@ import { LoginForm, LoginViaSSOForm, OneTimePasscodeForm, apiClient, useAppName 
 import React, { useEffect, useState } from 'react';
 import { Helmet } from 'react-helmet-async';
 import { useQuery, useQueryClient } from 'react-query';
-import { Navigate } from 'react-router-dom';
+import { Navigate, useLocation } from 'react-router-dom';
 import LoginPage from 'src/components/LoginPage';
 
 import { login as loginAction, logout } from 'src/ducks/auth/authSlice';
 import { ROUTE_HOME, ROUTE_USER_DISABLED } from 'src/routes/constants';
 import { useAppDispatch, useAppSelector } from 'src/store';
+
+type RedirectLocationState = {
+    from?: {
+        pathname: string
+        search?: string
+        hash?: string
+    }
+};
 
 const Login: React.FC = () => {
     /* Hooks */
@@ -34,6 +42,11 @@ const Login: React.FC = () => {
 
     const authState = useAppSelector((state) => state.auth);
     const appName = useAppName();
+
+    const location = useLocation();
+
+    const from = (location.state as RedirectLocationState | null)?.from;
+    const redirectTo = from ? `${from.pathname}${from.search ?? ''}${from.hash ?? ''}` : ROUTE_HOME;
 
     const [useSSO, setUseSSO] = useState(false);
 
@@ -81,7 +94,7 @@ const Login: React.FC = () => {
     /* Implementation */
 
     // Redirect if already logged in
-    if (authState.sessionToken !== null && authState.user !== null) return <Navigate to={ROUTE_HOME} replace />;
+    if (authState.sessionToken !== null && authState.user !== null) return <Navigate to={redirectTo} replace />;
 
     if (listSSOProvidersQuery.isLoading) {
         return (

### CODE REVIEW

**Comment by ZyphenSVC** [comment]

File: cmd/ui/src/views/Login.tsx [hunk 1]
```diff
@@ -19,13 +19,21 @@ import { LoginForm, LoginViaSSOForm, OneTimePasscodeForm, apiClient, useAppName
...
 import { ROUTE_HOME, ROUTE_USER_DISABLED } from 'src/routes/constants';
 import { useAppDispatch, useAppSelector } from 'src/store';
 
+type RedirectLocationState = {
+    from?: {
+        pathname: string;
+        search?: string;
+        hash?: string;
+    };
+};
```
maffkipp, line 35: `react-router-dom` provides a `Path` type that we could use here:

```suggestion
type RedirectLocationState = {
    from?: Path;
};
```

cmd/ui/src/views/Login.tsx
```diff
@@ -81,7 +94,7 @@ const Login: React.FC = () => {
     /* Implementation */
 
     // Redirect if already logged in
-    if (authState.sessionToken !== null && authState.user !== null) return <Navigate to={ROUTE_HOME} replace />;
+    if (authState.sessionToken !== null && authState.user !== null) return <Navigate to={redirectTo} replace />;
```
maffkipp, line 97: The `to` prop here has a type of `string | Partial<Path>`, so we could also just pass `from` directly and skip the concatenation above:
```suggestion
    if (authState.sessionToken !== null && authState.user !== null) return <Navigate to={from ?? ROUTE_HOME} replace />;
```

maffkipp (changes requested):
This change works as expected for username/password auth but does not preserve deep link params across the SSO or reset password flows from what I was able to test. I think it would be helpful to add support for those as well if possible!

PR #2938, opened 2026-06-29
feat(Data Quality): add `environment_kind_id` and `environment_kind_display_name` to available-domains response BED-8685

sirisjo:
## Description

We want to enrich the /available-domains response with the`environment_kind_id` and `environment_kind_display_name` to be able to populate Data Quality stats per environment on the Data Quality page.

To do this, I expanded the GetEnvironmentKindsAndEnvironmentExtensionDisplayNames function from the OpenGraph Service to return complete environment data rather than just the extension display name. The DB was already returning all of the data, but the service function was stripping it down to just the display name. The display name is still accessible to the existing functionality, but now we also have access to the other fields we need.

## Motivation and Context

Resolves BED-8685

## How Has This Been Tested?

Tests updated, and tested locally. 

## Screenshots (optional):

Screenshot 2026-06-29 at 15 54 17

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Environment results now include additional environment kind details: `environment_kind_id` and `environment_kind_display_name` (read-only).
  * OpenGraph environment listings expose these fields alongside existing environment information.
* **Bug Fixes**
  * Environment listing and filtering now consistently derive environment properties from schema environment metadata.
* **Tests**
  * Updated API/service tests to match the revised environment selector payload.
* **Documentation**
  * Updated the OpenAPI schema to include `environment_kind_id` and `environment_kind_display_name`.

### DIFF
--- a/cmd/api/src/model/search.go
+++ b/cmd/api/src/model/search.go
@@ -36,6 +36,7 @@ type EnvironmentSelector struct {
 	Type               string    `json:"type"`
 	Name               string    `json:"name"`
 	ObjectID           string    `json:"id"`
+	SchemaExtensionID  *int32    `json:"schema_extension_id,omitempty"`
 	Collected          bool      `json:"collected"`
 	ImpactValue        *int      `json:"impactValue,omitempty"`
 	HygieneAttackPaths *int64    `json:"hygiene_attack_paths,omitempty"` // caution: if value is bigger than maxsafeint, the UI will truncate the value
--- a/cmd/api/src/model/search.go
+++ b/cmd/api/src/model/search.go
@@ -36,7 +36,7 @@ type EnvironmentSelector struct {
 	Type               string    `json:"type"`
 	Name               string    `json:"name"`
 	ObjectID           string    `json:"id"`
-	SchemaExtensionID  *int32    `json:"schema_extension_id,omitempty"`
+	ExtensionID        *int32    `json:"extension_id,omitempty"`
 	Collected          bool      `json:"collected"`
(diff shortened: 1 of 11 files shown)


cmd/api/src/model/search.go
```diff
@@ -42,6 +42,12 @@ type EnvironmentSelector struct {
 	Exposures          Exposures `json:"exposures,omitempty"`
 }
 
+type EnvironmentProperties struct {
+	Type            string  `json:"type"`
+	KindId          *int32  `json:"environment_kind_id,omitempty"`
```
mvlipka: Let's use the `null` package here

cmd/api/src/model/search.go
```diff
@@ -42,6 +42,12 @@ type EnvironmentSelector struct {
 	Exposures          Exposures `json:"exposures,omitempty"`
 }
 
+type EnvironmentProperties struct {
```
mvlipka, line 46: Debating if we really need this since we're embedding the struct onto the EnvironmentSelector, but don't have strong feelings about it at the moment

mvlipka (changes requested):
Just get in the changes for the `null` and we're good to go!

cmd/api/src/model/search.go
sirisjo: Yeah I could go either way here too, but thought it was a nice way to group the environment-related properties together & have a concrete type to return from `resolveEnvProperties`

mvlipka (approved):

PR #2946, opened 2026-06-30
fix(data quality, db): migration to rename old data_quality_stats_table and create with new `data_quality_stats` table BED-8616

sirisjo:
## Description

Upon adding this table in a previous release, we discovered that there exists an old `data_quality_stats` table in the test and production environments from a long time ago that is not tracked in any of our migration files. When running the migration in the test environment, the new table was not created because the `data_quality_stats` table already existed. We need to drop that old table first, and then we can recreate it with the new columns that we want.

## Motivation and Context

Resolves BED-8616

## How Has This Been Tested?

Ran the migration locally.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated how data quality statistics are stored, including richer metric metadata and improved timestamp tracking to preserve metric history more reliably.
  * Improved migration safety by recreating the statistics storage cleanly and restoring the previous state if changes are rolled back.
  * Added support for an optional related “kind” reference and ensured efficient access to recent entries via a timestamp-based index.


cmd/api/src/database/migration/migrations/20260630135900_v9_drop_and_recreate_data_quality_stats.sql
```diff
@@ -0,0 +1,40 @@
...
+-- WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+-- See the License for the specific language governing permissions and
+-- limitations under the License.
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- +goose Up
+
+-- Drop the old data_quality_stats table that is not currently used anywhere but still exists in prod
+DROP TABLE IF EXISTS data_quality_stats;
```
elikmiller: Should we instead rename/archive the table instead of deleting it?

elikmiller (changes requested):

cmd/api/src/database/migration/migrations/20260630135900_v9_drop_and_recreate_data_quality_stats.sql
sirisjo: renamed to `data_quality_stats_old` 🫡

mvlipka (approved):
Verified the indexes are correct and that we don't have an associated aggregation table in `test`
Great catch!

elikmiller (approved):

PR #2954, opened 2026-07-02
feat: add multi-destination pathfinding with drag-and-drop reordering

julian1j:
Enables chaining up to 4 nodes (1 source + 3 destinations) in pathfinding queries. Each consecutive pair fires a parallel shortestpath API call and results are merged client-side. Includes drag-and-drop reordering, node removal, and URL persistence via tertiarySearch/quaternarySearch params.

## Description

 Enables chaining up to 4 nodes (1 source + 3 destinations) in pathfinding queries, similar to Google Maps multi-stop routing                                                                 
  - Each consecutive node pair fires a parallel shortestpath API call; results are merged client-side                                                                                           
  - Drag-and-drop reordering, inline node removal, and "Add Destination" button
  - URL persistence via new tertiarySearch/quaternarySearch params                                                                                                                               
  - Swap button retained for 2-node pathfinding; hidden when 3+ nodes are active 
  
## Motivation and Context

Resolves BED-8075

## How Has This Been Tested?

Tested locally on CE with standard example dataset. 

## Screenshots (optional):
two_nodes
three_nodes
four_nodes

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Multi-stop pathfinding in Explore now supports up to four ordered waypoints.
  * The pathfinding UI is dynamic, letting users add, remove, and reorder waypoint nodes (drag-and-drop and keyboard).
* **Bug Fixes**
  * Waypoint selections stay in sync with URL parameters, including higher-order waypoints.
  * Multi-leg path results are merged correctly across consecutive waypoint pairs.
  * Clearing higher-order waypoints now clears their related query fields.
* **Chores**
  * Updated ignore rules to exclude the local `.claude/` directory.


packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts
```diff
@@ -14,131 +14,218 @@
...
+        setNodes((prev) => {
+            const next = [...prev];
+            while (next.length <= index) next.push(emptyNode());
+            next[index] = { ...next[index], ...update };
+            return next;
+        });
+    }, []);
+
+    const syncNodeFromParam = useCallback(
+        (index: number, param: string | null, data: any) => {
```
specter-flq: We should consider adding a type (SearchResults | undefined) to the data param, that would help remove the any type for the node below and the need for type casting (as SearchValue | undefined ) and can safely remove that

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts
```diff
@@ -14,131 +14,218 @@
...
-                setDestinationSelectedItem(matchedNode);
-            }
-        } else {
-            setDestinationSearchTerm('');
-            setDestinationSelectedItem(undefined);
-        }
-    }, [secondarySearch, destinationSearchData]);
+    useEffect(() => {
+        syncNodeFromParam(3, quaternarySearch, data3);
+    }, [syncNodeFromParam, quaternarySearch, data3]);
```
specter-flq: We should maybe consider reducing from 4 useEffects to one,so everything could look something like this:

 ```
const params = [primarySearch, secondarySearch, tertiarySearch, quaternarySearch];
const searchData = [data0, data1, data2, data3];

const syncNodeFromParam = useCallback(
    (index: number, param: string | null, data: SearchResults | undefined) => {
        if (param && data) {
            const matchedNode = Object.values(data).find((node) => node.objectid === param);
            if (matchedNode) {
                updateNode(index, { searchTerm: matchedNode.name, selectedItem: matchedNode });
            }
        } else if (!param) {
            updateNode(index, emptyNode());
        }
    },
    [updateNode]
);

const syncAllNodesFromParams = () => {
    params.forEach((param, index) => syncNodeFromParam(index, param, searchData[index]));
};

// Sync URL params to node state
useEffect(() => {
    syncAllNodesFromParams();
}, [syncNodeFromParam, ...params, ...searchData]);
```

File: packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts [hunk 3]
```diff
@@ -14,131 +14,218 @@
...
 
-    // Handle syncing each search field up to query params to trigger a graph query. Should trigger pathfinding if both have been selected and node
-    // if only one has been selected
-    const handleSourceNodeSelected = (selected?: SearchValue) => {
-        const objectId = selected?.objectid ?? '';
-        const term = selected?.name ?? objectId;
+    // Keep extraNodeCount in sync with URL params
+    useEffect(() => {
+        const count = quaternarySearch ? 2 : tertiarySearch ? 1 : 0;
```
specter-flq: nit: wonder if we can make this ternary a bit more readable, one suggestion could be to do something like:
```
const extraCount = [tertiarySearch, quaternarySearch].filter(Boolean).length;
setExtraNodeCount(extraCount);
```

The caveat here is that it assumes that if we have a quaternary we would have the tertiary, to get the right extra count of 2. I think that will always be true because we have a cleanup function for params, but not a fan of assumptions. 

If not we could always put a comment on top of the terniary that says: "if quaternary then 2, if not and tertiary then 1 if none 0" 

Also should consider changing the variable name to two words  like extraCount, just to keep to descriptive readability to scan quicker.

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts
```diff
@@ -14,131 +14,218 @@
...
+    const { data: data2 } = useSearch(kw2, t2);
+    const { data: data3 } = useSearch(kw3, t3);
+
+    const updateNode = useCallback((index: number, update: Partial<PathfindingNode>) => {
+        setNodes((prev) => {
+            const next = [...prev];
+            while (next.length <= index) next.push(emptyNode());
+            next[index] = { ...next[index], ...update };
+            return next;
+        });
```
specter-flq: nice to have: we should consider using more descriptive two word var names here. for example instead of update use updatedNode, for prev -> previousNodes, next -> newNodes

Also may help to leave quick one line comments on top  of the while loop and the line below to give a quick scannable way of reading it. i.e Pad the array with empty nodes to prepare for new ones, update the new list with new values

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts
```diff
@@ -14,131 +14,218 @@
...
         }
     };
 
-    // Handle changes internal to the search form that should not trigger a graph query. Each param should sync independently
-    const handleSourceNodeEdited = (edit: string) => {
-        setSourceSelectedItem(undefined);
-        setSourceSearchTerm(edit);
+    const handleReorderNodes = (fromIndex: number, toIndex: number) => {
+        const currentNodes = nodes.slice(0, totalNodeCount);
+        const [moved] = currentNodes.splice(fromIndex, 1);
```
specter-flq: nit: would consider giving moved a two word var name to make it a bit more descriptive

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts
```diff
@@ -14,131 +14,218 @@
...
 
-    const handleDestinationNodeEdited = (edit: string) => {
-        setDestinationSelectedItem(undefined);
-        setDestinationSearchTerm(edit);
+    const handleAddNode = () => {
+        if (totalNodeCount < MAX_NODES) {
+            setExtraNodeCount((prev) => prev + 1);
+            setNodes((prev) => {
+                const next = [...prev];
+                while (next.length < 2 + extraNodeCount + 1) next.push(emptyNode());
```
specter-flq: maybe we can consider making this:

```
  const targetNodeCount = totalNodeCount + 1;
  while (next.length < targetNodeCount) next.push(emptyNode());
```

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -14,23 +14,37 @@
...
     handleDestinationNodeEdited: (edit: string) => void;
     handleSourceNodeSelected: (selected: SearchValue) => void;
     handleDestinationNodeSelected: (selected: SearchValue) => void;
+    handleNodeEdited: (index: number) => (edit: string) => void;
+    handleNodeSelected: (index: number) => (selected: SearchValue) => void;
     handleSwapPathfindingInputs: () => void;
+    handleReorderNodes: (fromIndex: number, toIndex: number) => void;
+    handleRemoveNode: (index: number) => void;
+    handleAddNode: () => void;
 };
```
specter-flq: We should consider moving these two types to usePathfindingSearch and export from there to use here. And that way in the case of the type PathfindingNode we don't have to declare it twice.

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/usePathfindingSearch.ts
```diff
@@ -14,131 +14,218 @@
...
-    const [destinationSelectedItem, setDestinationSelectedItem] = useState<SearchValue | undefined>(undefined);
+const MAX_NODES = 4;
+
+const SEARCH_PARAM_KEYS = ['primarySearch', 'secondarySearch', 'tertiarySearch', 'quaternarySearch'] as const;
 
-    const { primarySearch, secondarySearch, setExploreParams } = useExploreParams();
+type PathfindingNode = {
+    searchTerm: string;
+    selectedItem: SearchValue | undefined;
+};
```
specter-flq: lets add an export to this to be able to use in PathfindingSearch file, also incluse the PathFindingSearchState so it would look like this:

```
export type PathfindingNode = {
    searchTerm: string;
    selectedItem: SearchValue | undefined;
};

export type PathfindingSearchState = {
    sourceSearchTerm: string;
    destinationSearchTerm: string;
    sourceSelectedItem: SearchValue | undefined;
    destinationSelectedItem: SearchValue | undefined;
    nodes: PathfindingNode[];
    totalNodeCount: number;
    maxNodes: number;
    handleSourceNodeEdited: (edit: string) => void;
    handleDestinationNodeEdited: (edit: string) => void;
    handleSourceNodeSelected: (selected: SearchValue) => void;
    handleDestinationNodeSelected: (selected: SearchValue) => void;
    handleNodeEdited: (index: number) => (edit: string) => void;
    handleNodeSelected: (index: number) => (selected: SearchValue) => void;
    handleSwapPathfindingInputs: () => void;
    handleReorderNodes: (fromIndex: number, toIndex: number) => void;
    handleRemoveNode: (index: number) => void;
    handleAddNode: () => void;
};
```

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +55,168 @@ const PathfindingSearch = ({
...
+        label: index === 0 ? 'Start Node' : 'Destination Node',
+        searchTerm: node.searchTerm,
+        selectedItem: node.selectedItem,
+        removable: index > 0 && totalNodeCount > 2,
+        autoFocus:
+            index === 0
+                ? !node.searchTerm
+                : index === 1
+                  ? !!(nodes[0]?.searchTerm && !node.searchTerm)
+                  : !node.searchTerm,
```
specter-flq: Maybe readability we can consider abstracting this to a function  something like:

```
  const shouldAutoFocus = (index: number, node: PathfindingNode): boolean => {
        if (index === 1) return !!(nodes[0]?.searchTerm && !node.searchTerm);
        return !node.searchTerm;
    };

    const visibleNodes = nodes.slice(0, totalNodeCount).map((node, index) => ({
        label: index === 0 ? 'Start Node' : 'Destination Node',
        searchTerm: node.searchTerm,
        selectedItem: node.selectedItem,
        removable: index > 0 && totalNodeCount > 2,
        autoFocus: shouldAutoFocus(index, node),
    }));
```

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +55,168 @@ const PathfindingSearch = ({
...
+                        draggable
+                        onDragStart={handleDragStart(index)}
+                        onDragEnter={handleDragEnter(index)}
+                        onDragLeave={handleDragLeave(index)}
+                        onDragOver={handleDragOver}
+                        onDrop={handleDrop(index)}
+                        onDragEnd={handleDragEnd}
+                        className={`relative flex items-center gap-1 rounded transition-all group ${
+                            dragIndex === index ? 'opacity-40' : ''
+                        } ${dragOverIndex === index ? 'ring-2 ring-primary ring-offset-1' : ''}`}>
```
specter-flq: We should consider using the cn() function to keep consistent with how we do dynamic classes in other files. It would look like this:

```
className={cn('relative flex items-center gap-1 rounded transition-all group', {
     'opacity-40': dragIndex === index,
     'ring-2 ring-primary ring-offset-1': dragOverIndex === index,
})}>
```

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +31,171 @@ const PathfindingSearch = ({
...
+                                title='Remove destination'>
+                                <FontAwesomeIcon icon={faTimes} size='sm' />
+                            </button>
+                        )}
+                    </div>
+                ))}
+                {totalNodeCount < maxNodes && (
+                    <button
+                        onClick={handleAddNode}
+                        className='flex items-center gap-1.5 text-xs text-neutral-500 hover:text-neutral-700 dark:hover:text-neutral-300 py-0.5 cursor-pointer'
```
specter-flq: Sorry missed this in previous feedback, but in dark mode this button is not visible because of color. We should add the `dark:text-common-white` and delete `dark:hover:text-neutral-300` as it also makes it "disappear" in dark.

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +31,171 @@ const PathfindingSearch = ({
...
+                                handleNodeSelected={handleNodeSelected(index)}
+                                inputValue={node.searchTerm}
+                                selectedItem={node.selectedItem || null}
+                                labelText={node.label}
+                            />
+                        </div>
+                        {node.removable && (
+                            <button
+                                onClick={() => handleRemoveNode(index)}
+                                className='absolute right-1 top-1/2 -translate-y-1/2 p-1 text-neutral-500 hover:text-neutral-700 dark:hover:text-neutral-300 z-10'
```
specter-flq: Sorry missed this in previous feedback, but in dark mode this button is not visible because of color. We should add the dark:text-common-white and delete dark:hover:text-neutral-300 as it also makes it "disappear" in dark.

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +31,171 @@ const PathfindingSearch = ({
...
+                            onKeyDown={(e) => {
+                                if (e.key === 'ArrowUp' && index > 0) {
+                                    e.preventDefault();
+                                    handleReorderNodes(index, index - 1);
+                                } else if (e.key === 'ArrowDown' && index < visibleNodes.length - 1) {
+                                    e.preventDefault();
+                                    handleReorderNodes(index, index + 1);
+                                }
+                            }}
+                            className='cursor-grab text-neutral-400 hover:text-neutral-600 dark:hover:text-neutral-300 opacity-0 group-hover:opacity-100 focus:opacity-100 transition-opacity'>
```
specter-flq: Sorry missed this in previous feedback, but in dark mode this button is not visible because of color. We should add the dark:text-common-white and delete dark:hover:text-neutral-300 as it also makes it "disappear" in dark.

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +31,171 @@ const PathfindingSearch = ({
...
+                    onSwapPathfindingInputs={handleSwapPathfindingInputs}
+                />
+            )}
             <EdgeFilter pathfindingFilterState={pathfindingFilterState} />
         </div>
     );
 };
 
-const SourceToBullseyeIcon = () => {
+const SourceToBullseyeIcon = ({ destinationCount }: { destinationCount: number }) => {
```
specter-flq: In the previous version the circles/bullseyes are aligned with the center of the input. In this change they are align with bottom of input, we should consider looking for a solution to centering them like they were before. 

With these changes:

Image

vs before

Image

julian1j: image
image

Bullseye icon/circles are now centered with the input.

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
julian1j: image

Drag-to-reorder, add destination, and the remove "x" button are all visible in dark mode with the latest commit.

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +31,171 @@ const PathfindingSearch = ({
...
+                        <div className='flex-grow'>
+                            <ExploreSearchCombobox
+                                autoFocus={node.autoFocus}
+                                handleNodeEdited={handleNodeEdited(index)}
+                                handleNodeSelected={handleNodeSelected(index)}
+                                inputValue={node.searchTerm}
+                                selectedItem={node.selectedItem || null}
+                                labelText={node.label}
+                            />
+                        </div>
```
specter-flq, line 196: When going over the suggested changes ( which worked well ty again for that ) noticed that when dragging it had the circle/bullseye icon and line as part of the "ghost" element being dragged. So we should consider only showing a drag of the input element itself. I have attached a video of how I thought it could look.

https://github.com/user-attachments/assets/465bdd3d-3f39-416a-946b-03dd1941714e

This had me thinking that we maybe should have UI/UX give a bit of input on this to make sure we are applying what they had in mind for list dragging and dropping in the app. Waiting to see what they think. 

Here is how the block of code looks for that element on my end:

```
                   <div
                        key={index}
                        onDragEnter={handleDragEnter(index)}
                        onDragLeave={handleDragLeave(index)}
                        onDragOver={handleDragOver}
                        onDrop={handleDrop(index)}
                        className='relative flex items-center gap-1 rounded group'>
                        <PathfindingNodeIcon isStartNode={index === 0} showConnector={index > 0} />
                        <div
                            draggable
                            onDragStart={handleDragStart(index)}
                            onDragEnd={handleDragEnd}
                            className={cn('relative flex flex-grow items-center gap-1', {
                                'opacity-40': dragIndex === index,
                            })}>
                            <div
                                role='button'
                                tabIndex={0}
                                aria-label={`Reorder ${node.label}, position ${index + 1} of ${visibleNodes.length}`}
                                aria-roledescription='sortable'
                                onKeyDown={(e) => {
                                    if (e.key === 'ArrowUp' && index > 0) {
                                        e.preventDefault();
                                        handleReorderNodes(index, index - 1);
                                    } else if (e.key === 'ArrowDown' && index < visibleNodes.length - 1) {
                                        e.preventDefault();
                                        handleReorderNodes(index, index + 1);
                                    }
                                }}
                                className='cursor-grab text-neutral-400 hover:text-neutral-600 dark:text-common-white dark:hover:text-neutral-light-5 opacity-0 group-hover:opacity-100 focus-visible:opacity-100 transition-opacity'>
                                <FontAwesomeIcon icon={faGripVertical} size='sm' />
                            </div>
                            <div
                                className={cn('flex-grow rounded transition-all', {
                                    'ring-2 ring-primary ring-offset-1': dragOverIndex === index,
                                })}>
                                <ExploreSearchCombobox
                                    autoFocus={node.autoFocus}
                                    handleNodeEdited={handleNodeEdited(index)}
                                    handleNodeSelected={handleNodeSelected(index)}
                                    inputValue={node.searchTerm}
                                    selectedItem={node.selectedItem || null}
                                    labelText={node.label}
                                />
                            </div>
                            {node.removable && (
                                <button
                                    onClick={() => handleRemoveNode(index)}
                                    className='absolute right-1 top-1/2 -translate-y-1/2 p-1 text-neutral-500 hover:text-neutral-700 dark:text-common-white dark:hover:text-neutral-light-5 z-10'
                                    aria-label='Remove destination'
                                    title='Remove destination'>
                                    <FontAwesomeIcon icon={faTimes} size='sm' />
                                </button>
                            )}
                        </div>
                    </div>
```

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
```diff
@@ -41,55 +31,171 @@ const PathfindingSearch = ({
...
+                            onKeyDown={(e) => {
+                                if (e.key === 'ArrowUp' && index > 0) {
+                                    e.preventDefault();
+                                    handleReorderNodes(index, index - 1);
+                                } else if (e.key === 'ArrowDown' && index < visibleNodes.length - 1) {
+                                    e.preventDefault();
+                                    handleReorderNodes(index, index + 1);
+                                }
+                            }}
+                            className='cursor-grab text-neutral-400 hover:text-neutral-600 dark:text-common-white dark:hover:text-neutral-light-5 opacity-0 group-hover:opacity-100 focus:opacity-100 transition-opacity'>
```
specter-flq: There is a small but noticeable scenario where after you drag and drop, this icon stays visible until you click away. The fix for this would be to use `focus-visible:opacity-100` instead of `focus:opacity-100`

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
julian1j: ah, yes, this was just introduced when i applied the alignment fix because the draggable element is the entire row div, and moving things around to fix alignment issues introduced this behavior. agree that it doesn't look ideal

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
julian1j: yup, good catch on that, just confirmed the fix locally

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
julian1j: I'm looking at two options; I can either snapshot a smaller subset and just detach the icons from the input that we want to "ghost", or I can take the icon out of the row that was created with the alignment fix and rearrange the panel as a multi-column grid, the drag/drop ghost would just grab everything in the right column in that case.

specter-flq: Check out the code block I added above, that one shows the option you mentioned about just detaching the icon to a outer wrapper. Feels like less of a lift. Also the outline on drag is only around the input in that block I added. Feels a bit off that an outline is around everything when it normally around an input/element. In any case going to get some guidance from UI/UX to confirm we are on the right track with what they want, making sure we are consistent with the visual change they are starting to implement in the app. Thank again for all your changes!

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/PathfindingSearch.tsx
julian1j: I'm pushing a fix for this and a proposed solution for the drag-and-drop ghosting that captured more than it should have. I just want y'all to have a working version that addresses all of the comments above and we can refine again if needed. This is what I have being grabbed in the "ghost" of the drag and drop in the latest commit: 

Screenshot 2026-08-04 at 9 48 41 AM

specter-flq: Adding for Vis. Thanks again for the changes. Was going over your latest commit changes, there are three things happening still with the ghost:

- When you type in a search and see results, if you try to drag it take the results with it as well (shown in video)
- If you add a destination and focus on it and try to drag it bring the empty text results with it (shown in video)
- When you hover another row to drop (as seen in video), there is an outline that goes around everything including the circle/bullseye and it is almost on top of the circle icon. I would consider only applying outline/highlight to the input as there is not much space to the left of the bullseye to give it a bit of whitespace

Tia!


https://github.com/user-attachments/assets/2c88a419-5a31-4e2d-b3da-fab7b4eaa868

specter-flq (approved):
LGTM! Thank you for the changes!

PR #2960, opened 2026-07-06
docs: improve esc14 texts

JonasBK:
## Description

Improve the abuse text for new edges

## Motivation and Context

Resolves BED-6155

## How Has This Been Tested?

Locally with this dataset: 

[20250611101235_BloodHound.zip](https://github.com/user-attachments/files/29698808/20250611101235_BloodHound.zip)

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Documentation

* Improved readability of ADCS ESC14 Scenario A guidance across Linux and Windows workflows.
* Standardized headings, paragraph structure, and spacing in security identity and public information help text.
* Clarified certificate requirements and expanded Linux guidance with fallback provider and X509SKI mapping details.
* Preserved existing links and execution commands while making instructional content easier to scan.


packages/javascript/bh-shared-ui/src/components/HelpTexts/AdcsEsc14ScenarioA/AdcsEsc14ScenarioA.tsx
```diff
@@ -27,17 +27,23 @@ export const AltSecIdentitiesBlurb = () => (
 
 export const AdcsEsc14ScenarioALinux: FC = () => (
     <>
-        <Typography variant='body1'> ADCS ESC14 Scenario A </Typography>
-        <Typography variant='body2'>
-            An attacker can add an explicit certificate mapping in the AltSecurityIdentities of the target referring to
-            a certificate in the attacker's possession, and then use this certificate to authenticate as the target.
+        <Typography variant='body1'>
+            <b>ADCS ESC14 Scenario A</b>
+        </Typography>
```
bfaulk96: No need to nest these, can just use className
```suggestion
        <Typography variant='body1' className='font-bold'>
            ADCS ESC14 Scenario A
        </Typography>
```

packages/javascript/bh-shared-ui/src/components/HelpTexts/AdcsEsc14ScenarioA/AdcsEsc14ScenarioA.tsx
```diff
@@ -27,17 +27,23 @@ export const AltSecIdentitiesBlurb = () => (
...
+        <Typography variant='body2' component='div'>
+            <p className='my-4'>
+                An attacker can add an explicit certificate mapping in the AltSecurityIdentities of the target referring
+                to a certificate in the attacker's possession, and then use this certificate to authenticate as the
+                target.
+            </p>
```
bfaulk96: Same here - you likely don't need nested elements. Try this, I believe it should look similar:
```suggestion
        <Typography variant='body2' component='div' className='my-4'>
                An attacker can add an explicit certificate mapping in the AltSecurityIdentities of the target referring
                to a certificate in the attacker's possession, and then use this certificate to authenticate as the
                target.
```

packages/javascript/bh-shared-ui/src/components/HelpTexts/AdcsEsc14ScenarioA/AdcsEsc14ScenarioA.tsx
```diff
@@ -63,26 +69,38 @@ export const AdcsEsc14ScenarioALinux: FC = () => (
...
+                set on computer objects. Computers have validated write permission to their own dNSHostName attribute by
+                default, but neither users nor computers can write to their own mail attribute by default.
+            </p>
+            <p className='my-4'>
+                The abuse is possible with the strong explicit certificate mappings X509IssuerSerialNumber, X509SKI, or
+                X509SHA1PublicKey. In this example, we use X509SHA1PublicKey.
+            </p>
+            <p className='my-4'>Get the SHA1 hash of the certificate using openssl:</p>
         </Typography>
-        <Typography variant='body2'>Get the SHA1 hash of the certificate using openssl:</Typography>
```
bfaulk96, line 85: I think the Doodle team wants people to favor using Typography over `<p>` elements. This could all be rewritten a bit clearer as the following to show actual intent (since that top level is meant to be a div container)
```suggestion
        <div>
            <Typography variant='body2' className='my-4'>
                If the enrollment fails with an error message stating that the Email or DNS name is unavailable and
                cannot be added to the Subject or Subject Alternate name, then it is because the enrollee principal does
                not have their mail or dNSHostName attribute set, which is required by the certificate template. The
                mail attribute can be set on both user and computer objects but the dNSHostName attribute can only be
                set on computer objects. Computers have validated write permission to their own dNSHostName attribute by
                default, but neither users nor computers can write to their own mail attribute by default.
            </Typography>
            <Typography variant='body2' className='my-4'>
                The abuse is possible with the strong explicit certificate mappings X509IssuerSerialNumber, X509SKI, or
                X509SHA1PublicKey. In this example, we use X509SHA1PublicKey.
            </Typography>
            <Typography variant='body2' className='my-4'>Get the SHA1 hash of the certificate using openssl:</Typography>
        </div>
```

packages/javascript/bh-shared-ui/src/components/HelpTexts/AdcsEsc14ScenarioA/AdcsEsc14ScenarioA.tsx
```diff
@@ -123,18 +141,25 @@ sha1 fingerprint=f61331a504cff8cb5e60c269632c31aa3032a54a`}
...
+            <Typography variant='body1'>
+                <b>ADCS ESC14 Scenario A</b>
+            </Typography>
+            <Typography variant='body2' component='div'>
+                <p className='my-4'>
+                    An attacker can add an explicit certificate mapping in the altSecurityIdentities of the target
+                    referring to a certificate in the attacker's possession, and then use this certificate to
+                    authenticate as the target.
+                </p>
             </Typography>
```
bfaulk96: I won't keep commenting on it, but similar thoughts to earlier on the rest of these file changes

bfaulk96 (changes requested):
Didn't want to keep commenting on the same thing, but the general gist is that we don't really want child `p` tag elements under `Typography` elements.. There are definitely exceptions, but generally we'd want `Typography` elements to be final, basic elements, rather than wrapper elements.

For the files in this PR:
- When there are more than one p tag, we should probably just make the outer Typography div a normal div, and change those p tags to Typography (I think with variant `body2` mostly in the examples in here)
- When there's just one p tag with a className, we should just add that className to the parent Typography div (or if that styling doesn't match exactly, using a normal div outside and the typography inside, like the case above).
- When we have a nested `<b>` tag inside of a Typography, just add the `font-bold` className to the typography instead

Appreciate the work to make these pages look nicer!!

JonasBK:
Thanks a lot for the review @bfaulk96! 
And sorry for all the n00b mistakes.. I think I got it fixed now. Let me know if there is anything I should change :)

bfaulk96:
> Thanks a lot for the review @bfaulk96! And sorry for all the n00b mistakes.. I think I got it fixed now. Let me know if there is anything I should change :)

No need to be sorry! We don't know what we don't know! Totally reasonable to do what you did without familiarity with the Typography component.

Just running this locally now to double check that everything looks good!

packages/javascript/bh-shared-ui/src/components/HelpTexts/WriteAltSecurityIdentities/General.tsx
bfaulk96: Testing this locally, you may want to add `className='my-4'` to this one as well - the paragraph is pretty close to the accordion header

Image

```suggestion
            <Typography variant='body2' className='my-4'>
```

packages/javascript/bh-shared-ui/src/components/HelpTexts/AdcsEsc14ScenarioA/AdcsEsc14ScenarioA.tsx
```diff
@@ -48,41 +55,55 @@ export const AdcsEsc14ScenarioALinux: FC = () => (
...
+                ).
+            </Typography>
+        </div>
+
+        <Typography variant='body2' className='my-4'>
+            If the attacker cannot obtain a suitable certificate from ADCS, they may be able to obtain one from another
+            PKI provider used in the target environment.
         </Typography>
+
+        <Typography variant='body1'> Execution </Typography>
```
bfaulk96: nit: Might consider adding `font-medium` or `font-semibold` to this for a little emphasis as a header?
```suggestion
        <Typography variant='body1' className='font-medium'>Execution</Typography>
```

JonasBK: Good call!

bfaulk96 (approved):
Looks great!

PR #2962, opened 2026-07-06
feat: KindInfo Storage Method Implementation: BED-8858

brandonshearin:
## Description
- Create method that inserts a slice of KindInfos to the db
- Get method that returns a slice of KindInfos for a given `kind_id`
- Goose migration to add the new table
- Sentinel errors and KindInfo struct added to the service layer of the graphdb onion vertical

## Motivation and Context
Resolves BED-8858
In order to support entity panels, we want to store net-new `KindInfo` objects into a table named `schema_kind_info` and implement Create/Read methods

## How Has This Been Tested?
- for the migration, i ran `just goose up` and `just goose down` to verify that the new `schema_kind_info` table is properly created/dropped.
- for the Go methods i added integration test coverage

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added support for entity kind “info” entries with titles, ordered display (by position then title), and JSON-based content.
  * Kind info is now fully supported: create, update, list, and delete for a schema kind.

* **Bug Fixes**
  * Improved handling for invalid inputs, including duplicate position, duplicate info key, and missing related kind records.
  * Listing now returns a stable, consistent order and an empty result when no entries exist.


cmd/api/src/database/migration/migrations/20260706203456_v9_create_kind_info_table.sql
```diff
@@ -0,0 +1,51 @@
...
+--
+-- SPDX-License-Identifier: Apache-2.0
+
+-- +goose Up
+
+CREATE TABLE IF NOT EXISTS schema_kind_info (
+    id SERIAL PRIMARY KEY,
+    kind_id INT NOT NULL REFERENCES kind (id),
+    node_kind_id INT REFERENCES schema_node_kinds (id) ON DELETE CASCADE,
+    relationship_kind_id INT REFERENCES schema_relationship_kinds (id) ON DELETE CASCADE,
```
mistahj67, line 23: Instead of having 2 separate FK's here for cascade delete, would it simplify this to just FK to the `schema_extension` id  / table? Then there only needs to be 1 column that works for either node or relationship kind

server/graphdb/internal/services/services.go
```diff
@@ -41,6 +43,29 @@ type Kind struct {
...
 
+// KindInfo holds the data associated with a single entity panel
+type KindInfo struct {
+	KindID             int32
+	NodeKindID         *int32
+	RelationshipKindID *int32
+	InfoKey            string
+	Title              string
+	Position           int32
+	Content            json.RawMessage
```
mistahj67: I think this struct is missing parts of the database columns like the createdAt / updatedAt / id fields

cmd/api/src/database/migration/migrations/20260706203456_v9_create_kind_info_table.sql
brandonshearin: I had the same question when i picked up this ticket.

The answer is no, because of the way extension upload works now, which is different from the MVP version in which an upload meant a full drop of the old version.

The current extension upload logic is a proper upsert, so registered node kinds and edge kinds get cleaned up when the latest extension no longer contains them.

This means we need to tie `schema_kind_info` lifecycle to the narrower `schema_node_kind` and `schema_relationship_kind` tables

cc @LawsonWillard

server/graphdb/internal/services/services.go
brandonshearin: Seeking guidance regarding these timestamp fields. They aren't exposed to any API consumer, so I left them out of the "services" contract. What do you think?

LawsonWillard:
Just a heads up, I'm not planning on migrating the existing Upsert Extension endpoint to the new api architecture as part of BED-8764 yet due to it being a LOT of moving pieces and a short deadline. 

Regardless, with how the current upsert works i think we'll need individual create, update and delete methods, the existing read one here should work for the new endpoints but will need to be ported to the old architecture. I can create them as part of BED-8764 though.

LawsonWillard:
Until we migrate the UpsertExtension endpoint we may want to delete the write path here and port it to the old architecture.

Long term, i think the upsert extension endpoint will be owned by a separate service. Something like `extensions` but we can refine once we look at moving the upsert endpoint to the new architecture. 

This is to differentiate extension management from graph read paths and leave us the flexibility when each service becomes more complicated down the line.

server/graphdb/internal/services/services.go
mistahj67: Well you are correct, every layer of this architecture affords a point where one can control exactly what is available to use. And that could be blocked at this layer, but I think of the service layer as the "bottom-most" layer and so should be the most descriptive. And then the view / handler layer, you can opt to not include those fields. 

Another way to think of this is the db layer consumes the service layer, so it will likely benefit from those fields. 

That said, as you pointed out, every layer is an opportunity to define the boundaries, yes, agreed.

cmd/api/src/database/migration/migrations/20260706203456_v9_create_kind_info_table.sql
mistahj67: Ah so the schema_extension record doesn't get deleted but the schema_X_kind record can still be deleted. Fair enough! I was operating on old architecture. An unfortunate complexity to have 2 separate FK columns and require 1 or the other, but it does seem unavoidable here unless a refactor was done to create a schema_kind table that generated an id used in all 3 of these tables. That's a larger lift for sure. 

Closing this concern, thank you for elucidating.

server/graphdb/internal/appdb/kindinfo.go
```diff
@@ -0,0 +1,165 @@
...
+
+		return fmt.Errorf("creating kind infos: %w", err)
+	}
+
+	return nil
+}
+
+// kindInfoRow is the package-local DB row type for a schema_kind_info entry.
+type kindInfoRow struct {
+	KindID             int32           `db:"kind_id"`
```
LawsonWillard, line 35: I think it may be best to return the row id here and include it in the service struct as well. May not be needed by the API at the moment but we will likely need it at some point in the future

server/graphdb/internal/appdb/kindinfo.go
```diff
@@ -0,0 +1,165 @@
...
+
+// constraint names from the schema_kind_info table (see the migration that creates it)
+// that are translated into services-layer sentinels.
+const (
+	constraintKindInfoKindIDFKey     = "schema_kind_info_kind_id_fkey"
+	constraintKindInfoUniquePosition = "schema_kind_info_unique_kind_position"
+	constraintKindInfoUniqueInfoKey  = "schema_kind_info_unique_kind_info_key"
+)
+
+func (s *Store) CreateKindInfos(ctx context.Context, kindInfos []services.KindInfo) error {
```
ddlees: This method is fine but the ticket calls out the following:

> Create a method to add all KindInfo entries for a given kind id
> - All entries MUST stay in sync with their respective node/relationship kind
> - If a node/relationship kind removes a KindInfo entry from its definition, that entry MUST be removed

We "can" move this requirement to the layers that manage node & relationship kinds. That will require the following:
- Open a new transaction
- Delete the existing node/relationship kind
- Deletion triggers the cascade delete of relevant kind info rows
- Recreate the node/relationship kind with the same id (I suppose you don't have to use the same id but retaining the identifier can be helpful)
- Create the relevant kind info rows
- End transaction

It's unclear to me what the appropriate pattern is for transactions at this or other layers. If this makes sense we can move the requirement, otherwise we'll either want to update this creation method to fulfill the requirements or create a delete method and another method that uses it to fulfill the requirements.

ddlees (dismissed):

server/graphdb/internal/appdb/kindinfo.go
brandonshearin: I understand the need you are articulating.

I think this DB method should be pure create. I think a higher-level function in the opengraph extension upload path should orchestrate the kind of thing you spelled out.

We probably need another ticket to wire this into the existing upload path.

brandonshearin: Lawson confirmed that this point of feedback will be covered in his work in BED8764

brandonshearin:
fresh start --

create, update, get KindInfo handlers defined in existing `OpenGraphSchema` interface for easy integration with extension upload and reconciliation path.

GetKindInfos() left in graphdb onion -- for easy integration with `api/v2/nodes/{node_id}` et al.

cmd/api/src/database/graphschema.go
```diff
@@ -1387,6 +1394,174 @@ func (s *BloodhoundDB) GetPrimaryDisplayKinds(ctx context.Context) (graphschema.
...
+
+	graphSchemaKindInfoConstraintKindIDFKey     = "schema_kind_info_kind_id_fkey"
+	graphSchemaKindInfoConstraintUniquePosition = "schema_kind_info_unique_kind_position"
+	graphSchemaKindInfoConstraintUniqueInfoKey  = "schema_kind_info_unique_kind_info_key"
+)
+
+func (s *BloodhoundDB) CreateKindInfo(ctx context.Context, kindID int32, nodeKindID, relationshipKindID *int32, kindInfo model.KindInfoInput) (model.GraphSchemaKindInfo, error) {
+	var (
+		content            = kindInfo.Content
+		createdGraphSchema = model.GraphSchemaKindInfo{
```
LawsonWillard: non blocking nit: is this needed if we're just overriding it below?

brandonshearin: sloppy AI. fixed it.

server/graphdb/internal/services/services.go
```diff
@@ -41,6 +43,25 @@ type Kind struct {
 	Name string
 }
 
+// KindInfo holds the data associated with a single entity panel
+type KindInfo struct {
+	ID int32
+
+	KindID int32 // dawgs kind id
```
urangel, line 50: question: Should we use graph.ID for this type instead of int32? dawgs kind ids look like they are defined as uint64

brandonshearin: you ask a fine question. i'm not sure that it really matters.

in the dawgs library itself, i see a lot of int16 kind ids.

in the file that you commented on, i see int32 kind ids. in the graphschema.go file where i implemented KindInfo crud, i see int32 as well. i lean towards keeping as int32 for now.

its anybodys game.

LawsonWillard: I think it may be best to leave it, all the other schema tables use int32 for KindID so it would cause a lot of annoying type conversions if this were different. Also the dawgs kind id column is SMALLINT so it cant hold anything higher than int32 anyways

urangel: huh. Do you happen to know why we use so many different types?

int32 is consistent with graphschema.go so that seems okay

LawsonWillard (approved):

urangel (approved):
lgtm

LawsonWillard:
Updated changes look good to me! 🚀

seanjSO (approved):
🚀

PR #2973, opened 2026-07-09
feat(data quality): count OpenGraph nodes by environment BED-8614

mvlipka:
## Description

This PR adds counting OpenGraph nodes for extensions. The counts are aggregated by each node's environment id

## Motivation and Context

Resolves: BED-8614

## How Has This Been Tested?

I registered multiple extensions and ingested their data and then verified the counts in the data_quality_stats table against a simple Cypher query in Explore
image
image

image
image

I was able to test aggregations by adding the individual environment kinds collected for a specific run_id and verifying that that aggregation adds up to the individual counts of multiple environments for that same run_id
image

Testing relationship counts:
image
image

Aggregations:
image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added OpenGraph data-quality statistics collection, including per-environment and aggregated node and relationship counts.

* **Bug Fixes**
  * OpenGraph metrics are now conditional on the relevant feature being enabled.
  * If OpenGraph stats are enabled but no stats are returned, metrics are not persisted.

* **Tests**
  * Added end-to-end coverage for saving OpenGraph metrics, feature-flag behavior, skip conditions, and detailed failure propagation.

### DIFF
--- a/cmd/api/src/services/dataquality/dataquality.go
+++ b/cmd/api/src/services/dataquality/dataquality.go
@@ -18,7 +18,6 @@ package dataquality
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log/slog"
 	"sync"
@@ -26,14 +25,12 @@ import (
 	"github.com/gofrs/uuid"
 	"github.com/specterops/bloodhound/cmd/api/src/database"
 	"github.com/specterops/bloodhound/cmd/api/src/database/types/nan"
-	"github.com/specterops/bloodhound/cmd/api/src/database/types/null"
 	"github.com/specterops/bloodhound/cmd/api/src/model"
 	"github.com/specterops/bloodhound/cmd/api/src/model/appcfg"
 	adAnalysis "github.com/specterops/bloodhound/packages/go/analysis/ad"
 	"github.com/specterops/bloodhound/packages/go/analysis/post"
 	"github.com/specterops/bloodhound/packages/go/bhlog/attr"
 	"github.com/specterops/bloodhound/packages/go/bhlog/measure"
-	"github.com/specterops/bloodhound/packages/go/graphschema"
 	"github.com/specterops/bloodhound/packages/go/graphschema/ad"
 	"github.com/specterops/bloodhound/packages/go/graphschema/azure"
 	"github.com/specterops/bloodhound/packages/go/graphschema/common"
@@ -497,7 +494,7 @@ func SaveDataQuality(ctx context.Context, db database.Database, graphDB graph.Da
 		}
 	}
 
-	// OpenGraph node counts
+	// OpenGraph node and relationship counts
 	if openGraphExtensionManagementFlag, err := db.GetFlagByKey(ctx, appcfg.FeatureOpenGraphExtensionManagement); err != nil {
 		return fmt.Errorf("could not get open graph extension management feature flag: %w", err)
 	} else if openGraphExtensionManagementFlag.Enabled {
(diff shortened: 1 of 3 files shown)


cmd/api/src/services/dataquality/dataquality.go
```diff
@@ -493,5 +497,235 @@ func SaveDataQuality(ctx context.Context, db database.DataQualityData, graphDB g
...
+			environmentStats, environmentAggregations := buildOpenGraphNodeCountStats(runID.String(), extension, environment, nodeKindCounts, kindIDs)
+			stats = append(stats, environmentStats...)
+			aggregations = append(aggregations, environmentAggregations...)
+		}
+	}
+
+	return stats, aggregations, nil
+}
+
+func getOpenGraphSourceKind(ctx context.Context, db database.Database, environment model.SchemaEnvironment) (model.Kind, error) {
```
sirisjo: **nit**: can we rename this function to convey that we're getting the environment source kind?

cmd/api/src/services/dataquality/dataquality.go
```diff
@@ -493,5 +497,235 @@ func SaveDataQuality(ctx context.Context, db database.DataQualityData, graphDB g
...
+	}
+
+	if len(kinds) == 0 {
+		return model.Kind{}, fmt.Errorf("could not get source kind for schema environment %d: %w", environment.ID, database.ErrNotFound)
+	}
+
+	return kinds[0], nil
+}
+
+func countOpenGraphNodesByEnvironment(ctx context.Context, graphDB graph.Database, environmentKind graph.Kind, sourceKind graph.Kind) (map[string]map[string]int, error) {
```
sirisjo: **nit**: the return type for this function isn't super clear, could we create a bespoke type for this to communicate what this is supposed to be more clearly?

cmd/api/src/services/dataquality/dataquality.go
```diff
@@ -493,5 +497,235 @@ func SaveDataQuality(ctx context.Context, db database.DataQualityData, graphDB g
...
+			continue
+		}
+
+		environmentIDs = append(environmentIDs, environmentID)
+	}
+
+	return environmentIDs, nil
+}
+
+func getOpenGraphCountKindIDs(ctx context.Context, db database.Database, countsByEnvironmentID map[string]map[string]int) (map[string]null.Int32, error) {
```
sirisjo: **nit**: same feedback here as the other function return type, I think it would be more readable to have this return type be an named type struct

cmd/api/src/services/dataquality/dataquality.go
```diff
@@ -493,5 +497,235 @@ func SaveDataQuality(ctx context.Context, db database.DataQualityData, graphDB g
...
+		return nil, fmt.Errorf("could not get data quality metric kinds: %w", err)
+	}
+
+	kindIDsByName := make(map[string]null.Int32, len(kinds))
+	for _, kind := range kinds {
+		kindIDsByName[kind.Name] = null.Int32From(kind.ID)
+	}
+
+	return kindIDsByName, nil
+}
```
sirisjo: **thought**: I'm wondering if there's any way to build this map while also calculating the counts themselves? Then you wouldn't have to loop back through the `countsByEnvironmentID` over again.

cmd/api/src/services/dataquality/dataquality.go
```diff
@@ -493,5 +497,235 @@ func SaveDataQuality(ctx context.Context, db database.DataQualityData, graphDB g
...
+			return fmt.Errorf("could not get open graph stats: %w", err)
+		} else {
+			if len(stats) == 0 {
+				return nil
+			}
+
+			if _, err := db.CreateDataQualityStats(ctx, stats); err != nil {
+				return fmt.Errorf("could not save open graph stats: %w", err)
+			} else if _, err := db.CreateDataQualityAggregations(ctx, aggregations); err != nil {
+				return fmt.Errorf("could not save open graph aggregations: %w", err)
```
sirisjo, line 511: **thought**: would it be beneficial to wrap these in a transaction, so that if the aggregation write fails, they both fail? Or is it ok to only write to one table?

sirisjo:
**question**: Are we going to do relationship counts in a separate ticket?

sirisjo (approved):
Comments are nits, tested this out locally, LGTM!

cmd/api/src/services/dataquality/dataquality.go
mvlipka: Yeah this should be fine, since one isn't dependent on the other existing

mvlipka:
@sirisjo great suggestions! I completely blanked on getting relationships in this ticket, good catch
I've gotten the changes in and I've also moved the opengraph logic into a new `opengraph.go` file to simplify the existing `dataquality.go` file

sirisjo:
@mvlipka aside from the strange failing test (looks unrelated), these new changes look good to go! I am seeing the node & relationship counts populate properly in my local DB 💯

PR #2998, opened 2026-07-16
fix: fix selectnodes BED-8284

bsheth711:
## Description

This PR fixes a bug with SelectNodes where we previously used the old Selector Node's Source when updating certification status. Additionally, we did not update Selector Node sources. This meant if the Source for a selected node was different than its existing Selector Node, we did not certify the Selector Node as expected. 

The PR also refactors SelectNode to compartmentalize certification logic and when we should update a Selector Node and batched DB operations. 

## Motivation and Context

Resolves [BED-8284](https://specterops.atlassian.net/browse/BED-8284)

## How Has This Been Tested?

This has been tested with through incremental changes and multiple rounds of testing. Additionally, extensive tests have been added for SelectNodes as well as the new batch database functions. **This has been manually tested with the affected dataset as well.**

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-8284]: https://specterops.atlassian.net/browse/BED-8284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Performance**
  * Improved selector-node processing by handling inserts, updates, and deletions in batches.
  * Reduced unnecessary database updates when selected node data has not changed.

* **Behavior**
  * Preserved manual and revoked certifications during selector reconciliation.
  * Improved handling of automatic certification changes and stale or removed selected nodes.

* **Tests**
  * Added comprehensive coverage for batch operations, certification history, batching limits, and selector-node lifecycle scenarios.


packages/go/analysis/agt.go
```diff
@@ -559,13 +559,76 @@ func fetchOldSelectedNodes(ctx context.Context, db database.Database, selectorId
...
+}
+
+func shouldUpdateSelectorNode(oldNode model.AssetGroupSelectorNode, newNode model.AssetGroupSelectorNode) bool {
+	if oldNode.Certified != newNode.Certified ||
+		oldNode.NodeName != newNode.NodeName ||
+		oldNode.NodePrimaryKind != newNode.NodePrimaryKind ||
+		oldNode.NodeEnvironmentId != newNode.NodeEnvironmentId ||
+		oldNode.NodeObjectId != newNode.NodeObjectId ||
+		oldNode.Source != newNode.Source {
+		return true
```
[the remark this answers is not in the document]
bsheth711: I'm going to stick with previous behavior here. We don't really care if CertifiedBy alone changes, and it shouldn't without the Certified status changing anyways.

cweidenkeller (approved):

cmd/api/src/database/assetgrouptags.go
```diff
@@ -729,32 +731,377 @@ func (s *BloodhoundDB) GetAssetGroupTagForSelection(ctx context.Context) ([]mode
...
-	return s.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
-		if result := tx.WithContext(ctx).Exec(fmt.Sprintf("INSERT INTO %s (selector_id, node_id, certified, certified_by, source, node_primary_kind, node_environment_id, node_object_id, node_name, created_at, updated_at) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, current_timestamp, current_timestamp) ON CONFLICT DO NOTHING", model.AssetGroupSelectorNode{}.TableName()), selectorId, nodeId, certified, certifiedBy, source, primaryKind, environmentId, objectId, displayName); result.Error != nil {
-			return CheckError(result)
-		} else if certified != model.AssetGroupCertificationPending {
-			bhdb := NewBloodhoundDB(tx, s.pool, s.idResolver, s.config)
-			return bhdb.CreateAssetGroupHistoryRecord(ctx, model.AssetGroupActorBloodHound, "", displayName, model.ToAssetGroupHistoryActionFromAssetGroupCertification(certified), assetGroupTagId, null.String{}, null.String{})
+func (s *BloodhoundDB) InsertSelectorNodes(ctx context.Context, nodes []model.AssetGroupSelectorNode) error {
+	const selectorNodeInsertBatchSize = 5000
```
mistahj67: What are your thoughts on pulling this into a single variable `selectorNodeBatchSize` at the top of this file? All are set to 5000 currently? Can differentiate if need be in the future?

cmd/api/src/database/assetgrouptags.go
```diff
@@ -729,32 +731,377 @@ func (s *BloodhoundDB) GetAssetGroupTagForSelection(ctx context.Context) ([]mode
...
+
+	var (
+		err             error
+		transaction     pgx.Tx
+		batchUpperBound int
```
mistahj67: Is there any looping concerns that the batchUpperBound is scoped outside of the for loop? Typically I'd err on the side of keeping it defined inside the for loop scope to ensure no looping shenanigans

cmd/api/src/database/assetgrouptags.go
```diff
@@ -729,32 +731,377 @@ func (s *BloodhoundDB) GetAssetGroupTagForSelection(ctx context.Context) ([]mode
...
+			return err
 		}
+	}
+
+	if err = transaction.Commit(ctx); err != nil {
+		return fmt.Errorf("committing selector node batch insert transaction: %w", err)
+	}
+
+	return nil
+}
```
mistahj67: suggestion:

All three operations are now identical in this logic and the operation signature also matches `(ctx context.Context, nodes []model.AssetGroupSelectorNode) error`

What do you think of abstracting this to a generic wrapper function that you then supply the operation innerFn to? 

```go
func (s *BloodhoundDB) SelectorNodesBatchWrite(ctx context.Context, op func (ctx context.Context, nodes []model.AssetGroupSelectorNode) error, nodes []model.AssetGroupSelectorNode) error {
	return nil
}
```

If you opt for this route, then my comment about moving the batch size can be disregarded as this will be the sole use case and it can be just defined in the above function.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -729,32 +731,377 @@ func (s *BloodhoundDB) GetAssetGroupTagForSelection(ctx context.Context) ([]mode
...
+
+type selectorNodeCertificationChange struct {
+	Target                  string `db:"target"`
+	AssetGroupCertification int32  `db:"asset_group_certification"`
+	AssetGroupTagId         int32  `db:"asset_group_tag_id"`
+}
+
+func (s selectorNodeCertificationChange) historyAction() model.AssetGroupHistoryAction {
+	return model.ToAssetGroupHistoryActionFromAssetGroupCertification(model.AssetGroupCertification(s.AssetGroupCertification))
+}
```
mistahj67: style nit: this is used in 1 spot much farther down, why not just inline this in the 1 spot it needs to happen vs creating a method requiring one to jump around to grok.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -729,32 +731,377 @@ func (s *BloodhoundDB) GetAssetGroupTagForSelection(ctx context.Context) ([]mode
...
+	)
+
+	if len(certificationChanges) == 0 {
 		return nil
-	})
+	}
+
+	for _, certificationChange := range certificationChanges {
+		assetGroupTagIds = append(assetGroupTagIds, certificationChange.AssetGroupTagId)
+		actions = append(actions, string(certificationChange.historyAction()))
```
mistahj67: nit: see above comment about just inlining this method

cmd/api/src/database/assetgrouptags.go
bsheth711: I like this idea

mistahj67 (approved):
Code LGTM. Performance boost is 🧑‍🍳 🤌 📈 Added testing 👍 

Verified that the bug was addressed as well and thank you for considering my feedback and patience with this one! 

🚀

PR #3010, opened 2026-07-20
feat: Add New GET Node Kind Endpoint - BED-8756

LawsonWillard:
## Description

- Adds new GET node kind endpoint to return node kind and entity panel information
- Adds unit, integration and e2e tests for the new endpoint
- Adds endpoint to the existing OpenAPI spec 

## Motivation and Context
Resolves: BED-8765

As a BloodHound user I want an endpoint that allows me to pull information for a specific node kind by its node kind identifier.

## How Has This Been Tested?

Adds new unit, integration and e2e tests for the endpoint
Manually tested against a local BH instance to ensure the endpoint is functioning as expected 

## Types of changes:

## Checklist:
  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Added an experimental endpoint for retrieving node kind details (including metadata, styling, extension info, and associated markdown content) under the `/api/v2/node-kinds/{node_kind_id}` path.
  * Updated the OpenAPI specification to document the new endpoint and related response/schema shapes.
* **Bug Fixes**
  * Node kind “position” values can now start at `0`.
  * If markdown content is malformed in node/relationship kind info, responses still succeed while reporting issues in logs.
* **Tests**
  * Added unit, integration, and end-to-end coverage for the new endpoint and supporting data/response handling.


server/extensions/internal/handlers/nodekind.go
```diff
@@ -0,0 +1,141 @@
...
+	} else if nodeKind, err := s.extensions.GetNodeKind(ctx, int32(id)); errors.Is(err, services.ErrNodeKindNotFound) {
+		responses.WriteError(ctx, http.StatusNotFound, "node kind not found", response)
+	} else if err != nil {
+		responses.WriteInternalServerError(ctx, err, response)
+	} else if nodeKindView, markdownErr := BuildNodeKindView(nodeKind); markdownErr != nil {
+		slog.WarnContext(ctx, "Failed to parse node kind info markdown content", attr.Error(markdownErr))
+		responses.WriteBasic(ctx, nodeKindView, http.StatusOK, response)
+	} else {
+		responses.WriteBasic(ctx, nodeKindView, http.StatusOK, response)
+	}
```
mistahj67, line 155: You might consider wrapping the nodeKind in an field as well that way if you ever need to extend this outside of the `node-kinds` you have that extensibility so something like:

```json
{
   "data": {
      "node_kind": { 
          nodeKindViewObj here
        }
     }
}
```

server/extensions/internal/routes/routes.go
```diff
@@ -0,0 +1,40 @@
...
+
+// Register attaches the extensions endpoints to the given router instance.
+//
+// Rate limiting is applied first (outermost middleware layer) so that all
+// requests — including unauthenticated ones — count against the per-IP limit
+// and can be rejected with 429 before the permissions check runs.
+func Register(routerInst *router.Router, handlerSet *handlers.Handlers, rateLimit func() mux.MiddlewareFunc) {
+	var (
+		permissions   = auth.Permissions()
+		nodeKindRoute = routerInst.GET(fmt.Sprintf("/api/v2/node-kinds/{%s}", handlers.URIPathVariableNodeKindID), handlerSet.GetNodeKindByID)
```
mistahj67, line 35: typically it's preferred to match the route with the resource it's getting, though this seems to be a bit of a "view" endpoint combining multiple tables. There's already an `extension-edges` precedence and this is limited to just the schema_nodes and not any other nodes which the generic node-kinds implies I think.

What do you think of naming this `api/v2/extension-nodes/{%s}` instead?

```suggestion
		nodeKindRoute = routerInst.GET(fmt.Sprintf("/api/v2/extension-nodes/{%s}", handlers.URIPathVariableNodeKindID), handlerSet.GetNodeKindByID)
```

server/extensions/internal/handlers/nodekind.go
LawsonWillard: Imo this is desirable but the other endpoints that return similar data for entity panels dont conform to this yet. All are marked as experimental and subject to change, so if we want we can update them to do so. I'll get a consensus to see if we want to update all the new endpoints to conform to this.

server/extensions/internal/routes/routes.go
LawsonWillard: I originally suggested `/extensions/node-kinds` for the endpoint which was rejected as I believe the intention is to separate node kinds from extensions here. Long term I think we want it to return information for all nodes as we will eventually be adding extensionless node kinds to the `schema_node_kind` table. 

I think we are also looking at deprecating the `extensions-edges` in the nearish future. 

This is a loose recollection from a refinement session so might be incorrect here. Let me confirm

brandonshearin:
when manual testing through the swagger docs --
1. i requested the endpoint for node_kind_2, which corresponds in my local data to the AD User kind
2. no extension information was populated.


I think the AD extension should be populated yeah? i dont see the extension field in the example response either.

LawsonWillard:
@brandonshearin yeah, currently its not returning extension information which it should. Trying to determine if/what the migration of extensionless node-kinds to the schema_node_kind table looks like and how itll affect this endpoint.

server/extensions/internal/appdb/kindinfo.go
```diff
@@ -0,0 +1,101 @@
...
+		Position:           row.Position,
+		Content:            row.Content,
+		CreatedAt:          row.CreatedAt.ValueOrZero(),
+		UpdatedAt:          row.UpdatedAt.ValueOrZero(),
+	}
+}
+
+// GetKindInfosByNodeKindID returns all KindInfo's associated with the given node kind id,
+// ordered by position then title. An empty slice is returned when no rows match.
+func (s *Store) GetKindInfosByNodeKindID(ctx context.Context, nodeKindID int32) ([]services.KindInfo, error) {
```
brandonshearin: in the graphdb onion, this method is implemented with a kindName rather than a nodeKindID parameter. this way, the method can be used by GetNodeKind() AND GetRelationshipKind(). what are your thoughts on making that change?

server/extensions/internal/handlers/nodekind.go
```diff
@@ -0,0 +1,141 @@
...
+// GetNodeKindByID returns the details of a node kind identified by its schema_node_kinds
+// row id. Returns 400 when the id is malformed, 404 when the node kind cannot be found,
+// and 200 with the node kind details otherwise.
+func (s Handlers) GetNodeKindByID(response http.ResponseWriter, request *http.Request) {
+	var (
+		ctx = request.Context()
+		raw = mux.Vars(request)[URIPathVariableNodeKindID]
+	)
+
+	if id, err := strconv.ParseInt(raw, 10, 32); err != nil {
```
brandonshearin, line 139: i think it might be better to explicitly return after each call to responses.Write* function.  that way, if this handler ever changes and the branching logic changes, we will still be returning immediately after writing an http response like we would expect.

server/extensions/internal/handlers/nodekind.go
```diff
@@ -0,0 +1,141 @@
...
+		responses.WriteError(ctx, http.StatusBadRequest, "node kind id is malformed", response)
+	} else if nodeKind, err := s.extensions.GetNodeKind(ctx, int32(id)); errors.Is(err, services.ErrNodeKindNotFound) {
+		responses.WriteError(ctx, http.StatusNotFound, "node kind not found", response)
+	} else if err != nil {
+		responses.WriteInternalServerError(ctx, err, response)
+	} else if nodeKindView, markdownErr := BuildNodeKindView(nodeKind); markdownErr != nil {
+		slog.WarnContext(ctx, "Failed to parse node kind info markdown content", attr.Error(markdownErr))
```
brandonshearin, line 149: im not sure we should be returning a 200 code here when the markdown fails. what do you think?

server/extensions/internal/appdb/kindinfo.go
LawsonWillard: Ahhh right right, good call out! I can update it to conform to that.

server/extensions/internal/handlers/nodekind.go
LawsonWillard: Thought the same at first but this mirrors the same logic found in the Get relationship/node endpoints in the graphdb onion so trying to keep the same functionality as that. 

With the new template rendering validation about to be picked up, this shouldnt really be an issue but i dont think we want to fail everything if one of the markdowns fails to render. Its less of an application failure and more of a extension issue which is where that should be resolved.

server/extensions/internal/handlers/nodekind.go
LawsonWillard: Checked in with the team, research and product already vetted the API response shape and dont feel its necessary/want the wrapped response so im going to leave it as is for now. With the experimental tag we can always come back later and add it.

server/extensions/internal/handlers/nodekind.go
LawsonWillard: Typically I prefer explicit over implicit but here I dont think it matters as much, unit tests should fully cover the endpoint to ensure any changes still return the expected output. That said I can add the return statements to make it more explicit.

server/extensions/internal/services/nodekind_test.go
```diff
@@ -0,0 +1,115 @@
...
+			wantErrContains: "fetching extension 7 for node kind 42",
+		},
+		{
+			name: "success_-_attaches_empty_extension_when_extension_not_found",
+			setupMock: func(databaseMock *mocks.MockDatabase) {
+				databaseMock.EXPECT().GetNodeKind(ctx, nodeKindID).Return(baseNodeKind, nil)
+				databaseMock.EXPECT().GetKindInfos(ctx, "User").Return(infos, nil)
+				databaseMock.EXPECT().GetExtension(ctx, extID).Return(services.Extension{}, services.ErrExtensionNotFound)
+			},
+			wantResult: services.NodeKind{ID: nodeKindID, Name: "User", DisplayName: "User", SchemaExtensionID: extID, Info: infos, Extension: services.Extension{}},
```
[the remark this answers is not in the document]
LawsonWillard: This is intended as we will be making the schema_node_kind's extension id fk nullable to allow for node kinds that are not attached to an extension to be added to the table

server/extensions/internal/handlers/nodekind.go
```diff
@@ -0,0 +1,152 @@
...
+// NodeKindView is the JSON shape returned by the node kind handlers.
+type NodeKindView struct {
+	NodeKindID    int32                   `json:"node_kind_id"`
+	Name          string                  `json:"name"`
+	DisplayName   string                  `json:"display_name"`
+	Description   string                  `json:"description"`
+	IsDisplayKind bool                    `json:"is_display_kind"`
+	Icon          string                  `json:"icon"`
+	Color         string                  `json:"color"`
+	Extension     ExtensionDetailsView    `json:"extension"`
```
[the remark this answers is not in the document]
LawsonWillard: Like below, this is intended as we will be making the schema_node_kind's extension id fk nullable to allow for node kinds that are not attached to an extension to be added to the table

server/extensions/internal/services/nodekind.go
```diff
@@ -41,17 +42,25 @@ type NodeKind struct {
...
+
 	if nodeKind, err := s.db.GetNodeKind(ctx, id); err != nil {
 		return NodeKind{}, err
-	} else if infos, err := s.db.GetKindInfosByNodeKindID(ctx, id); err != nil {
-		return NodeKind{}, fmt.Errorf("fetching kind infos for node kind %d: %w", id, err)
+	} else if nodeKind.Info, err = s.db.GetKindInfos(ctx, nodeKind.Name); err != nil {
+		return NodeKind{}, fmt.Errorf("fetching kind infos for node kind %s: %w", nodeKind.Name, err)
 	} else {
-		nodeKind.Info = infos
+		if nodeKind.SchemaExtensionID != 0 {
```
brandonshearin, line 59: Help me understand this defensive guard --

Anything in the schema_node_kinds table will be owned by an extension. The only way for a record to get into this table is to reference a valid extension ID. Why is this defensive guard necessary?

server/extensions/internal/handlers/nodekind.go
```diff
@@ -128,14 +135,18 @@ func (s Handlers) GetNodeKindByID(response http.ResponseWriter, request *http.Re
 
 	if id, err := strconv.ParseInt(raw, 10, 32); err != nil {
 		responses.WriteError(ctx, http.StatusBadRequest, "node kind id is malformed", response)
+		return
 	} else if nodeKind, err := s.extensions.GetNodeKind(ctx, int32(id)); errors.Is(err, services.ErrNodeKindNotFound) {
 		responses.WriteError(ctx, http.StatusNotFound, "node kind not found", response)
+		return
 	} else if err != nil {
 		responses.WriteInternalServerError(ctx, err, response)
```
brandonshearin, line 146: missed a return here

server/extensions/internal/handlers/nodekind.go
```diff
@@ -44,6 +43,7 @@ type NodeKindView struct {
 	IsDisplayKind bool                    `json:"is_display_kind"`
 	Icon          string                  `json:"icon"`
 	Color         string                  `json:"color"`
+	Extension     ExtensionDetailsView    `json:"extension"`
```
brandonshearin: if an extension doesn't exist for a node kind, we want the API response to look like:
```
{
  name: "User",
  extension: null,
  // ... other fields
}
```

specifically, the extension field should take a null value, and i think rabbit is saying that in this "empty" response scenario , the current implementation  is sending back an extension object with zero values for all fields.

server/extensions/internal/services/nodekind.go
LawsonWillard: For now each node kind must always have an extension, however, I think the plan is to make the extension id fk column nullable so it can support schemaless node kinds. This way we can get rid of the custom node kind table and consolidate it all to the schema node kinds. 

This guard preempts that, no need for an extra db look up if the schema extension id is 0

server/extensions/internal/handlers/nodekind.go
LawsonWillard: I thought we were planning on returning the 0 value fields, but I can update it to return null!

brandonshearin (commented):
looks great mr willard.

brandonshearin (approved):
ship it!

PR #3013, opened 2026-07-20
feat: improve node kind display labels BED-9025

justin-prime1:
## Description

Updates node type labels in Privilege Zones and related asset group UI surfaces to use human-readable names instead of raw graph kind values.

The display labels now come from `schema_node_kinds.display_name` via `/api/v2/graphs/kinds`, rather than a frontend-only hardcoded map. Raw graph kind values are still preserved for icons, test IDs, filtering, sorting inputs, and API calls.

## Motivation and Context

Resolves https://specterops.atlassian.net/browse/BED-9025

Raw graph kind labels like `AZApp`, `AZServicePrincipal`, `CertTemplate`, and `AIACA` are difficult for first-time users to understand. Moving the labels into schema metadata gives the UI a shared backend-backed source of truth while keeping raw kinds available for graph/query behavior.

## Implementation Notes

- Adds `node_kinds` metadata to `/api/v2/graphs/kinds`
- Adds a migration to update built-in `schema_node_kinds.display_name` values
- Updates OpenAPI and JS client response types
- Updates Asset Group and Privilege Zone node-kind displays to consume graph kind metadata
- Keeps unknown/custom node kinds falling back to their raw value

## How Has This Been Tested?

- `just prepare-for-codereview`
- `go test ./cmd/api/src/api/v2 -run TestResources_ListKinds`
- `yarn workspace bh-shared-ui test src/utils/nodeKindDisplay.test.ts src/components/AssetGroupEdit/AssetGroupEdit.test.tsx src/components/AssetGroupFilters/AssetGroupFilters.test.tsx src/views/PrivilegeZones/Details/ObjectCountPanel.test.tsx src/views/PrivilegeZones/Details/ObjectsAccordion.tag.test.tsx`
- `yarn workspace bh-shared-ui check-types`
- `yarn workspace js-client-library build`
- `git diff --check`

## Types of changes

## Checklist:

  - Associated ticket: https://specterops.atlassian.net/browse/BED-9025
  - OpenAPI docs were updated
  - Relevant tests passed locally

### DIFF
--- a/packages/javascript/bh-shared-ui/src/utils/nodeKindDisplay.ts
+++ b/packages/javascript/bh-shared-ui/src/utils/nodeKindDisplay.ts
@@ -14,46 +14,18 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { ActiveDirectoryNodeKind, AzureNodeKind } from '../graphSchema';
+import type { GraphKindsResponse } from 'js-client-library';
 
-const NODE_KIND_DISPLAY_LABELS: Record<string, string> = {
-    [ActiveDirectoryNodeKind.Entity]: 'Entity',
-    [ActiveDirectoryNodeKind.User]: 'User',
-    [ActiveDirectoryNodeKind.Computer]: 'Computer',
-    [ActiveDirectoryNodeKind.Group]: 'Group',
-    [ActiveDirectoryNodeKind.GPO]: 'Group Policy Object',
-    [ActiveDirectoryNodeKind.OU]: 'Organizational Unit',
-    [ActiveDirectoryNodeKind.Container]: 'Container',
-    [ActiveDirectoryNodeKind.Domain]: 'Domain',
-    [ActiveDirectoryNodeKind.LocalGroup]: 'AD Local Group',
-    [ActiveDirectoryNodeKind.LocalUser]: 'AD Local User',
-    [ActiveDirectoryNodeKind.AIACA]: 'AIA Certificate Authority',
-    [ActiveDirectoryNodeKind.RootCA]: 'Root Certificate Authority',
-    [ActiveDirectoryNodeKind.EnterpriseCA]: 'Enterprise Certificate Authority',
-    [ActiveDirectoryNodeKind.NTAuthStore]: 'NTAuth Store',
-    [ActiveDirectoryNodeKind.CertTemplate]: 'Certificate Template',
-    [ActiveDirectoryNodeKind.IssuancePolicy]: 'Issuance Policy',
-    [AzureNodeKind.Entity]: 'Azure Entity',
-    [AzureNodeKind.VMScaleSet]: 'Azure VM Scale Set',
-    [AzureNodeKind.App]: 'Azure Application',
-    [AzureNodeKind.Role]: 'Azure Role',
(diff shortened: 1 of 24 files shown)


justin-prime1:
Local verification update:

- `yarn lint` passes
- `cmd/ui && yarn build` passes
- Focused Privilege Zones tests pass:
  - `yarn test src/views/PrivilegeZones/Details/RulesAccordion.test.tsx src/views/PrivilegeZones/Details/ObjectsAccordion.tag.test.tsx`
  - previously ran the broader touched shared-ui test set as listed in the PR body

The `build-ui` check appears to be failing during `yarn audit`, not during lint or build. Local `yarn audit` currently reports existing dependency advisories in `axios`, `brace-expansion`, `js-yaml`, and `tar`/`node-gyp`, which look unrelated to this UI label change.

The `run-go-integration-tests` failure also appears unrelated to this frontend-only change, but I do not have access to the hosted job logs from this environment.

justin-prime1:
Follow-up update:

- Restored the missing `focusedOptionIconStyles` shared-ui export expected by the parent BHE UI.
- Fixed the unrelated Explore > Cypher > Saved Queries header color so it uses normal contrast text instead of inheriting purple from the text button variant.
- Added a focused test assertion for the Saved Queries header contrast class.
- Verified the isolated local app renders successfully and Saved Queries computes to normal contrast text.

packages/javascript/bh-shared-ui/src/utils/nodeKindDisplay.ts
```diff
@@ -0,0 +1,59 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+import { ActiveDirectoryNodeKind, AzureNodeKind } from '../graphSchema';
+
+const NODE_KIND_DISPLAY_LABELS: Record<string, string> = {
```
mistahj67: While this would definitely work, hardcoding a map unfortunately creates maintainability tech debt and something that risks falling out of sync rather quickly. Coupling that with OG work and it gets a bit worse.

But there is an already created solution that answers all of that, just have to decide on how the UI should receive it. The schema-node-kinds table has a `display_name` column, that would allow for dynamic / maintainable and OG-safe friendly names for kinds. 

I originally had planned on surfacing that column in the `/kinds` endpoint for use by the UI when I added filtering to that endpoint. But it was decided to wait to determine if a separate endpoint is a better fit. Whatever is decided, I have some backend changes that are stashed that might help!

mistahj67 (changes requested):
Thanks for taking a crack at this, you are not the first one 😅 

I set to blocking because I think this may need an API ticket / work and then the UI side can consume from there instead. Let me know which direction to take and I can assist!

packages/javascript/bh-shared-ui/src/utils/nodeKindDisplay.ts
urangel: Agree with pivoting to using display names from the schema_node_kinds table. We will need a migration to update the display names of AD/Azure if we want them to align with the display names chosen in this changeset.

PR #3029, opened 2026-07-22
chore: Support Bundle Type Updates (BED-7971)

bfaulk96:
### This is just a PR into another feature branch, not in need of serious review*

- **chore: update support bundle summary type to match OpenAPI schema**

## Description

Updated types to reflect what already exists in our OpenAPI schemas for client support bundle summary

## Motivation and Context

Used in: BED-7971: 

This is all for in-progress work, so these changes do not currently affect anything (also, this is a PR into a feature branch)

## Types of changes

## Summary by CodeRabbit

* **Refactor**
  * Updated support-bundle status reporting with a consolidated operation model.
  * Renamed the client support-bundle field to `support_bundle_summary`.
  * Added standardized operation states, including queued, running, succeeded, failed, and canceled.
  * Support-bundle summaries now include current and most recently completed operation details.


packages/javascript/js-client-library/src/responses.ts
```diff
@@ -345,28 +345,36 @@ export type GetExportQueryResponse = AxiosResponse<Blob>;
...
 export type GetClientResponse = PaginatedResponse<Client[]>;
 
-export type SupportBundleOperationStatus = 'queued' | 'running' | 'succeeded' | 'failed' | 'cancelled';
+export enum SupportBundleOperationStatus {
+    QUEUED = 'queued',
+    RUNNING = 'running',
+    SUCCEEDED = 'succeeded',
+    FAILED = 'failed',
+    CANCELLED = 'cancelled',
+}
```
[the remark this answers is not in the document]
bfaulk96: @catsiller Where does the team tend to stand on this? I don't feel strongly either way, so happy to update to this suggestion if you agree it's preferred.

packages/javascript/js-client-library/src/responses.ts
```diff
@@ -345,28 +345,36 @@ export type GetExportQueryResponse = AxiosResponse<Blob>;
...
+    created_at: string;
+    started_at: string | null;
+    completed_at: string | null;
+    execution_time: string;
 };
 
-export type SupportBundleStatus = {
-    latest_completed: SupportBundleCompletedOperation | null;
-    active_operation: SupportBundleActiveOperation | null;
+export type SupportBundleSummary = {
```
catsiller: thoughts on rename to `SupportBundleSummaryStatus`? getting a little long but maybe a little bit more specific

packages/javascript/js-client-library/src/responses.ts
```diff
@@ -345,28 +345,36 @@ export type GetExportQueryResponse = AxiosResponse<Blob>;
...
 
-export type SupportBundleOperationStatus = 'queued' | 'running' | 'succeeded' | 'failed' | 'cancelled';
+export enum ManagementOperationStatus {
+    QUEUED = 'queued',
+    RUNNING = 'running',
+    SUCCEEDED = 'succeeded',
+    FAILED = 'failed',
+    CANCELED = 'canceled',
+}
 export type SupportBundleArtifactStatus = 'pending' | 'complete' | 'failed' | null;
```
catsiller: ohhh wondering if we should do something similar here as we did for export enum ManagementOperationStatus {...}

packages/javascript/js-client-library/src/responses.ts
```diff
@@ -345,28 +345,36 @@ export type GetExportQueryResponse = AxiosResponse<Blob>;
 
 export type GetClientResponse = PaginatedResponse<Client[]>;
 
-export type SupportBundleOperationStatus = 'queued' | 'running' | 'succeeded' | 'failed' | 'cancelled';
+export enum ManagementOperationStatus {
+    QUEUED = 'queued',
+    RUNNING = 'running',
+    SUCCEEDED = 'succeeded',
+    FAILED = 'failed',
+    CANCELED = 'canceled',
```
catsiller, line 353: are we updating `cancelled` to `canceled`?

catsiller: ps - really like this type update to the block

packages/javascript/js-client-library/src/responses.ts
bfaulk96: @catsiller We're actually not using this type at all right now. I left it in because I wasn't sure if it was something that would be used somewhere when the backend is complete. We can just delete it for now and add something back in if necessary?

packages/javascript/js-client-library/src/responses.ts
bfaulk96: > are we updating `cancelled` to `canceled`?

Yep! The OpenAPI schema has `canceled` with one `l`, so I wanted to make sure that those lined up appropriately. Ahh, gotta love English 😅

packages/javascript/js-client-library/src/responses.ts
bfaulk96: So I updated this based on the attribute name on the client object being `support_bundle_summary` - I'm open to renaming it, though!

packages/javascript/js-client-library/src/responses.ts
catsiller: okay to remove

catsiller (approved):

PR #3049, opened 2026-07-24
feat: audit log middleware BED-8595

stephanieslamb:
## Description

This PR is a basic level POC for adding an audit log middleware. 
Current functionality includes: 
- Audit logs for all methods
- If audit intent fails, the request also fails

Improvements:
- `fields` are not populated.  The middleware currently sets it as empty
- Retention is hardcoded to 3 months
- Legacy clean up

## Motivation and Context

Resolves BED-8595

This change is needed to move audit logs out of transactions in the DB layer into a middleware that is called with every request. 

## How Has This Been Tested?

Unit and integration tests have been added.  

## Screenshots (optional):
Screenshot 2026-08-19 at 10 24 16 AM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## New Features
- Added comprehensive API audit logging with request details, actors, request IDs, and success or failure outcomes.
- Unauthenticated requests are attributed anonymously, with sensitive values automatically redacted.

## Performance & Maintenance
- Added monthly audit-log partitioning, retention cleanup, safe migration backfills, and record preservation.

## Reliability
- Requests fail safely when audit recording cannot start; later audit-write failures do not alter responses.
- Handler panics record failure outcomes before being propagated.
- Improved response streaming and flushing through compression and logging middleware.


cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+        FROM pg_partitioned_table pt
+        JOIN pg_class c ON c.oid = pt.partrelid
+        WHERE c.relname = 'audit_logs' AND pt.partstrat = 'r'
+    ) THEN
+        RETURN;
+    END IF;
+
+    DROP TABLE IF EXISTS audit_logs_partitioned;
+    CREATE TABLE audit_logs_partitioned (
+        id                  BIGINT NOT NULL DEFAULT nextval('audit_logs_id_seq'),
```
mvlipka, line 97: Nit: `BIGSERIAL` works here

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+        created_at          TIMESTAMPTZ NOT NULL DEFAULT now(),
+        action              TEXT NOT NULL,
+        actor_id            TEXT,
+        actor_name          TEXT,
+        actor_email         VARCHAR(330) DEFAULT NULL::character varying,
+        request_id          TEXT,
+        source_ip_address   TEXT,
+        status              VARCHAR(15) DEFAULT 'intent' CHECK (status IN ('intent', 'success', 'failure')),
```
mvlipka: I think an `enum` would make sense here, to me they're a bit easier to maintain than constraints going forward

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+        commit_id           TEXT,
```
mvlipka, line 106: I know are `commit_id` is currently `TEXT`, but is there appetite for converting it to `UUID` while moving to the new table?

cmd/api/src/api/middleware/audit.go
```diff
@@ -0,0 +1,154 @@
...
+			slog.ErrorContext(outcomeCtx, "Failed to write audit failure row", attr.Error(failureErr))
+		}
+	} else if successErr := auditService.Success(outcomeCtx, commitID, entry); successErr != nil {
+		slog.ErrorContext(outcomeCtx, "Failed to write audit success row", attr.Error(successErr))
+	}
+}
+
+// anonymousActorName is the actor name for unauthenticated requests, keeping
+// them attributed (by source IP) rather than dropped.
+const anonymousActorName = "anonymous"
```
mvlipka: Maybe a nit: anonymous to me indicates that a user does not wish to let their identity be known
`unknown` may be better wording here? If the user is unauthenticated, we simply don't know who they were

cmd/api/src/api/middleware/audit_test.go
```diff
@@ -0,0 +1,370 @@
...
+	"github.com/specterops/bloodhound/cmd/api/src/database/types/null"
+	"github.com/specterops/bloodhound/cmd/api/src/model"
+	"github.com/specterops/bloodhound/server/audit"
+	"github.com/stretchr/testify/require"
+)
+
+// fakeAuditService is a hand-rolled test double for the middleware.AuditService
+// port. It records every call and can be configured to return errors so the
+// best-effort behavior of the middleware can be exercised.
+type fakeAuditService struct {
```
mvlipka: I noticed we started calling mock services `fake`
We do it in a lot of the new architecture, so not requesting a change here, but curious about the move from `mock*` to `fake*` as I don't see it in the ADR

cmd/api/src/daemons/gc/data_pruning.go
```diff
@@ -18,22 +18,35 @@ package gc
...
 	"time"
 
 	"github.com/specterops/bloodhound/cmd/api/src/database"
+	"github.com/specterops/bloodhound/packages/go/bhlog/attr"
+	"github.com/specterops/bloodhound/server/audit"
 )
 
+// defaultAuditRetentionMonths bounds how many months of audit_logs partitions
+// are retained. Partitions whose entire range is older than this window are
+// dropped. TODO(audit retention): source this from appcfg once the
```
mvlipka: Can we get the TODO defined on a separate line so it stands out more and IDEs can catalog it

mvlipka (commented):
I just have some nits, not necessarily requesting changes as these don't affect functionality

I think some of the LOC can be reduced by switching to table driven tests and removing these test-only functions

Lots of very wordy comments, I don't really have a suggestion on that, but something to be aware of. Especially in the tests

cmd/api/src/api/middleware/audit.go
```diff
@@ -0,0 +1,154 @@
...
+
+// AuditMiddleware records the intent/success/failure lifecycle of every API
+// request. It writes an intent row before the handler runs and a success or
+// failure row afterward based on the response status; a failed intent write
+// rejects the request with a 500. Route templates in excludedRoutes are skipped.
+func AuditMiddleware(auditService AuditService, muxRouter *mux.Router, excludedRoutes ...string) mux.MiddlewareFunc {
+	exclusions := make(map[string]bool, len(excludedRoutes))
+	for _, route := range excludedRoutes {
+		exclusions[route] = true
+	}
```
mistahj67: not for this round, but it might be nice to have a similar mechanism to the permissions, FF, param parsing, whereby devs can tack onto the endpoint itself a way to .SkipAudit() or whatever. Not sure if that's even doable.

cmd/api/src/api/middleware/audit.go
mistahj67: I concur, `unknown` has been used in the past for this scenario and has precedence.

mistahj67: Should this be fallback in the datalayer instead given an empty actor? Would allow for less onus on the callers to know about this edge case.

cmd/api/src/api/middleware/audit_test.go
mistahj67: nice catch, would prefer being consistent here please

cmd/api/src/api/middleware/audit_test.go
```diff
@@ -0,0 +1,370 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package middleware_test
```
mistahj67, line 17: I believe the goal is to move towards more table driven tests, if there's time to convert these and the other test files in this change set, I think that'd be a nice improvement

cmd/api/src/api/middleware/logging.go
```diff
@@ -97,6 +99,43 @@ func (s *responseRecorder) WriteHeader(statusCode int) {
...
+}
+
+// Hijack forwards to the delegate when it supports http.Hijacker so that
+// connection upgrades (e.g. websockets) still work through the recorder.
+func (s *responseRecorder) Hijack() (net.Conn, *bufio.ReadWriter, error) {
+	if hijacker, ok := s.delegate.(http.Hijacker); ok {
+		return hijacker.Hijack()
+	}
+	return nil, nil, http.ErrNotSupported
+}
```
mistahj67: I'm less familiar with this method / use case, but if it's not strictly needed for the audit logs, i'd not add more than necessary if possible. same for the flusher above.

cmd/api/src/daemons/gc/data_pruning.go
```diff
@@ -18,22 +18,35 @@ package gc
...
+// retention_months parameter lands (Step I) instead of the constant.
+const defaultAuditRetentionMonths = 3
+
 // Daemon holds data relevant to the data daemon
 type Daemon struct {
-	exitC chan struct{}
-	db    database.Database
+	exitC           chan struct{}
```
mistahj67: This data pruning daemon does not follow the best practices W.R.T. daemon orchestration and the stop channels. Consider fixing that in this PR. 

The broken window is to "Separate the stop signal from daemon completion."

cmd/api/src/daemons/gc/data_pruning.go
```diff
@@ -49,23 +62,43 @@ func (s *Daemon) Start(ctx context.Context) {
...
+// partitions older than the retention window. Failures are logged and do not
+// stop the daemon; the next tick retries. It is a no-op when no maintainer is
+// configured.
+func (s *Daemon) sweepAuditPartitions(ctx context.Context) {
+	if s.auditMaintainer == nil {
+		return
+	}
+
+	now := time.Now().UTC()
+	if err := s.auditMaintainer.PreCreateNextPartition(ctx, now); err != nil {
```
mistahj67: naming nit:  the action is to create another partition, the fact it's done first is a logic choice and shouldn't be part of the name as the function itself does nothing to ensure it's done "first"
```suggestion
	if err := s.auditMaintainer.CreateNextPartition(ctx, now); err != nil {
```

cmd/api/src/daemons/gc/data_pruning.go
```diff
@@ -18,22 +18,35 @@ package gc
...
 	"github.com/specterops/bloodhound/cmd/api/src/database"
+	"github.com/specterops/bloodhound/packages/go/bhlog/attr"
+	"github.com/specterops/bloodhound/server/audit"
 )
 
+// defaultAuditRetentionMonths bounds how many months of audit_logs partitions
+// are retained. Partitions whose entire range is older than this window are
+// dropped. TODO(audit retention): source this from appcfg once the
+// retention_months parameter lands (Step I) instead of the constant.
+const defaultAuditRetentionMonths = 3
```
mistahj67, line 35: suggestion: consider using a true time.Duration here and passing it down instead of an integer semantically labeled as "months"

cmd/api/src/daemons/gc/data_pruning.go
```diff
@@ -49,23 +62,43 @@ func (s *Daemon) Start(ctx context.Context) {
 	defer close(s.exitC)
 	defer ticker.Stop()
 
-	// prune sessions and collections once when the daemon starts up
+	// prune sessions and collections and maintain audit partitions once when the daemon starts up
 	s.db.SweepSessions(ctx)
 	s.db.SweepAssetGroupCollections(ctx)
+	s.sweepAuditPartitions(ctx)
```
mistahj67: suggestion: abstract this into a s.runDataPruning that you can then call in both places as both need to be kept in lockstep

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+    IF EXISTS (
+        SELECT 1
+        FROM pg_partitioned_table pt
+        JOIN pg_class c ON c.oid = pt.partrelid
+        WHERE c.relname = 'audit_logs' AND pt.partstrat = 'r'
+    ) THEN
+        RETURN;
+    END IF;
+
+    DROP TABLE IF EXISTS audit_logs_partitioned;
```
mistahj67, line 95: any re-run of this will wipe data, is it certain that this should be dropped? why not a create table if not exists?

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+    CREATE TABLE audit_logs_partitioned (
+        id                  BIGINT NOT NULL DEFAULT nextval('audit_logs_id_seq'),
+        created_at          TIMESTAMPTZ NOT NULL DEFAULT now(),
+        action              TEXT NOT NULL,
+        actor_id            TEXT,
+        actor_name          TEXT,
+        actor_email         VARCHAR(330) DEFAULT NULL::character varying,
```
mistahj67: is the 330 limit mirrored on the users table? if not, this could result in failure to write if an email is longer than 330 characters

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
mistahj67: agreed and there's been a push to leverage pg enums in newer work as well.

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+        request_id          TEXT,
+        source_ip_address   TEXT,
+        status              VARCHAR(15) DEFAULT 'intent' CHECK (status IN ('intent', 'success', 'failure')),
+        commit_id           TEXT,
+        fields              JSONB,
+        source              VARCHAR(20) DEFAULT 'middleware',
```
mistahj67: same concern here with the 20 character cap, if source is developer supplied, there's a risk to breaking writing records if they are longer than 20 characters

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
```diff
@@ -0,0 +1,271 @@
...
+
+-- Phase 2: monthly partitions + default. Child partition names match the final
+-- audit_logs parent name so the GC daemon's partition maintenance
+-- (audit_logs_YYYY_MM) stays aligned after the rename below. Skipped entirely if
+-- a prior run already completed the swap (the staging table is then gone).
+-- +goose StatementBegin
+DO $$
+DECLARE
+    start_date    DATE := '2024-01-01';
+    end_date      DATE := '2026-08-01';
```
mistahj67: Is there a concern this needs to be set to 09-01 as the next release will be in september?

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
mistahj67, line 1: I don't have full context but at a glance this in place swap / backfill plus the upcoming gc expired partitions will result in audit logs being dropped. Is there concern for this all happening in this PR vs running a parallel table soaked for a bit and then swapping over and dropping the table later to ensure recoverability?

server/audit/internal/appdb/appdb.go
```diff
@@ -0,0 +1,105 @@
...
+	)
+	// Store SQL NULL for empty fields instead of jsonb 'null'.
+	if len(record.Fields) == 0 {
+		fieldsArg = nil
+	}
+	insertBuilder.InsertInto(tableAuditLogs)
+	insertBuilder.Cols(
+		"created_at", "action", "actor_id", "actor_name", "actor_email",
+		"request_id", "source_ip_address", "status", "commit_id", "fields", "source",
+	)
```
mistahj67: Consider constants for these / spreading that constant string array

server/audit/internal/appdb/partitions.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	return nil
+}
+
+// DropExpiredPartitions drops every monthly partition whose entire range is older
+// than the retention window (retentionMonths before asOf). Drops are idempotent
+// and the default partition is never touched.
+func (s *Store) DropExpiredPartitions(ctx context.Context, asOf time.Time, retentionMonths int) error {
+	var (
+		cutoff = firstOfMonth(asOf).AddDate(0, -retentionMonths, 0)
```
mistahj67, line 60: Is it worth having a check that the asOf time is not in the future?

server/modules/modules.go
```diff
@@ -77,4 +87,15 @@ func Register(deps Deps) {
...
 	extensions.Register(deps.Router, deps.Pool, deps.RateLimitMiddleware)
+
+	// Audit middleware records the intent/success/failure lifecycle of every
+	// request. It is attached post-routing so the authenticated actor set by the
+	// auth middleware is available on the request context. The health check
+	// route carries no audit value and is excluded. The Maintainer is returned
+	// so the entrypoint can hand it to the GC daemon to manage the audit_logs
+	// partitions.
+	auditService, auditMaintainer := audit.Register(deps.Pool)
+	deps.Router.UsePostrouting(middleware.AuditMiddleware(auditService, deps.Router.MuxRouter(), "/health"))
```
mistahj67: Needing to supply the routes as hardcoded strings at this point feels like it'll get a bit convoluted quickly and risks violating separation of concerns for the slices. I wonder if there should be a slice defined way to expose them to be consumed or what a possible solution looks like where this doesn't become unwieldy

cmd/api/src/api/middleware/audit.go
stephanieslamb: Moved this to the service layer.

cmd/api/src/database/migration/migrations/20260707000001_v9_audit_log_partitioning.sql
stephanieslamb: updated the migration so the old table can soak.  will create a ticket to drop old table the next release if this is route is agreed upon.

PR #3062, opened 2026-07-28
fix: adjust tooltip z-index to fix hidden node labels in Explore search/pathfinding - BED-9077

julian1j:
Hovering over a node kind icon in search or pathfinding rendered the label underneath the results list instead of over it. This is because TooltipContent used z-50, which sits below the app's overlay layers. the Explore search dropdown alone uses z-index 1300. 

Proposed fix changes this to z-[1500], matching the other floating content in doodle-ui (Select content, DatePicker popover, Dialog content).

## Description

Single file, one line modification, change z-50 to z-[1500]

## Motivation and Context

Resolves BED-9077

## How Has This Been Tested?

Tested locally in CE. Search for a node in Explore, hover over the node icon, you should see a tooltip/label for the node type. 

## Screenshots (optional):
Before: 
image

After: 
image

## Types of changes

## Checklist:

- [x ] I have met the contributing prerequisites
  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip visibility by ensuring tooltips appear above other interface elements when overlapping content.

### DIFF
--- a/packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
+++ b/packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
@@ -78,7 +78,7 @@ const TooltipContent = React.forwardRef<React.ElementRef<typeof TooltipPrimitive
                 sideOffset={sideOffset}
                 className={cn(
                     'TooltipContent',
-                    'z-50 overflow-hidden rounded-md border bg-neutral-light-2 px-3 py-1.5 text-xs text-popover-foreground shadow-md animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
+                    'z-[1500] overflow-hidden rounded-md border bg-neutral-light-2 px-3 py-1.5 text-xs text-popover-foreground shadow-md animate-in fade-in-0 zoom-in-95 data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=closed]:zoom-out-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2',
                     className,
                     widthOptions[contentWidth]
                 )}


bfaulk96 (approved):
Makes sense to me. I actually ran into a related issue on the "Manage Clients" page last week.

elikmiller:
It may be preferable to ensure the parent container has a separate stacking context instead of changing the z index of the tooltip at the component library level
https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Positioned_layout/Stacking_context

bfaulk96:
> It may be preferable to ensure the parent container has a separate stacking context instead of changing the z index of the tooltip at the component library level https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Positioned_layout/Stacking_context

I think Tooltips are rendered in a React portal, which takes them out of the component stack? I had a similar issue to the one shared here, where tooltips that were defined around components in a Menu were showing behind the menu because of that..

I could definitely be wrong on that assumption, but that's at least how MUI's tooltips work (I imagine Radix – our underlying tooltip provider – does something similar)

elikmiller:
> > It may be preferable to ensure the parent container has a separate stacking context instead of changing the z index of the tooltip at the component library level https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Positioned_layout/Stacking_context
> 
> I think Tooltips are rendered in a React portal, which takes them out of the component stack? I had a similar issue to the one shared here, where tooltips that were defined around components in a Menu were showing behind the menu because of that..
> 
> I could definitely be wrong on that assumption, but that's at least how MUI's tooltips work (I imagine Radix – our underlying tooltip provider – does something similar)

That sounds like a good assumption. I suppose as long as our design system/component library is not prescriptive about z-indicies then this change is fine. The other solution that occurs to me would be to isolate the react root element into a separate stacking context wholesale via https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/isolation

I bring it up because I'm not confident that this change to the component library won't trigger visual regressions in other areas of the app. But we can squash them if we find them.

bfaulk96:
@elikmiller FWIW, MUI uses 1500 z-index for their tooltips, as there are very few circumstances where you *wouldn't* want a tooltip on the top layer. For that reason, I feel like that edge-case should become the exception, rather than needing to regularly set high z-index via styling

bfaulk96:
I just checked the Radix website for their tooltip component. Interestingly, it has `z-index: auto` on the tooltip. It uses a React portal and puts it at the `body` root.

So it does seem like we'd need to set our own z-index potentially.

PR #3065, opened 2026-07-28
fix: respond with source kind only as fallback BED-9059

urangel:
## Description

Fixes kind resolution in some API responses.

## Motivation and Context

Resolves BED-9059

Primary kind resolution was resolving to source kinds when a more specific kind was available. Source kinds lack icon information and so the UI is unable to render the appropriate display with the information provided in responses where only one kind is returned. Primary kind resolution is extended to handle source kinds based on the DB values instead of just the code based constants for AD an Azure.

## How Has This Been Tested?

Unit tests added. Local check

## Screenshots (optional):

image

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Expanded primary display-kind resolution to include source kinds alongside custom kinds.
  - Added explicit identification of source/base kinds and improved how primary display kinds are selected.
  - Updated open graph kind handling to prefer matching display kinds, with a fallback to source kinds when no display kind exists.

- **Bug Fixes**
  - Improved primary display-kind map initialization and augmentation to ensure source kinds are always incorporated.

- **Tests**
  - Added coverage for open graph precedence and fallback behavior.
  - Updated test setups to reflect the new source-kind flag behavior.

### DIFF
--- a/cmd/api/src/database/graphschema.go
+++ b/cmd/api/src/database/graphschema.go
@@ -1366,23 +1366,34 @@ func (s *BloodhoundDB) DeletePrincipalKind(ctx context.Context, environmentId in
 // GetPrimaryDisplayKinds - returns a map of all node kinds that are display kinds. custom_node_kinds is the single source
 // of truth for display node kinds. Schema-backed display kinds are mirrored there on extension upsert, and schemaless kinds encountered
 // during ingest are also upserted.
+//
+// Source kinds (from the source_kinds registry, e.g. Base, AZBase, or an OpenGraph source kind like GithubBase) are also
+// included in the map and flagged with IsSourceKind. This allows PrimaryDisplayKind to generically treat any source kind as
+// a base/fallback kind rather than relying on hardcoded knowledge of the built-in AD and Azure base kinds.
 func (s *BloodhoundDB) GetPrimaryDisplayKinds(ctx context.Context) (graphschema.PrimaryDisplayKinds, error) {
+	var (
+		primaryDisplayKinds = make(graphschema.PrimaryDisplayKinds)
+		isSourceKind        = false
+	)
+
 	if customNodeKinds, err := s.GetCustomNodeKinds(ctx); err != nil {
 		return nil, err
 	} else {
-		var primaryDisplayKinds = make(graphschema.PrimaryDisplayKinds)
 		for _, customNodeKind := range customNodeKinds {
-			primaryDisplayKinds[graph.StringKind(customNodeKind.KindName)] = graphschema.DisplayKind{
-				Name: customNodeKind.KindName,
-				Icon: graphschema.DisplayNodeIcon{
-					Name:  customNodeKind.Config.Icon.Name,
-					Type:  customNodeKind.Config.Icon.Type,
-					Color: customNodeKind.Config.Icon.Color,
-				},
-			}
+			primaryDisplayKinds.Add(customNodeKind.KindName, customNodeKind.Config.Icon.Name, customNodeKind.Config.Icon.Color, customNodeKind.Config.Icon.Type, isSourceKind)
 		}
-		return primaryDisplayKinds, nil
 	}
+
+	if sourceKinds, err := s.GetSourceKinds(ctx); err != nil {
+		return nil, err
+	} else {
+		isSourceKind = true
+		for _, sourceKind := range sourceKinds {
+			primaryDisplayKinds.Add(sourceKind.Name, "", "", "", isSourceKind)
+		}
+	}
+
+	return primaryDisplayKinds, nil
 }
 
 // entity panel vars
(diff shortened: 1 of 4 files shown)


cmd/api/src/database/graphschema.go
```diff
@@ -1384,8 +1389,20 @@ func (s *BloodhoundDB) GetPrimaryDisplayKinds(ctx context.Context) (graphschema.
 				},
 			}
 		}
-		return primaryDisplayKinds, nil
 	}
+
+	if sourceKinds, err := s.GetSourceKinds(ctx); err != nil {
+		return nil, err
```
mistahj67, line 1388: What do you think of if an error happens, this still returns the primaryDisplayKinds?

cmd/api/src/database/graphschema.go
```diff
@@ -1384,8 +1389,20 @@ func (s *BloodhoundDB) GetPrimaryDisplayKinds(ctx context.Context) (graphschema.
...
+	} else {
+		for _, sourceKind := range sourceKinds {
+			primaryDisplayKinds[graph.StringKind(sourceKind.Name)] = graphschema.DisplayKind{
+				Name:         sourceKind.Name,
+				IsSourceKind: true,
+			}
```
mistahj67: The primaryKindMap data structure has a `.Add` method on it. What do you think of leveraging that here as well as with the customKinds above? And adding `isSourceKind` to that method as well?

mistahj67: Is there a concern that this resets the sourceKind icon information if it is already present and sets it to empty values? Might need to update the `.Add` method to perform more of a merge in the case a kind already is defined in the map 🤔

packages/go/graphschema/primarykind_test.go
```diff
@@ -58,4 +58,26 @@ func Test_PrimaryNodeKind(t *testing.T) {
 		}, graph.Kinds{graph.StringKind("Hero"), graph.StringKind("Villain")})
 		require.Equal(t, graph.StringKind("Hero"), primaryKind)
 	})
+
+	t.Run("returns open graph display kind over its source kind", func(t *testing.T) {
```
mistahj67, line 62: Would you consider swapping these examples to the dogpark schema that is loaded / defined already?

mistahj67 (commented):
Thanks for addressing this miss! Couple questions to consider

cmd/api/src/database/graphschema.go
urangel: That sounds reasonable. With a nil err too?

cmd/api/src/database/graphschema.go
urangel: > The primaryKindMap data structure has a .Add method on it. What do you think of leveraging that here as well as with the customKinds above? And adding isSourceKind to that method as well?

I did this and I think I prefer how it is written currently. The multi param, empty string param, and bool param feel a little clunkier and more opaque. I don't mind swapping it out though if you want

```go
	if customNodeKinds, err := s.GetCustomNodeKinds(ctx); err != nil {
		return nil, err
	} else {
		for _, customNodeKind := range customNodeKinds {
			primaryDisplayKinds.Add(customNodeKind.KindName, customNodeKind.Config.Icon.Name, customNodeKind.Config.Icon.Color, customNodeKind.Config.Icon.Type, false)
		}
	}

	if sourceKinds, err := s.GetSourceKinds(ctx); err != nil {
		return primaryDisplayKinds, nil
	} else {
		for _, sourceKind := range sourceKinds {
			primaryDisplayKinds.Add(sourceKind.Name, "", "", "", true)
		}
	}
```

> s there a concern that this resets the sourceKind icon information if it is already present and sets it to empty values? Might need to update the .Add method to perform more of a merge in the case a kind already is defined in the map 🤔

This is a valid concern but not something that I think should be solved here. What I think should happen is that Source Kinds get defined similar to other extension defined kinds to avoid the possible but out of regular opengraph flow situation.

elikmiller:
@coderabbitai review

cmd/api/src/database/graphschema.go
mistahj67: On second thought, lets just leave it as nil. I'll resolve

cmd/api/src/database/graphschema.go
mistahj67: yea fair ultimately up to you, I try to stick to similar one way paths to increase maintainability but this does get hit with readability pain :/

urangel: I made the updates and handled the merging along with adding tests for the updated Add method 🤝 Thanks for the suggestions!

mistahj67 (approved):
code lgtm, testing locally as well 🚀 thanks for fixing this miss!

PR #3111, opened 2026-08-06
chore(api): remove unused request signing helpers - BED-9255

### LINKED ISSUE #3098: Bug: request signature helper signs URL.Path while the server validates RequestURI
## Description:

The request signing helper that BloodHound ships and the server-side validator disagree about
what the signature covers, so a request signed with the shipped helper is rejected as soon as
it carries a query string.

Line numbers below are at tag `v9.5.1`.

`ValidateRequestSignature` builds the expected digest from `request.RequestURI`
([`cmd/api/src/api/auth.go:355`](https://github.com/SpecterOps/BloodHound/blob/v9.5.1/cmd/api/src/api/auth.go#L355)),
which includes the query string.

`SignRequestAtTime` builds it from `request.URL.Path`
([`cmd/api/src/api/signature.go:160`](https://github.com/SpecterOps/BloodHound/blob/v9.5.1/cmd/api/src/api/signature.go#L160)),
which does not.

For a request without parameters the two strings are identical and nothing shows. Add one
parameter and the two sides compute different digests, and the server answers
`401 signature digest mismatch`.

I suspect the reason is that `http.Request.RequestURI` is only populated on the server side.
On a client-side request it is empty, so reaching for `URL.Path` is the natural thing to write.
The equivalent of the server's value on a client request is `URL.RequestURI()`.

Yes, if a PR would be welcome. I am happy to leave it with you instead if you would rather fix
it yourselves.

## Component(s) Affected:

- API

## Steps to Reproduce:

1. Create an API token.
2. Sign a request for an endpoint that takes query parameters, for example
   `GET /api/v2/graphs/shortest-path?start_node=A&end_node=B&only_traversable=true`, using
   `api.SignRequest`.
3. Send it.
4. The server answers `401` with `signature digest mismatch`. The same code against an
   endpoint with no query string succeeds.

The disagreement can also be shown without a server, by calling the shared digest function
both ways:

```go
req, _ := http.NewRequest(
    http.MethodGet,
    "https://bloodhound.example/api/v2/graphs/shortest-path?start_node=A&end_node=B",
    nil,
)
datetime := time.Now().Format(time.RFC3339)

// What SignRequestAtTime passes today.
fromPath, _ := api.NewRequestSignature(
    ctx, sha256.New, tokenKey, datetime, req.Method, req.URL.Path, nil,
)

// What ValidateRequestSignature will compare it against, since the server's
// request.RequestURI is the full target.
fromTarget, _ := api.NewRequestSignature(
    ctx, sha256.New, tokenKey, datetime, req.Method, req.URL.RequestURI(), nil,
)

// fromPath != fromTarget
```

## Expected Behavior:

A request signed with the helper the project distributes is accepted by the server, whether or
not it carries a query string.

## Actual Behavior:

It is accepted only when there is no query string. With one parameter the server answers `401`
with `digest validation failed: signature digest mismatch`
([`cmd/api/src/api/auth.go:365`](https://github.com/SpecterOps/BloodHound/blob/v9.5.1/cmd/api/src/api/auth.go#L365)),
which is accurate and gives no hint that the URI is the part that differs.

## Environment:

BloodHound CE v9.5.1, running locally in Docker. Client written in Go against the same signing
scheme. The graph driver should not be relevant, since this sits in the authentication layer
before any query runs.

## Additional Information:

The Python client published alongside the documentation (`docs/assets/apiclient.py` in
`SpecterOps/bloodhound-docs`) signs the whole URI it is handed, and several of its examples
carry query strings, for example
`/api/v2/domains/{id}/details?finding=...&skip=0&limit=0&Accepted=eq:False` and
`/api/v2/posture-stats?from=...&to=...`. Those calls work, which lines up with the server
signing `RequestURI`.

So the documented Python client and the server agree with each other, and the Go helper is the
one that differs. That may be why this has stayed unnoticed: anyone following the documented
client never meets it, and it takes a Go caller reaching for `SignRequestAtTime` to run into
it.

I found it while writing a third-party Go client
([bhgraph](https://github.com/saluc28/bhgraph), if the context is useful), so I came at it
from the outside rather than through the shipped helper. That client signs `URL.RequestURI()`
and is accepted, which is what pointed at the direction of the mismatch.

Have `SignRequestAtTime` sign the full request target, so that both sides cover the same
string:

```go
// cmd/api/src/api/signature.go:160
request.Method, request.URL.RequestURI(), tee
```

Two things that may be worth weighing:

The alternative, having the server validate `request.URL.Path`, would leave the query string
outside the signature. Parameters would then be modifiable without invalidating it, so the
client-side change looks like the safer direction.

On blast radius: I searched the repository and `SignRequest` and `SignRequestAtTime` appear
only in `cmd/api/src/api/signature.go` and its test. There seem to be no internal callers, so
the change would affect external clients using the helper, and those are the ones currently
failing on any request with parameters.

## Related Issues:

I searched the tracker for `signature`, `RequestURI`, `SignRequest` and `digest mismatch` and
did not find this reported.

## Contributor Checklist:

saluc28:
## Description

Removes `SignRequest` and `SignRequestAtTime` from `cmd/api/src/api/signature.go`, along with the two tests that exercised them and the imports that were left unused once they were gone.

`NewRequestSignature` is untouched. `ValidateRequestSignature` in `auth.go` still calls it, and the internal tests that cover it are unchanged, as is `TestSelfDestructingTempFile`.

## Motivation and Context

Resolves BED-9255
Resolves #3098

Both helpers built their digest from `request.URL.Path`, while the server validates against `request.RequestURI`. The two strings are identical until a request carries a query string, and from that point anything signed with them is rejected with `401 digest validation failed: signature digest mismatch`.

Neither function has a caller anywhere in the repository, and removal rather than repair is the direction the team chose in https://github.com/SpecterOps/BloodHound/issues/3098#issuecomment-5197456132.

So this PR deliberately does not restore signing for requests carrying a query string. It drops the helpers that could not do it, which is why `go build ./...` still passes across the whole module.

## How Has This Been Tested?

- `go build ./...` across the whole module, which is what confirms nothing else depended on either function
- `go test ./cmd/api/src/api/`
- `go vet ./cmd/api/src/api/`
- `gofmt -s -l` reports nothing for either file touched
- Searched the repository for any remaining reference to `SignRequest` or `SignRequestAtTime`, and none is left outside the tests that were removed

## Types of changes

## Checklist:

  - Associated an issue: #3098
  - Assignee and labels are not settable from an outside fork, so I have left them to you
  - No API endpoint, OpenAPI document or database migration is affected
  - The only GoDoc removed is the one attached to the two deleted functions
  - Removed the two tests that covered the removed functions


saluc28:

wes-mil (approved):

wes-mil:
@saluc28 Hey, sorry there's a vuln with some UI dependencies. After it's patched in main, I'll need you to rebase/merge main into this branch in order to move it forward

saluc28:
@wes-mil thanks for the heads up. I have merged `main` into the branch now that the `js-yaml` bump has landed, so `build-ui` should be happy on the next run.

For the record, the earlier `run-go-integration-tests` failure was `TestAGT_FetchNodesFromSeeds_ParentExpansion_with_limit` in `packages/go/analysis`, which does not depend on `cmd/api/src/api` either directly or transitively, so it is unrelated to this change

wes-mil:
Thank you! I will get this merged tomorrow :smile:

PR #3130, opened 2026-08-10
fix: Prevent stale pathfinding results from hiding cypher search widget (BED-9216)

dcairnsspecterops:
## Description

Previously, when the user arrived from a successful pathfinding search and switched to the cypher tab, a cypher call would be made, which would cause the search box to close in the UI. 

Now, we disable to call if there is no cypher query. We also added the tab to the query key to prevent the pathfinding nodes from hydrating in the explore page.

## Motivation and Context

https://github.com/user-attachments/assets/4362343d-006a-4dbd-8dd1-388e6c2a604d

Resolves BED-9216

## How Has This Been Tested?

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Explore search behavior after graph searches.
  - The Explore search widget remains open when the Cypher tab is active without a query.
  - The widget closes automatically when search results include nodes.
  - Single-node results are selected automatically, while multi-node results clear the current selection.
  - Empty results no longer trigger selection or widget changes.
  - Re-running multi-node queries consistently updates the current selection.


packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -89,6 +92,7 @@ export const useExploreGraph = (options: ExploreGraphQueryOptions = {}) => {
         },
         ...rest,
         ...userSettings,
+        enabled: !cypherQueryIsEmpty && rest.enabled,
```
benwaples: Would this change, plus the associated above, be better suited in `bh-shared-ui/src/hooks/useExploreGraph/queries/cypherSearch.ts`?

For context, we broke it up that way because each `searchType` has unique criteria for the query so it seemed reasonable to break each up and reduce the cognitive overhead in this file!

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
```diff
@@ -75,6 +75,9 @@ export const useExploreGraph = (options: ExploreGraphQueryOptions = {}) => {
 
     const queryConfig = query.getQueryConfig();
 
+    const cypherSearchExists = params.cypherSearch;
+    const cypherQueryIsEmpty = !cypherSearchExists && params.exploreSearchTab === 'cypher';
```
benwaples: I was curious why adding this check works but the [existing check](https://github.com/SpecterOps/bloodhound-enterprise/blob/678b63eaba30255ffca2772da7853093809c3fe9/cmd/ui/src/views/Explore/ExploreSearch/ExploreSearch.tsx#L137-L138) upstream fails to block the extra search because the upstream should be enough to block useExploreGraph from running again. Ultimately this check before running the query doesnt hurt tho and im not opposed, but I was concerned there was a problem upstream so I took the opportunity to investigate!

From what I can tell the query is being refetched when clicking on the cypher tab, which seems unintentional because the query key remains the same (IIRC thats the control mechanism for saying when a query should refetch). I _think_ its being refetched when the `CypherSearchInner` component gets mounted, meaning throwing on a `refetchOnMount: false` field to the query options stopped it from refetching and closing the search area. 

I cant imagine we are intending to refetch on mount in that case, so I feel that blocking the fetch saves us a network query still! But if we are intending to do that, totally ignore.

benwaples (commented):
A couple comments/questions, but not a blocker as is!

Do you mind rebasing off main and we can cherry pick this into stage if we need it for release?

TY!!

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
dcairnsspecterops: The interesting problem here is that:
1) we use `searchType` to decide when to launch a cypher call:
```
    const performSearch = (query?: string) => {
        setExploreParams({
            searchType: 'cypher',
            cypherSearch: encodeCypherQuery(query ?? cypherQuery),
        });
    };
```
2) This bug occurs when the tab is 'cypher', but the searchType is still 'pathfinding', since a cypher query has not been launched (and therefore searchType has not been updated)

image

3) Therefore, we are in `usePathfindingSearch` when this bug occurs.

I think we could move the logic to `usePathfindingSearch` if that feels right to you?

benwaples: tysm for laying this out!! Hmm very interesting situation... remaking the pathfinding query when clicking the cypher tab does seem unintentional 🤔 

Agreed we should block that query from happening but I think using the enabled field to check for cypher searches from within `usePathfindingSearch` might be covering up an issue higher up? How do you feel about the suggested `refetchOnMount: false` addition below? What might be happening is the `CypherSearchInner` component gets remounted when the tab changes to cypher which triggers that instance of the `useExploreGraph` to refetch the existing pathfinding query and in turn executing the onSuccess callback.

packages/javascript/bh-shared-ui/src/hooks/useExploreGraph/useExploreGraph.tsx
benwaples: Thank you for entertaining my questions! It does appear that the `refetchOnMound`: false introduces a race condition and will occasionally still close the search panel :/

benwaples (approved):
Ahhh i love the specificity and readability this adds, thank you!

Before merging do you mind updating the description to reflect the current changes? That way if we come back to this in the future the description matches the changes!

TheNando (approved):

PR #3161, opened 2026-08-16
fix: return session and local group completeness under their own keys

amaralazizy:
### Description

`GetDatabaseCompleteness` (`GET /api/v2/completeness`) returns the two completeness percentages under each other's names:

```go
result["LocalGroupCompleteness"] = userSessionCompleteness // session value
result["SessionCompleteness"]    = localGroupCompleteness  // local group value
```

`FetchUserSessionCompleteness` is the fraction of active users with at least one session, and `FetchLocalGroupCompleteness` is the fraction of active computers with at least one local-admin relationship. The handler assigns each to the other key, so an API consumer reading `SessionCompleteness` gets local group completeness, and vice versa.

The historical stats path in `cmd/api/src/services/dataquality/dataquality.go` (lines 228–245, 262–271) assigns the same two values correctly, so the two code paths disagree with each other. The UI data quality view reads the historical path, which is why this is not visible in the product UI.

The assignment appears unchanged since the initial public commit of the repository.

### Motivation and Context

The endpoint is part of the documented public API (`packages/go/openapi/src/paths/data-quality.completeness.yaml`), and any consumer using it for collection-quality reporting gets the two figures swapped. I hit this while measuring collection completeness against a local BHCE v9.5.1 instance through the API — every prediction was inverted, and it traced back to this handler.

### How Has This Been Tested?

Added `TestResources_GetDatabaseCompleteness_ResponseKeys` in `cmd/api/src/api/v2/dataquality_test.go`. The existing test mocks `ReadTransaction` to return without running the delegate, so the mapping itself was never exercised. The new test drives the delegate with a mocked transaction over a graph where the two metrics differ (4 active users, 1 with a session → `0.25`; 2 active computers, 1 with an admin → `0.5`) and asserts the response body.

Against the current code the test fails exactly as expected:

```
- "LocalGroupCompleteness": (float64) 0.5,
- "SessionCompleteness": (float64) 0.25
+ "LocalGroupCompleteness": (float64) 0.25,
+ "SessionCompleteness": (float64) 0.5
```

`go test ./cmd/api/src/api/v2/...` passes with the fix; `gofmt` clean. I was not able to run `just prepare-for-codereview` in full, but this change touches no models, mocks, or route definitions, and the OpenAPI schema for this endpoint is an untyped `additionalProperties: number` map, so no spec change is required.

### Note on compatibility

This changes observable API output. Anyone who has compensated for the transposition on the client side would need to drop that workaround — flagging it in case you would rather land it behind a release note.

### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

### Checklist

- [x] My code follows the code style of this project.
- [x] I have read the **CONTRIBUTING** document.
- [x] I have added tests to cover my changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected database completeness results so session and local-group completeness values are displayed in their appropriate fields.
  * Added validation to ensure completeness responses return accurate, distinct values.

### DIFF
--- a/cmd/api/src/api/v2/dataquality.go
+++ b/cmd/api/src/api/v2/dataquality.go
@@ -56,13 +56,13 @@ func (s Resources) GetDatabaseCompleteness(response http.ResponseWriter, request
 		if userSessionCompleteness, err := ad.FetchUserSessionCompleteness(tx); err != nil {
 			return err
 		} else {
-			result["LocalGroupCompleteness"] = userSessionCompleteness
+			result["SessionCompleteness"] = userSessionCompleteness
 		}
 
 		if localGroupCompleteness, err := ad.FetchLocalGroupCompleteness(tx); err != nil {
 			return err
 		} else {
-			result["SessionCompleteness"] = localGroupCompleteness
+			result["LocalGroupCompleteness"] = localGroupCompleteness
 		}
 
 		return nil
--- a/cmd/api/src/api/v2/dataquality_test.go
+++ b/cmd/api/src/api/v2/dataquality_test.go
@@ -26,6 +26,7 @@ import (
 	"net/url"
 	"reflect"
 	"testing"
+	"time"
 
 	"github.com/specterops/bloodhound/cmd/api/src/utils"
 	"github.com/specterops/bloodhound/cmd/api/src/utils/test"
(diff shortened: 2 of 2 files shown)


amaralazizy:

urangel (approved):
Thank you!

PR #3164, opened 2026-08-17
feat: added required fields indicator BED-7225

Holocraft:
## Description

This adds functionality to the Input component to set required field indicators. It also combines the Label component into the Input and adds accessibility features to Inputs.

## Motivation and Context

Resolves [BED-7225](https://specterops.atlassian.net/browse/BED-7225)

## How Has This Been Tested?

Manually

## Screenshots (optional):
Screenshot 2026-08-17 at 11 20 14 AM

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

[BED-7225]: https://specterops.atlassian.net/browse/BED-7225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added support for input labels, required indicators, helper text, and error messages.
  * Added configurable styling for input fields, labels, helper text, and error messages.
  * Improved accessibility with generated IDs and associated validation messaging.

* **Documentation**
  * Updated input examples to demonstrate optional, required, helper text, outlined, and validation states.

### DIFF
--- a/packages/javascript/doodle-ui/src/components/Input/Input.tsx
+++ b/packages/javascript/doodle-ui/src/components/Input/Input.tsx
@@ -17,6 +17,7 @@ import { cva, VariantProps } from 'class-variance-authority';
 import * as React from 'react';
 import { Label } from '../Label';
 import { cn } from '../utils';
+import { Typography } from '../Typography';
 
 export const InputVariants = cva(
     'flex h-10 w-full text-base text-main placeholder:text-input-placeholder-text disabled:cursor-not-allowed disabled:opacity-50 file:border-0 file:bg-transparent file:pr-3 file:text-sm file:font-medium file:text-main file:cursor-pointer',
@@ -58,8 +59,8 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
             error = false,
             helperText,
             errorMessage,
-            type,
             variant,
+            intent,
             className,
             fieldClassName,
             labelClassName,
@@ -74,7 +75,7 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
     ) => {
         // need ID for fields with labels so screen readers have correct accessible name
         const generatedId = React.useId();
-        const isComposed = label !== undefined && label !== null;
+        const isComposed = !!label;
         const inputId = isComposed ? id ?? generatedId : id;
         // generate IDs for supporting text and combine them with any caller-provided IDs so assistive technology can describe the input.
         const helperTextId = `${inputId}-helper-text`;
@@ -92,11 +93,10 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
                 {...props}
                 ref={ref}
                 id={inputId}
-                type={type}
                 required={required}
-                className={cn(InputVariants({ variant, className }))}
+                className={cn(InputVariants({ variant, intent }), className)}
                 aria-describedby={isComposed ? describedBy || undefined : ariaDescribedBy}
-                aria-invalid={isComposed && error ? true : ariaInvalid}
+                aria-invalid={!!(isComposed && error) || ariaInvalid}
                 aria-errormessage={isComposed && error && errorMessage ? errorMessageId : ariaErrorMessage}
             />
         );
@@ -119,14 +119,17 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
                 </Label>
                 {input}
                 {helperText && (
-                    <p id={helperTextId} className={cn('text-sm text-main', helperTextClassName)}>
+                    <Typography id={helperTextId} className={cn('text-sm text-main', helperTextClassName)}>
                         {helperText}
-                    </p>
+                    </Typography>
                 )}
                 {error && errorMessage && (
-                    <p id={errorMessageId} role='alert' className={cn('text-error text-sm', errorMessageClassName)}>
+                    <Typography
+                        id={errorMessageId}
+                        role='alert'
+                        className={cn('text-error text-sm', errorMessageClassName)}>
                         {errorMessage}
-                    </p>
+                    </Typography>
                 )}
             </div>
         );
--- a/packages/javascript/doodle-ui/src/components/Input/Input.tsx
+++ b/packages/javascript/doodle-ui/src/components/Input/Input.tsx
@@ -15,6 +15,7 @@
 // SPDX-License-Identifier: Apache-2.0
 import { cva, VariantProps } from 'class-variance-authority';
 import * as React from 'react';
+import { Label } from '../Label';
 import { cn } from '../utils';
 
 export const InputVariants = cva(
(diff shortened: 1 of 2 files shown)


packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+            'aria-describedby': ariaDescribedBy,
+            'aria-invalid': ariaInvalid,
+            'aria-errormessage': ariaErrorMessage,
+            ...props
+        },
+        ref
+    ) => {
+        // need ID for fields with labels so screen readers have correct accessible name
+        const generatedId = React.useId();
+        const isComposed = label !== undefined && label !== null;
```
bfaulk96: Is an empty string label something we want to allow? If not, it might make more sense to check for truthiness: `const isComposed = !!label;`

packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+        const input = (
+            <input
+                {...props}
+                ref={ref}
+                id={inputId}
+                type={type}
+                required={required}
+                className={cn(InputVariants({ variant, className }))}
+                aria-describedby={isComposed ? describedBy || undefined : ariaDescribedBy}
+                aria-invalid={isComposed && error ? true : ariaInvalid}
```
bfaulk96: Really small nitpick, but this may read clearer without the ternary? `!!(isComposed && error) || ariaInvalid`

packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+        return (
+            <div className={cn('grid gap-1.5', fieldClassName)}>
+                <Label htmlFor={inputId} className={labelClassName}>
+                    {label}{' '}
+                    {required ? (
+                        <span aria-hidden='true' className='text-status-error-main'>
+                            *
+                        </span>
+                    ) : (
+                        <span>(Optional)</span>
```
bfaulk96: Should this be put behind a param as well? Seems pretty verbose in some cases to have `(Optional)` behind every optional label?

Holocraft: I thought about that too but it's one of the requirements in the ticket 🤷

packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+                    )}
+                </Label>
+                {input}
+                {helperText && (
+                    <p id={helperTextId} className={cn('text-sm text-main', helperTextClassName)}>
+                        {helperText}
+                    </p>
```
KillahDillah: Thoughts on using the `Typography` component here?

packages/javascript/doodle-ui/src/components/Input/Input.tsx
Holocraft: Yes, we want to allow fields without labels.

packages/javascript/doodle-ui/src/components/Input/Input.tsx
Holocraft: I was also thinking about this being a screen reader only thing but we'd have to ask design about this

packages/javascript/doodle-ui/src/components/Input/Input.tsx
Holocraft: Can do. Do we have a protocol for when to use Typography? Like every instance of h1's p's etc?

packages/javascript/doodle-ui/src/components/Input/Input.tsx
bfaulk96: I definitely think it's reasonable to always have `(Optional)` from a screen reader perspective, but tbh I imagine people will just use custom labels to avoid this if they don't want the lengthy label shown all the time.

Either way, I won't hold ya back over this!

packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+        {
+            id,
+            label,
+            required = false,
+            error = false,
+            helperText,
+            errorMessage,
+            type,
+            variant,
+            className,
```
TheNando, line 65: It looks like the `intent` prop is leaking and not actually being used. You can see it in the Time Input story. Let's destructure it to prevent the leak, then add it to the classname below.
```suggestion
            variant,
            intent,
            className,
```

packages/javascript/doodle-ui/src/components/Input/Input.tsx
Holocraft: Can do. Do we have a protocol for when to use Typography? Like every instance of h1's p's etc?

packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+            .join(' ');
+
+        const input = (
+            <input
+                {...props}
+                ref={ref}
+                id={inputId}
+                type={type}
+                required={required}
+                className={cn(InputVariants({ variant, className }))}
```
TheNando: Not really sure what intent was added for, but if we want its style to actually apply, this would be the simplest way to do it.

Also, just a nit-pick, but worth note, passing `className` to the cva variant functions is mostly a no-op, as it just passes through. It is more precise to just pass className as an additional param, as it better expresses intent.

```suggestion
                className={cn(InputVariants({ variant, intent }), className)}
```

packages/javascript/doodle-ui/src/components/Input/Input.tsx
```diff
@@ -37,11 +38,100 @@ export const InputVariants = cva(
...
+const Input = React.forwardRef<HTMLInputElement, InputProps>(
+    (
+        {
+            id,
+            label,
+            required = false,
+            error = false,
+            helperText,
+            errorMessage,
+            type,
```
TheNando: Type is destructured, but it's not actually used anywhere except to be added back as a prop. May as well remove it an let it get passed with `...props`.

TheNando (changes requested):

packages/javascript/doodle-ui/src/components/Input/Input.tsx
TheNando: At the app/shared component level, ideally it will be 95% Typography. Within Doodle UI, we might expect that to be a bit lower. The obvious cases like Headers, Captions, and Subtitles, we have the variants. Body1 and Body2 map to `<p>` tags, so in this case where it's a `<p>` and it has `text-sm`, Body2 seems appropriate.

But, checking under the hood here for this question brought something to my attention. The `text-main` class isn't actually applying any style change. You can see in the screenshot below that only the `text-sm` is reflecting in the styles. Out of scope for this, but we'll need to address that. 

image

packages/javascript/doodle-ui/src/components/Input/Input.tsx
Holocraft: This is a great catch. This means that `intent` has always leaked and has never been used as far as I can tell with history.

bfaulk96 (approved):

TheNando (approved):

PR #3176, opened 2026-08-19
feat: DB support for creating PZ rules via OG extension management (BED-9331)

benwaples:
## Description
Adds DB helpers for creating, updating, and deleting rules from extension management. These changes add a rule_key and extension_id to the asset_group_tag_selector model. Those fields shouldnt be undateable from API routes for now so the create/update selector handlers identify those fields as read only.

This changeset also includes a fix to then OpenAPI spec create selector documentation.

Shoutout to @LawsonWillard for helping with GetByExtensionId and fixing tests!!

## Motivation and Context

Resolves BED-9331

Supporting OpenGraph Extension Management integration with Privilege Zone rules.

## How Has This Been Tested?

Unit test, integration tests, and manually regression testing the API endpoint handler changes.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Asset-group tag selectors support extension ownership and rule keys.
  * Selector responses include read-only `extension_id` and `rule_key` fields.
  * OpenGraph-managed selectors support updates, retrieval, removal, and lifecycle tracking.
  * Selectors can be retrieved by extension ID.

* **Bug Fixes**
  * Improved handling for duplicate selector names and rule keys.
  * Prevented attempts to set read-only ownership fields during creation or updates.

* **Documentation**
  * Updated API schemas for selector creation requests, general requests, and responses.

### DIFF
--- a/cmd/api/src/api/v2/assetgrouptags.go
+++ b/cmd/api/src/api/v2/assetgrouptags.go
@@ -60,6 +60,8 @@ const (
 
 	includeProperties = true
 	excludeProperties = false
+
+	assetGroupTagSelectorReadOnlyFieldError = "request contains read-only fields"
 )
 
 func maybeAddEnvironmentIdFilter(filters []graph.Criteria, environmentIds []string) []graph.Criteria {
@@ -87,6 +89,12 @@ type assetGroupTagSelectorRequest struct {
 	AutoCertify *model.SelectorAutoCertifyMethod `json:"auto_certify"`
 	Description *string                          `json:"description"`
 	Disabled    *bool                            `json:"disabled"`
+	ExtensionId *int32                           `json:"extension_id"`
+	RuleKey     *string                          `json:"rule_key"`
+}
+
+func (s assetGroupTagSelectorRequest) hasReadOnlyFields() bool {
+	return s.ExtensionId != nil || s.RuleKey != nil
 }
 
 func (s Resources) GetAssetGroupTags(response http.ResponseWriter, request *http.Request) {
@@ -209,6 +217,8 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 		api.HandleDatabaseError(request, response, err)
 	} else if err := json.NewDecoder(request.Body).Decode(&createSelectorRequest); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if createSelectorRequest.hasReadOnlyFields() {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, assetGroupTagSelectorReadOnlyFieldError, request), response)
 	} else if errs := validation.Validate(createSelectorRequest.AssetGroupTagSelector); len(errs) > 0 {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
 	} else if actor, isUser := auth.GetUserFromAuthCtx(bhctx.FromRequest(request).AuthCtx); !isUser {
@@ -232,7 +242,15 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 			description = *createSelectorRequest.Description
 		}
 
-		if selector, err := s.DB.CreateAssetGroupTagSelector(request.Context(), assetTagId, actor, createSelectorRequest.Name, description, false, true, autoCertify, createSelectorRequest.Seeds); errors.Is(err, database.ErrDuplicateAGTagSelectorName) {
+		if selector, err := s.DB.CreateAssetGroupTagSelector(request.Context(), actor, model.AssetGroupTagSelector{
+			AssetGroupTagId: assetTagId,
+			Name:            createSelectorRequest.Name,
+			Description:     description,
+			IsDefault:       false,
+			AllowDisable:    true,
+			AutoCertify:     autoCertify,
+			Seeds:           createSelectorRequest.Seeds,
+		}); errors.Is(err, database.ErrDuplicateAGTagSelectorName) {
 			api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusConflict, api.ErrorResponseAssetGroupTagSelectorDuplicateName, request), response)
 		} else if err != nil {
 			api.HandleDatabaseError(request, response, err)
@@ -273,6 +291,8 @@ func (s *Resources) UpdateAssetGroupTagSelector(response http.ResponseWriter, re
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "selector is not part of asset group tag", request), response)
 	} else if err := json.NewDecoder(request.Body).Decode(&selUpdateReq); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if selUpdateReq.hasReadOnlyFields() {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, assetGroupTagSelectorReadOnlyFieldError, request), response)
 	} else {
 		// we can update DisabledAt on a default selector
 		if selUpdateReq.Disabled != nil {
--- a/cmd/api/src/api/v2/assetgrouptags.go
+++ b/cmd/api/src/api/v2/assetgrouptags.go
@@ -60,8 +60,6 @@ const (
 
 	includeProperties = true
 	excludeProperties = false
-
-	assetGroupTagSelectorReadOnlyFieldError = "request contains read-only fields"
 )
 
 func maybeAddEnvironmentIdFilter(filters []graph.Criteria, environmentIds []string) []graph.Criteria {
@@ -218,7 +216,7 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 	} else if err := json.NewDecoder(request.Body).Decode(&createSelectorRequest); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
 	} else if createSelectorRequest.hasReadOnlyFields() {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, assetGroupTagSelectorReadOnlyFieldError, request), response)
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorAssetGroupTagSelectorReadOnlyField, request), response)
 	} else if errs := validation.Validate(createSelectorRequest.AssetGroupTagSelector); len(errs) > 0 {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
 	} else if actor, isUser := auth.GetUserFromAuthCtx(bhctx.FromRequest(request).AuthCtx); !isUser {
@@ -292,7 +290,7 @@ func (s *Resources) UpdateAssetGroupTagSelector(response http.ResponseWriter, re
 	} else if err := json.NewDecoder(request.Body).Decode(&selUpdateReq); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
 	} else if selUpdateReq.hasReadOnlyFields() {
-		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, assetGroupTagSelectorReadOnlyFieldError, request), response)
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorAssetGroupTagSelectorReadOnlyField, request), response)
 	} else {
 		// we can update DisabledAt on a default selector
 		if selUpdateReq.Disabled != nil {
--- a/cmd/api/src/api/v2/assetgrouptags.go
+++ b/cmd/api/src/api/v2/assetgrouptags.go
@@ -91,7 +91,7 @@ type assetGroupTagSelectorRequest struct {
 	RuleKey     *string                          `json:"rule_key"`
 }
 
-func (s assetGroupTagSelectorRequest) hasReadOnlyFields() bool {
+func (s assetGroupTagSelectorRequest) hasExtensionOnlyFields() bool {
 	return s.ExtensionId != nil || s.RuleKey != nil
 }
 
@@ -215,7 +215,7 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 		api.HandleDatabaseError(request, response, err)
 	} else if err := json.NewDecoder(request.Body).Decode(&createSelectorRequest); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
-	} else if createSelectorRequest.hasReadOnlyFields() {
+	} else if createSelectorRequest.hasExtensionOnlyFields() {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorAssetGroupTagSelectorReadOnlyField, request), response)
 	} else if errs := validation.Validate(createSelectorRequest.AssetGroupTagSelector); len(errs) > 0 {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, errs.Error(), request), response)
@@ -289,7 +289,7 @@ func (s *Resources) UpdateAssetGroupTagSelector(response http.ResponseWriter, re
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusNotFound, "selector is not part of asset group tag", request), response)
 	} else if err := json.NewDecoder(request.Body).Decode(&selUpdateReq); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
-	} else if selUpdateReq.hasReadOnlyFields() {
+	} else if selUpdateReq.hasExtensionOnlyFields() {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorAssetGroupTagSelectorReadOnlyField, request), response)
 	} else {
 		// we can update DisabledAt on a default selector
--- a/cmd/api/src/api/v2/assetgrouptags_test.go
+++ b/cmd/api/src/api/v2/assetgrouptags_test.go
@@ -528,8 +528,16 @@ func TestResources_CreateAssetGroupTagSelector(t *testing.T) {
 					mockGraphDb.EXPECT().
 						PrepareCypherQuery(gomock.Any(), int64(-3)).
 						Return(queries.PreparedQuery{}, nil).Times(1)
-					mockDB.EXPECT().
-						CreateAssetGroupTagSelector(gomock.Any(), 1, user, testSelector.Name, testSelector.Description, false, true, testSelector.AutoCertify, testSelector.Seeds).
+					mockDB.EXPECT().CreateAssetGroupTagSelector(
+						gomock.Any(), user, model.AssetGroupTagSelector{
+							AssetGroupTagId: 1,
+							Name:            testSelector.Name,
+							Description:     testSelector.Description,
+							IsDefault:       false,
+							AllowDisable:    true,
+							AutoCertify:     testSelector.AutoCertify,
+							Seeds:           testSelector.Seeds,
+						}).
 						Return(model.AssetGroupTagSelector{}, errors.New("failure")).Times(1)
 					mockDB.EXPECT().GetAssetGroupTag(gomock.Any(), 1).
 						Return(model.AssetGroupTag{}, nil).Times(1)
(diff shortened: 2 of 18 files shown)


cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -232,7 +242,7 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 			description = *createSelectorRequest.Description
 		}
 
-		if selector, err := s.DB.CreateAssetGroupTagSelector(request.Context(), assetTagId, actor, createSelectorRequest.Name, description, false, true, autoCertify, createSelectorRequest.Seeds); errors.Is(err, database.ErrDuplicateAGTagSelectorName) {
+		if selector, err := s.DB.CreateAssetGroupTagSelector(request.Context(), assetTagId, actor, createSelectorRequest.Name, description, false, true, autoCertify, createSelectorRequest.Seeds, null.String{}); errors.Is(err, database.ErrDuplicateAGTagSelectorName) {
```
benwaples: This ensures the existing public routes are not able to create a rule with rule_key, _for now_. It very possibly could be a use case in the future but I figure safe guard against it for now

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -209,6 +217,8 @@ func (s *Resources) CreateAssetGroupTagSelector(response http.ResponseWriter, re
 		api.HandleDatabaseError(request, response, err)
 	} else if err := json.NewDecoder(request.Body).Decode(&createSelectorRequest); err != nil {
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, api.ErrorResponsePayloadUnmarshalError, request), response)
+	} else if createSelectorRequest.hasReadOnlyFields() {
+		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusBadRequest, assetGroupTagSelectorReadOnlyFieldError, request), response)
```
benwaples: Is this overkill?

cmd/api/src/database/assetgrouptags.go
```diff
@@ -59,6 +59,41 @@ type AssetGroupTagSelectorData interface {
...
+	if strings.Contains(result.Error.Error(), "duplicate key value violates unique constraint \"asset_group_tag_selectors_unique_name_asset_group_tag\"") {
+		return fmt.Errorf("%w: %v", ErrDuplicateAGTagSelectorName, result.Error)
+	} else if strings.Contains(result.Error.Error(), "duplicate key value violates unique constraint \"idx_asset_group_tag_selectors_rule_key\"") {
+		return fmt.Errorf("%w: %v", ErrDuplicateAGTagSelectorRuleKey, result.Error)
+	}
+
+	return CheckError(result)
+}
+
+func validateOpenGraphAssetGroupTagSelectorInput(input model.AssetGroupTagSelector) error {
```
benwaples: This also might be overkill, especially at the DB function layer, thoughts?

cmd/api/src/database/assetgrouptags.go
```diff
@@ -215,6 +252,171 @@ func (s *BloodhoundDB) DeleteAssetGroupTagSelector(ctx context.Context, user mod
...
+	if err := validateOpenGraphAssetGroupTagSelectorInput(input); err != nil {
+		return model.AssetGroupTagSelector{}, err
+	} else if selector.AssetGroupTagId != model.AssetGroupTierZeroPosition {
+		return model.AssetGroupTagSelector{}, fmt.Errorf("extension-owned selectors must remain in tier zero")
+	}
+
+	if input.DisabledAt.Valid {
+		disabledAt = null.TimeFrom(time.Now())
+		disabledBy = null.StringFrom(model.AssetGroupActorOpenGraphExtensionManagement)
+	}
```
benwaples: This is something the robot recommended so im not confident in it. But I think the another useful variation could be that we have a single key, `disabled: true | false`, from the extension management perspective and that key would be able to set these fields values

cmd/api/src/database/db.go
```diff
@@ -46,16 +46,17 @@ var (
...
-	ErrDuplicateUserPrincipal      = errors.New("duplicate user principal name")
-	ErrDuplicateEmail              = errors.New("duplicate user email address")
-	ErrDuplicateCustomNodeKindName = errors.New("duplicate custom node kind name")
-	ErrDuplicateKindName           = errors.New("duplicate kind name")
-	ErrDuplicateGlyph              = errors.New("duplicate glyph")
-	ErrPositionOutOfRange          = errors.New("position out of range")
+	ErrDuplicateAGName               = errors.New("duplicate asset group name")
+	ErrDuplicateAGTag                = errors.New("duplicate asset group tag")
+	ErrDuplicateAGTagSelectorName    = errors.New("duplicate asset group tag selector name")
+	ErrDuplicateAGTagSelectorRuleKey = errors.New("duplicate asset group tag selector rule key")
```
benwaples, line 52: this is what got added

cmd/api/src/database/assetgrouptags.go
```diff
@@ -105,15 +141,14 @@ func (s *BloodhoundDB) CreateAssetGroupTagSelector(ctx context.Context, assetGro
...
 		bhdb := NewBloodhoundDB(tx, s.pool, s.idResolver, s.config)
 		if result := tx.Raw(fmt.Sprintf(`
-			INSERT INTO %s (asset_group_tag_id, created_at, created_by, updated_at, updated_by, name, description, is_default, allow_disable, auto_certify)
-			VALUES (?, NOW(), ?, NOW(), ?, ?, ?, ?, ?, ?)
-			RETURNING id, asset_group_tag_id, created_at, created_by, updated_at, updated_by, disabled_at, disabled_by, name, description, is_default, allow_disable, auto_certify`,
+				INSERT INTO %s (asset_group_tag_id, created_at, created_by, updated_at, updated_by, name, description, is_default, allow_disable, auto_certify, rule_key)
+				VALUES (?, NOW(), ?, NOW(), ?, ?, ?, ?, ?, ?, ?)
+				RETURNING id, asset_group_tag_id, created_at, created_by, updated_at, updated_by,
+					disabled_at, disabled_by, name, description, is_default, allow_disable,
+					auto_certify, rule_key, extension_id`,
```
LawsonWillard, line 165: Im seeing extension id added here but not in the parameters or model above

LawsonWillard: oh wait, i just saw that there's a separate one for OG

cmd/api/src/api/v2/assetgrouptags.go
LawsonWillard: I think this is needed for now

cmd/api/src/database/assetgrouptags.go
LawsonWillard: Doesnt hurt and validation will definitely happen before the DB layer, though might want to move the error to the model package so we can key off it and return a 400

cmd/api/src/database/assetgrouptags.go
```diff
@@ -215,6 +252,178 @@ func (s *BloodhoundDB) DeleteAssetGroupTagSelector(ctx context.Context, user mod
...
+
+		return bhdb.CreateAssetGroupHistoryRecord(ctx, model.AssetGroupActorOpenGraphExtensionManagement, "", selector.Name, model.AssetGroupHistoryActionUpdateSelector, selector.AssetGroupTagId, null.String{}, null.String{})
+	}); err != nil {
+		return model.AssetGroupTagSelector{}, err
+	}
+
+	return selector, nil
+}
+
+func (s *BloodhoundDB) DeleteOpenGraphAssetGroupTagSelector(ctx context.Context, ruleKey string) error {
```
LawsonWillard: I think the reconcile function will require delete by id if thats feasible, it can be adjusted if not

cmd/api/src/database/assetgrouptags.go
benwaples: We spoke offline and the plan is for me to share all the validations we do in the endpoint handers prior to insert/update/delete so they can be mimicked in the extension management side

cmd/api/src/database/assetgrouptags.go
benwaples: yup! I want to block extension_id from being inserted or updated via the public API routes

cmd/api/src/database/assetgrouptags.go
benwaples: ahh thank you for the context, Ill see if I can plumb through the OG actor to the existing delete function because otherwise these would be identical.

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -552,15 +560,62 @@ func TestResources_CreateAssetGroupTagSelector(t *testing.T) {
...
+					apitest.SetURLVar(input, api.URIPathVariableAssetGroupTagID, "1")
+					apitest.BodyString(input, `{"name":"TestSelector","extension_id":7,"seeds":[{"type":1,"value":"S-1-5-21-1234"}]}`)
+				},
+				Setup: func() {
+					mockDB.EXPECT().GetAssetGroupTag(gomock.Any(), 1).
+						Return(model.AssetGroupTag{}, nil).Times(1)
+				},
+				Test: func(output apitest.Output) {
+					apitest.StatusCode(output, http.StatusBadRequest)
+				},
```
bsheth711, line 594: Consider adding: apitest.BodyContains()

cmd/api/src/api/v2/assetgrouptags_test.go
```diff
@@ -552,15 +560,62 @@ func TestResources_CreateAssetGroupTagSelector(t *testing.T) {
...
+						AutoCertify: model.SelectorAutoCertifyMethodDisabled,
+					})
+				},
+				Setup: func() {
+					mockDB.EXPECT().GetAssetGroupTag(gomock.Any(), 1).
+						Return(model.AssetGroupTag{}, nil).Times(1)
+				},
+				Test: func(output apitest.Output) {
+					apitest.StatusCode(output, http.StatusBadRequest)
+				},
```
bsheth711, line 619: Consider adding: apitest.BodyContains()

cmd/api/src/database/assetgrouptags.go
```diff
@@ -57,6 +57,43 @@ type AssetGroupTagSelectorData interface {
...
+}
+
+// The create/update paths share the same selector unique constraints; keep their error mapping in one place.
+func checkAssetGroupTagSelectorMutationError(result *gorm.DB) error {
+	if result.Error == nil {
+		return nil
+	}
+
+	var pgErr *pgconn.PgError
+	if errors.As(result.Error, &pgErr) && pgErr.Code == "23505" {
```
bsheth711: Can we put "23505" in a named constant rather than a magic string

cmd/api/src/database/assetgrouptags.go
```diff
@@ -82,43 +119,39 @@ func insertSelectorSeeds(tx *gorm.DB, selectorId int, seeds []model.SelectorSeed
...
-			VALUES (?, NOW(), ?, NOW(), ?, ?, ?, ?, ?, ?)
-			RETURNING id, asset_group_tag_id, created_at, created_by, updated_at, updated_by, disabled_at, disabled_by, name, description, is_default, allow_disable, auto_certify`,
+				INSERT INTO %s (
+					asset_group_tag_id, extension_id, rule_key, created_at, created_by, updated_at, updated_by,
+					disabled_at, disabled_by, name, description, is_default, allow_disable, auto_certify
+				)
+				VALUES (?, ?, ?, NOW(), ?, NOW(), ?, ?, ?, ?, ?, ?, ?, ?)
+				RETURNING id, asset_group_tag_id, created_at, created_by, updated_at, updated_by,
+					disabled_at, disabled_by, name, description, is_default, allow_disable,
+					auto_certify, rule_key, extension_id`,
```
bsheth711, line 165: Something I've started using recently: pgx.StrictNamedArgs. It makes the SQL more readable IMO and prevents misuse with arguments when using prepared statements. For example usage see: https://github.com/SpecterOps/BloodHound/blob/3b9cf24569a3f5a43ecf0d8fe63b4f61e5940d90/cmd/api/src/database/assetgrouptags.go#L950-L951

cmd/api/src/database/assetgrouptags.go
```diff
@@ -215,6 +251,59 @@ func (s *BloodhoundDB) DeleteAssetGroupTagSelector(ctx context.Context, user mod
...
+	if err := s.AuditableTransaction(ctx, auditEntry, func(tx *gorm.DB) error {
+		bhdb := NewBloodhoundDB(tx, s.pool, s.idResolver, s.config)
+		if result := tx.Raw(fmt.Sprintf(`
+			UPDATE %s
+			SET updated_at = NOW(), updated_by = ?, name = ?, description = ?, disabled_at = ?, disabled_by = ?,
+				auto_certify = ?, is_default = ?, allow_disable = ?
+			WHERE rule_key = ? AND extension_id = ?
+			RETURNING id, asset_group_tag_id, created_at, created_by, updated_at, updated_by,
+				disabled_at, disabled_by, name, description, is_default, allow_disable,
+				auto_certify, rule_key, extension_id`,
```
bsheth711, line 319: Consider using pgx.StrictNamedArgs

cmd/api/src/database/assetgrouptags_integration_test.go
```diff
@@ -1185,6 +1347,136 @@ func TestDatabase_GetAssetGroupTagSelectors(t *testing.T) {
 	})
 }
 
+func TestDatabase_GetAssetGroupTagSelectorsByExtensionId(t *testing.T) {
+
```
bsheth711, line 1442: This test could be parallel--use t.Parallel()

cmd/api/src/database/assetgrouptags_integration_test.go
```diff
@@ -1185,6 +1347,136 @@ func TestDatabase_GetAssetGroupTagSelectors(t *testing.T) {
...
+			assert.Equal(t, expected.AllowDisable, selector.AllowDisable)
+			assert.Equal(t, expected.IsDefault, selector.IsDefault)
+			assert.Equal(t, model.AssetGroupTierZeroPosition, selector.AssetGroupTagId)
+			assert.Equal(t, model.AssetGroupActorOpenGraphExtensionManagement, selector.CreatedBy)
+			assert.Equal(t, model.AssetGroupActorOpenGraphExtensionManagement, selector.UpdatedBy)
+			assert.NotZero(t, selector.ID)
+		}
+	}
+
+	tests := []testCase{
```
bsheth711, line 1497: These subtests use the same testSuite but perform mutation. I would either reset the state between subtests or use separate test suites for each test so that they can all expect a clean slate. Also, not sure I would go with a table driven test for an integration test.

cmd/api/src/database/assetgrouptags_integration_test.go
LawsonWillard: ~~Not opposed to adding it, however I dont think our integration test setup makes use of parallel unless thats changed recently.~~

Im wrong, took a look again and seems like parallel should be available at the top level test suite.

cmd/api/src/database/assetgrouptags.go
```diff
@@ -215,6 +251,59 @@ func (s *BloodhoundDB) DeleteAssetGroupTagSelector(ctx context.Context, user mod
...
+			return ErrNotFound
+		} else if result := tx.Exec(fmt.Sprintf("DELETE FROM %s WHERE selector_id = ?", model.SelectorSeed{}.TableName()), selector.ID); result.Error != nil {
+			return CheckError(result)
+		} else if seeds, err := insertSelectorSeeds(tx, selector.ID, input.Seeds); err != nil {
+			return err
+		} else {
+			selector.Seeds = seeds
+		}
+
+		return bhdb.CreateAssetGroupHistoryRecord(ctx, model.AssetGroupActorOpenGraphExtensionManagement, "", selector.Name, model.AssetGroupHistoryActionUpdateSelector, selector.AssetGroupTagId, null.String{}, null.String{})
```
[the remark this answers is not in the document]
LawsonWillard: I believe this is intentional, however, while im not super familiar with seeds, reconcile can probably handle the upsert process rather than needing to always delete then re create if needed, otherwise we can keep it as is!

cmd/api/src/database/assetgrouptags_integration_test.go
LawsonWillard: ~~I can add a tear down for these tests so it resets state between each!~~

LawsonWillard: Not opposed to moving it out of table driven, but they do keep all tests for a given function together/organized and using the setup, assert and teardown funcs still gives the flexibility needed for quirks in individual test cases.

cmd/api/src/api/v2/assetgrouptags.go
```diff
@@ -87,6 +89,12 @@ type assetGroupTagSelectorRequest struct {
 	AutoCertify *model.SelectorAutoCertifyMethod `json:"auto_certify"`
 	Description *string                          `json:"description"`
 	Disabled    *bool                            `json:"disabled"`
+	ExtensionId *int32                           `json:"extension_id"`
+	RuleKey     *string                          `json:"rule_key"`
+}
+
+func (s assetGroupTagSelectorRequest) hasReadOnlyFields() bool {
```
LawsonWillard: Maybe rename to haseExtensionOnlyFields? I feel like read only might be a bit too generic

cmd/api/src/database/migration/migrations/20260814142955_v9_add_asset_group_tag_selector_extension_ownership.sql
```diff
@@ -0,0 +1,56 @@
...
+            ADD CONSTRAINT asset_group_tag_selectors_extension_id_fkey
+            FOREIGN KEY (extension_id)
+            REFERENCES schema_extensions(id)
+            ON DELETE CASCADE;
+    END IF;
+END $$;
+-- +goose StatementEnd
+
+-- index for looking up selectors by rule_key and ensuring uniqueness of rule_key
+CREATE UNIQUE INDEX IF NOT EXISTS idx_asset_group_tag_selectors_rule_key
```
LawsonWillard, line 60: I think we may want to add extension_id here, this way rule keys are unique to an extension, not all of BloodHound. I think we can also get rid of the WHERE clause as postgres treats NULL as unique unless specified with `NULLS NOT DISTINCT`

cmd/api/src/database/assetgrouptags_integration_test.go
LawsonWillard: I originally had these running with a new test suite / db per sub test, i've lifted the setup to run just once and added a teardown func to ensure there's no mutation that could affect other tests.

cmd/api/src/database/assetgrouptags_integration_test.go
LawsonWillard: should be added now!

cmd/api/src/api/v2/assetgrouptags_test.go
benwaples: added, thank you!

cmd/api/src/api/v2/assetgrouptags_test.go
benwaples: added!

cmd/api/src/database/assetgrouptags.go
benwaples: Yes Im not opposed to fewer magic strings! This is repeated in other places so Ill see if I can refactor to a shared variable

cmd/api/src/database/assetgrouptags.go
benwaples: Love the suggestion!

Looks like GORM doesnt like the defined type `pgx.StrictNamedArgs` but does accept the underlying type of `map[string]any(sqlArgs)` so I went with that so I didnt have to refactor these two helpers to pgx

cmd/api/src/database/assetgrouptags.go
benwaples: updated with `map[string]any` to support GORM and not migrate these to PGX for now

cmd/api/src/api/v2/assetgrouptags.go
benwaples: love the specificity, updated!

cmd/api/src/database/migration/migrations/20260814142955_v9_add_asset_group_tag_selector_extension_ownership.sql
benwaples: Agreed it would be a gotcha if a rule failed to insert because the `rule_key` was already used by another extension, great catch! 

I do want to note that adding `extension_id` to this unique index, and the fact that `extension_id` is nullable, allows multiple rows with the same rule_key and extension_id = null because null is seen as unique. To counteract this, what do you think about adding a check constraint to ensure both are null or both are not null?

And the where clause is useful in this case because it narrows the scope of the index and keeps rules without a rule_key out of the index, which shooould make the looks ups with `rules_key`, `extensions_id` a little faster.

cmd/api/src/database/assetgrouptags.go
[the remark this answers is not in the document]
benwaples: updated to match the existing update method where we dont touch the seeds if seeds are not passed! So we should be good to not have to handle this on the reconcile side!

cmd/api/src/database/migration/migrations/20260814142955_v9_add_asset_group_tag_selector_extension_ownership.sql
LawsonWillard: Yeah i think the constraint is worth it! Adds a defensive layer that I think is needed. 

Forgot about partial indexes ahahah, yeah I agree leaving it in would be best!

benwaples: added the check and kept the where clause!

bsheth711 (approved):

cmd/api/src/database/assetgrouptags_integration_test.go
```diff
@@ -80,6 +118,220 @@ func TestDatabase_CreateAssetGroupTagSelector(t *testing.T) {
...
+	_, err = dbInst.UpdateAssetGroupTagSelector(testCtx, testActor.ID.String(), "", selector)
+	require.NoError(t, err)
+
+	gotSelector, err := dbInst.GetAssetGroupTagSelectorBySelectorId(testCtx, selector.ID)
+	require.NoError(t, err)
+	require.Equal(t, extension.ID, gotSelector.ExtensionId.Int32)
+	require.Equal(t, "test.rule", gotSelector.RuleKey.String)
+}
+
+func TestDatabase_OpenGraphAssetGroupTagSelectorHelpers(t *testing.T) {
```
bsheth711, line 256: Nit: Some of the requires in these test should be asserts. Use require for making sure test setup is correct, use assert for what you are actually testing.

cmd/api/src/database/assetgrouptags_integration_test.go
```diff
@@ -80,6 +118,220 @@ func TestDatabase_CreateAssetGroupTagSelector(t *testing.T) {
...
+	input.Name = "test selector with same rule key in other extension"
+	_, err = createOpenGraphSelectorForTest(dbInst, testCtx, otherExtension.ID, input)
+	require.NoError(t, err)
+
+	input.Name = "test selector with duplicate rule key in same extension"
+	_, err = createOpenGraphSelectorForTest(dbInst, testCtx, extension.ID, input)
+	require.ErrorIs(t, err, database.ErrDuplicateAGTagSelectorRuleKey)
+}
+
+func TestDatabase_UpdateAssetGroupTagSelectorPreservesExtensionManagedFields(t *testing.T) {
```
bsheth711, line 219: Nit: Some of the requires in these test should be asserts. Use require for making sure test setup is correct, use assert for what you are actually testing.

cmd/api/src/database/assetgrouptags_integration_test.go
```diff
@@ -80,6 +118,220 @@ func TestDatabase_CreateAssetGroupTagSelector(t *testing.T) {
...
+		Name:            "test selector with extension id only",
+		Description:     "test description",
+		AllowDisable:    true,
+		AutoCertify:     model.SelectorAutoCertifyMethodDisabled,
+		Seeds:           testSeeds,
+	})
+	assertAssetGroupTagSelectorConstraintError(t, err, assetGroupTagSelectorsRuleKeyExtensionIDCheckConstraint)
+}
+
+func TestDatabase_CreateAssetGroupTagSelectorRuleKeyIsUniquePerExtension(t *testing.T) {
```
bsheth711, line 185: Nit: Some of the requires in these test should be asserts. Use require for making sure test setup is correct, use assert for what you are actually testing.

cmd/api/src/database/assetgrouptags_integration_test.go
```diff
@@ -80,6 +118,220 @@ func TestDatabase_CreateAssetGroupTagSelector(t *testing.T) {
 	require.Equal(t, model.AssetGroupHistoryActionCreateSelector, history[0].Action)
 }
 
+func TestDatabase_CreateAssetGroupTagSelectorRuleKey(t *testing.T) {
```
bsheth711, line 121: Nit: Some of the requires in these test should be asserts. Use require for making sure test setup is correct, use assert for what you are actually testing.

bsheth711:
LGTM aside from the nit on require vs assert. Should be a quick fix and then looks ready to merge! 🚀

LawsonWillard (approved):
Agree with Ben's nits but otherwise looks good to me!

cmd/api/src/database/assetgrouptags_integration_test.go
benwaples: Updated to use assert where appropriate, TIL about the difference in these tests. thank you!

cmd/api/src/database/assetgrouptags_integration_test.go
benwaples: Updated to use assert where appropriate, TIL about the difference in these tests. thank you!

cmd/api/src/database/assetgrouptags_integration_test.go
benwaples: Updated to use assert where appropriate, TIL about the difference in these tests. thank you!

cmd/api/src/database/assetgrouptags_integration_test.go
benwaples: Updated to use assert where appropriate, TIL about the difference in these tests. thank you!

PR #3212, opened 2026-08-26
fix: Broken Object Level Auth ETAC Bypass -relay-targets Bed-8843

Useinovski:
## Description

Applied ETAC filtering to GetEdgeRelayTargets() for users with ETAC enabled.

## Motivation and Context

Resolves BED-8843

Added ETAC filtering to GetEdgeRelayTargets()

## How Has This Been Tested?

Unit test. Tested via bruno

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **Bug Fixes**
  - Relay-target requests now validate access to both source and target nodes before returning results.
  - Invalid or inaccessible nodes continue to produce appropriate HTTP 400 or 404 responses.
  - LDAP relay-target responses now correctly respect graph visibility settings, including hidden-node filtering.
  - When ETAC filtering is disabled, LDAP responses include the full set of accessible nodes.


cmd/api/src/api/v2/edge.go
```diff
@@ -65,11 +67,23 @@ func (s *Resources) GetEdgeRelayTargets(response http.ResponseWriter, request *h
 		api.WriteErrorResponse(request.Context(), api.BuildErrorResponse(http.StatusInternalServerError, fmt.Sprintf("Error getting composition for edge: %v", err), request), response)
 	} else if primaryDisplayKinds, err := s.DB.GetPrimaryDisplayKinds(request.Context()); err != nil {
 		api.HandleDatabaseError(request, response, err)
+	} else if user, isUser := auth.GetUserFromAuthCtx(bhctx.FromRequest(request).AuthCtx); !isUser {
```
mvlipka: Nit: can we move this to be the first check
Although very unlikely to happen, it'd be best to avoid all the database queries if there was an error parsing the auth context

cmd/api/src/api/v2/edge_test.go
```diff
@@ -566,6 +570,34 @@ func TestResources_GetEdgeRelayTargets(t *testing.T) {
 				mocks.mockDb.EXPECT().GetPrimaryDisplayKinds(gomock.Any())
 			},
 		},
+		{
+			name: "Successful Request With ETAC Enabled",
```
mvlipka: It doesn't look like this test does anything as we're not returning any data from the graph queries
It's returning nil, so it's not actually testing the filtering

Would also like a test for when ETAC is disabled but a user has an ETAC control on them as well

mvlipka (changes requested):
Logic looks good, but not sure the test is actually testing the logic

cmd/api/src/api/v2/edge_test.go
Useinovski: Tests updated for both ETAC disabled and enabled. I'm not sure what you mean by disabled ETAC but a user still has ETAC control?

stephanieslamb (approved):
Looks like effective tests were added.

mvlipka (approved):

PR #3220, opened 2026-08-28
chore: add prioritization_rank type BED-8620

LucasParraF:
## Description

Added prioritization_rank type to UnifiedFindings

## Motivation and Context

Resolves BED-8620

## How Has This Been Tested?

Tested locally

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **Improvements**
  * Updated protected asset score responses to include time windows, environments, asset-group information, and score data in a clearer structure.
  * Unified findings can now represent an unavailable prioritization rank as either omitted or `null`, improving compatibility with incomplete or unranked results.


cweidenkeller (approved):

specter-flq (approved):

packages/javascript/js-client-library/src/responses.ts
```diff
@@ -488,6 +488,7 @@ export type UnifiedFinding = {
     status: string;
     first_seen: string;
     last_seen: string;
+    prioritization_rank: number;
```
maffkipp: I think we might want to make this field optional and nullable to force consumers to handle those cases, on testing this I am seeing `null` values get populated in the table. my local data is probably more broken than normal, but it is possible!

kpowderly: yes, archived findings & un-ranked findings will return the prioritization_rank as null when the flag is enabled.

we can omit-empty later if that makes it easier for the front end

maffkipp (approved):

dcairnsspecterops:
image


I'm curious what we want to show when the user views null values? This can happen before processing happens for example

PR #3240, opened 2026-09-02
fix(azure): AZAddSecret must not target managed identities (#943)

### LINKED ISSUE #943: Bug: Bloodhound CE:  Edge AZAddSecret on Managed Identities
## Description:
Bloodhound creates the edge _AZAddSecret_ to Managed Identities

No

## Component(s) Affected:
Bloodhound Edges

## Steps to Reproduce:
1. Create a new user in Entra ID
2.  Assign the role Application Administrator to the user (scoped to the whole tenant)
3. Create an Azure resource with a managed identity (exampel VM)
4. Collect the data with AzureHound and import them
5. Check if there are any paths between the user and the managed identity

## Expected Behavior:
The edge _AZAddSecret_ should no be created for Managed Identities

## Actual Behavior /Screenshots/Code Snippets/Sample Files:
The user _test.fec_ has the role Application Administrator and therefore can add secrets to app registrations and enterprise applications.
However, BloodHound create the edge _AZAddSecrets_ for Managed Identities as well:

While it is possible to add secrets to enterprise applications, it is not possible to manage the secrets of managed identities:

## Environment Information:
Bloodhound CE: 6.1.0
Neo4j: 4.4.38
PostgreSQL: 16.4 (Debian 16.4-1.pgdg120+2)
GraphDB version: v6.1.0
API Version: v6.1.0
AzureHound: v2.2.1

## Potential Solution:
Check if the service principal is a managed identity and do not create the edge. The information if the service principal is a Managed Identity is already present at the node.

## Contributor Checklist:

earbona23:
Fixes #943.

## The problem

`postAddSecret` (`packages/go/analysis/azure/post.go`) paired every AddSecret-capable role with **every** application and service principal in the tenant:

```go
for _, role := range addSecretRoles {
    for _, target := range tenantAppsAndSPs {   // no discrimination
        sink.Submit(ctx, post.EnsureRelationshipJob{FromID: role.ID, ToID: target.ID, Kind: azure.AddSecret})
    }
}
```

A managed identity is ingested as a plain `AZServicePrincipal`, so it was swept in and given an inbound `AZAddSecret` edge. That edge is bogus: a managed identity's credentials are issued and rotated by Azure — there is no `addPassword`/`addKey` on it — so an Application Administrator cannot mint one. The path renders as a privilege-escalation route that cannot be walked.

## The fix

The discriminator is subtle and worth stating, because it will misdirect anyone who assumes otherwise: **there is no managed-identity node kind.** `AZManagedIdentity` is a *relationship* (`resource --AZManagedIdentity--> ServicePrincipal`), so the only way to recognise a managed identity is an inbound `AZManagedIdentity` edge. A fix written as a node-kind or property check silently matches nothing.

`managedIdentityServicePrincipalIDs` collects the end nodes of `AZManagedIdentity` edges, and `postAddSecret` skips those targets. The exclusion is applied **in `postAddSecret`, not in the shared `TenantApplicationsAndServicePrincipals` query** — `postAddOwner` (post.go L620) uses that same query, and whether `AZAddOwner` to a managed identity is equally bogus is a *different* question. I have left AddOwner untouched here and am happy to look at it separately rather than change it as a side effect.

@rvazarkar named `packages/go/analysis/azure/post.go#addSecret` as the location on the issue — this is that change.

## Test

`AZAddSecretHarness` gains a VM-backed managed identity (a VM, a managed-identity SP, and the `AZManagedIdentity` edge between them). This makes the regression load-bearing:

| | targets | AddSecret edges |
|---|---|---|
| without the fix | AZApp, AZServicePrincipal, **AZManagedIdentitySP** (3) × 2 roles | **6** — test fails |
| with the fix | AZApp, AZServicePrincipal (2) × 2 roles | **4** — test passes |

`RoleAddSecret` still asserts exactly 4 edges, and additionally asserts that no `AZAddSecret` edge targets the managed-identity SP.

## What I verified, and what I could not

- `go build` and `go vet` pass on the changed packages.
- The integration test **binary compiles** with the harness and test changes (`go test -tags integration -c`).
- I could **not run the integration suite locally** — it needs the Postgres + graph-DB harness and my environment cannot pull the images. The edge-count logic above is deterministic from the harness, and this repo's CI runs the integration tests, so I am relying on that to confirm execution. If a maintainer runs `RoleAddSecret` and sees anything other than 4, I would want to know immediately.

Small, single-purpose change: 3 files, +50/-5, no change to the shared query or to AddOwner.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Azure permission analysis by preventing managed identity service principals from being incorrectly identified as targets for secret-add capabilities.
  - Azure-managed identities are now handled according to their credential lifecycle, reducing false-positive access relationships and producing more accurate security findings.


earbona23:

PR #3251, opened 2026-09-02
refactor(a11y): update explore and shared a11y tests - BED-9440

TheNando:
## Description

* **Tests**
  * Expanded accessibility coverage across Explore navigation, search, graph controls, pathfinding, saved queries, uploads, date ranges, selectors, and entity panels.
  * Added accessibility validation for no-data upload dialogs and saved-query actions, including edit, sharing, and deletion dialogs.
  * Improved test reliability with consistent waits, targeted checks for dialogs and menus, and deterministic test scenarios.
  * Added coverage for empty states, filtering, search results, calendar interactions, and pathfinding workflows.

## Motivation and Context

Resolves BED-9440

## How Has This Been Tested?

Manually tested running Playwright locally

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Tests

* Standardized accessibility testing across Explore and shared UI workflows.
* Improved test reliability with consistent navigation waits, deterministic test data, and focused checks for dialogs, menus, controls, and popovers.
* Added coverage for the no-data upload dialog and saved-query actions, including editing, sharing, and deletion.
* Updated accessibility coverage for search, pathfinding, graph controls, uploads, date ranges, selectors, navigation, and entity panels.

### DIFF
--- a/cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
+++ b/cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
@@ -14,56 +14,118 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { expect, expectNoAccessibilityViolations, test } from '../../fixtures';
+import { Page, TestInfo } from '@playwright/test';
+import { hideBySelector, test } from 'bh-playwright-testing';
 
-test.describe('WCAG A/AA Violations - Explore - Cypher Tab', () => {
-    test.beforeEach(async ({ page }) => {
-        await page.goto('/ui/explore');
+const installSaveQueryStub = async (page: Page, testInfo: TestInfo) => {
+    const queryName = `a11y-export-${testInfo.project.name}-${Date.now()}`;
 
-        const cypherTab = page.getByRole('tab', { name: 'Cypher' });
-        await cypherTab.click();
-        await page.getByRole('textbox', { name: 'Cypher Editor' }).waitFor({ state: 'visible' });
+    const savedQuery = {
+        id: 1,
+        name: queryName,
+        description: '',
+        query: 'MATCH (n) RETURN n LIMIT 10',
+        user_id: '0e712979-d6e9-4496-9f31-f96f5565873e',
+        scope: 'owned',
+    };
+
+    const self = {
+        data: {
+            sso_provider_id: null,
+            AuthSecret: {
+                digest_method: 'argon2',
+                expires_at: '2026-11-15T14:32:05.868773Z',
+                id: 1,
+            },
+            first_name: 'BloodHound',
+            last_name: 'Dev',
+            email_address: 'spam@example.com',
+            principal_name: 'admin',
+            last_login: '2026-08-24T17:12:33.222043Z',
+            is_disabled: false,
+            all_environments: true,
+            environment_targeted_access_control: [],
+            eula_accepted: true,
+            id: '0e712979-d6e9-4496-9f31-f96f5565873e',
+        },
+    };
+
+    // Fetch the real /api/v2/self response and override the id so the logged-in
+    // user (self) owns the stubbed saved query and can edit it. The rest of the
+    // self payload is preserved so the page keeps loading normally.
+    await page.route('**/api/v2/self', async (route) => {
+        const request = route.request();
+        const pathname = new URL(request.url()).pathname;
+
+        if (request.method() === 'GET' && pathname === '/api/v2/self') {
+            return route.fulfill({ json: self });
+        }
+
+        return route.fallback();
     });
 
-    test('Empty query', async ({ page, makeAxeBuilder }, testInfo) => {
+    // Stub the saved queries list so a single owned query already exists, and the
+    // per-query permissions lookup opened by the Edit/Share dialog. Any other
+    // saved-query requests fall through to the real API.
+    await page.route('**/api/v2/saved-queries**', async (route) => {
+        const request = route.request();
+        const pathname = new URL(request.url()).pathname;
+
+        if (request.method() === 'GET' && pathname === '/api/v2/saved-queries') {
+            return route.fulfill({
+                json: { data: [savedQuery], count: 1, limit: 10, skip: 0 },
+            });
+        }
+
+        if (request.method() === 'GET' && pathname === `/api/v2/saved-queries/${savedQuery.id}/permissions`) {
+            return route.fulfill({
+                json: { data: { query_id: savedQuery.id, public: false, shared_to_user_ids: [] } },
+            });
+        }
+
+        return route.fallback();
+    });
+};
+
+test.describe('WCAG A/AA Violations - Explore - Cypher Tab', () => {
+    test.beforeEach(async ({ goAndWaitFor, page }, testInfo) => {
+        await goAndWaitFor('/ui/explore?exploreSearchTab=cypher', page.getByRole('textbox', { name: 'Cypher Editor' }));
+    });
+
+    test('Empty query', async ({ page, checkA11y }) => {
         const cypherEditor = page.getByRole('textbox', {
             name: 'Cypher Editor',
         });
 
-        await cypherEditor.waitFor({ state: 'visible' });
+        await cypherEditor.waitFor();
 
-        const results = await makeAxeBuilder().include('#content-wrapper').analyze();
-
-        await expectNoAccessibilityViolations(testInfo, results, { page });
+        await checkA11y();
     });
 
-    test('With full query', async ({ page, makeAxeBuilder }, testInfo) => {
+    test('With full query', async ({ page, checkA11y }) => {
         const query = 'MATCH (n) RETURN n LIMIT 10';
         const cypherEditor = page.getByRole('textbox', {
             name: 'Cypher Editor',
         });
 
-        await cypherEditor.waitFor({ state: 'visible' });
+        await cypherEditor.waitFor();
 
         await cypherEditor.fill(query);
 
-        const results = await makeAxeBuilder().include('#content-wrapper').analyze();
-
-        await expectNoAccessibilityViolations(testInfo, results, { page });
+        await checkA11y();
     });
 
-    test('Tag Results to Zone dialog', async ({ page, makeAxeBuilder }, testInfo) => {
+    test('Tag Results to Zone dialog', async ({ page, checkA11y }) => {
         const query = 'MATCH (n) RETURN n LIMIT 10';
         const cypherEditor = page.getByRole('textbox', {
             name: 'Cypher Editor',
         });
 
-        await cypherEditor.waitFor({ state: 'visible' });
-
+        await cypherEditor.waitFor();
         await cypherEditor.fill(query);
 
         const tagButton = page.getByRole('button', { name: 'Tag' });
-        await tagButton.waitFor({ state: 'visible' });
+        await tagButton.waitFor();
         await tagButton.click();
 
         await page.getByRole('button', { name: 'Zone' }).click();
@@ -71,7 +133,10 @@ test.describe('WCAG A/AA Violations - Explore - Cypher Tab', () => {
         const dialog = page.getByRole('dialog', {
             name: 'Tag Results to Zone',
         });
-        await dialog.waitFor({ state: 'visible' });
+        await dialog.waitFor();
+
+        await hideBySelector(page, '[data-radix-popper-content-wrapper]');
+        await hideBySelector(page, '#content-wrapper');
 
         const selectZoneControl = dialog.getByRole('combobox');
         const cancelButton = dialog.getByRole('button', { name: 'Cancel' });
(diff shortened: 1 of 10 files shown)


cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
```diff
@@ -14,91 +14,155 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-import { expect, expectNoAccessibilityViolations, test } from '../../fixtures';
+import { Page, TestInfo } from '@playwright/test';
```
TheNando, line 17: These tests have each been updated with the following changes:
* goAndWaitFor fixture which combines separate page.goto and locator().waitFor() calls into one
* Replace all `.waitFor({ state: 'visible' });` with `.waitFor();` as visible state is the default option
* Replace the `makeAxeBuilder()` and `expectNoAccessibilityViolations` combo that appears on each test with the new `checkA11y` fixture which provides sane defaults.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
```diff
@@ -14,91 +14,155 @@
...
+import { hideBySelector, test } from 'bh-playwright-testing';
+
+const installSaveQueryStub = async (page: Page, testInfo: TestInfo) => {
```
TheNando, line 20: This test module also has expanded coverage and better stubs that more accurately reflect real data.

cmd/ui/tests/a11y/Explore/entity-info-panel.a11y.spec.ts
```diff
@@ -172,11 +172,14 @@ test.describe('WCAG A/AA Violations - Explore - Entity Information Panel', () =>
 
         await infoPanel.getByText('RELATED_USER_0@PLAYWRIGHT.LOCAL').waitFor({ state: 'visible' });
 
-        const results = await makeAxeBuilder().include('#content-wrapper').analyze();
-        await expectNoAccessibilityViolations(testInfo, results, { page });
+        await hideBySelector(page, '[data-testid="explore_search-container"]');
```
TheNando, line 175: Tests that check dialogs mostly now will hide the content under the dialog. This often prevents "incomplete" test states which give false positives.

cmd/ui/tests/a11y/Explore/explore-graph-controls.a11y.spec.ts
```diff
@@ -98,94 +100,91 @@ const installExploreGraphRoute = async (page: Page) => {
     });
 };
 
-const loadExploreGraph = async (page: Page) => {
```
TheNando, line 101: This ended up being redundant.

cmd/ui/tests/a11y/Explore/pathfinding.a11y.spec.ts
```diff
@@ -13,22 +13,40 @@
 // limitations under the License.
 //
 // SPDX-License-Identifier: Apache-2.0
-import { expect, expectNoAccessibilityViolations, test } from '../../fixtures';
+import { Page } from '@playwright/test';
+import { expect, test } from 'bh-playwright-testing';
+
+const installPathfindingStub = async (page: Page) => {
```
TheNando, line 19: There was a lot of cleanup on this module that was needed due to a missing stub which is provided here.

cmd/ui/tests/a11y/Explore/pathfinding.a11y.spec.ts
```diff
@@ -68,28 +85,24 @@ test.describe('WCAG A/AA Violations - Explore - Pathfinding Tab', () => {
...
 
         const noResultsMessage = `No results found for "${searchTerm}"`;
         await expect(page.getByText(noResultsMessage)).toBeVisible();
 
-        const results = await makeAxeBuilder().include('#content-wrapper').analyze();
-        await expectNoAccessibilityViolations(testInfo, results, { page });
+        await checkA11y();
     });
 
-    test('Destination Node', async ({ page, makeAxeBuilder }, testInfo) => {
```
TheNando, line 81: Many of these vars are only used once so they have been refactored into chains instead.

cmd/ui/tests/a11y/Explore/pathfinding.a11y.spec.ts
```diff
@@ -111,21 +124,15 @@ test.describe('WCAG A/AA Violations - Explore - Pathfinding Tab', () => {
...
-        const destinationNodeField = page.getByLabel('Destination Node');
-        await destinationNodeField.click();
-        await destinationNodeField.fill(searchTerm);
+        // Pathfinding autofocus opens the Start node popup over the Destination Node field.
+        await page.getByRole('textbox', { name: 'Start Node' }).press('Escape');
+        await page.getByRole('textbox', { name: 'Destination Node' }).fill(searchTerm);
+        await page.getByText('DESTINATION TEST RESULT').waitFor();
 
-        const searchResult = page.getByRole('option').filter({ hasText: searchResultName });
-        await expect(searchResult).toBeVisible();
```
TheNando, line 122: All assertions have been removed in favor of waitFors which are more expressive of intent.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
```diff
@@ -209,968 +276,126 @@ test.describe('WCAG A/AA Violations - Explore - Cypher Tab', () => {
...
-                name: 'Query Name',
-                exact: true,
-            });
-            const saveButton = saveQueryDialog.getByRole('button', {
-                name: 'Save',
-                exact: true,
-            });
+        // Select the dots action from the saved query
+        await page.getByTestId('saved-query-action-menu-trigger').click();
+        await page.getByRole('button', { name: 'Edit/Share' }).click();
```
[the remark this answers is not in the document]
TheNando: I will address this in my next commit.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
```diff
@@ -209,968 +276,127 @@ test.describe('WCAG A/AA Violations - Explore - Cypher Tab', () => {
...
-        }
-
-        if (testError !== undefined) {
-            throw testError;
-        }
-
-        if (cleanupError !== undefined) {
-            throw cleanupError;
-        }
+        await checkA11y();
```
JayC-Specter, line 400: Should we wait for the Delete Query dialog here and scope the scan to it? checkA11y() defaults to #content-wrapper, but that wrapper is hidden above, so I think this can pass without actually scanning the dialog. Could we use checkA11y({ include: '[role="dialog"]' }) here like the other dialog tests?

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
```diff
@@ -14,91 +14,155 @@
...
+        user_id: '0e712979-d6e9-4496-9f31-f96f5565873e',
+        scope: 'owned',
+    };
+
+    const self = {
+        data: {
+            sso_provider_id: null,
+            AuthSecret: {
+                digest_method: 'argon2',
+                expires_at: '2026-11-15T14:32:05.868773Z',
```
JayC-Specter, line 37: Should we avoid using a real near-term expiration date here? Once we get past November 15, 2026 , won't this user will be treated as having an expired password and these saved-query tests will start redirecting away from Explore.

cmd/ui/tests/a11y/Explore/pathfinding.a11y.spec.ts
```diff
@@ -177,34 +178,21 @@ test.describe('WCAG A/AA Violations - Explore - Pathfinding Tab', () => {
...
-        const destinationNodeField = page.getByLabel('Destination Node');
-        await destinationNodeField.click();
-        await destinationNodeField.fill(destinationResultName);
+        await page.getByRole('textbox', { name: 'Destination Node' }).fill(destinationResultName);
+        await page.getByRole('option').filter({ hasText: destinationResultName }).click();
 
-        const destinationResult = page.getByRole('option').filter({ hasText: destinationResultName });
-        await expect(destinationResult).toBeVisible();
-        await destinationResult.click();
+        await page.getByRole('button', { name: 'Swap start and destination' }).waitFor();
```
JayC-Specter, line 189: Since this test is specifically covering the enabled state, should we keep the toBeEnabled() check here? The swap button is still rendered while disabled, so waitFor() only confirms that it's visible and doesn't guarantee Axe runs against the intended enabled state.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
```diff
@@ -209,968 +276,127 @@ test.describe('WCAG A/AA Violations - Explore - Cypher Tab', () => {
...
+        await hideBySelector(page, '#content-wrapper');
 
-        if (cleanupError !== undefined) {
-            throw cleanupError;
-        }
+        await checkA11y({ include: '[role="dialog"]' });
     });
 
-    test('Saved Queries - Filter by Platforms', async ({ page, makeAxeBuilder }, testInfo) => {
+    test('Saved Queries - With filter', async ({ page, checkA11y }) => {
```
JayC-Specter, line 314: I noticed this refactor removes the existing saved-query export, Categories filter, and Source filter scenarios, while this replacement only covers the Platforms filter. Was that coverage intentionally removed?

JayC-Specter (changes requested):

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
bfaulk96, line 1: Just want to double-check... There is a ton of deleted code here (try/catch/finally blocks). Is this possible because of the endpoint mocking?

cmd/ui/tests/a11y/Explore/entity-info-panel.a11y.spec.ts
bfaulk96, line 1: Just personal curiosity on this file and some others, what is `hideBySelector` and why do we need it in several places now?

cmd/ui/tests/a11y/Explore/pathfinding.a11y.spec.ts
```diff
@@ -111,21 +124,15 @@ test.describe('WCAG A/AA Violations - Explore - Pathfinding Tab', () => {
...
-        // Pathfinding autofocus opens the Start Node popup over the Destination Node field.
-        await page.getByLabel('Start Node').press('Escape');
-
-        const destinationNodeField = page.getByLabel('Destination Node');
-        await destinationNodeField.click();
-        await destinationNodeField.fill(searchTerm);
+        // Pathfinding autofocus opens the Start node popup over the Destination Node field.
+        await page.getByRole('textbox', { name: 'Start Node' }).press('Escape');
+        await page.getByRole('textbox', { name: 'Destination Node' }).fill(searchTerm);
+        await page.getByText('DESTINATION TEST RESULT').waitFor();
```
bfaulk96, line 130: Probably best not to hardcode this - we define a variable on line 107 above
```suggestion
        await page.getByText(searchResultName).waitFor();
```

bfaulk96 (commented):
Impressive how much you were able to simplify some of these file! 🔥

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
bfaulk96: Agreed, we should probably dynamically create the date so it's *always* consistently in the future

TheNando: Good catch. I need to be better about checking the fine details of the mock data.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
TheNando: I'm not sure if a previous PR change broke this or what, but neither of the tests with the try/catch blocks were working as desired. They were relying on the failures caused by missing stubs to get to a testable state, and that state didn't really reflect the data we needed to see. I rewrote them from scratch with updated stubs.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
TheNando: Considering that each of these filters is using the same base component with just differing data, it seems redundant to test each of them. We just want to ensure that interacting with one of the components gets to a filtered data state with no a11y issues.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
TheNando: 👀 Yeah. I'll update this on Tuesday.

cmd/ui/tests/a11y/Explore/entity-info-panel.a11y.spec.ts
TheNando: It's an imperfect solution to a weird problem with Playwright. I noticed that many times when there was a dialog or content floating over other content, it would always return a pass, even if I intentionally tried to cause a fail by, for example, setting a low contrast color. So false positives. I noticed in the large generated JSON data that it was assessing these tests with "incomplete" results each time and when I forced a fail, the incomplete always included the contrast issue, but it would still pass the test.

After a little research, I found that there was data in the console that stated something to the effect of "can't assess due to overlapping content". So I tried "turning off" the content underneath with this custom hide fixture, and that got the test to fail with the proper a11y violations. But it isn't fool proof, and it only works easily when the underlying content is not a parent of the dialog. Good enough for now, but I'm still looking for a better solution.

cmd/ui/tests/a11y/Explore/pathfinding.a11y.spec.ts
TheNando: Hmm. Yeah. That is a sensible assertion. We might not need to explicitly test the disabled state button apart from the button component test, but until we have component lib tests, this is a good call.

cmd/ui/tests/a11y/Explore/cypher.a11y.spec.ts
JayC-Specter: Was removing the Destination Node tests intentional? The enabled-controls test still selects a destination, but it looks like we're no longer checking the open, results, and no-results states with Axe.

TheNando: Yes, it was. I removed the destination node tests on purpose as they are using the same components as the start nodes so they provide no additional value.

JayC-Specter (approved):

PR #3264, opened 2026-09-04
feat(a11y): a11y scan hardening (BED-9440)

bfaulk96:
## Description

Stacked on #3251.

- Upgrade `axe-core` and `@axe-core/playwright` to 4.13.
- Make the shared accessibility helper report incomplete results and provide an opt-in `failOnIncomplete` check, while keeping incomplete scans non-fatal by default.
- Scope dialog and entity-panel scans to their visible UI rather than hiding underlying application content.
- Replace FileDrop's invisible, absolute file-input overlay with a native visible button; add interaction coverage for clicking its icon and instructional text.
- Disable React Query Devtools only for the Playwright Vite mode and refine Quick Upload file-status styling.

Related: BED-9440

## How Has This Been Tested?

- `yarn test:a11y tests/a11y/Shared/quick-upload.a11y.spec.ts` — 17 passed across Chromium/Firefox in light and dark modes.
- `yarn workspace bh-shared-ui check-types`
- `yarn workspace bh-shared-ui lint`
- `yarn workspace bh-shared-ui check-format`
- `yarn test:a11y --project=chromium-light` — Quick Upload, no-data dialog, and entity-panel coverage passed; 81 tests passed overall, with 34 unrelated existing failures elsewhere in the suite.

## Screenshots (optional)

Screenshot 2026-09-04 at 10 35 17 AM
Screenshot 2026-09-04 at 12 30 11 PM
Screenshot 2026-09-04 at 12 30 25 PM
Screenshot 2026-09-04 at 12 31 47 PM
Screenshot 2026-09-04 at 12 30 34 PM

## Summary by CodeRabbit

- **Accessibility**
  - Improved accessibility checks for dialogs, upload flows, and entity panels.
  - Accessibility reports now include incomplete results and can fail when those results are detected.

- **Bug Fixes**
  - Improved FileDrop interaction, including clicking, dragging, disabled states, and file selection.
  - Progress indicators now display more accurately across upload states.
  - Upload-in-progress messaging is now presented as a clear informational alert.

- **Testing**
  - Added comprehensive coverage for FileDrop behavior and accessibility scenarios.


cmd/ui/src/main.tsx
```diff
@@ -80,7 +80,7 @@ const main = async () => {
     root.render(
         <Provider store={store}>
             <QueryClientProvider client={queryClient}>
-                <ReactQueryDevtools position='bottom-right' />
+                {import.meta.env.MODE !== 'playwright' && <ReactQueryDevtools position='bottom-right' />}
```
bfaulk96, line 83: We have no reason to debug with react query devtools when running playwright tests, which should resemble production. The RQ Devtools button floats on the page, which can potentially cause stacking issues with the a11y checks

cmd/ui/playwright.a11y.config.ts
```diff
@@ -43,7 +43,7 @@ const browsers = [
 ] as const;
 
 const webServer = {
-    command: `yarn dev --host ${hostname} --port ${port || '3000'}`,
+    command: `yarn dev --mode playwright --host ${hostname} --port ${port || '3000'}`,
```
bfaulk96, line 46: We use this mode to know not to render the RQ Dev tools

packages/javascript/bh-shared-ui/src/components/FileDrop/FileDrop.tsx
bfaulk96, line 1: This component ultimately looks exactly like it did before, with a couple of changes to *how* it is designed:

- The button is now the full-size element, rather than having a full-size div with a button absolutely positioned over top of it (This was causing a11y stacking issues which prevented tests from properly checking failures)
- Button now directly contains the icon and text, which have `pointer-events-none` so that they don't interfere with drag/drop detections

packages/javascript/bh-shared-ui/src/components/FileStatusListItem/FileStatusListItem.tsx
bfaulk96, line 1: This also looks similar to how it did before, with some meaningful changes:
- Percentage is now a purple color (primary in light mode, near-white purple in dark) to help differentiate the percentage from the filename
- Percentage or "Failed to Upload" message are now right-aligned to provide a better visual experience (see screenshots)
- On error or completion, we use the background color of the row itself rather than a percentage bar, for a11y stacking improvements
- Used flex gaps and spacing for more uniform spacing than previously
- Added padding to the right side now that we have right-aligned text

packages/javascript/bh-shared-ui/src/components/FileUploadDialog/FileUploadDialog.tsx
```diff
@@ -108,10 +108,9 @@ const FileUploadDialog: React.FC<{
                 )}
 
                 {currentlyUploading && (
-                    <div>
-                        <p>Upload in progress.</p>
-                        <p>You can continue using the platform&mdash;we will alert you once the upload is complete.</p>
-                    </div>
+                    <Alert title='Upload in Progress' className='mb-2' variant='info'>
+                        You can continue using the platform&mdash;we will alert you once the upload is complete.
+                    </Alert>
```
bfaulk96, line 113: This block of plain text previously looked pretty bad popping up in the dialog; This Alert feels a lot more natural as an informational popup (see screenshot)

cmd/ui/src/main.tsx
TheNando: Nice

packages/javascript/bh-playwright-testing/src/axe.ts
```diff
@@ -185,6 +188,7 @@ export const test = base.extend<AxeFixtures, TestOptions>({
                 include = a11yDefaultInclude,
                 exclude,
                 disableRules,
+                failOnIncomplete = false,
```
TheNando, line 191: I'm partially thinking that if we go this route, we may as well go all-in on failing incompletes, just make it on all the time so the dev doesn't have to think about it. Though, unless we can find and address what causes the incomplete when it happens, we'll likely still end up needing some mechanism of explicitly accepting or dismissing an incomplete when they come up. We don't necessarily need to figure that out for this PR.

TheNando (commented):
I still want to poke around a little more tomorrow, but I have some feedback. This is a good sized PR, so the following notes are in the spirit of the not-always-convenient "ideal" state.

Ideally:

* we'd handle package updates in isolation instead of mixing them with features. This is crucial for major versions with break change migrations

* we'd have a tighter scoped, atomic PR just addressing the hardening mentioned in the PR name and have a separate PR for the FileDrop coverage. This one is definitely a nit pick.

If the PR were large, I'd be more adamant about these concerns, but this is all low-risk, so no prob.

packages/javascript/bh-playwright-testing/src/axe.ts
bfaulk96: Yeah, I found one edge case to be transparent gradients. For example, the loading bar on file upload, when at, say, 50%, has both a transparent color on part of it and a fully transparent half, which disallows proper checking. In this particular case, it wasn't really something we needed to test, since I could successfully test the 0% and 100% which gives us the accessibility of both color states.

bfaulk96: As for defaulting to *not* fail, I actually originally did have it the other way (the boolean was `ignoreIncomplete` and defaulted to false), but it resulted in like 85+ failures or something crazy and I didn't want to affect existing tests so heavily. Perhaps that is the right route though, using the opposite default and just going through and setting existing tests to the legacy way.

TheNando (approved):

PR #3277, opened 2026-09-08
feat(Nav): auto-expand/contract when resizing at breakpoint - BED-9547

TheNando:
## Description

* Add automatic expand and collapse when resizing through large width breakpoint
    * When the user resizes the browser width to from less than 1280px to greater than, the nav menu will automatically expand
    * When the user resizes the browser width to from greater than 1280px to less than, the nav menu will automatically contract
    * If the nav is manually expanded/contracted after that, the setting will keep on refresh.

## Motivation and Context

Resolves BED-9547

## How Has This Been Tested?

Manual test and unit tests

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Navigation now automatically expands on extra-large screens and collapses on smaller screens.
  - Navigation responds when the browser crosses the extra-large breakpoint while preserving user-selected state when the breakpoint does not change.
  - Responsive navigation state is preserved during initial page loads and browser refreshes.

- **Bug Fixes**
  - Improved navigation behavior during window resizing.
  - Improved media-query handling when browser media-query APIs are unavailable.

### DIFF
--- a/packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
+++ b/packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
@@ -26,8 +26,10 @@ export const BREAKPOINTS = {
     '3xl': '1920px',
 };
 
+const isMatchMediaAvailable = (): boolean => typeof window !== 'undefined' && typeof window.matchMedia === 'function';
+
 const getMediaQueryMatches = (query: string): boolean => {
-    if (typeof window === 'undefined' || typeof window.matchMedia !== 'function') return false;
+    if (!isMatchMediaAvailable()) return false;
 
     return window.matchMedia(query).matches;
 };
@@ -36,7 +38,7 @@ export const useMediaQuery = (query: string): boolean => {
     const [matches, setMatches] = useState(() => getMediaQueryMatches(query));
 
     useEffect(() => {
-        if (typeof window.matchMedia !== 'function') return;
+        if (!isMatchMediaAvailable()) return;
 
         const mediaQueryList = window.matchMedia(query);
 
--- a/packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
+++ b/packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
@@ -26,13 +26,21 @@ export const BREAKPOINTS = {
     '3xl': '1920px',
 };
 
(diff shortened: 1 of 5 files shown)


packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
```diff
@@ -26,13 +26,21 @@ export const BREAKPOINTS = {
     '3xl': '1920px',
 };
 
+const getMediaQueryMatches = (query: string): boolean => {
+    if (typeof window === 'undefined' || typeof window.matchMedia !== 'function') return false;
```
bfaulk96: nit: can be simplified
```suggestion
    if (typeof window?.matchMedia !== 'function') return false;
```

packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
```diff
@@ -26,13 +26,21 @@ export const BREAKPOINTS = {
...
+
+    return window.matchMedia(query).matches;
+};
+
 export const useMediaQuery = (query: string): boolean => {
-    const [matches, setMatches] = useState(false);
+    const [matches, setMatches] = useState(() => getMediaQueryMatches(query));
 
     useEffect(() => {
+        if (typeof window.matchMedia !== 'function') return;
```
bfaulk96: This should probably be the same check as what is used in `getMediaQueryMatches`
```suggestion
        if (typeof window?.matchMedia !== 'function') return;
```

bfaulk96 (approved):
Two nitpicks, but looks good otherwise, and works well running it locally

packages/javascript/bh-shared-ui/src/hooks/useMediaQuery/useMediaQuery.tsx
TheNando: There is a small advantage to the explicit check. With the optional chain, we'll correctly check for the feature. But it doesn't check for window existing. This is mostly a moot point since our app should have either a real window or a stub (Vitest). But I'd still lean towards the explicit pattern because it gives us built-in SSR compatibility should we ever decide to go that way. But I'll totally update the other condition to match.

image

bfaulk96: Makes sense! My bad!

PR #3282, opened 2026-09-09
chore: Improve cross extension findings - BED-9566

brandonshearin:
## Description
This migration changes the meaning of `schema_findings.environment_id`.
Before this change, it referenced `schema_environments.id`. After this change, it references the global DAWGS `kind.id` for the environment kind. This is required so a relationship finding can reference an environment declared by another extension, including when the "dependent" extension is loaded later.

**migration risk**
(imo low risk, with a known recovery path)
A rollback will fail if a schema_finding names an environment_kind whose owning extension is never installed, or is later removed. In either case, there is no "schema_environment" mapping to rollback to.

As currently written, the migration's Down path rejects rollback in that case. If the feature must be removed after deployment, the recovering from this exception would need manual remediation on any tenant that violates the condition above. Manual remediation would involve removing any schema_finding created that wouldn't have a corresponding schema_environment mapping.

Builtin extensions aren't victims of this problem, and embedded extensions aren't either (because all of these extensions have schema_environment mappings). So at the end of the day it really is an edge case, but due to the manual remediation required for a true rollback I want to call it out.

## Motivation and Context
Resolves BED-9566
It is hard to load extensions which contain hybrid findings. This work decouples extensions such that the upload order no longer matters.  It achieves this decoupling by removing the dependency that a relationship finding definition had on another extensions environment definitions. This coupling was fragile and unneccessary.

## How Has This Been Tested?
- the extension schema upsert unit and integration tests have been brought in to line with the new data model.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

* **New Features**
  * Relationship findings can automatically register previously unknown environment kinds.
  * Findings can reference environment kinds supplied by other extensions.
  * Creating and updating findings no longer requires environment kinds to be configured in advance.

* **Bug Fixes**
  * Improved environment-kind resolution for relationship findings.
  * Finding references are preserved more safely when environments are removed.

### DIFF
--- a/cmd/api/src/database/migration/migrations/20260908174043_v9_all_cross_extension_findings.sql
+++ b/cmd/api/src/database/migration/migrations/20260908174043_v9_all_cross_extension_findings.sql
@@ -1,0 +1,44 @@
+-- +goose Up
+-- schema_findings.environment_id now stores the environment kind ID from dawgs rather than
+-- the schema_environments ID. This permits findings to reference environment kinds defined by another extension.
+ALTER TABLE schema_findings
+    DROP CONSTRAINT IF EXISTS schema_findings_environment_id_fkey;
+
+UPDATE schema_findings sf
+SET environment_id = se.environment_kind_id
+FROM schema_environments se
+WHERE sf.environment_id = se.id;
+
+ALTER TABLE schema_findings
+    ADD CONSTRAINT schema_findings_environment_id_fkey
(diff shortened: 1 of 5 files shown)


LawsonWillard (approved):
Love the docs on the down migration reasoning 🙇 , PR looks good to me! 🚀

PR #3292, opened 2026-09-11
fix: Failure to create cypher query audit log: context canceled BED-8525

AD7ZJ:
## Description

If a request context is cancelled after the audit log intent is written, it's possible for the audit log result to not get written since it's running under the same context. 

## Motivation and Context

Resolves https://specterops.atlassian.net/browse/BED-8525

## How Has This Been Tested?

Tests have been added to demonstrate the issue for the cypher endpoint is resolved. 

## Types of changes

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved audit-log reliability for graph queries and mutations when requests are canceled or interrupted.
  - Audit records now capture failed outcomes by default and update to successful only after execution completes successfully.
  - Audit-log updates use a bounded timeout, preventing them from delaying requests indefinitely.
  - Original graph query errors are preserved even if recording the audit outcome encounters an error.
  - Ensured audit outcomes remain recorded even after the originating request is canceled.

### DIFF
--- a/cmd/api/src/api/v2/cypherquery.go
+++ b/cmd/api/src/api/v2/cypherquery.go
@@ -36,7 +36,7 @@ import (
 	"github.com/specterops/dawgs/util"
 )
 
-const auditLogOutcomeTimeout = time.Second
+const auditLogOutcomeTimeout = time.Second * 30
 
 var errUnauthorizedGraphMutation = errors.New("unauthorized graph mutation")
 
--- a/cmd/api/src/api/v2/cypherquery.go
+++ b/cmd/api/src/api/v2/cypherquery.go
@@ -17,11 +17,13 @@
 package v2
 
 import (
+	"context"
 	"errors"
 	"log/slog"
 	"maps"
 	"net/http"
 	"slices"
+	"time"
 
 	"github.com/specterops/bloodhound/cmd/api/src/api"
 	"github.com/specterops/bloodhound/cmd/api/src/auth"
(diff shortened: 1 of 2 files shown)


cmd/api/src/api/v2/cypherquery.go
```diff
@@ -148,7 +152,10 @@ func (s Resources) CypherQuery(response http.ResponseWriter, request *http.Reque
 	auditLogEntry.Status = model.AuditLogStatusFailure
 
 	defer func() {
-		if err = s.DB.AppendAuditLog(request.Context(), auditLogEntry); err != nil {
+		auditContext, cancelAudit := context.WithTimeout(context.WithoutCancel(request.Context()), auditLogOutcomeTimeout)
```
urangel, line 155: Nice! This looks to fix the issue.

I think it also applies the auditLogOutcomeTimeout to successful queries that are not canceled. Is that expected?

AD7ZJ: Yeah, I debated on having the timeout at all but I thought it's better just in case the write to the DB hung for some reason, the handler wouldn't get stalled indefinitely.

urangel: Okay that makes sense to me to prevent the handler from stalling. 

I might suggest going with 10 or 30 seconds for the timeout to align with the warn/error thresholds set in api/src/database/db.go

AD7ZJ: That's a good idea, thanks for pointing that out!

urangel: Going to ahead and resolve and approve so you are set to merge after making this update 🚀

urangel (approved):

PR #3306, opened 2026-09-13
fix(analysis): handle missing RestrictOutboundNTLM property as Windows default

sanjio3:
Fixes #2534

## Summary

When the `RestrictSendingNtlmTraffic` registry key is absent (the default Windows configuration — the "Network security: Restrict NTLM: Outgoing NTLM traffic to remote servers" policy never configured), the `restrictoutboundntlm` node property was left unset during ingest. `NewNTLMCache` treated any property read error as fail-closed, skipping those computers from `UnprotectedComputersCache` and silently dropping all four `CoerceAndRelayNTLMTo*` edge types in most deployments (a regression from #1787 which changed the registry fields to nullable pointers).

Per [Microsoft's documentation](https://learn.microsoft.com/en-us/previous-versions/windows/it-pro/windows-10/security/threat-protection/security-policy-settings/network-security-restrict-ntlm-outgoing-ntlm-traffic-to-remote-servers), an undefined value is equivalent to "Allow all" — so a collected-but-default computer should be treated as unrestricted.

## Changes

- `packages/go/ein/ad.go`: in `ConvertComputerToNode`, when NTLM registry data was collected but `RestrictSendingNtlmTraffic` is nil (key absent = Windows default), explicitly set `restrictoutboundntlm` to `false` so the analysis can distinguish a collected-but-default computer from one whose registry data was never collected.
- `packages/go/analysis/ad/ntlm.go`: `NewNTLMCache` keeps its fail-closed behavior for computers where the property is genuinely missing (registry data not collected or unreachable), preventing false positives. Added docstring.

## Test plan

- `go build ./packages/go/analysis/ad/ ./packages/go/ein/` and `go vet` -> pass
- Existing integration harnesses explicitly set `RestrictOutboundNTLM` on every computer and are unaffected
- Computers with collected-but-default registry data now resolve to `false` (unrestricted); computers with no collected registry data keep the property absent and remain fail-closed

## Summary by CodeRabbit

- **Bug Fixes**
  - Computer analysis now records outbound NTLM restriction as `false` when the corresponding registry setting is absent, improving consistency in results.
  - NTLM relay analysis now excludes computers when outbound restriction data was not collected, preventing potentially unsafe relay paths.

- **Documentation**
  - Clarified how missing NTLM registry values are interpreted during computer analysis.
  - Documented NTLM relay cache handling when outbound traffic restriction data is unavailable.

- **Tests**
  - Expanded coverage for present and absent NTLM restriction settings.
  - Added validation for safe handling of incomplete computer data.


sanjio3:

chryzsh:
I filed the original issue (#2534), so thanks for picking this up. I went through the change and I think it's right.

One thing: putting the fix in ingest rather than in NewNTLMCache (like the earlier #3305) covers more than it looks. Three places read restrictoutboundntlm and fail closed when it's missing:

- packages/go/analysis/ad/ntlm.go:96 in NewNTLMCache, which decides which computers get edges
- packages/go/analysis/ad/ntlm.go:564 in GetCoerceAndRelayNTLMtoSMBEdgeComposition
- packages/go/analysis/ad/ntlm.go:764 in GetCoercionTargetsForCoerceAndRelayNTLMtoSMB

If you only patch the cache you get the edges, but the composition and coercion target panels will still be empty for those same computers. Setting the property to false at ingest fixes all three at once, since they all just read the property.

A few small things:

- packages/go/ein/ad.go:110 looks dead now. Both branches of the new if/else overwrite it, so that = nil is never relevant. Might as well delete it, otherwise the file says two different things about the default a few lines apart.
- The new test checks the property gets set to false, which is the important part. However, there is no checks on the rest of the path, i.e. that a computer ingested without the key actually comes out with a CoerceAndRelayNTLMTo* edge on it. The existing harnesses set RestrictOutboundNTLM by hand, so they never touch ConvertComputerToNode at all. Testing that properly probably means a new fixture pair under cmd/api/src/services/graphify/fixtures/, which is arguably more work than the fix itself. Up to you all. I'm only bringing it up because #1787 got to release through that exact missing test.

sanjio3:
@chryzsh thanks for the close read — that's exactly right, and I've made the fix you flagged.

The `RestrictOutboundNTLM = nil` assignment in the registry clear block was indeed dead: both branches of the `RestrictSendingNtlmTraffic` check now assign it (the parity comparison when present, and `false` when the key is absent), so the initial nil never survives. I've removed that line and tightened the function comment so the file now says one consistent thing about the default — a collected-but-absent key maps to `false`, not `nil`. The other eight registry values keep their nil-clearing because they have no such override.

Pushed to the PR as `182b3d24`.

On the e2e edge test — agreed it's the higher-value coverage and the #1787 precedent is a fair warning. That path needs a new fixture pair under `cmd/api/src/services/graphify/fixtures/` plus the full ingest→analysis path, which is more than this PR should carry. I'd rather keep this PR to the targeted fix. I'm happy to pick the fixture-based edge test up as its own follow-up PR if that's useful — let me know if you'd like me to, or if it's better kept for a maintainer.

sanjio3:
@chryzsh quick follow-up on the e2e coverage side.

I've added an integration test on this PR — `TestPostNTLMRelaySMBMissingRestrictOutboundNTLMFailsClosed` in `packages/go/analysis/ad/ntlm_integration_test.go` (pushed as `dd78c26`) — that exercises the analysis layer for the fail-closed side of the issue #2534.

It loads the `NTLMCoerceAndRelayNTLMToSMB` harness, drops `restrictoutboundntlm` off one of the target computers (simulating a node whose registry data was never collected, i.e. the property is absent — which is the state `ConvertComputerToNode` intentionally leaves untouched), then runs the NTLM post-processing. The assertion is that `NewNTLMCache` excludes that computer and it does *not* gain a `CoerceAndRelayNTLMToSMB` edge, while the computer whose property is present-`false` keeps its edge. That locks in the no-false-positive guarantee the ingest fix's design relies on (and which the pre-merge risk note called out).

To be clear about scope: this does *not* cover the positive ingest path (missing key → property `false` → edge generated), which is precisely the fixture-pair test you described under `cmd/api/src/services/graphify/fixtures/`. That still needs the new fixture pair plus a full ingest→analysis run, and I still intend for it to be a separate follow-up PR rather than growing this one.

PR #3307, opened 2026-09-13
fix(ui): preserve deep link path after login - BED-5483

### LINKED ISSUE #2806: Bug: UI/UX Feedback: Unauthenticated deeplinks do not redirect

sanjio3:
Fixes #2806

## Summary

When an unauthenticated user opens a deep link (e.g. `/explore`) and then logs in, the app always redirected to the home page, losing the originally requested path.

## Changes

- `Login.tsx`: read `location.state.from` after a successful login and navigate there first; fall back to the home route when the saved path is missing or is the login route itself
- `Login.test.tsx`: new tests covering deep-link redirect, missing state, and login-route fallback

## Test plan

- Vitest: 3 new tests pass (package-level suite green)

## Summary by CodeRabbit

- **New Features**
  - Authenticated users visiting the login page are redirected to their intended destination when available, including query parameters and URL fragments.
  - Redirects safely fall back to the home page when no destination is provided or the destination is the login page itself.

- **Tests**
  - Added coverage for authenticated-user redirects, including deep links, missing navigation state, login-route fallback behavior, and preservation of URL details.


sanjio3:

PR #3309, opened 2026-09-13
test(analysis): guard service principal app id resolution against regressions

### LINKED ISSUE #1846: Bug: Service Principal App ID node property displaying Service Principal Object ID instead of App ID
## Description:

The node entity panel for AZServicePrincipal objects displays the object ID of the service principal in the App ID property instead of the application ID. 

Yes

## Component(s) Affected:

- API

## Steps to Reproduce:

Ingest an azurehound payload for an Entra ID tenant with a service principal (e.g., specterdev), search for a service principal in the Explore tab, and click a service principal node. Note that the object ID and app ID are the same and that the app ID does not reflect the application ID noted in Entra ID. 

## Expected Behavior:

The application ID should be displayed in the app ID property of the service principal node entity panel. 

## Actual Behavior:

The object ID is displayed in the app ID property of the service principal node entity panel instead of the application ID. 

## Environment Information:

BloodHound: 8.1.0

Collector: AzureHound 2.4.1

OS: Windows 11

## Additional Information:

https://specterops.atlassian.net/browse/BED-6431

The issue is that the FetchServicePrincipalApplications function called by getServicePrincipalAppID function (https://github.com/SpecterOps/BloodHound/blob/main/packages/go/analysis/azure/service_principal.go#L58 ) gets the end node (the service principal) instead of the application (the end node) on this line: https://github.com/SpecterOps/BloodHound/blob/0c823b12f27b29aa36df0342a49769e20a107e99/packages/go/analysis/azure/queries.go#L567 

Changing FetchEndNodes to FetchStartNodes resolves the issue.

## Contributor Checklist:

- [X] I have searched the issue tracker to ensure this bug hasn't been reported before or is not already being addressed.
- [X] I have provided clear steps to reproduce the issue.
- [X] I have included relevant environment information details.

sanjio3:
Closes #1846

## Summary

The original report was that the Service Principal panel showed the Object ID where the App ID should be. While investigating, the production query already resolves via `FetchStartNodes` on the current main, so the underlying bug is fixed upstream. This PR adds regression tests to lock in the correct behavior and prevent future regressions.

## Changes

- Add 133 lines of regression tests guarding the Service Principal App ID resolution

## Test plan

- `go test ./packages/go/analysis/ad/ -run TestServicePrincipal` passes

## Summary by CodeRabbit

- **Tests**
  - Added coverage for Azure service principal analysis.
  - Verified that linked application nodes are returned correctly through service principal relationships.
  - Verified that application IDs are read from the linked application’s object ID.


sanjio3:

PR #3326, opened 2026-09-16
chore: Icon/Icon Button refactor - BED-6062

KillahDillah:
## Description

## Motivation and Context

Resolves BED-6062

Update and refactor Icon and IconButton component. Also refactors some of Tooltip where Icon is being used.
Previously, Icon rendered a button or a tooltip, ignoring aria-labels.

## How Has This Been Tested?
Visually and new tests have been added.

## Types of changes

- Database Migrations

## Checklist:

  - Assigned myself to this PR
  - Open API docs

## Summary by CodeRabbit

- **New Features**
  - Added accessible icon and icon-button controls with variant, sizing, and tooltip support.
  - Added stable IDs and field names to date-range inputs for improved form integration.

- **Improvements**
  - Buttons now default to non-submit behavior.
  - Standardized icon, button, and tooltip presentation across shared interface components.
  - Improved dialog focus handling, tooltip sizing, and control accessibility.
  - Updated help links and filtering controls for clearer interaction.


packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -113,6 +100,7 @@ const Tooltip: React.FC<TooltipProps> = (props) => {
         contentWidth,
         contentProps = {},
     } = props;
+    const defaultTriggerLabel = !props.children && typeof tooltip === 'string' ? tooltip : undefined;
```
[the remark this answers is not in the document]
bfaulk96: > Use a generic fallback label, or require triggerProps['aria-label'] for this supported API path.

I like the idea of requiring aria-label if tooltip isn't a string and there are no children, but that sounds like some complicated TS type gymnastics

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
```diff
@@ -521,25 +514,21 @@ export const TagForm: FC = () => {
...
                                                                 align: 'end',
-                                                            }}>
-                                                            <span>
-                                                                <AppIcon.Info />
-                                                            </span>
-                                                        </Tooltip>
-                                                    </FormLabel>
+                                                            }}
+                                                        />
+                                                    </div>
```
bfaulk96, line 531: Wait, how is this tooltip (and the other one in this component) triggered? It's not wrapping any content; does it just attach itself to the parent or something?

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,36 @@
...
+    'aria-label': string;
+    children: React.ReactElement;
+    className?: string;
+}
+
+export const Icon: React.FC<IconProps> = ({ 'aria-label': ariaLabel, children, className }) => {
+    const accessibleIcon = React.cloneElement(children, { 'aria-label': ariaLabel } as React.HTMLAttributes<HTMLElement>);
+
+    return (
+        <Tooltip tooltip={ariaLabel} contentProps={{ side: 'bottom', align: 'start' }}>
```
bfaulk96: Might be nice to have an optional param like `tooltipPosition` for scenarios where the tooltip would need to be elsewhere? Though I suppose forced consistency may be intentional?

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Tooltip } from '../Tooltip';
+import { cn } from '../utils';
+
+export interface IconProps {
+    'aria-label': string;
+    children: React.ReactElement;
+    className?: string;
+}
+
+export const Icon: React.FC<IconProps> = ({ 'aria-label': ariaLabel, children, className }) => {
```
bfaulk96: I worry about using `aria-label` as the tooltip label; the main reason being that it's sort of hidden functionality. While not the biggest deal, I could see capitalization being treated differently in a tooltip vs an aria-label, and an unaware developer could accidentally make a bad-looking tooltip that would have otherwise been fine as a screen reader label.. Having `label` and `aria-label` separate would also allow for using `ReactNode` as your `label` component. I understand this would require a *lot* of refactoring, though... It just kind of sits wrong with me to use a prop specifically for accessibility to show UI content.

Some fun type gymnastics like this could achieve something helpful:
```tsx
type CommonProps = {
  ... // All the other props
};

type AccessibleLabelProps =
  | {
      label: string;
      'aria-label'?: string;
    }
  | {
      label?: Exclude<ReactNode, string>;
      'aria-label': string;
    };

type IconProps = CommonProps & AccessibleLabelProps;
```

packages/go/analysis/ad/entity_details_integration_test.go
KillahDillah, line 1: I am trying to use rebase more than merge - this was something that came in with the rebase, and it's causing errors. what happened here?

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
KillahDillah: This is a great point - and something to visit in BED-7769 as this one is already spiraling. I might have included way too much with Tooltip in this Icon refactor.

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
KillahDillah: Hmmm - good point. Maybe something to discuss? I've gone a bit overboard on the Tooltip changes, maybe this is something that either is merged in with this current PR, or I can work it out of this branch and revisit it in the Tooltip refactor BED-7769. What do you think?

packages/javascript/bh-shared-ui/src/components/GraphMenu/GraphMenu.tsx
```diff
@@ -16,29 +16,23 @@
...
-                triggerProps={{ asChild: true, className: 'pointer-events-auto' }}
-                contentProps={{ className: 'dark:bg-neutral-4 dark:border-neutral-5 dark:text-white' }}>
-                <MenuTrigger asChild>
-                    <IconButton aria-label={label} data-testid={testId}>
-                        <FontAwesomeIcon icon={icon} />
-                    </IconButton>
-                </MenuTrigger>
-            </Tooltip>
+            <MenuTrigger asChild>
+                <IconButton aria-label={label} data-testid={testId}>
```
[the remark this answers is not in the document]
KillahDillah: GraphMenu tooltips are displaying - visually verified that they are

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
KillahDillah: So - I set the default Tooltip trigger to be the AppIcon.Info. Check the Tooltip component for TooltipTrigger. This is yet another Tooltip improvement that could be part of the Tooltip refactor. But felt like it was closely related to the Icon refactor. Went a little overboard.

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
bfaulk96: I kind of lean toward working it *out* of this change.. I feel like there may be scenarios where we don't necessarily want a tooltip for an `Icon` or `IconButton`, so having a `label` that would differentiate that we specifically want to show a tooltip makes it a cleaner distinction between accessible or just providing additional details.

Or maybe a happy medium would be adding the `label` support for the spots where you already refactored to remove the surrounding tooltip, but then make a note in that tooltip refactor to change the `aria-label` fields to `label` in other icons/iconbuttons where we wanted a tooltip?

I could kind of go either way, though. How do you feel?

packages/javascript/bh-shared-ui/src/views/PrivilegeZones/Save/TagForm/TagForm.tsx
bfaulk96: Ah, makes sense. I hadn't made it that far yet. Thanks!

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.test.tsx
bfaulk96, line 1: I'm a little confused by this change - the `searchResultsMenu` variable uses that getByTestId you've replaced it with (for both of these tests)

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
```diff
@@ -0,0 +1,177 @@
...
+            table: {
+                category: 'Accessibility',
+                type: {
+                    summary: 'string',
+                },
+            },
+        },
+        tooltip: {
+            description:
+                'Optional visible tooltip content. Defaults to the aria-label; override it only when additional context is needed.',
```
bfaulk96: Ohhh see I almost feel like tooltip should _only_ show if this is provided? not fall back to aria-label?

packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -28,30 +29,15 @@ type TriggerProps = React.ComponentPropsWithoutRef<typeof TooltipPrimitive.Trigg
 
 const TooltipTrigger = React.forwardRef<React.ElementRef<typeof TooltipPrimitive.Trigger>, TriggerProps>(
     (props, ref) => {
-        const { className, ...rest } = props;
-        const asChild = !!props.children;
+        const { children, asChild = !!children, className, type, ...rest } = props;
```
bfaulk96: I'm sure I'm just missing something, but why `asChild = !!children`? Logically to me, it's confusing that we'd mark this as a child simply because it has children

packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -113,6 +104,20 @@ const Tooltip: React.FC<TooltipProps> = (props) => {
         contentWidth,
         contentProps = {},
     } = props;
+    const defaultTriggerLabel = !props.children && !renderTrigger
```
bfaulk96: We use `props.children` in at least two places - we should just destructure that like the other props above

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.test.tsx
KillahDillah: Not sure what the deal is - but when I use the variables, the tests fail. This way, the tests don't fail.

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
KillahDillah: So - I did this because I noticed in a few spots that we used the same exact string for aria-label as we did the tooltip. Felt redundant. Also didn't want a screen-reader user to have a different experience than a sighted user. 

It's also ok to use a FontAwesomeIcon or AppIcon without using this Icon component and wrap that in a Tooltip. So, maybe this is renamed to something other than `Icon`? 

In the previous use of `Icon` it was used to either render an IconButton, or render an Icon with a Tooltip, but also no aria-label support. I was going off of that.

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
KillahDillah: There's places in our UI where the Tooltip isn't used with an Icon and it gives no meaning of what the Icon is actually doing. Leaving the possibility of a user being confused to what the Icon is actually being used for. This thought came from the BofA audit where differing colors came off as confusing without a legend. I felt like we should be using the Tooltips and aria-labels in more places than we actually are.

Maybe I have the wrong idea?

packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
KillahDillah: OK - this took me a second to untangle some code - but I made a fix for this.

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
bfaulk96: Hmm, I agree with you on that overall point I think. Just don't love the idea of forcing a tooltip *every* time... Maybe make it opt-out since it's more of an edge case that you *wouldn't* want a tooltip? Like `hideTooltip` or something? That way it would still default to clearer buttons and icons unless there was a good reason not to have a tooltip on them?

packages/javascript/bh-shared-ui/src/components/GraphControls/GraphControls.test.tsx
bfaulk96: Roger that 🫡

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
bfaulk96: > I noticed in a few spots that we used the same exact string for aria-label as we did the tooltip. Felt redundant

I was leaning toward not requiring an `aria-label` if a `label` was provided (and using the `label` as `aria-label` internally).

That said, while I don't *love* using the aria-label prop for a tooltip, I do think it's an easy win without requiring opt-in from every UI dev using icons, so I don't feel that strongly – you can leave as is 😄

packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
```diff
@@ -28,28 +29,20 @@ type TriggerProps = React.ComponentPropsWithoutRef<typeof TooltipPrimitive.Trigg
 
 const TooltipTrigger = React.forwardRef<React.ElementRef<typeof TooltipPrimitive.Trigger>, TriggerProps>(
     (props, ref) => {
-        const { className, ...rest } = props;
-        const asChild = !!props.children;
+        const { children, asChild = children == null || React.isValidElement(children), className, type, ...rest } =
```
bfaulk96, line 32: So what is the purpose/goal of `asChild` here? because with the new logic, if not passed directly, it's true if children is `null`, `undefined`, or an element.

I'm sorry for asking about this again, just not fully grasping what that variable is meaning/trying to accomplish in general...

By name I'd expect it to mean that we treat this as a child if:

```tsx
<Button>
  <TooltipTrigger />
  Hello world
</Button>
```
But that doesn't seem to track with the `|| React.isValidElement(children)` condition

bfaulk96: Not saying what you have is wrong or needs changed – truly just asking for clarification since I'm lost, ha

KillahDillah: I'm still working on this - But this is being suggested over and over again by AI, stating that without it it can cause an error with Radix rendering a button within a button. Not quite finished yet though.

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
KillahDillah: I like that idea. Will make that change.

packages/go/analysis/ad/entity_details_integration_test.go
KillahDillah: manually removed

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
bfaulk96: Also something we can add *if* needed in the future, so consider at your own discretion. Feel free to resolve this if it's not something you'd want to add now

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
KillahDillah: Oh - I understand, like, the reverse of how I have it?

bfaulk96: Yep, exactly! But as I mentioned, that gets a little weirder on types, because you *have* to have either an aria-label or a label at all times. But either way, I'm ultimately okay with the route you took here

packages/javascript/doodle-ui/src/components/Tooltip/Tooltip.tsx
KillahDillah: OK - after messing with this, trying to understand and then pulling Nando in. I understand why AI was so adamant about this change.

asChild is used to override Radix's default trigger of a `button`. This is needed because I've set a new default to an icon (which should not render as a button). asChild must be `true` because the default is passing in `children`.  `React.isValidElement(children)` is added as a safer type check so a string can't be passed in, only a JSX element.

bfaulk96: Alright, makes sense. Thanks for the clarification!

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
KillahDillah: Change made

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
```diff
@@ -0,0 +1,275 @@
...
+                <div className='flex flex-col items-center gap-4'>
+                    <IconButton aria-label='Open Menu' size={18} variant={variant}>
+                        <Icon label='Open Menu'>
+                            <FontAwesomeIcon icon={faEllipsisVertical} />
+                        </Icon>
+                    </IconButton>
+                    Enabled
+                </div>
+                <div className='flex flex-col items-center gap-4'>
+                    <IconButton aria-label='Delete item' disabled size={18} variant={variant}>
```
bfaulk96: nit: I know it's just a pointless example, but should probably say something like `Open filters` given the app icon we're using

Image

bfaulk96: There are several other examples like this, as well

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,63 @@
...
+export const Icon: React.FC<IconProps> = ({
+    'aria-hidden': ariaHidden = false,
+    'aria-label': ariaLabel,
+    children,
+    className,
+    hideTooltip = false,
+    label,
+}) => {
+    const accessibleLabel = ariaLabel ?? label;
+    const tooltipLabel = label ?? accessibleLabel;
```
bfaulk96: Should probably use `||` to account for empty string as well (I'd think if there was an empty string, we'd want to default to the other? (Also makes it hard to play with some of this behavior in Doodle UI witthout)
```suggestion
    const accessibleLabel = ariaLabel || label;
    const tooltipLabel = label || accessibleLabel;
```

bfaulk96 (commented):
Just a couple more comments, sorry

packages/javascript/doodle-ui/src/components/IconButton/IconButton.stories.tsx
KillahDillah: changed!

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
KillahDillah: done

bfaulk96 (approved):
Tested locally; looks good!

packages/javascript/bh-shared-ui/src/components/CreateUserDialog/CreateUserDialog.tsx
```diff
@@ -57,9 +59,17 @@ const CreateUserDialog: React.FC<{
             </DialogTrigger>
             <DialogPortal>
                 <DialogOverlay>
-                    <DialogContent maxWidth='lg' className='!bg-transparent shadow-none max-h-screen overflow-y-auto'>
+                    <DialogContent
+                        maxWidth='lg'
+                        className='!bg-transparent shadow-none max-h-screen overflow-y-auto'
```
TheNando: This isn't the best solution. The real issue here is that we have a card within a dialog. The dialog already is providing a visual surface container, so it is redundant to also have a cards.

packages/javascript/bh-shared-ui/src/components/CreateUserDialog/CreateUserDialog.tsx
```diff
@@ -57,9 +59,17 @@ const CreateUserDialog: React.FC<{
...
+                        onOpenAutoFocus={(event) => {
+                            event.preventDefault();
+                            dialogTitleRef.current?.focus();
+                        }}>
```
TheNando: I'm not sure it makes sense to set a visual focus on a hidden element. I'm guessing this aimed at visual impairment accessibility? We might want to revisit this. Other accessibility guidelines state keyboard focus should be visible when the dialog opens.

packages/javascript/bh-shared-ui/src/views/Explore/ExploreSearch/CypherSearch.tsx
```diff
@@ -382,6 +382,7 @@ const CypherSearchInner = ({
                             </Button>
                             <SaveQueryActionMenu saveAs={handleSaveAs} />
 
+                            {/* below is a link, rather than a Button, as it launches to a new page */}
```
TheNando: While this may feel a little awkward, it is correct. Though, if we find more of these, we may consider creating an `IconLink` component.

packages/javascript/doodle-ui/src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,63 @@
...
+          label: string;
+      }
+    | {
+          'aria-label': string;
+          label?: never;
+      };
+
+export type IconProps = IconBaseProps & IconLabelProps;
+
+export const Icon: React.FC<IconProps> = ({
```
TheNando: Hmmm. I got a lot of feelings about this component. I'm not convinced it's worth it.

PR #3329, opened 2026-09-17
feat: add collector job type client support BED-9716

bfaulk96:
## Description

Adds the JavaScript client-library support needed by Managed Collection forms to list collector job types, including their display metadata and icon fields.

## Motivation and Context

Resolves BED-9716

The Collection Plan form needs the available built-in job types to populate its service selector and retain the selected type by ID.

## How Has This Been Tested?

- unit tests
- manual testing in BHE

## Types of changes

## Checklist:

  - [ ] Assigned myself to this PR
  - [ ] Added the appropriate labels
  - [x] Associated an issue: BED-9716
  - [ ] Read the Contributing guide
  - [ ] Open API docs (no endpoint contract changes in this PR)
  - [x] Code comments / JSDocs where applicable
  - [x] Added/updated tests to cover my changes
  - [x] All new and existing focused tests passed

## New Features

* Added JavaScript client access to collector job type data with pagination and detailed configuration information.
* Added reusable form validation for required fields, names, and descriptions, including configurable length, formatting, and whitespace rules.
* Made collector job type information available through the shared client library for application integrations.

## Tests

* Added coverage for required-field, name, and description validation behaviors, including custom configuration boundaries and whitespace handling.


packages/javascript/js-client-library/src/types.ts
```diff
@@ -1018,6 +1018,28 @@ export interface CollectorJobSecret {
...
 
+export interface CollectorJobType {
+    id: number;
+    key: string;
+    display_name: string;
+    icon?: string;
+    icon_color?: string;
+    params_version: string;
+    params_schema: Record<string, unknown>;
+    backoff_strategy: 'exponential' | 'fixed';
```
specter-flq, line 1027: I would have to check the contract to see if we use these somewhere else, but just from here I wonder if we should consider adding a type for these values instead of just adding the strings here.

packages/javascript/bh-shared-ui/src/utils/formValidationRules.ts
specter-flq, line 1: ❤️

packages/javascript/bh-shared-ui/src/utils/formValidationRules.test.ts
specter-flq, line 1: Debatable one, gonna give it more thought, but wanted to throw it out there. Feels like this is testing if react hook form is doing its job. And I know in the past we have not tested things from libraries and just relied on their tests. On the basis that we picked to use that so we are trusting them to have a working product in a way.

packages/javascript/js-client-library/src/types.ts
bfaulk96: I'd normally agree with you on this, but there isn't really a reason for the frontend to be using this variable - it's just included in the type to match the API schema.

I suppose in the future maybe? But at that point, we could always change it to an enum when we needed to use it?

packages/javascript/bh-shared-ui/src/utils/formValidationRules.test.ts
bfaulk96: I'm torn on this. 100% agree that if we trust a library enough to use it, we shouldn't need to be testing the library. On the other hand though, I'm not sure having tests that just expect a specific object output are all that helpful. Using RHF helps us visualize whether our helper function built the correct object to do what we are claiming it does for RHF, which seems valuable (it's more of "test that our helper results in a config that does what we want with RHF" rather than "test that RHF works")

packages/javascript/js-client-library/src/types.ts
specter-flq: yup that works, if we see it somewhere else we can always abstract to reuse

packages/javascript/bh-shared-ui/src/utils/formValidationRules.test.ts
specter-flq: Yup agreed agreed that's a good argument for it.

specter-flq (approved):

bfaulk96:
Just leaving myself a reminder to remove the `icon`/`icon_color` fields

Update: done ✅

